PR #32, opened 2025-10-31
First cut at helm chart.

ktpedre:
This deploys the atlas-ui-3 only with no authentication. A future update will add a nginx reverse proxy gateway that interfaces with an oauth2/oidc provider.


garland3:
We should move this to the scripts folder and update the readme.md

PR #273, opened 2026-01-29
Updating to the file-extractor-mock to use banyan-ingest for pdf text extraction

Yang-Ho:
Adding banyan-ingest as the first pdf text extractor that is tried before defaulting to pypdf. 

Users must specify a nemotron-parse endpoint url (https://build.nvidia.com/nvidia/nemotron-parse)

* Also includes a quick fix related to f stings


garland3:
Superseded by #277 which creates a separate banyan-extractor-mock service and fixes several issues (process_document filepath vs BytesIO, get_output_as_markdown vs get_content_list, missing json import, adds PPTX support).

PR #496, opened 2026-04-04
fix: file GET/DELETE 404 for slash-containing S3 keys

garland3:
## Summary

- **Bug:** File operations (GET, DELETE) on keys containing slashes (e.g. `users/agarlan@sandia.gov/generated/report.pdf`) returned 404 in production. The FastAPI route `{file_key}` only matches a single path segment, so the router stopped at the first `/`.
- **Fix:** Changed GET and DELETE route decorators from `{file_key}` to `{file_key:path}`, which captures the full multi-segment S3 key.
- **Route reorder safeguard:** Moved the greedy `{file_key:path}` catch-all routes to the end of the router, after all specific `/files/...` endpoints (healthz, list, download, stats). Without this, the greedy path parameter would swallow `/files/download/...` requests.

## What changed

**`atlas/routes/files_routes.py`** (1 file)

| Route | Before | After |
|-------|--------|-------|
| `GET /files/{file_key}` | `{file_key}` (single segment) | `{file_key:path}` (multi-segment) |
| `DELETE /files/{file_key}` | `{file_key}` (single segment) | `{file_key:path}` (multi-segment) |
| Route order | GET/DELETE before download | GET/DELETE moved after download |

## Validation

- 122 file-related tests pass (`pytest atlas/tests/ -k "file"`)
- Ruff linting clean
- No behavioral change for single-segment keys
- Download endpoint routing verified (was broken before reorder)

## Follow-up concern

`ChatArea.jsx` and `AllFilesView.jsx` use `encodeURIComponent()` on file keys, which encodes `/` → `%2F`. Depending on how Starlette handles `%2F` in path matching, those frontend components may still 404 for slash-containing keys even after this backend fix. `FilesPage.jsx` passes keys raw and is unaffected. This is a separate pre-existing issue worth investigating.

## Test plan

- [ ] Verify file delete works for keys like `users/user@example.com/generated/file.pdf`
- [ ] Verify file GET works for the same slash-containing keys
- [ ] Verify `/files/download/...` still works (route ordering)
- [ ] Verify single-segment keys still work (no regression)


garland3:
Closing — redoing from sandialabs/main as a feature branch directly on the sandialabs repo. Replacement PR incoming.

PR #535, opened 2026-04-17
Fall back to sync call when MCP tool refuses task-augmented execution

garland3:
## Summary
- Tool calls were failing with `McpError: FunctionTool '...' does not support task-augmented execution` (`fastmcp/server/tasks/routing.py:63`) whenever the server advertised task capability at the capability-negotiation layer but the specific tool declared `tasks.mode="forbidden"`. The LLM saw no usable tool result and typically hallucinated around it ("I don't have a calculator…").
- Root cause: `MCPToolManager.call_tool` used server-level `_supports_tasks` as a blanket signal and sent `task=True` for every tool on a task-capable server.
- Fix (`atlas/modules/mcp_tools/client.py`): wrap the task-mode `call_tool` in a targeted try/except. On the forbidden-error message we record the `(server, tool)` pair in a new `_tool_task_forbidden` set and fall through to the existing synchronous call path. Subsequent invocations short-circuit to sync mode directly. Unrelated errors still propagate unchanged.
- Helper `_is_task_forbidden_error` matches on the documented fastmcp error message text (not the concrete exception class, since that varies across fastmcp versions).

## Why a string match?
`McpError`'s payload is an `ErrorData` with `code=METHOD_NOT_FOUND` — the same code fastmcp uses for "requires task-augmented execution" (the inverse mode). Discriminating by code alone would merge two distinct conditions. The message text is the narrow, stable signal.

## Test plan
- [x] `TestTaskForbiddenFallback::test_task_forbidden_falls_back_to_sync_and_returns_result` — first attempt uses `task=True`, raises; second attempt drops `task` and returns the sync result; cache populated.
- [x] `TestTaskForbiddenFallback::test_forbidden_cache_skips_task_mode_on_subsequent_call` — pre-seeded cache makes only one non-task call; no retry needed.
- [x] `TestTaskForbiddenFallback::test_non_forbidden_error_is_not_swallowed` — a `RuntimeError` in task-mode propagates; cache stays clean.
- [x] Existing `TestAdaptiveTaskPolling` tests still green (immediate, no-task-support, cancellation, timeout-then-poll).
- [x] Full backend suite: **1219 passed, 17 skipped**.
- [ ] Manual smoke: exercise the calculator `evaluate` tool from the app on a conversation and confirm the model now receives a real tool result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### PR #TBD - 2026-04-16
+- **Fix**: Tool calls failed with `McpError: FunctionTool '...' does not support task-augmented execution` when the server advertised task capability but the individual tool declared `tasks.mode="forbidden"`. `MCPToolManager.call_tool` now catches that specific error, falls back to a synchronous (non-task) call, and caches the `(server, tool)` pair so subsequent invocations skip task mode directly. Unrelated errors still propagate unchanged.
+
 ### PR #533 - 2026-04-15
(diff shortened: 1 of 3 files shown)


garland3:

PR #559, opened 2026-04-25
fix(mcp): scope per-user HTTP clients per conversation

garland3:
## Summary

- Cache FastMCP HTTP `Client` instances by `(user_email, server_name, conversation_id)` instead of `(user_email, server_name)`. Sharing one client across conversations accumulates FastMCP's reentrant nesting counter and, once the underlying session task dies, prevents reconnection with `"nesting counter should be 0 when starting new session, got N"`.
- Per-conversation keying also gives each chat its own MCP session ID, so stateful HTTP servers (e.g. the 3D-printer sim's per-session `PrinterService`) stay isolated across the same user's conversations rather than leaking state between chats.
- `_get_user_client`, `_get_or_create_user_http_client`, and `get_prompt` accept `conversation_id`; `_invalidate_user_client` wipes every `(user, server, *)` entry on token revocation; `release_sessions` only evicts the current conversation's entries so other conversations keep their live sessions; `prompt_override_service` forwards `user_email` + `conversation_id` into `get_prompt`.

## Review-driven fixes added in this PR

- **`handle_reset_session` releases the old conversation's MCP sessions** (Tier-1 A). Each reset previously orphaned the `(user, server, old_conv_id)` cache entries and `MCPSessionManager` sessions because the cache key now includes `conversation_id`. `handle_reset_session` now captures the old `conversation_id` before tearing down the session and invokes `tool_manager.release_sessions` (best-effort via `getattr`) before generating the new id.
- **`get_prompt` mirrors `call_tool`'s auth routing** (Tier-2 E). For HTTP MCP servers with `auth_type` of oauth/jwt/bearer/api_key, `get_prompt` now routes through `_get_user_client` (per-user token) instead of `_get_or_create_user_http_client` (admin token), and raises `AuthenticationRequiredException` with the same OAuth start URL the tool path emits.
- **`_get_or_create_user_http_client` rejects falsy `conversation_id`** (Tier-1 B). Drops the `Optional` default and raises `ValueError` so future regressions surface immediately rather than silently aliasing every caller for `(user, server)` into one shared cache slot.
- **Test coverage** for the actual behavior, not just cache-key shape (Tier-1 C, D):
  - New `tests/test_mcp_session_isolation_integration.py` drives a faithful `FakeFastMCPClient` (with real `__aenter__`/`__aexit__`/nesting counter) through the real `MCPSessionManager` to verify cross-conversation independence and recover-from-dead-session, plus a documentation test that reproduces the pre-fix `nesting counter should be 0` error on a deliberately shared client.
  - New unit tests for `release_sessions` per-conversation eviction (covers the previously untested behavior change in this PR).
  - New unit tests for `get_prompt` auth routing on auth-typed and unauth-typed servers, including the missing-token and missing-user-email error paths.
  - New unit tests for `handle_reset_session` calling `release_sessions(old_conv_id)` and tolerating exceptions from it.

## Disposition of remaining review items

Tier-3 items are pre-existing conditions this PR makes more visible but does not regress; opened as follow-ups so this PR stays focused:

- Issue #560 — `_invalidate_user_client` should release live MCP sessions on token revocation.
- Issue #561 — `MCPSessionManager.acquire` is not user-scoped; chat path trusts WS-supplied `conversation_id` without ownership validation.
- Issue #562 — bound `_user_clients` cache (LRU/TTL) and add defensive cleanup on ungraceful WebSocket disconnect.

## Test plan

- [x] `pytest atlas/tests/test_mcp_client_auth.py atlas/tests/test_mcp_per_user_http_clients.py atlas/tests/test_multi_prompt.py atlas/tests/test_mcp_session_isolation_integration.py atlas/tests/test_chat_history.py atlas/tests/test_attach_file_flow.py` — all pass locally.
- [x] Faithful (non-mock) integration-style test reproduces the pre-fix `nesting counter should be 0` failure mode on a shared client and asserts that the post-fix per-conversation cache prevents it. Covers the original manual repro's intent (two conversations on the same user driving a stateful HTTP MCP server) without depending on a live `3dprintsim` server.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,29 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### PR #559 - 2026-04-25
+- MCP cross-conversation isolation: cache FastMCP HTTP `Client` instances by
+  `(user_email, server_name, conversation_id)` so each conversation gets its
+  own MCP session ID and FastMCP nesting counter. Fixes the
+  "nesting counter should be 0" reconnect failure that surfaced after a
+  shared client's session task died, and isolates stateful HTTP servers
+  (e.g. the per-session `PrinterService`) across the same user's conversations.
+- `handle_reset_session` now releases the previous conversation's MCP
+  sessions and per-conversation HTTP clients before generating a new
+  `conversation_id` — previously each "New chat" click orphaned the old
+  `(user, server, old_conv_id)` cache entries and `MCPSessionManager` sessions.
+- `get_prompt` now mirrors `call_tool`'s auth routing on HTTP MCP servers
+  with `auth_type` of oauth/jwt/bearer/api_key: requests go through the
+  user's stored token instead of the admin/server-default token, and missing
+  tokens raise `AuthenticationRequiredException` with the OAuth start URL.
+- `_get_or_create_user_http_client` now requires a non-empty
+  `conversation_id` (a `None` value would alias every caller into one
+  shared cache slot, recreating the bug this cache exists to prevent).
+- Added integration-style tests using a faithful `FakeFastMCPClient` that
+  drives the real `MCPSessionManager` and reproduces the pre-fix nesting
+  counter failure mode, plus unit coverage for `release_sessions`'s
+  per-conversation eviction.
+
 ### PR #557 - 2026-04-22
 - MCP task-augmented execution fixes: discovery-time seeding of task-forbidden
   cache from per-tool execution.taskSupport metadata (SEP-1686), and runtime
--- a/atlas/application/chat/preprocessors/prompt_override_service.py
+++ b/atlas/application/chat/preprocessors/prompt_override_service.py
@@ -65,7 +65,11 @@ class PromptOverrideService:
                     meta["conversation_id"] = conversation_id
 
                 prompt_obj = await self.tool_manager.get_prompt(
-                    server, prompt_name, meta=meta if meta else None
+                    server,
+                    prompt_name,
+                    meta=meta if meta else None,
+                    user_email=user_email,
+                    conversation_id=conversation_id,
                 )
                 prompt_text = self._extract_prompt_text(prompt_obj)
 
--- a/atlas/application/chat/service.py
+++ b/atlas/application/chat/service.py
@@ -392,8 +392,25 @@ class ChatService:
         does not overwrite the previous one (session_id stays the
         same for the lifetime of the WebSocket connection).
         """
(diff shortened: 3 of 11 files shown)


garland3:
Pushed five commits in response to the multi-agent review. Summary of disposition:

**Fixed in this PR (Tier-1 + Tier-2 blockers):**

| Item | Commit | What changed |
|---|---|---|
| A — `handle_reset_session` leaks old conversation | `89afdf1` | Captures the old `conversation_id` and calls `tool_manager.release_sessions(old_conv_id, user_email=user_email)` (best-effort via `getattr`) before generating the new id. Tested via two new unit tests in `test_chat_history.py` (happy path + swallows-errors). |
| E — `get_prompt` may bypass per-user auth | `57e38da` | Confirmed: `get_prompt` only checked `_is_http_server`, not `_requires_user_auth`, so prompts on oauth/jwt/bearer/api_key servers ran under the admin token. Now mirrors `call_tool`'s routing; raises `AuthenticationRequiredException` with the OAuth start URL when no user token is available. Three new tests in `test_mcp_client_auth.py`. |
| B — silent `conversation_id=None` footgun | `a597638` | `_get_or_create_user_http_client` now requires a non-empty `conversation_id` and raises `ValueError` on falsy input. New test asserts the guard rejects both `None` and `""`. |
| D — `release_sessions` per-conversation eviction had zero coverage | `a597638` | Added `test_release_sessions_evicts_only_target_conversation` and `test_release_sessions_without_user_email_only_releases_session_manager`. |
| C — mocked tests can't verify the actual fix | `d9e84c3` | New `test_mcp_session_isolation_integration.py` uses a `FakeFastMCPClient` with real `__aenter__`/`__aexit__`/nesting counter, drives the real `MCPSessionManager`, and verifies cross-conversation independence + recover-from-dead-session. Includes a documentation test that reproduces the pre-fix `nesting counter should be 0` error on a deliberately shared client to prove the fake's failure shape matches FastMCP's. |
| CHANGELOG | `1a10068` | Added the missing `## [Unreleased]` entry for #559. |

**Deferred to follow-up issues (Tier-3 — pre-existing, not regressions):**

- #560 — `_invalidate_user_client` should also close live MCP sessions on token revocation.
- #561 — `MCPSessionManager.acquire` is not user-scoped; WS-supplied `conversation_id` is trusted without ownership validation. Pre-existing security concern this PR makes more visible.
- #562 — Bound `_user_clients` cache (LRU/TTL) and defensive cleanup on ungraceful WebSocket disconnect.

The manual repro from the original PR body has been replaced by the integration-style test in `test_mcp_session_isolation_integration.py`, which exercises the same scenario (two conversations, server-side session death, recovery) without depending on a live `3dprintsim` server.

PR #578, opened 2026-05-01
fix(packaging): ship prompt templates as atlas package resources

### LINKED ISSUE #572: System prompt markdown files not included in PyPI package
### Description
Atlas version: v0.1.5
`atlas-server` appears to rely on prompt files such as:
``` 
prompts/system_prompt.md
```
but those files are not being installed when the package is installed from PyPI with `pip install atlas-chat`.

The repository contains a top-level prompts/ directory with files like:
``` 
prompts/system_prompt.md
prompts/agent_system_prompt.md
...
```

However, after installing from PyPI, those files are not present in site-packages, and the `atlas-server` logs:
```
atlas.modules.prompts.prompt_provider - WARNING - Prompt template not found: <pythonpath>/site-packages/prompts/system_prompt.md
```

### Likely cause
pyproject.toml currently only packages the atlas* packages:

``` 
[tool.setuptools.packages.find]
where = ["."]
include = ["atlas*"]
```

and package data is only defined for atlas:
``` 
[tool.setuptools.package-data]
atlas = [
    ...
]
```

Since prompts/ is a top-level directory and not part of the atlas package, it is not included in the built wheel.

### Suggested fix
Long term, it would be cleaner to move prompts/ under atlas/ and load the files as package resources.

A smaller packaging fix would be to make prompts/ an installable package and include its *.md files in the wheel.

### Repro
```bash
pip install atlas-chat
```
Then inspect site-packages and note that prompts/system_prompt.md is missing.

garland3:
Closes #572. Replaces #573 (which took the shortcut of making top-level `prompts/` an installable package while leaving the brittle `__file__/../../..` walk in `PromptProvider`).

## Why the previous behaviour broke
`PromptProvider.__init__` resolved its base path by walking three directories up from `__file__` to land at the "repo root", then joining `prompt_base_path` ("prompts") onto it. That worked in dev because `prompts/` happened to sit next to the source tree. Once installed from PyPI the walk landed in `site-packages/`, where no `prompts/` directory exists — so `atlas-server` logged `Prompt template not found` for every default template (`system_prompt.md`, `agent_*.md`, `tool_synthesis_prompt.md`).

## What this PR changes
- **Move `prompts/*.md` → `atlas/prompts/*.md`.** Single source of truth that ships with the wheel.
- **`PromptProvider` now uses `importlib.resources.files("atlas") / "prompts"`** by default. The `__file__`-walk is gone.
- **`prompt_base_path` keeps its filesystem-override role.** Default value (`"prompts"` or empty) → bundled resources. Anything else → filesystem path (absolute, or relative to CWD), opening the door for operators to drop in their own templates without touching the package.
- **`pyproject.toml`**: declare `prompts/*.md` under `atlas` `package-data`. No top-level `prompts` package — the templates are owned by the atlas package, period.
- **New `atlas/tests/test_prompt_packaging.py`**: asserts every expected template exists as an `importlib.resources` resource, exercises the bundled-resource branch from a cwd with no `prompts/` folder present (the exact failure mode from the issue), and verifies the absolute-path filesystem override still works.

## Verification
- Full backend suite: `1412 passed, 16 skipped` (one pre-existing collection error in `test_admin_mcp_server_management_routes.py` unrelated to this change — `from main import app` resolves to the pptx_generator main module).
- Built a wheel locally; confirmed `atlas/prompts/*.md` are present in the archive.
- Installed the wheel into a clean venv, ran `PromptProvider().get_system_prompt(...)` from `/tmp` (no source tree, no `prompts/` in cwd) — returns the formatted template, `base_path` resolves to `site-packages/atlas/prompts/`.

## Test plan
- [ ] CI passes
- [ ] Manual: `pip install` the built wheel, run `atlas-server`, confirm no `Prompt template not found` warnings on startup
- [ ] Manual (override): set `prompt_base_path` to a directory containing custom templates, confirm overrides take effect


garland3:
## Summary
Fixes installed-from-wheel breakage where `PromptProvider` walked `__file__/../../..` to find `prompts/`, which landed in `site-packages/` and caused "Prompt template not found" warnings on startup. Moves `prompts/*.md` under `atlas/prompts/` so they ship as package data, and switches `PromptProvider` to load templates via `importlib.resources.files("atlas") / "prompts"` by default. Keeps `prompt_base_path` as a filesystem override (absolute or CWD-relative) when set to anything other than the `"prompts"` sentinel, and adds `atlas/tests/test_prompt_packaging.py` to lock in both the bundled-resource and override paths.

## Verdict
Looks good.

---

PR #571, opened 2026-04-28
chore(scripts): add s3-prune-old-files.sh for cron-based S3 retention

garland3:
## Summary

Adds a cron-friendly script for deleting S3 objects older than N days, intended to run inside the Atlas app container against the same `S3_*` env it already uses.

The PR has been substantially reworked in response to the original review:

- **Now Python/boto3, not bash**, so it does **not** require the `aws` CLI (which is not installed in the image). `boto3` is already a runtime dependency of Atlas.
- **`scripts/` is now COPY'd into both `Dockerfile` and `Dockerfile-test`** — the original bash script wouldn't have shipped in the image at all.
- **List failures are loud**: a failed `list_objects_v2` call returns a distinct exit code (`2`) instead of silently exiting `0` with `scanned=0` (the bash version's process-substitution failure mode flagged by Codex P1).
- **Pagination is explicit** via the boto3 paginator — buckets with >1000 objects are handled.
- **Argument parsing is robust**: `--days` / `--prefix` without a value now produce a clean argparse error, not a `set -u` "unbound variable" traceback.
- **Versioning / object-lock behavior is documented**: the script issues unversioned `DeleteObject`, which on a versioned bucket creates a delete marker (does not free space) and on object-locked objects will be reported as failures. Recommends an S3 Lifecycle rule for those cases.
- **16 regression tests** in `atlas/tests/test_s3_prune_script.py` cover every failure mode above (list failure, partial delete failure, dry-run safety, paginator iteration, missing args, naive datetime).

### Files

- `scripts/s3_prune_old_files.py` — new Python script
- `scripts/s3-prune-old-files.sh` — removed (replaced)
- `atlas/tests/test_s3_prune_script.py` — new regression suite
- `Dockerfile`, `Dockerfile-test` — `COPY scripts/ ./scripts/`

### Exit codes

| Code | Meaning |
| --- | --- |
| 0 | success (including dry-run) |
| 1 | config / argument error |
| 2 | S3 listing failed — destructive op aborted, nothing was inspected |
| 3 | one or more deletes failed (partial work may have completed) |

## Cron usage

```
# Daily at 03:15 UTC, prune anything older than 30 days, log to file.
15 3 * * * docker exec atlas-app python3 /app/scripts/s3_prune_old_files.py --days 30 >> /var/log/s3-prune.log 2>&1

# Always test first:
docker exec atlas-app python3 /app/scripts/s3_prune_old_files.py --days 30 --dry-run
```

## Test plan

[...]

PR #579, opened 2026-05-04
chore(scripts): add S3 prune script (Python/boto3) for cron retention

garland3:
Replaces #571 (which was opened from a fork and could not pass the GHCR push step due to fork-token permissions). Same branch, now pushed to the upstream repo so CI can complete cleanly.

## Summary

Adds a cron-friendly script for deleting S3 objects older than N days, intended to run inside the Atlas app container against the same `S3_*` env it already uses.

- **Python/boto3**, not bash, so it does **not** require the `aws` CLI (which is not installed in the image). `boto3` is already a runtime dependency of Atlas.
- **`scripts/` is now COPY'd into both `Dockerfile` and `Dockerfile-test`** — without this the script would not have shipped in the image at all.
- **List failures are loud**: a failed `list_objects_v2` returns a distinct exit code (`2`) instead of silently exiting `0` with `scanned=0` (the bash version's process-substitution failure mode).
- **Pagination is explicit** via the boto3 paginator — buckets with >1000 objects are handled.
- **Argument parsing is robust**: `--days` / `--prefix` without a value produce a clean argparse error, not a `set -u` "unbound variable" traceback.
- **Versioning / object-lock behavior is documented**: the script issues unversioned `DeleteObject`, which on a versioned bucket creates a delete marker (does not free space) and on object-locked objects will be reported as failures. Recommends an S3 Lifecycle rule for those cases.
- **16 regression tests** in `atlas/tests/test_s3_prune_script.py` cover every failure mode above (list failure, partial delete failure, dry-run safety, paginator iteration, missing args, naive datetime).

### Files

- `scripts/s3_prune_old_files.py` — new Python script
- `scripts/s3-prune-old-files.sh` — removed (replaced)
- `atlas/tests/test_s3_prune_script.py` — new regression suite
- `Dockerfile`, `Dockerfile-test` — `COPY scripts/ ./scripts/`

### Exit codes

| Code | Meaning |
| --- | --- |
| 0 | success (including dry-run) |
| 1 | config / argument error |
| 2 | S3 listing failed — destructive op aborted, nothing was inspected |
| 3 | one or more deletes failed (partial work may have completed) |

## Cron usage

```
# Daily at 03:15 UTC, prune anything older than 30 days, log to file.
15 3 * * * docker exec atlas-app python3 /app/scripts/s3_prune_old_files.py --days 30 >> /var/log/s3-prune.log 2>&1

# Always test first:
docker exec atlas-app python3 /app/scripts/s3_prune_old_files.py --days 30 --dry-run
```

## Test plan

- [x] `python3 scripts/s3_prune_old_files.py --help` renders the documented usage block.
- [x] `python3 scripts/s3_prune_old_files.py --days` produces a clean argparse error (not a traceback).
- [x] `pytest atlas/tests/test_s3_prune_script.py` — 16 passed locally; full backend suite passes in CI debug + production modes.
- [x] `ruff check` clean.
- [ ] In a non-prod env: run `--dry-run --days 30` against a test bucket and confirm the listed objects match a manual `list_objects_v2 LastModified < cutoff` query.
- [ ] Run without `--dry-run` against the same bucket and confirm only the expected objects are deleted; rerun and confirm `deleted=0`.


garland3:
## Summary
Replaces a bash S3-retention script with a Python/boto3 implementation that deletes objects older than N days, intended for cron use inside the Atlas app container. Adds explicit pagination, distinct exit codes for list vs. delete failures, robust argparse handling, and 16 regression tests. Updates `Dockerfile` and `Dockerfile-test` to `COPY scripts/` so the script actually ships in the image.

## Verdict
Looks good.

---

PR #603, opened 2026-05-12
Close subprocess transports on event-loop teardown

garland3:
## Summary

CI was emitting `PytestUnraisableExceptionWarning` lines on every run (Python 3.12.13 and 3.13.2), all tracing back to `BaseSubprocessTransport.__del__` ending in `RuntimeError: Event loop is closed`. Tests still passed, but the warnings were noisy and the named test rotated run-to-run — classic GC-attribution red herring.

## Root cause

`ProcessManager.launch` (atlas/modules/process_manager/manager.py) registers `_wait_and_finalize`, and `cancel` registers an inner `_kill_if_still_alive`. Both sit on `await asyncio_proc.wait()`. When a test ends with the child still running, pytest-asyncio's per-test loop is torn down and those tasks are cancelled mid-wait, so asyncio never calls `_process_exited` on the `_UnixSubprocessTransport`. The transport is still reachable (through `pm._asyncio_procs` and the stream readers) with `_closed=False` and a stale `_loop`. A later GC pass fires `BaseSubprocessTransport.__del__` which, on Python 3.12, unconditionally calls `proto.pipe.close()` → `loop.call_soon(...)` on the closed loop → `RuntimeError: Event loop is closed`. pytest surfaces it as `PytestUnraisableExceptionWarning` against whatever test happened to be running at GC time. CPython [gh-114177](https://github.com/python/cpython/issues/114177) added a loop-is-closed guard to that branch in 3.13 which kills most 3.13 occurrences, but residual paths still trip.

Both a real cleanup bug in our code *and* a Python-version interaction — the test-time leak is what gives `__del__` anything to run against, and 3.12 is just stricter than 3.13 about the consequences.

## Fix

Handle `CancelledError` inside the two long-lived tasks that await `asyncio_proc.wait()`. On cancellation:
1. `os.killpg(getpgid(pid), SIGKILL)` to take down the whole process group (matches the rest of the module — `launch()` always starts children with `start_new_session=True`, so killing only the session leader would orphan descendants); fall back to `asyncio_proc.kill()` if the group kill fails.
2. `asyncio_proc._transport.close()` while the loop is still alive — this flips `transport._closed=True` so the later `__del__` short-circuits before touching the closed loop. Narrow exception scope to `(RuntimeError, OSError)` with debug logging so a real bug isn't swallowed silently.
3. Pop the entry from `_asyncio_procs`, re-raise `CancelledError`.

The shared cleanup is extracted into a single private helper `_kill_and_close_transport()` so the two call sites can't drift.

## Regression tests (`atlas/tests/test_process_manager.py`)
- `test_kill_and_close_transport_marks_transport_closed` — unit test of the helper.
- `test_wait_and_finalize_closes_transport_on_cancellation` — simulates pytest-asyncio's loop teardown by locating the `_wait_and_finalize` task in `asyncio.all_tasks()` and cancelling it.
- `test_loop_teardown_does_not_emit_unraisable_warning` — end-to-end: cancel + drop refs + `gc.collect()` inside `warnings.catch_warnings(record=True)`, asserts zero `unclosed transport` / `Event loop is closed` warnings.

All three tests fail against pristine `main` and pass with this PR — confirmed by running them with the manager.py change reverted.

## Verification

Full suite under `PYTHONDEVMODE=1 PYTHONWARNINGS=always` (which exposes `ResourceWarning: unclosed transport <_UnixSubprocessTransport ...>` before GC converts it to the unraisable form):

**Python 3.13.12**

| Metric | Before | After |
| --- | --- | --- |
| `_UnixSubprocessTransport` leak warnings | 7 | **0** |
| pytest warnings total | 33 | 1 |
| Tests | 1469 passed / 17 skipped | 1472 passed / 17 skipped (3 new) |

**Python 3.12.3** (after fix)

- `_UnixSubprocessTransport` leaks: **0**
- `PytestUnraisableExceptionWarning` / "Event loop is closed": **0**
- 1472 passed / 17 skipped

Residual warnings (1 on 3.13, 2 on 3.12) are unrelated infrastructure: an OTEL JSONL exporter file kept open across the test session, pytest-asyncio's own loop-teardown finalizer, and on 3.12 a couple of `subprocess.Popen.__del__` "still running" warnings (separate code path from the asyncio transport issue this PR fixes). Captured as low-priority follow-ups in `INVESTIGATION_unraisable_warning.md`.

## Review response

Inline comments from Codex / Copilot were all addressed in commit `6d54a53`:
- **Codex P2 — process-group kill on teardown:** now uses `os.killpg(getpgid(pid), SIGKILL)` to match `cancel()`'s semantics.
- **Copilot — narrow `except Exception`:** narrowed to `(RuntimeError, OSError)` with `logger.debug(...)` on failure.
- **Copilot — extract helper:** new `_kill_and_close_transport()` shared by both call sites.
- **Copilot — regression test:** three new tests added (see above).
- **Copilot — investigation doc wording:** updated to reference this PR.

## Test plan
- [x] `PYTHONDEVMODE=1 PYTHONWARNINGS=always pytest atlas/tests` clean on 3.13.12
- [x] `PYTHONDEVMODE=1 pytest atlas/tests` clean on 3.12.3
- [x] New regression tests fail against pristine `main`, pass with this PR
- [x] All five originally-CI-named tests still pass with zero warnings under 3.12 dev mode
- [ ] CI green on this PR

### DIFF
--- a/INVESTIGATION_unraisable_warning.md
+++ b/INVESTIGATION_unraisable_warning.md
@@ -1,0 +1,117 @@
+# PytestUnraisableExceptionWarning investigation
+
+Branch: `investigate/unraisable-py313-warning` (local-only, no PR.)
+
+## Symptom
+CI on Python 3.12.13 and 3.13.2 emits ~3–5
+`PytestUnraisableExceptionWarning` lines that all trace back to
+`BaseSubprocessTransport.__del__` ending in
+`RuntimeError: Event loop is closed`. Tests still pass. Different runs
+attribute the warning to different tests (the user-reported set spans
+`test_agent_portal_groups`, `test_agent_portal_polish`,
+`test_process_manager::test_get_missing_raises`,
+`test_websocket_auth_header`, `test_rag_citations`,
+`test_streaming_token_flow`) — i.e. the named test is whichever one
+happened to be running when GC ran, not the one that leaked.
+
+## Reproduction
+Running the named tests in isolation never emits the warning. Running
+the full suite under `PYTHONDEVMODE=1 PYTHONWARNINGS=always` on Python
+3.13.12 surfaces 7 `ResourceWarning: unclosed transport
+<_UnixSubprocessTransport pid=… running stdout=… stderr=…>` entries plus
+a clutch of follow-on pipe / file warnings — this is the same leak the
+CI run trips on, just before GC has had a chance to convert it to the
+unraisable form.
+
+## Root cause
+`atlas.modules.process_manager.manager.ProcessManager.launch` spawns
+each subprocess via `asyncio.create_subprocess_exec` and creates two
+background tasks per process: `_pump_stream`/`_pump_pty` and
+`_wait_and_finalize` (which awaits `proc.wait()` and pops the entry
+from `_asyncio_procs` once the child exits). `cancel` additionally
+schedules a fire-and-forget `_kill_if_still_alive` task.
+
+Several tests — both bare `pm = ProcessManager()` cases and the HTTP
+fixture in `test_agent_portal_polish.py` / `test_agent_portal_groups.py`
+— end while a child is still running. pytest-asyncio's per-test event
+loop is then torn down: pending tasks receive `CancelledError`,
+`asyncio_proc.wait()` aborts, and the asyncio subprocess transport
+never has `_process_exited` invoked. Result: the
+`_UnixSubprocessTransport` is reachable (still referenced through the
+about-to-be-dropped `ProcessManager`) but `_closed` is still `False`
+and `_loop` points at the now-closed loop.
+
+When a later test triggers a GC pass, `BaseSubprocessTransport.__del__`
+fires:
+* On Python 3.12 it unconditionally calls `proto.pipe.close()`, which
+  schedules `loop.call_soon(self._call_connection_lost, exc)` on the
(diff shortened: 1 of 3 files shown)


garland3:
## Summary
Adds `CancelledError` handling to the two long-lived tasks in `atlas/modules/process_manager/manager.py` that await `asyncio_proc.wait()` (`_wait_and_finalize` and the inner `_kill_if_still_alive`). On cancellation the change kills the child process, closes the asyncio subprocess transport while the loop is still alive, and removes the entry from `_asyncio_procs`. The goal is to eliminate `PytestUnraisableExceptionWarning: Event loop is closed` noise that surfaced from `BaseSubprocessTransport.__del__` after pytest-asyncio tore down its per-test loop with live subprocess transports still reachable.

## Verdict
Looks good.

---

garland3:
## Summary
Fixes `PytestUnraisableExceptionWarning` noise traced to `BaseSubprocessTransport.__del__` running against a closed event loop. `ProcessManager._wait_and_finalize` and the inner `_kill_if_still_alive` now catch `CancelledError`, force-kill the process group, and close the asyncio subprocess transport synchronously so its later `__del__` short-circuits. Shared cleanup is extracted into a `_kill_and_close_transport()` helper, and three regression tests cover the helper, the cancelled-wait path, and the end-to-end GC behavior.

## Verdict
Looks good.

---

PR #630, opened 2026-06-04
fix(chat): block sending while WebSocket is disconnected (#448)

### LINKED ISSUE #448: don't allow sending  a message if WebSocket is disconnected
If you start a conversation, WebSocket disconnects, then try to send a message, then it hangs, but should not have let you send the message when disconnected.

garland3:
Closes #448.

## Problem
Starting a conversation, letting the WebSocket disconnect, then trying to send a message would add the user's message bubble and show the "Thinking..." spinner — but the message never reached the backend, so the UI hung forever. The user should not have been allowed to send while disconnected.

## Root cause
`handleSubmit` and the send button already guarded on `isConnected`, but two gaps remained:
1. Follow-up suggestion buttons called `sendChatMessage()` directly with no connection guard.
2. `sendChatMessage` added the user message and set `isThinking(true)` *before* calling `sendMessage()`, which silently no-op'd when the socket was closed → permanent "Thinking…" with no reply.

## Fix
- `WSContext.sendMessage` now returns `true`/`false` based on whether the socket was actually open.
- `ChatContext.sendChatMessage` guards on `isConnected` up front **and** on the send result (catching a socket that drops mid-call), and only mutates the UI after the message is on the wire. This makes the central funnel safe for every caller, not just the form submit. Returns a boolean.
- A toast notifies the user when a send is rejected, and a disconnected banner above the input explains why sending is blocked.
- `handleSubmit` keeps the user's typed text if the send is rejected, so nothing is lost.

## Testing
- New regression test `ws-disconnect-blocks-send.test.jsx` (banner + disabled send when offline, Enter does not send when offline, normal send when online).
- Full frontend suite: 429 tests passing.
- Lint clean on changed files.

### DIFF
--- a/frontend/src/components/ChatArea.jsx
+++ b/frontend/src/components/ChatArea.jsx
@@ -1,7 +1,7 @@
 import { useState, useRef, useEffect, useCallback } from 'react'
 import { useChat } from '../contexts/ChatContext'
 import { useWS } from '../contexts/WSContext'
-import { Send, Paperclip, X, Square, FileText, FileSearch, FileX, Search, Image, Wrench } from 'lucide-react'
+import { Send, Paperclip, X, Square, FileText, FileSearch, FileX, Search, Image, Wrench, WifiOff } from 'lucide-react'
 import Message from './Message'
 import WelcomeScreen from './WelcomeScreen'
 import encodeFileKeyPath from '../utils/encodeFileKeyPath'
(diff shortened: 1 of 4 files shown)


garland3:
## Summary
This PR fixes a chat hang that occurred when the WebSocket dropped: a user
could send a message while disconnected, see their bubble and a "Thinking…"
spinner, but the message never reached the backend. `WSContext.sendMessage`
now returns whether the socket was actually open, `ChatContext.sendChatMessage`
guards on connection state up front and on the send result (returning a boolean
and only mutating the UI after the message is on the wire), and `ChatArea`
surfaces a disconnected banner plus a toast while preserving the user's typed
text on a rejected send. Touches three components/contexts plus a new
regression test (+160 / -20 across 4 files).

## Verdict
Looks good.

---

PR #645, opened 2026-06-10
test: make suite hermetic against developer .env / MCP_TOKEN_ENCRYPTION_KEY

garland3:
## Problem

`AppSettings` is configured to load `../.env` (`atlas/modules/config/settings.py`). During tests that file is the *developer's* local `.env`, and pydantic-settings honored it. So `test_token_storage.py::TestRequiresEncryptionKey::test_refuses_to_start_without_encryption_key` — which deletes `MCP_TOKEN_ENCRYPTION_KEY` from the environment and expects construction to raise — read the key straight back from `.env`. The result: the test **passed in CI** (no `.env`) but **failed for any developer with `MCP_TOKEN_ENCRYPTION_KEY` in their `.env`**. Test outcomes depended on per-machine `.env` state.

## Fix

Disable env-file loading for the entire test session in `atlas/tests/conftest.py`:

```python
AppSettings.model_config["env_file"] = None
```

`AppSettings` now reads only the process environment, which tests fully control via `monkeypatch` / `os.environ`. This is a **test-isolation guard only** — runtime behavior is unchanged; the app still loads `.env` normally in production.

## Tests

New `atlas/tests/test_env_isolation.py`:
- asserts the conftest guard stays active (`env_file is None`);
- reproduces the leak end to end — writes a sentinel `.env` at the `../.env` location, chdirs, clears the env var, and asserts the on-disk value does not leak into `AppSettings`. Verified this test fails if the guard is removed.

## Verification

- `test_token_storage.py` + `test_env_isolation.py`: 36 passed, with and without a populated `.env` present.
- Full backend suite: same 147 pre-existing failures (unrelated telemetry-route 503s) on this branch as on baseline `main` — this change introduces **zero new failures** and adds 2 passing tests.

### DIFF
--- a/atlas/tests/conftest.py
+++ b/atlas/tests/conftest.py
@@ -30,6 +30,24 @@ if str(project_root) not in sys.path:
 _TELEMETRY_TMPDIR = tempfile.mkdtemp(prefix="atlas-test-telemetry-")
 os.environ.setdefault("APP_LOG_DIR", _TELEMETRY_TMPDIR)
 
+# --- .env isolation ------------------------------------------------------
+# ``AppSettings`` is configured to load ``../.env`` (see
+# ``atlas/modules/config/settings.py``). During tests that file is the
+# *developer's* local .env, whose contents must never influence the suite —
+# otherwise results depend on each contributor's machine. The concrete bug
+# this guards against: a developer with ``MCP_TOKEN_ENCRYPTION_KEY`` set in
+# their .env made ``test_refuses_to_start_without_encryption_key`` (which
+# deletes the env var, then expects construction to fail) pass locally for
+# CI but fail on their machine, because pydantic-settings re-read the key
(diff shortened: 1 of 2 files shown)


garland3:
## Summary
This PR makes the backend test suite hermetic against a developer's local `.env` file. `AppSettings` is configured to load `../.env`, which let pydantic-settings read real values (notably `MCP_TOKEN_ENCRYPTION_KEY`) back into tests that expected those variables to be absent — causing outcomes to depend on per-machine `.env` state. The fix disables env-file loading for the whole test session in `atlas/tests/conftest.py` (`AppSettings.model_config["env_file"] = None`) so settings come only from the process environment that tests control. A new `atlas/tests/test_env_isolation.py` asserts the guard stays active and reproduces the leak end to end. Runtime behavior is unchanged; production still loads `.env` normally.

## Verdict
Looks good.

---

PR #648, opened 2026-06-12
Direct PDF to Model Support

kevin-pedretti:
## Summary

Adds a `supports_pdf` model capability so uploaded PDFs are sent to the LLM as **inline base64 document content blocks** instead of being text-extracted into the files manifest. This lets PDF-capable models read the document directly. It mirrors the existing `supports_vision` flow across the same five layers.

## Configuration

```yaml
models:
  claude-sonnet-bedrock:
    model_url: "bedrock/anthropic.claude-sonnet-4-..."
    model_name: "claude-sonnet"
    supports_pdf: true
```

Defaults to `false`. Exposed on `/api/config` and `/api/config/shell` per model.

## How it works

- **Config** (`modules/config/models.py`): `ModelConfig.supports_pdf`.
- **API** (`routes/config_routes.py`): `supports_pdf` added to both config endpoints.
- **File processing** (`application/chat/utilities/file_processor.py`): stores raw `pdf_b64`/`pdf_mime_type` on eligible refs, clears stale PDF data each turn, and **skips text extraction** for natively-sent PDFs to avoid duplicate tokens.
- **Message building** (`preprocessors/message_builder.py`): attaches PDFs as `file` content blocks on the last user message and excludes them from the manifest. Block order is documents, then text, then images.
- **Orchestrator** (`orchestrator.py`): `_model_supports_pdf()` threads the flag into the file processor and message builder.

The content block uses the LiteLLM OpenAI-spec `file` format, which LiteLLM maps to each provider's native document format (a Bedrock Converse `document` block for Claude):

```json
{"type": "file", "file": {"file_data": "data:application/pdf;base64,...", "format": "application/pdf"}}
```

## Limits

| Limit | Value |
|-------|-------|
| Max PDF size | 20 MB base64 (≈15 MB raw) |
| Max pages per PDF | 100 |
| Max PDFs per request | 5 |

PDFs over any limit fall back to the text-extraction / manifest path with a user warning. The binding ceiling is Bedrock's hard 20 MB total request payload (not a per-document cap, which no longer applies to PDFs on Claude 4+).

## Provider note: Bedrock citations

On Bedrock's Converse API, full **visual** PDF understanding (charts, scanned pages, layout) requires citations to be enabled; without it Converse falls back to basic text extraction. The portable LiteLLM `file` block does not currently expose a citations toggle, so on Bedrock this delivers text-extraction-mode document reading. Emitting a provider-native document block with `citations: {enabled: true}` (or using InvokeModel) is a documented follow-up.

[...]

PR #651, opened 2026-06-13
Direct PDF to Model Support

garland3:
Sandia-hosted branch for the **Direct PDF to Model Support** work originally proposed by @kevin-pedretti in #648. Opened here so CI runs with full repository permissions (the contributor fork PR cannot push the org Docker image or post the security-summary comment). Original authorship is preserved in the commit history.

## Summary

Adds a `supports_pdf` model capability so uploaded PDFs are sent to the LLM as inline base64 **document content blocks** (LiteLLM `file` blocks, mapped to a Bedrock Converse `document` block for Claude) instead of being text-extracted into the files manifest, mirroring the existing `supports_vision` plumbing across config, config routes, file processor, message builder, and orchestrator.

## Review follow-ups addressed on top of the original commit

- **Content preservation on demotion & follow-up turns:** native PDFs now extract text up front as a *durable fallback*. It is excluded from the manifest on the turn the PDF is sent natively (no token duplication) but survives when the inline document block is later dropped — fixing (a) follow-up turns losing PDF access after `pdf_b64` is cleared and (b) count-demoted PDFs (#6+) collapsing to name-only entries.
- **Aggregate payload guard:** new 18 MB aggregate inline-payload budget (PDFs + vision images) keeps several mid-sized PDFs from exceeding Bedrock's ~20 MB total request limit; over-budget PDFs fall back to extracted text with a user warning.
- **Cheaper, non-blocking guards:** per-document size check runs before the `pypdf` page count, and page counting runs off the event loop via `asyncio.to_thread`.
- **Honest warnings** when extraction is unavailable ("a file reference (no extractable text)").
- **Docs / CHANGELOG / PR-validation script** reconciled with the actual fallback behavior.

## Limits

| Limit | Value |
|-------|-------|
| Max PDF size (per document) | 20 MB base64 (≈15 MB raw) |
| Max pages per PDF | 100 |
| Max PDFs per request | 5 |
| Max aggregate inline payload | 18 MB base64 |

## Testing

- `atlas/tests/test_pdf_document_support.py` — storage, limits, manifest exclusion, message building, **demotion preserves extracted text**, **aggregate-payload demotion**, **follow-up-turn fallback**, stale cleanup.
- Full backend suite green locally: 1571 passed, 17 skipped.

Closes #648 once merged here.


garland3:
## ✅ End-to-end verification (real Anthropic API)

Ran the merged branch live through the chat UI against the **real Anthropic API** (`claude-sonnet-4`, configured with `supports_pdf: true`).

**Test design (un-guessable facts):** a single-page PDF was generated containing facts the model cannot know a priori — a secret code `MAGENTA-FALCON-4417`, the name *Dr. Penelope Quist*, and the codeword *PORCELAIN OTTER*. It was uploaded with `extractMode: none`, so the document was sent **natively as a `file` content block — no text extraction**.

**Backend confirmation (native path, not extraction):**
```
message_builder - "Attached 0 image(s) and 1 PDF(s) to user message for model"
```

**Result:** the model quoted all three PDF-only facts back exactly — proof it read the uploaded PDF directly.

Native PDF E2E: claude-sonnet-4 answering from an uploaded PDF

**Automated tests:** full backend suite green locally (1571 passed, 17 skipped), including new coverage for excess-PDF demotion preserving extracted text, the aggregate-payload guard, and the follow-up-turn text fallback.

_E2E path: Playwright-driven Chrome → `localhost:8020` (worktree build of this branch) → select `claude-sonnet-4` → upload PDF → ask → streamed answer. Screenshot also added to the feature doc (`docs/developer/pdf-document-support-2026-06-12.md`)._

Issue #622, opened 2026-05-31
Opt-in chat capture + rollback for fine-tuning data


garland3:
Add an **opt-in** mode (off by default, also gated by a system feature flag) where Atlas captures full LLM I/O — prompts, completions, tool calls, tool results, available-tool list — for users who voluntarily turn it on. Goal: generate training data to fine-tune a small customized LLM.

The high-value piece is a **rollback-with-forced-tool** flow: when the model picks the wrong tool, the user clicks ↶ on that turn, selects the correct tool from the available list, and Atlas replays the turn with \`selected_tools=[X], tool_choice_required=True\` (both fields already exist on \`ChatRequest\`). Both the wrong and corrected versions are saved as a DPO-style \`(rejected, chosen)\` pair. Secondary signals: editing a tool result before the model sees it, and a one-tap end-of-conversation rating. Silence is **not** treated as success.

Full design sketch: https://gist.github.com/garland3/8e57640820b128ef1a3509af5f450b2b

PR #652, opened 2026-06-14
feat(chat): rewind and edit a previous prompt (#142)

### LINKED ISSUE #142: allow editing of the last previous message and resubmit

garland3:
## Summary

Implements **issue #142** — go back to any earlier prompt in a conversation, optionally edit it, and resubmit from that point. The targeted prompt and everything after it are dropped, and the (edited) prompt re-runs as a fresh turn, keeping a single linear thread (**overwrite-in-place**).

Submitting an unchanged prompt simply regenerates from that point, so this also covers "regenerate from here" for free.

### UX
- A subtle pencil button appears on your own messages on hover, matching the existing copy button (hidden until hover).
- Clicking it opens an inline editor prefilled with the message. **Send** resubmits; **Cancel** / **Esc** backs out; **Enter** sends, **Shift+Enter** newlines.
- Rewind is blocked while a response is still streaming.

### Live demo (real LLM, `groq-gpt-oss-120b`)

**1. A real answer, with the edit affordance on hover**

affordance

**2. Edit the prompt in place** (changed to ask about *additive* primary colors; the old answer is still shown below, about to be replaced)

editor

**3. Resubmitted — fresh answer replaces the old turn**

after

## How it works

Messages are addressed by **user-message ordinal** (0-based position among `user` messages), not absolute transcript position — the frontend renders extra system/tool rows with no backend counterpart, so counting user messages is the one scheme both sides agree on.

**Backend**
- `ConversationHistory.truncate_at_user_index(n)` drops the Nth user message and everything after it, **returning the removed turns**.
- `ChatOrchestrator.execute(..., rewind_to_user_index=...)` applies the truncation *before* appending the new prompt. Out-of-range/negative index is a safe no-op.
- New `ChatRequest.rewind_to_user_index`; `main.py` reads it off the `chat` WebSocket message. The existing full-upsert persistence rewrites the saved conversation to match the truncated thread automatically.

**Frontend**
- `Message.jsx` renders the pencil + inline editor and calls `onRewind(userIndex, newContent)`.
- `ChatArea.jsx` computes each user message's ordinal while mapping the transcript.
- `ChatContext.jsx` `rewindAndResubmit` truncates the local transcript and adds `rewind_to_user_index` to the `chat` payload.

## Scope decision

Scoped to **user prompts** with **overwrite-in-place** (no branch tree). This is the 80% of the value at a fraction of the cost and is forward-compatible with conversation branching later. See `docs/rewind-edit/README.md`.

## Relationship to #622

The same rewind primitive is what #622 (opt-in capture + forced-tool DPO replay) builds on: rewind to a turn, replay with `selected_tools=[X]` + `tool_choice_required=True`. `truncate_at_user_index` returns the removed ("rejected") trajectory specifically so a future capture layer can record it with no further plumbing. Capture/consent itself is out of scope here.

## Test plan
- [x] `atlas/tests/test_conversation_history_rewind.py` — truncation by user ordinal (middle, first, out-of-range, negative, ignores tool/system rows)
- [x] `atlas/tests/test_orchestrator_rewind.py` — orchestrator truncates-then-appends; no-rewind and out-of-range paths
- [x] `frontend/src/test/message-rewind-edit.test.jsx` — affordance visibility, editor prefill, Send/Cancel/Esc, empty-edit guard (8 tests)
- [x] `test/pr-validation/test_pr652_rewind_edit_prompt.sh` — end-to-end multi-turn rewind through the real orchestrator + front-to-back wiring checks (10/10 pass)
- [x] Verified in a live browser with a real LLM: generate -> edit -> regenerate, old turn replaced (see screenshots)
- [x] `npm run lint` clean (no new warnings); frontend build clean; full frontend suite green (442 tests)

Closes #142.


garland3:
## Summary
This PR adds an inline "rewind and resubmit" edit flow to the chat UI: a user can edit an earlier prompt, which truncates the conversation at that turn (locally and in saved server history) and resends the edited message. It wires a new `rewind_to_user_index` field from the frontend through the WebSocket handler and chat DTO into the orchestrator, where `ConversationHistory.truncate_at_user_index` drops the targeted user turn and everything after it. The change spans ~16 files across frontend React components and backend domain/orchestrator/transport layers and ships with tests, screenshots, and docs.

## Top issues
1. **Local transcript truncated before the send is validated** — frontend/src/contexts/ChatContext.jsx:427. `rewindAndResubmit` deletes the targeted prompt and everything after it from the on-screen transcript *before* calling `sendChatMessage`, which can bail out (no model, disconnected socket, or a drop between check and send) and return `false` without sending or restoring anything. The conversation tail is destroyed in the UI with nothing on the wire; for server-saved conversations the backend keeps its full history, producing a frontend/backend desync that misaddresses the next rewind. (high)
2. **The new `rewindAndResubmit` logic has no test** — frontend/src/contexts/ChatContext.jsx:412. The only non-trivial new frontend logic is untested: the component suite mocks `ChatContext`, so the real callback never runs. Both the streaming guard (a headline behavior) and the local-transcript truncation/ordinal counting — the part most likely to drift from the backend — are completely uncovered. (high)
3. **User-message ordinal contract is reimplemented in three places** — frontend/src/components/ChatArea.jsx:781, frontend/src/contexts/ChatContext.jsx:427, atlas/domain/messages/models.py:130. The "find the Nth user message, ignore other rows" algorithm is hand-coded independently in all three spots. Any future change to what counts as a renderable user row must be applied identically in each, or the wrong prompt is silently dropped — a corruption no single-layer test or type check will catch. (medium)
4. **Edited text is silently discarded when blocked mid-stream** — frontend/src/components/Message.jsx:92. `submitEdit` calls `setIsEditing(false)` before `onRewind`, so editing a prior prompt and clicking Send while a reply is streaming closes the editor and throws the edited text away — the user sees only a transient toast. Gate the close on the boolean `onRewind` returns, or disable Send while a response is in flight. (medium)
5. **Edit affordance is invisible to keyboard focus and unreachable on touch** — frontend/src/components/Message.jsx:553. The pencil button uses `opacity-0 group-hover:opacity-100`, so a Tab-focused control is fully transparent (focus ring on an invisible element) and touch users have no hover to reveal it. As the feature's only entry point, the new functionality is hard to reach without a mouse. Reveal on focus (`focus-within`/`group-focus-within`) and keep it discoverable on small viewports. (medium)

## Evidence
Evidence: 9/10. For a large, user-facing change (~16 files, +808/-18 across frontend and backend) the demonstration is broad and proportionate: end-to-end screenshots walk the primary flow through the affordance, the prefilled editor, and the resubmitted result against a live server and model; automated tests cover every layer (domain truncation incl. first/middle/out-of-range/negative and tool/system-row cases, orchestrator routing, and the frontend affordance); an integration harness asserts the field is wired front-to-back; and lint/build/full frontend suite are reported clean. The one notable gap is the "blocked while streaming" guard — it is implemented and described but neither screenshotted nor tested, since the component test renders `Message` in isolation and never reaches the `ChatContext` guard. Add a unit test around `rewindAndResubmit` for the streaming-blocked path (or a screenshot of the blocked toast).

## Detail

### Security
- **Unvalidated `rewind_to_user_index` reaches domain logic untyped** — atlas/main.py:542. The value is taken straight from the inbound WebSocket frame and passed through the orchestrator into `truncate_at_user_index` with no coercion; the `Optional[int]` annotations are not enforced at runtime. A non-numeric JSON value (`"0"`, `[]`, `{}`) is not `None`, so the rewind branch runs and the `if user_index < 0:` comparison raises an unhandled `TypeError`, aborting the chat turn; `true` coerces to `1` and floats silently no-op. Impact is bounded to the caller's own authenticated session (no IDOR) — robustness/DoS-of-self. Coerce at the trust boundary (`isinstance(idx, int) and not isinstance(idx, bool)`) or guard the domain method. (low)

### Correctness
- **Rewind correctness depends on an unenforced FE/BE user-message-count invariant** — frontend/src/components/ChatArea.jsx:787. The scheme assumes the Nth rendered `role === 'user'` row maps to the Nth backend `MessageRole.USER` entry, but that is never asserted or reconciled. If the two ever differ (suppressed/optimistic user rows, future synthetic turns), the ordinal silently truncates at the wrong point with no error surfaced, since there is no echo-back to verify against. Have the backend return the post-truncation user-message count or include a consistency token (e.g. hash of the targeted message) for `truncate_at_user_index` to verify; at minimum add a test pinning the 1:1 assumption. (low)
- **A no-op rewind is logged as routine, not flagged** — atlas/application/chat/orchestrator.py:176. When `rewind_to_user_index` addresses no existing user message, `truncate_at_user_index` returns `[]` and the orchestrator logs "removed 0 message(s)" at info, then appends normally — exactly the signature of a FE/BE ordinal desync, but invisible on dashboards. Emit a `logger.warning` when `rewind_to_user_index is not None and len(removed) == 0`. (low)

### Architecture
- **Inline editor re-implements composer keyboard/textarea behavior** — frontend/src/components/Message.jsx:99. The new editor hand-rolls Enter-to-submit / Shift+Enter-newline plus auto-growing rows and styling that duplicate the main composer, creating two independent definitions of prompt-textarea behavior that can drift (IME guards, submit chord, max-rows, paste). It also adds an editing state machine onto an already-overloaded `Message.jsx`. Factor the textarea behavior into a shared hook/component used by both, or at least extract a colocated `MessageEditor`. (low)

### Tests & observability
- **Streaming guard on the edit affordance (`canRewind`) is untested** — frontend/src/components/Message.jsx:67. No test renders a user message with `_streaming: true` to confirm the pencil is hidden, so dropping `!message._streaming` would pass the suite while re-enabling mid-stream edits. Add a case asserting the affordance is absent when `_streaming` is true. (medium)
- **Keyboard submit/newline paths only half-covered** — frontend/src/components/Message.jsx:101. Only the empty-Enter guard is asserted; nothing covers non-empty Enter calling `onRewind` or Shift+Enter *not* submitting. Add both assertions. (low)
- **No test pins the returned "removed trajectory" contract relied on by #622** — atlas/tests/test_orchestrator_rewind.py. The removed-turns return value is a deliberate forward-compatibility commitment for the capture/replay layer, but nothing asserts the orchestrator preserves it end to end. Add an orchestrator-level assertion or a comment-backed test making the contract explicit. (low)

### UX & docs
- **Focus is lost after Cancel/Esc/Send closes the editor** — frontend/src/components/Message.jsx:87. Opening the editor focuses the textarea, but closing it returns focus to `<body>` rather than the trigger, dropping keyboard/screen-reader users at the top of the document. Restore focus to the pencil button on close. (medium)
- **No confirmation before an edit drops the rest of the conversation** — frontend/src/components/Message.jsx:584. Submitting permanently removes the targeted prompt and everything after it with no undo; the only warning is an easy-to-miss helper line. Add a lightweight confirm for multi-turn rewinds, or make the dropped count explicit in the helper text. (low)
- **Editor height counts newlines only, ignoring wrapped lines** — frontend/src/components/Message.jsx:579. `rows` is sized by `\n` count, so a long single-line prompt renders at 2 rows. Estimate wrapped lines or autosize to `scrollHeight` (capped at 10). (low)
- **No in-UI hint for the editor's keyboard shortcuts** — frontend/src/components/Message.jsx:99. Enter/Shift+Enter/Esc behavior is undocumented in the UI; add an optional helper hint to match the affordance's discoverability bar. (info)

## Verdict
Request changes.

---

garland3:
## Summary
This PR adds an inline "edit & resubmit" affordance to chat prompts: a pencil button on user messages opens an inline editor, and resubmitting rewinds the conversation to that turn — truncating the targeted prompt and everything after it on both the frontend transcript and the backend `ConversationHistory`. The frontend introduces a shared `userMessageOrdinal` util, a `rewindAndResubmit` callback in `ChatContext`, and editor UI in `Message.jsx`; the backend adds a `rewind_to_user_index` field to `ChatRequest`, orchestrator coercion/routing, and a `truncate_at_user_index` domain method. The change spans 18 files (+1148/-21) across the frontend, transport, application, and domain layers.

## Top issues
1. **`rewindAndResubmit` callback is entirely untested** — frontend/src/contexts/ChatContext.jsx:425 — The only non-trivial new frontend logic has no coverage; the component suite mocks `ChatContext`, so the real streaming guard and empty-content guard never run and could be dropped with the suite staying green.
2. **Local-transcript truncation ordering invariant is untested** — frontend/src/contexts/ChatContext.jsx:401 — The "truncate only after the send is confirmed on the wire" logic (the guard against destroying the conversation tail on a failed/disconnected send) is unasserted; a refactor reintroducing the data-loss bug would pass every test.
3. **Edit affordance is unreachable on touch devices** — frontend/src/components/Message.jsx:569 — The pencil is revealed only via `group-hover`, and it is the sole entry point to the feature; mobile/tablet users cannot discover or reach rewind/edit at all.
4. **Destructive rewind gives no count of what will be dropped and no undo** — frontend/src/components/Message.jsx:600 — Resubmitting permanently removes the targeted prompt and every later turn (including persisted history) with only a static "replaces the messages below" line — no count, no confirmation, no undo.
5. **Inline editor re-implements composer textarea behavior** — frontend/src/components/Message.jsx:99,579 — Enter-to-submit and autosize are hand-rolled instead of shared with the main composer, and have already drifted (composer autosizes off `scrollHeight`; editor counts `\n` only), forcing dual maintenance of "how a prompt textarea behaves."

## Evidence
Evidence: 9/10. For a large, user-facing change the demonstration is broad and proportionate: end-to-end screenshots of the primary flow against a live server and real model (hover affordance, prefilled editor, resubmitted answer replacing the old turn); layered automated tests covering domain truncation across first/middle/out-of-range/negative ordinals and interleaved tool/system rows, orchestrator routing including a parametrized non-integer/bool/float index hardening case, the shared ordinal contract in isolation, and 13 affordance-behavior cases; an integration harness driving a multi-turn conversation through the real orchestrator; plus a clear description and a `docs/rewind-edit/README.md` walkthrough with documented limitations. The notable gaps are the two undemonstrated paths that also surface as the top test findings: the streaming-blocked guard (described as a headline guarantee but never directly exercised, since the context is mocked) and the failure/disconnected-send path that is supposed to leave the transcript intact. The author should add unit coverage around `rewindAndResubmit` asserting it returns `false` and toasts while streaming without truncating, and that truncation happens only after a confirmed send.

## Detail
### Security
- **Destructive truncation guarded only at the orchestrator layer** — atlas/domain/messages/models.py:114 (low) — `truncate_at_user_index` is an irreversible mutation whose only runtime guard is `_coerce_user_index` in the orchestrator; the DTO field and method parameter are untyped at runtime, so any future caller reaching the method without coercing could let a `bool` index target ordinal 1 or raise `TypeError`. Impact is confined to the caller's own authenticated session (no cross-session/IDOR exposure). Suggest making the method self-defending (return `[]` unless given a genuine non-bool `int`) as defense-in-depth behind the existing coercion.
- **Untyped wire field validated deep in the stack** — atlas/main.py:542 (info) — `rewind_to_user_index=data.get("rewind_to_user_index")` forwards the raw JSON value into `ChatRequest` with no edge validation; safe today only because the orchestrator coerces. Centralizing coercion at the WebSocket handler or a `ChatRequest` factory would make the trust boundary explicit.

### Correctness
- **In-range FE/BE ordinal desync silently truncates the wrong prompt** — atlas/application/chat/orchestrator.py:206 (low) — The contract relies on the frontend rewindable-user ordinal matching the backend `MessageRole.USER` count. Only out-of-range indices warn; an in-range index that drifted to address a *different* user turn truncates successfully and silently. The invariant holds today (agent follow-ups never enter `session.history`) but nothing enforces it. Suggest carrying a cheap consistency token (hash/prefix of the targeted prompt, or expected pre-truncation count) verified before cutting, and at minimum a test pinning the 1:1 FE/BE user-count assumption.
- **Non-integer rewind index degrades to append-as-duplicate, not a no-op** — atlas/application/chat/orchestrator.py:200 (low) — `_coerce_user_index` maps any non-`int` value to `None`, after which the edited content is appended as a new trailing turn while the original remains — a silent duplicate, not the intended overwrite. Defensive-only today (the sender always emits an integer). If stricter behavior is wanted, treat a present-but-uncoercible index as an error rather than silently appending.
- **Out-of-range rewind gives the caller no signal** — atlas/application/chat/orchestrator.py:189 (low) — The orchestrator logs a WARNING but `main.py` receives no indication the rewind was ignored, so the client cannot surface the ordinal mismatch. Suggest returning a status/domain error the WebSocket handler can relay as a non-fatal notification.

### Architecture
- **Editing state machine accretes onto an already-large `Message` component** — frontend/src/components/Message.jsx:66-127 (low) — `Message.jsx` (~629 lines) already mixes rendering, copy, citation, streaming, and tool display; this PR folds a full editing state machine (state, refs, focus `useEffect`, start/cancel/submit/keydown handlers) inline. Suggest extracting a colocated `MessageEditor` component, which also yields a direct unit-test seam.
- **User-message-ordinal scheme duplicated across the FE/BE boundary on an unstated invariant** — frontend/src/utils/userMessageOrdinal.js:33 ↔ atlas/domain/messages/models.py:130 (low) — The frontend excludes agent-loop rows (`isRewindableUserMessage`) while the backend counts every `MessageRole.USER` row; they agree only because agent-input rows are never persisted — an implicit, untested, load-bearing invariant. Suggest echoing back the post-truncation count (or a hash) for reconciliation, and/or a single cross-boundary test, plus cross-referencing comments in both files.
- **Inline IIFE still recomputes user ordinals in `ChatArea`** — frontend/src/components/ChatArea.jsx:787 (medium) — After the welcome extraction of `userMessageOrdinal`, `ChatArea` still contains an inline ordinal-assignment loop, so any future rule change (e.g. a new `_agentInput` flag) must be applied in two places. Suggest moving the assignment into the shared utility (e.g. a `mapWithUserIndices` helper) so render and truncation consume one implementation.

### Tests & observability
- **Orchestrator desync/no-op warnings emitted but never asserted** — atlas/tests/test_orchestrator_rewind.py (medium) — The two `logger.warning` lines (non-integer index, out-of-range "possible FE/BE desync") are the only operator signal that the ordinal contract drifted, yet the rewind tests assert only resulting history state; the level could be downgraded to `info` with no failure. Suggest `caplog` assertions pinning a WARNING on both paths and asserting the routine truncation path does *not* warn.
- **End-to-end script validates frontend wiring by grep, not behavior** — test/pr-validation/test_pr652_rewind_edit_prompt.sh:148 (low) — The "front-to-back" sections verify the frontend half only by grepping for source strings plus the isolated component test; no layer drives `rewindAndResubmit` → `sendChatMessage` → truncation. Suggest running the `ChatContext` tests (once they exist) from the script, or softening the "front-to-back" wording.
- **No test spans the streaming-blocked producer/consumer contract** — frontend/src/components/Message.jsx:92 ↔ ChatContext.jsx:434 (low) — `submitEdit` keeps the editor open when `onRewind` returns `false`, and the component test stubs that return value, but nothing asserts the `ChatContext` guard is what actually returns `false` mid-stream. Suggest asserting `rewindAndResubmit` returns the literal `false` while streaming so the `accepted !== false` check stays valid.

### Performance
- **Editor height recomputes by allocating a split array on every keystroke** — frontend/src/components/Message.jsx:579 (low) — `editValue.split('\n').length` runs on every render and allocates a throwaway substring array proportional to the (unbounded, user-pasteable) input on the render hot path. Suggest counting newlines without materializing an array, or memoizing the row count with `useMemo` keyed on `editValue`.

### UX & docs
- **Copy button no longer matches the pencil's focus-reveal** — frontend/src/components/Message.jsx:579 (medium) — The pencil gains `focus:opacity-100 focus-visible:opacity-100` but the adjacent copy button (now a sibling in the same flex container) still reveals only on `group-hover`, so a keyboard user sees a focus ring on an invisible copy control. Suggest adding the same focus-reveal classes to the copy button.
- **Inline editor height ignores wrapped lines** — frontend/src/components/Message.jsx:595 (low) — Sizing by `\n` count renders a long single-paragraph prompt at just 2 rows, forcing scroll-in-a-tiny-box exactly when reviewing the text to resubmit matters most. Suggest autosizing to `scrollHeight` (capped ~10 rows) on mount and input.
- **Focus drops to `<body>` after a successful resubmit** — frontend/src/components/Message.jsx:108 (low) — `restoreFocusRef` is set only in `cancelEdit`, so on an accepted Send the trigger unmounts and focus falls back to `<body>`, dropping keyboard/screen-reader users to the top of the document mid-conversation. Suggest moving focus to a stable target (the streaming response region or the composer input).

## Verdict
Request changes.

---

garland3:
## Summary
Adds an inline "edit and resubmit an earlier prompt" (rewind) flow: a per-message editor on user turns, an ordinal scheme shared across the render and truncation paths, and a `rewind_to_user_index` field threaded from the WebSocket handler through the orchestrator to domain-level history truncation.

## Verdict
Request changes.

## Required changes
1. Add a `rewindAndResubmit` unit test at `frontend/src/contexts/ChatContext.jsx:430` — the existing component suite mocks `ChatContext`, so the real callback never runs. Assert it returns `false` and toasts (without calling `sendMessage`) while `isThinking`/`isSynthesizing`/`isStreaming` is true, returns `false` on empty/whitespace content, and forwards `{ rewindToUserIndex }` on the happy path.
2. Add a send-then-truncate ordering test at `frontend/src/contexts/ChatContext.jsx:394` — when `sendMessage` returns falsy (dropped socket) with a rewind index set, assert the transcript is left intact and the function returns `false`; pair with a success case asserting truncation happens only after `sent` is truthy. This pins the data-loss guard a future refactor is most likely to break.
3. Only close the editor on a successful rewind at `frontend/src/components/Message.jsx:92` — `submitEdit` calls `setIsEditing(false)` before/regardless of `onRewind`, so when the streaming guard returns `false` the user's typed text is discarded into a transient toast. Gate the close on the boolean `onRewind` returns.
4. Make the edit affordance reachable on touch/pointer-only devices at `frontend/src/components/Message.jsx:577` — the pencil only reveals via `group-hover`, so mobile/tablet users cannot discover or invoke the feature's sole entry point. Render it persistently (or behind a visible overflow menu) under `@media (hover: none)`.
5. Restore focus after an accepted Send at `frontend/src/components/Message.jsx:97` — `restoreFocusRef` is set only in `cancelEdit`, so on a successful resubmit focus falls back to `<body>`. Move focus to a stable target (composer or response region) on success.
6. Add a test pinning `_agentInput: true` onto agent-loop answers at `frontend/src/contexts/ChatContext.jsx:561` — only the consumer half (`isRewindableUserMessage` skipping the flag) is tested; if `answerAgentQuestion` stopped setting it, agent answers would re-enter the ordinal count and rewind would silently truncate the wrong turn with every test still green.
7. Memoize the ordinal wrapping at `frontend/src/components/ChatArea.jsx:787` — `withUserOrdinals(messages)` runs two full passes plus N wrapper allocations on every render, including each streamed token. Wrap in `useMemo(() => withUserOrdinals(messages), [messages])`.

## Evidence
Evidence: 9/10. For a large, multi-layer change the demonstration is broad and proportionate — e2e screenshots of the happy path against a live server, layered domain/orchestrator/ordinal tests, and an integration harness. The one gap is the headline "blocked while streaming" guard and the "truncate only after confirmed send" data-loss guard: both live in `rewindAndResubmit`/`sendChatMessage` and are neither screenshotted nor directly tested (the component suite mocks the context). Required changes 1–2 close it.

## Notes
- Cross-boundary ordinal contract relies on an unstated invariant (FE excludes `_agentInput` rows; BE counts every `MessageRole.USER`); consider an echo-back consistency token or a cross-boundary test — `atlas/application/chat/orchestrator.py:206`.
- An out-of-range or drifted rewind index silently appends/truncates with no user-visible signal — `atlas/application/chat/orchestrator.py:189`.
- Consider extracting a `MessageEditor` component to keep editing state out of the ~640-line `Message.jsx:66`.

---

garland3:
## Summary
Adds edit-and-resubmit ("rewind") for user prompts: an inline editor on each user message that, on submit, truncates the conversation from that prompt onward (in the UI and in server history via a new user-ordinal index) and re-runs the turn. Spans the React UI, WebSocket transport, orchestrator, and domain history model.

## Verdict
Request changes.

## Required changes
1. Restore focus on accepted submit at `frontend/src/components/Message.jsx:118` — `restoreFocusRef` is set only in `cancelEdit`, so a successful Send unmounts the editor and drops keyboard/screen-reader focus to `<body>` just as the new turn streams in. Set `restoreFocusRef.current = true` before closing (or move focus to the composer/response region).
2. Make the FE↔BE user-ordinal contract self-checking at `frontend/src/utils/userMessageOrdinal.js:22` ↔ `atlas/domain/messages/models.py:130` — the frontend counts rewindable rows (excluding `_agentInput`) while `truncate_at_user_index` counts every `USER` row; they agree only by an unenforced invariant, and a future `USER` row reaching `session.history` would silently truncate the wrong prompt. Carry a consistency token (targeted-prompt hash/prefix or expected user count) verified before truncation, and add one cross-boundary test pinning the 1:1 mapping over an interleaved transcript.
3. Surface the blast radius before a destructive rewind at `frontend/src/components/Message.jsx:611` — replace the static "replaces the messages below" helper with the concrete count (e.g. "removes this and N later messages") and add a lightweight confirm for multi-turn rewinds (skip it when only the last turn is affected).
4. Add a `ChatArea`-level render test for the ordinal→prop glue at `frontend/src/components/ChatArea.jsx:786` — `userIndex` and `onRewind` wiring is untested between two isolated suites; render a mixed transcript (system/user/assistant/tool/`_agentInput`/user) and assert only rewindable user rows expose the edit affordance and that the second real prompt calls `rewindAndResubmit` with ordinal `1`.

## Evidence
Evidence: 9/10. Broad, proportionate proof for a large change — three live end-to-end screenshots of the happy path, layered domain/orchestrator/component tests, a front-to-back integration script, and new context tests pinning the streaming-blocked and data-loss guards. The one gap is visual coverage of the non-happy-path states (blocked-while-streaming toast, touch/`hover:none` entry point), both currently shown only by tests/CSS.

## Notes
- Align the pencil's `aria-label` with its destructive `title` warning — `frontend/src/components/Message.jsx:581`.
- Rewind Send silently no-ops with the editor left open (no toast) when no model is selected — `frontend/src/contexts/ChatContext.jsx:333` / `Message.jsx:92`.
- Curb `Message.jsx` growth by extracting a `MessageEditor` and sharing the composer's autosize/submit-chord behavior via a hook — `frontend/src/components/Message.jsx:66`.

---

PR #463, opened 2026-03-22
feat: reasoning content streaming for reasoning LLMs

garland3:
## Summary

- Stream chain-of-thought reasoning tokens from reasoning-capable LLMs (o3, Qwen3, GPT-OSS via vLLM) to the frontend in real-time
- Collapsible reasoning UI section auto-expands during streaming with blinking cursor, collapses when complete
- Reasoning content persists across page reloads via message metadata
- Includes monkey-patch for [LiteLLM issue #20246](https://github.com/BerriAI/litellm/issues/20246) — vLLM `delta.reasoning` not mapped to `reasoning_content` in streaming mode (see `docs/developer/reasoning-content-streaming-2026-03-22.md` for removal instructions)

### Backend changes
- `ReasoningToken` dataclass for individual streaming reasoning chunks
- `ReasoningBlock` yielded from `stream_plain`/`stream_with_tools` with full accumulated reasoning
- All mode runners (plain, RAG, tools, agent) handle reasoning tokens and persist to metadata
- `streaming_helpers.py` error path now sends error events to frontend instead of failing silently

### Frontend changes
- `reasoning_token` / `reasoning_content` WebSocket event handlers with 30ms buffered flush
- `STREAM_REASONING_TOKEN` / `STREAM_REASONING_END` reducer actions in `useMessages`
- Collapsible reasoning section in `Message.jsx` with auto-scroll during streaming
- `ChatContext.jsx` persists `reasoning_content` in local save metadata

## Test plan
- [ ] Send a message using a reasoning model (e.g. o3, Qwen3, local vLLM GPT-OSS) and verify reasoning section appears and streams in real-time
- [ ] Verify reasoning section auto-collapses after streaming completes and can be toggled open/closed
- [ ] Verify reasoning persists after page reload
- [ ] Test tools mode: reasoning should display both before tool calls and during post-tool synthesis
- [ ] Test agent mode: reasoning should display and persist in message metadata
- [ ] Test with non-reasoning models: no reasoning section should appear
- [ ] Run `pytest atlas/tests/test_reasoning_content.py`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #655, opened 2026-06-14
feat: reasoning content streaming for reasoning LLMs (replaces #463)

garland3:
## Summary

Replaces #463 — rebuilt cleanly as a single commit on top of current `main` so the history is easy to review. The net diff is byte-identical to #463; all review feedback already addressed on that branch is carried forward, and the now-redundant `eslint-disable` in `ChatContext.jsx` (the effect dep was changed from `messages?.length` to `messages`) was removed.

Streams chain-of-thought reasoning tokens from reasoning-capable LLMs (o3, Qwen3, GPT-OSS via vLLM) to the frontend in real-time. A collapsible reasoning UI section auto-expands during streaming with a blinking cursor and collapses when complete. Reasoning content persists across page reloads via message metadata.

Includes a monkey-patch for [LiteLLM issue #20246](https://github.com/BerriAI/litellm/issues/20246) — vLLM `delta.reasoning` not mapped to `reasoning_content` in streaming mode (see `docs/developer/reasoning-content-streaming-2026-03-22.md` for removal instructions).

### Backend changes
- `ReasoningToken` dataclass for individual streaming reasoning chunks
- `ReasoningBlock` yielded from `stream_plain`/`stream_with_tools` with full accumulated reasoning
- All mode runners (plain, RAG, tools, agent) handle reasoning tokens and persist to metadata
- `streaming_helpers.py` error path now sends error events to the frontend instead of failing silently

### Frontend changes
- `reasoning_token` / `reasoning_content` WebSocket event handlers with 30ms buffered flush
- `STREAM_REASONING_TOKEN` / `STREAM_REASONING_END` reducer actions in `useMessages`
- Collapsible per-message reasoning section in `Message.jsx` with auto-scroll during streaming
- `ChatContext.jsx` persists `reasoning_content` in local save metadata

## Test plan
- [ ] Send a message using a reasoning model (e.g. o3, Qwen3, local vLLM GPT-OSS) and verify the reasoning section appears and streams in real-time
- [ ] Verify the reasoning section auto-collapses after streaming completes and can be toggled open/closed
- [ ] Verify reasoning persists after page reload
- [ ] Test tools mode: reasoning should display both before tool calls and during post-tool synthesis
- [ ] Test agent mode: reasoning should display and persist in message metadata
- [ ] Test with non-reasoning models: no reasoning section should appear
- [x] `pytest atlas/tests/test_reasoning_content.py atlas/tests/test_streaming_token_flow.py` — 30 passed
- [x] Frontend: `vitest run` reasoning suites — 27 passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### PR #655 - 2026-06-14
+- **Feature**: Real-time reasoning content streaming for reasoning-capable LLMs (o3, Qwen3, GPT-OSS via vLLM). Chain-of-thought reasoning tokens stream to the frontend live, displayed in a collapsible per-message section that auto-expands during streaming and persists in message metadata across page reloads. Includes a monkey-patch for LiteLLM issue [#20246](https://github.com/BerriAI/litellm/issues/20246) (vLLM `delta.reasoning` not mapped to `reasoning_content` in streaming mode); see `docs/developer/reasoning-content-streaming-2026-03-22.md` for removal instructions. Replaces PR #463.
+
 ### PR #654 - 2026-06-14
 - **UI theme**: Defaulted first-time users to dark mode and improved the auto-approve tools warning contrast in light mode.
 
--- a/atlas/application/chat/agent/agentic_loop.py
+++ b/atlas/application/chat/agent/agentic_loop.py
@@ -21,6 +21,7 @@ from typing import Any, Dict, List, Optional
 
 from atlas.interfaces.llm import LLMProtocol, LLMResponse
 from atlas.interfaces.tools import ToolManagerProtocol
+from atlas.modules.llm.models import ReasoningBlock, ReasoningToken
 from atlas.modules.prompts.prompt_provider import PromptProvider
 
 from ..utilities import error_handler, tool_executor
@@ -88,6 +89,7 @@ class AgenticLoop(AgentLoopProtocol):
 
         steps = 0
         final_answer: Optional[str] = None
+        last_reasoning: Optional[str] = None
         use_streaming = streaming and event_publisher
 
         while steps < max_steps:
@@ -110,12 +112,14 @@ class AgenticLoop(AgentLoopProtocol):
 
             if not llm_response.has_tool_calls():
                 final_answer = llm_response.content or ""
+                last_reasoning = getattr(llm_response, 'reasoning_content', None)
                 break
 
             # Model chose to call tools -- execute all in parallel, then loop
             tool_calls = [tc for tc in (llm_response.tool_calls or []) if tc is not None]
             if not tool_calls:
                 final_answer = llm_response.content or ""
+                last_reasoning = getattr(llm_response, 'reasoning_content', None)
                 break
 
             messages.append({
@@ -168,6 +172,7 @@ class AgenticLoop(AgentLoopProtocol):
             final_answer=final_answer,
             steps=steps,
             metadata={"agent_mode": True, "strategy": "agentic"},
+            reasoning_content=last_reasoning,
         )
 
     async def _call_llm(
@@ -229,10 +234,25 @@ class AgenticLoop(AgentLoopProtocol):
         accumulated_content = ""
         final_response: Optional[LLMResponse] = None
         is_first = True
+        sent_reasoning = False
 
         try:
             async for item in stream:
-                if isinstance(item, str):
+                if isinstance(item, ReasoningToken):
+                    # Stream reasoning tokens to frontend in real-time
+                    sent_reasoning = True
+                    await event_publisher.send_json({
+                        "type": "reasoning_token",
+                        "token": item.token,
+                    })
+                elif isinstance(item, ReasoningBlock):
+                    # Final complete reasoning block
+                    sent_reasoning = True
+                    await event_publisher.send_json({
+                        "type": "reasoning_content",
+                        "content": item.content,
+                    })
+                elif isinstance(item, str):
                     await event_publisher.publish_token_stream(
                         token=item, is_first=is_first, is_last=False,
                     )
@@ -250,8 +270,12 @@ class AgenticLoop(AgentLoopProtocol):
         if final_response is None:
             final_response = LLMResponse(content=accumulated_content)
 
-        # If the response is text-only (no tools), close the stream
-        if not final_response.has_tool_calls() and accumulated_content:
+        # Close the stream if we emitted any text or reasoning, so the frontend
+        # clears the _streaming flag before tool-call messages are added. This
+        # also covers the reasoning-only-then-tool-calls case: otherwise the
+        # reasoning message stays _streaming and the post-tool final answer
+        # appends to that stale pre-tool message instead of a fresh one.
+        if accumulated_content or sent_reasoning:
             await event_publisher.publish_token_stream(
                 token="", is_first=False, is_last=True,
             )
--- a/atlas/application/chat/agent/protocols.py
+++ b/atlas/application/chat/agent/protocols.py
@@ -21,6 +21,7 @@ class AgentResult:
     final_answer: str
     steps: int
     metadata: Dict[str, Any]
+    reasoning_content: Optional[str] = None
 
 
 @dataclass
--- a/atlas/application/chat/agent/streaming_final_answer.py
+++ b/atlas/application/chat/agent/streaming_final_answer.py
@@ -5,6 +5,7 @@ import logging
 from atlas.application.chat.utilities.error_handler import classify_llm_error
 from atlas.interfaces.events import EventPublisher
 from atlas.interfaces.llm import LLMProtocol
+from atlas.modules.llm.models import ReasoningBlock, ReasoningToken
 
 logger = logging.getLogger(__name__)
 
@@ -30,6 +31,21 @@ async def stream_final_answer(
         async for token in llm.stream_plain(
             model, messages, temperature=temperature, user_email=user_email,
         ):
+            # Reasoning-capable models interleave reasoning markers with text
+            # tokens; forward them as reasoning events instead of appending to
+            # the answer (str concatenation on a marker would raise).
+            if isinstance(token, ReasoningToken):
+                await event_publisher.send_json({
+                    "type": "reasoning_token",
+                    "token": token.token,
+                })
+                continue
+            if isinstance(token, ReasoningBlock):
+                await event_publisher.send_json({
+                    "type": "reasoning_content",
+                    "content": token.content,
+                })
+                continue
             await event_publisher.publish_token_stream(
                 token=token, is_first=is_first, is_last=False,
             )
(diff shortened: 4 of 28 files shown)


garland3:
## Summary
Adds end-to-end streaming of model reasoning ("chain-of-thought") content: a LiteLLM monkey-patch surfaces a `reasoning` delta field, new typed stream markers carry it through the plain/RAG/tools/agent runners, and a collapsible per-message reasoning section renders it in the frontend with auto-expand, a streaming cursor, and reload persistence.

## Verdict
Request changes.

## Required changes
1. In `stream_and_accumulate` at `atlas/application/chat/modes/streaming_helpers.py:97-112`, stop shipping a successful fallback result as an error. When `fallback_fn()` returns content, deliver it via `publish_chat_response(message=accumulated, ...)`; only emit `{"type":"error", ...}` when the fallback failed or was absent. Right now a valid recovered answer is rendered as a system "Error: …" bubble while history persists the same text as a normal assistant message.
2. Avoid the per-token serialization tax in `_inject_reasoning_content` at `atlas/modules/llm/litellm_caller.py` — the global patch calls `delta.model_dump()` on every chunk of every model. Read the field by attribute first (`getattr(delta, 'reasoning', None)`, then `__pydantic_extra__`) and `continue` when absent, so non-reasoning streams pay nothing.
3. Add unit coverage for the reasoning monkey-patch at `atlas/modules/llm/litellm_caller.py:55-130` — it is currently exercised only by a live e2e test that is skipped without a local vLLM endpoint. Test `_inject_reasoning_content` against a fake chunk for: reasoning present, `reasoning_content` already set (no overwrite), no `reasoning` key (no-op), empty choices, and an assignment that raises (swallowed/logged).
4. Forward trailing reasoning from the synthesis stream in `_stream_synthesis` at `atlas/application/chat/modes/tools.py:304` — `ReasoningToken`/`ReasoningBlock` items yielded after the final `LLMResponse` are currently extracted only as stored metadata and never sent to the frontend.
5. Guard the reasoning-panel auto-scroll at `frontend/src/components/Message.jsx:530-533` — only pin to bottom when already near it (`scrollHeight - scrollTop - clientHeight < threshold`), so a user scrolling up to read earlier reasoning is not yanked back down on each token.
6. Expose toggle state on the reasoning control at `frontend/src/components/Message.jsx:516` — add `aria-expanded` and `aria-controls` (with an `id` on the panel) so assistive tech can announce open/closed state.
7. Add demonstration evidence: at least one recording or screenshot set of the reasoning section streaming live, auto-collapsing on completion, toggling, and surviving a reload — and check off (or annotate) the six manual test-plan items, which are all currently unchecked.

## Evidence
Evidence: 6/10. The backend and frontend reducer/handler logic are well unit-tested for a change this size, but the user-facing visual behavior is undemonstrated — no screenshots or recording of the reasoning UI, and the manual test plan is entirely unchecked.

## Notes
- Reasoning reaches the frontend via two divergent encodings (standalone `reasoning_*` events vs. a `reasoning_content` field on `chat_response`); the marker→event translation is also copy-pasted across `agentic_loop.py`, `tools.py`, and `streaming_helpers.py` — worth centralizing behind one publisher helper.
- `stream_and_accumulate` return type was widened from `str` to a bare `tuple`; annotate as `Tuple[str, Optional[str]]` (`streaming_helpers.py:20`).
- Loopback detection in `litellm_caller.py:620` uses substring matching; parse the URL and compare the hostname against an explicit loopback set instead.

---

garland3:
Thanks for the thorough review. Addressed the feedback in the (amended) single commit — summary below.

### Correctness
- **Successful fallback rendered as an error** (`streaming_helpers.py`, AGENT-REVIEW-BOT-3 #1 / Codex): when the stream fails but `fallback_fn()` recovers a valid answer, it's now delivered via `publish_chat_response(...)` (matching the existing no-token fallback path); the `error` frame is only sent when the fallback also fails or is absent. Added two unit tests covering both branches.
- **Reasoning-only → tool-call streams not closed** (`agentic_loop.py`, Codex): now track whether any reasoning was emitted and close the stream (`is_last`) before returning a tool-call response, so the post-tool answer doesn't append to the stale pre-tool message.
- **Agent final-answer streaming treated `ReasoningToken` as a string** (`streaming_final_answer.py`, Codex): now forwards `ReasoningToken`/`ReasoningBlock` as reasoning events instead of concatenating them (which would have raised and dropped to the non-streamed fallback). Added a unit test.

### Performance
- **Per-token `model_dump()`** (`litellm_caller.py`, BOT #2 / Copilot): `_inject_reasoning_content` now reads `reasoning` via a cheap attribute lookup, then `__pydantic_extra__`, and `continue`s when absent — non-reasoning streams pay nothing. Added 6 unit tests for the monkey-patch (reasoning present, present-via-extra, already-set no-overwrite, no-op, empty choices, swallowed assignment error) — previously only covered by the skipped live e2e test (BOT #3).

### Frontend
- **`reasoning_content` ignored `data.content`** (Copilot / BOT): `STREAM_REASONING_END` now accepts the backend's authoritative full text and reconciles the streamed accumulation; handler passes `data.content` through.
- **Orphan reasoning flush timer**: `endTokenStream()` now flushes and clears the reasoning buffer/timer, so a stream that ends (error/`response_complete`/final token) before a `reasoning_content` event can't fire a late orphan flush.
- **Auto-scroll yank** (BOT #5): the reasoning panel only pins to the bottom when the user is already near it.
- **A11y / `type="button"`** (BOT #6 / Copilot): toggle now has `type="button"`, `aria-expanded`, and `aria-controls` pointing at an `id`'d panel.

### Docs / types / changelog
- Added `ReasoningToken` to the `stream_with_rag` / `stream_with_rag_and_tools` return annotations (Copilot); `streaming_helpers` return type is now `tuple[str, str | None]` and its docstring notes the reasoning markers.
- Moved the changelog entry under `## [Unreleased]` as **PR #655** per the AGENTS.md convention (Copilot), and corrected the doc's LiteLLM version reference to `v1.83.x` (matches `pyproject.toml`).
- Fixed the misleading "preserve all metadata fields" comment in `ChatContext.jsx` to describe the explicit allowlist (Copilot).

### Not changed (with rationale)
- **"Synthesis trailing reasoning never sent to frontend" (BOT #4)** — this looks like a false positive: `_stream_synthesis` delegates to `stream_and_accumulate`, which *does* forward `ReasoningToken`/`ReasoningBlock` to the frontend; the cited `tools.py:304` is the `tool_calls` dict construction, not synthesis code.
- **Demonstration screenshots/recording (BOT #7)** — the live reasoning UI needs a running vLLM/o3 reasoning endpoint that isn't available in this environment, so I can't produce authentic captures. The reducer/handler/UI behavior is covered by the frontend unit tests; flagging the manual visual items as still needing a human pass on real hardware.
- **Loopback substring matching / centralizing the marker→event publisher** (BOT notes) — pre-existing code and a cross-cutting refactor respectively; out of scope for this PR, worth a follow-up.

All backend reasoning tests (39) and the full frontend suite (486) pass locally; ruff/eslint clean.

garland3:
## Summary
Adds end-to-end reasoning-content streaming: a LiteLLM stream patch surfaces
`reasoning` deltas, new `ReasoningToken`/`ReasoningBlock` markers flow through
the plain/RAG/tools/agentic runners and the WebSocket protocol, and a
collapsible per-message reasoning panel renders them in the frontend.

## Verdict
Request changes.

## Required changes
1. Verify the fallback return value before declaring recovery at
   `atlas/application/chat/modes/streaming_helpers.py:103-122` — `recovered`
   is set `True` merely because `fallback_fn()` didn't raise, so an empty
   fallback result publishes a blank assistant message and suppresses the
   error frame. Gate on the value (e.g. `recovered = bool(<fallback result>)`)
   and fall through to the error branch when it's falsy.
2. Key the no-content fallback on missing content only at
   `atlas/application/chat/modes/streaming_helpers.py:76` — the added
   `and not reasoning_content` guard skips recovery when a model emits
   reasoning but no answer, persisting an empty assistant message. Restore
   `if not accumulated and fallback_fn`.
3. Honor the reasoning toggle during streaming at
   `frontend/src/components/Message.jsx:515,522,530` — `showExpanded` is
   forced open while streaming, so a mid-stream click silently flips hidden
   state and inverts the promised auto-collapse. Either disable the toggle
   while streaming, or drive `showExpanded` from `!reasoningCollapsed` and
   collapse once on completion.
4. Restrict the stream wrapper to async streams at
   `atlas/modules/llm/litellm_caller.py:77-98` — `_patched_csw_init`
   installs the async-only `_ReasoningMappedStream` on every
   `CustomStreamWrapper`, breaking any sync `stream=True` path
   (no `__next__`, no attribute proxying). Wrap only when
   `hasattr(raw_stream, "__anext__")`, or add `__iter__`/`__next__` plus a
   `__getattr__` passthrough.
5. Persist the reasoning that was actually displayed in tools mode at
   `atlas/application/chat/modes/tools.py:223-228,335-349` — only
   `synthesis_reasoning` is stored, so pre-tool-call reasoning shown live is
   lost on reload. Persist `accumulated_reasoning` (or fall back to it when
   synthesis has none) so reloaded history matches what the user saw.
6. Add tests for the reasoning-only-then-tool-calls paths that this PR
   deliberately handles but never exercises: a `stream_with_tools` /
   `stream_plain` case yielding reasoning then tool calls (no content) at
   `atlas/modules/llm/litellm_streaming.py:114-115,248-249`, and the
   matching `is_last` stream-close assertion in `AgenticLoop`
   (`atlas/application/chat/agent/agentic_loop.py:238-280`) and
   `ToolsModeRunner`.
7. Add a `Message.jsx` render test for the reasoning panel
   (`frontend/src/components/Message.jsx:513-551`) asserting toggle behavior,
   `aria-expanded`/`aria-controls` tracking, and the near-bottom auto-scroll
   guard — the two behaviors added this revision ship with no component-level
   coverage.

## Evidence
Evidence: 6/10. The backend/reducer logic is well-tested for a change this
size (39 backend + full frontend suites passing), but the feature's entire
value is visual and undemonstrated — the live e2e is skipped without a vLLM
endpoint and all six manual visual checks are unchecked. Add a screenshot
set or recording (or a component-level visual harness for `Message.jsx`)
covering streaming / auto-collapse / toggle / post-reload.

## Notes
- Reduce the typed/untyped split: reasoning travels both as
  `reasoning_token`/`reasoning_content` events and as a field on
  `chat_response`; converging on one canonical wire representation would
  shrink the frontend ingestion surface (`streaming_helpers.py:113-121`).
- `raw_events: List[Dict]` on the public `ChatResult` exposes the internal
  event stream and is re-scanned to derive reasoning (`atlas/atlas_client.py:23-36,156-170`);
  prefer a first-class accumulated field.
- Substring loopback check at `atlas/modules/llm/litellm_caller.py:627`
  (`"localhost" not in model_url`) misclassifies hosts like
  `localhost.attacker.example`; compare the parsed hostname instead.

---

garland3:
Superseded by #781.

This branch was ~230 commits behind `main` and no longer applied — several of the files it touched (`streaming_helpers.py`, the tools-mode synthesis path, and message persistence, which has since moved into `buildPersistedMessage`) had been refactored underneath it. #781 reproduces the same feature against current `main` as a single reviewable commit, with tests and the developer doc carried forward.

Two unrelated drive-by changes from this branch were intentionally left out of #781 and are noted there: the `streaming_helpers.py` error-path rework (main has since added `on_error_message` to that code) and the localhost `OPENAI_API_KEY` heuristic plus `e2e_tests_live.sh` rewrite. Worth reopening separately if still wanted.

PR #658, opened 2026-06-18
fix(quality): clear CodeQL code-quality alerts

garland3:
## Summary

Triage + cleanup of the CodeQL **code-quality** alerts on `main`. This fixes the 12 genuine findings in source and leaves the rest documented as false positives / by-design (dismissed in the code-scanning UI).

### Fixed in code
- **`py/empty-except` (8)** — added `logger.debug(...)` to best-effort `except` blocks instead of silently `pass`ing, so failures are diagnosable:
  - `atlas/core/otel_config.py` — span-file `chmod`, shutdown `fsync`, log-file `chmod`
  - `atlas/core/telemetry.py` — sidecar dir `chmod`, error-path `os.close`, sidecar file `chmod`
  - `atlas/modules/mcp_tools/client.py` — server task-support probe, `tool_task_completed` callback
- **`py/unused-global-variable` (1)** — removed the unused module-level `logger` (and now-unused `import logging`) in `atlas/routes/telemetry_routes.py` (0 references in the module).
- **`py/implicit-string-concatenation-in-list` (1)** — `atlas/modules/llm/litellm_caller.py`: made the RAG citation instruction an explicit `+` concatenation (string value unchanged) to remove the missing-comma ambiguity.
- **`js/unused-local-variable` + `js/useless-assignment-to-local` (2)** — `frontend/src/components/AgentPortal.jsx`: removed the dead `STREAM_COLORS` constant (the live copy lives in `agent-portal/Pane.jsx`) and replaced a useless `next` initializer with a ternary.

### Not changed (dismissed as non-issues)
- **`py/unused-global-variable` ×3** — `_ephemeral_warned`, `_HMAC_WARNED` (warn-once flags), `_resolved_path` (memoization cache). These are read on subsequent calls; CodeQL's per-call view can't see it. Deleting them would break behavior → false positives.
- **`py/print-during-import` ×3** — `atlas/atlas_chat_cli.py`. Intentional CLI bootstrap output (version / missing-env-file diagnostics).

## Validation
- `ruff check` on all changed Python: clean
- Targeted tests: `test_telemetry_routes`, `test_telemetry_spans`, `test_mcp_*`, capabilities/audit-log suites — **138 passed, 1 skipped**
- Frontend: `eslint` 0 errors, `vite build` succeeds

### DIFF
--- a/atlas/core/otel_config.py
+++ b/atlas/core/otel_config.py
@@ -101,8 +101,10 @@ class JSONLSpanExporter(SpanExporter):
             self._fh = self.file_path.open("a", encoding="utf-8")
         try:
             os.chmod(self.file_path, 0o600)
-        except OSError:
-            pass
+        except OSError as e:
+            logging.getLogger(__name__).debug(
+                "Could not restrict span file permissions on %s: %s", self.file_path, e
+            )
 
     def export(self, spans: list[ReadableSpan]) -> SpanExportResult:  # noqa: D401
         with self._lock:
@@ -145,8 +147,10 @@ class JSONLSpanExporter(SpanExporter):
                     self._fh.flush()
                     try:
                         os.fsync(self._fh.fileno())
-                    except (OSError, ValueError):
-                        pass
+                    except (OSError, ValueError) as e:
+                        logging.getLogger(__name__).debug(
+                            "fsync of span file failed during shutdown: %s", e
+                        )
                     self._fh.close()
                 except Exception as e:  # noqa: BLE001
                     logging.getLogger(__name__).debug(
@@ -281,8 +285,10 @@ class OpenTelemetryConfig:
         # filesystems without POSIX modes.
         try:
             os.chmod(self.log_file, 0o600)
-        except OSError:
-            pass
+        except OSError as e:
+            logging.getLogger(__name__).debug(
+                "Could not restrict log file permissions on %s: %s", self.log_file, e
+            )
         root.addHandler(file_handler)
         root.setLevel(self.log_level)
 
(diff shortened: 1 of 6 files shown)


garland3:
## Summary
A behavior-preserving CodeQL code-quality cleanup across 6 files (+32/-33): replacing silent `except: pass` blocks with `logger.debug(...)`, removing an unused logger and import, making an implicit string concatenation explicit, deleting a dead JS constant, and swapping a mutable `let` for a ternary.

## Verdict
Request changes.

## Required changes
1. Promote the permission-restriction failures from `debug` to `warning` at `atlas/core/telemetry.py:320` and `:342`, and `atlas/core/otel_config.py:104` and `:289` — when these `chmod(..., 0o600/0o700)` calls fail, the span/sidecar files holding tool output and trace data are left at default (potentially group/world-readable) mode, and a `debug`-level signal is invisible in production. To avoid noise on non-POSIX mounts, log `warning` only when `os.name == "posix"`, else `debug`.
2. Add at least one negative test per new except category at `atlas/core/telemetry.py:319-342`, `atlas/core/otel_config.py:104-291`, and `atlas/modules/mcp_tools/client.py:2132-2341` — e.g. monkeypatch `os.chmod` to raise `OSError` and assert the function still returns and emits the log record (via `caplog`); none of the eight new branches are currently exercised, so a future refactor would regress them silently.

## Evidence
Evidence: 8/10. Well-evidenced for a behavior-preserving cleanup — `ruff`/`eslint` clean, targeted suites at 138 passed/1 skipped, and `vite build` succeeds. The one gap: the "citation-instruction string value unchanged" claim at `atlas/modules/llm/litellm_caller.py:1047-1053` is asserted but not demonstrated; confirm the concatenated result is byte-identical (a snapshot/assertion on the assembled block would close this).

## Notes
- Where feasible, create sensitive files with the restrictive mode atomically via `os.open(..., 0o600)` rather than chmod-after-open, eliminating the umask window — `atlas/core/telemetry.py:write_tool_output_sidecar` already does this.

---

PR #663, opened 2026-06-19
Fix Bedrock tool-calling failure on streaming follow-up requests

garland3:
## Problem

AWS Bedrock tool calling fails on **follow-up requests** within an agent turn (the second LLM call, after a tool has been executed and its result appended). The request errors out with:

```
Unable to convert openai tool calls=[...] to bedrock tool calls.
Received error=argument of type 'types.SimpleNamespace' is not iterable
```

## Root cause

In streaming mode, the streaming caller (`litellm_streaming.py`) accumulated tool-call fragments and emitted them as `types.SimpleNamespace` objects. The agentic/react/act/think_act loops append these objects verbatim onto the assistant message and re-send them on the next turn.

litellm's Bedrock Converse transformer (`_convert_to_bedrock_tool_call_invoke`) reads tool calls with **dict** syntax (`"function" in tool`, `tool["id"]`, `tool["function"]`). `SimpleNamespace` supports only attribute access, so the Bedrock follow-up request blows up during message transformation. Other providers normalize the objects differently, so the failure is Bedrock-specific.

The `tools` mode already had a local `SimpleNamespace -> dict` workaround (with a comment describing exactly this), but the agent loops never got it.

## Fix

Build the accumulated streaming tool calls with litellm's own `ChatCompletionMessageToolCall` / `Function` types instead of `SimpleNamespace`. These support **both**:

- attribute access (`tc.function.name`) — used by the tool executor, and
- dict access (`tool["id"]`) — used by the Bedrock transform,

matching the non-streaming path exactly and fixing every agent loop centrally rather than just one mode.

## Testing

- Added `test_stream_with_tools_emits_bedrock_compatible_tool_calls`, which drives the real `stream_with_tools` with a mocked streaming response and asserts the emitted tool calls survive the Bedrock Converse transform while still supporting attribute access. (Fails on the old `SimpleNamespace` code.)
- `tests/test_streaming_token_flow.py`, `test_multi_tool_calling.py`, `test_agentic_loop.py` pass; broader tool/agent/streaming subset (442 tests) green.

Scope is limited to the bug and its regression test.

### DIFF
--- a/atlas/modules/llm/litellm_streaming.py
+++ b/atlas/modules/llm/litellm_streaming.py
@@ -7,10 +7,10 @@ These methods are mixed into LiteLLMCaller via LiteLLMStreamingMixin.
 import asyncio
 import logging
 import time
-from types import SimpleNamespace
 from typing import Any, AsyncGenerator, Dict, List, Optional, Union
 
 from litellm import acompletion
+from litellm.types.utils import ChatCompletionMessageToolCall, Function
 
 from atlas.core.metrics_logger import log_metric
 from atlas.core.telemetry import set_attrs, start_span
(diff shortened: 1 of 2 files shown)


garland3:
## Summary
Replaces the `SimpleNamespace` tool-call objects emitted by the streaming
path with litellm's typed `ChatCompletionMessageToolCall`/`Function`, so
streamed tool calls survive the Bedrock Converse transform and match the
non-streaming path. Adds one regression test.

## Verdict
Request changes.

## Required changes
1. Reconcile the now-stale normalization comment at `atlas/application/chat/modes/tools.py:267-280` — streaming no longer yields `SimpleNamespace`, so the comment is factually wrong; either rewrite it to describe the new litellm type as a deliberate defensive conversion, or drop the manual dict conversion so the tools runner and the agentic/react loops share one documented re-send shape.
2. Guard the brittle private import at `atlas/tests/test_streaming_token_flow.py:499` with `pytest.importorskip(...)` (or a try/except `pytest.skip`) — `_convert_to_bedrock_tool_call_invoke` is an underscore-prefixed internal symbol that a litellm bump can rename, turning the regression into a confusing collection failure.
3. Add a multi-tool-call test variant in `atlas/tests/test_streaming_token_flow.py:507` streaming `index=0` and `index=1` and asserting two well-formed `toolUse` blocks with correct id/name mapping — the per-index `sorted()` accumulation branch (the part most likely to break on parallel calls) is currently unexercised.
4. Pass `id=tc["id"] or None` when constructing the tool call at `atlas/modules/llm/litellm_streaming.py:234` — an empty-string id is preserved verbatim and yields an empty Bedrock `toolUseId`, which fails the follow-up request this PR aims to fix; `None` lets litellm auto-generate a valid id.

## Evidence
Evidence: 9/10. Strong for a non-visual bug fix — root cause is precisely described and a targeted regression test drives the real transform; the only gap is no pasted before/after pytest output to make the "fails before, passes after" claim self-verifying.

## Notes
- Consider isolating the hand-built `ChatCompletionMessageToolCall(...)` behind a small module-private helper at `atlas/modules/llm/litellm_streaming.py:234-241` to centralize the upstream-type coupling.

---

PR #666, opened 2026-06-22
Fix transfer MCP write_file_to_disk for session files; drop dead config

garland3:
## Problem

Writing a tool-produced **session file** (e.g. a SolidWorks STEP export) to local disk silently failed. Across multiple models, the assistant would loop on `read_file_from_disk` and eventually give up with "I'm unable to locate the exported file" — even though the file existed in session storage.

### Root cause

The backend's session-file injection (`inject_context_into_args` in `tool_executor.py`) only rewrites a parameter named **exactly** `filename`/`file_names` into a tokenized download URL (strict matching since `f4b60a1`). `write_file_to_disk` only accepted inline `content`/`content_is_base64` and declared no `filename`, so there was **no path** for it to obtain a session artifact's bytes. The model can't produce 20 KB of binary inline, so it fell back to `read_file_from_disk` (the developer's local disk, where the artifact doesn't exist) and looped. Mock S3 was never involved.

## Fix

- Add a `filename` parameter to `write_file_to_disk`. The backend rewrites it to a download URL; the tool fetches the bytes over HTTP with the same `MCP_TRANSFER_MAX_BYTES` cap and `Content-Length` fast-fail as the reader, then writes them verbatim (mirrors the `file_viewer` fetch pattern).
- Support a **directory destination** by appending the source name (from the backend-supplied `original_filename`).
- Inline `content`/base64 still works; an unresolved bare name returns a clear error instead of failing silently.

## Default access guards (footgun prevention)

These tools are auto-approved on dev machines, so they ship with sane guards **on by default** that still stay out of the way of normal project work:

- **Primary root** defaults to the **home directory** (`MCP_TRANSFER_BASE_DIR` to override); relative paths anchor there and access outside it is denied.
- **Whitelist extra dirs** (e.g. network mounts outside home) via `MCP_TRANSFER_ALLOWED_DIRS` — `os.pathsep`-separated, e.g. `/projects:/mnt`.
- **Hidden paths blocked**: dotfiles/dot-directories below a root (`~/.ssh`, `~/.aws`, `.env`, `.git`, …) are denied to protect credentials/configs.
- **Escape hatches**: `MCP_TRANSFER_ALLOW_HIDDEN=true` permits hidden paths; `MCP_TRANSFER_ALLOW_ANY_PATH=true` disables the guards entirely.
- Paths are **resolved (symlinks + `..`) before** the guards run, so neither can escape an allowed root.

All transfer env vars are documented in `.env.example`.

## Cleanup

- Remove the dead `INCLUDE_FILE_CONTENT_BASE64` setting and its unreachable helper (never wired into injection), plus the `.env.example` entry.
- Correct `docs/developer/mcp-file-io.md`: fetching the rewritten `filename` URL is the only supported mechanism for receiving session files (base64 content is not auto-injected).

## Testing

- `bash test/atlas_tests.sh` → **1647 passed, 17 skipped**.
- `test_mcp_transfer.py` cases: backend-fetch + URL resolution, directory naming, size-cap rejection, missing-source, unresolved-name, outside-root denied, `ALLOW_ANY_PATH` bypass, `ALLOWED_DIRS` whitelist, hidden-path denied, `ALLOW_HIDDEN` permit, and home-as-default-root.


garland3:
## Summary
Adds a `filename` parameter to the transfer MCP `write_file_to_disk` tool so session-produced artifacts can be written to local disk: the backend rewrites the name to a tokenized download URL and the tool fetches the bytes (reusing the reader's size cap and `Content-Length` fast-fail), supports a directory destination, and removes the dead `INCLUDE_FILE_CONTENT_BASE64` setting and its helper.

## Verdict
Looks good.

## Evidence
No separate evidence assessment was produced for this review. The PR description reports the full suite passing (1642 passed, 17 skipped) and adds `test_mcp_transfer.py` cases covering backend fetch + URL resolution, directory-destination naming, the size-cap rejection, missing-source, and unresolved-name paths, which is proportionate to a change of this size.

---

garland3:
## Summary
Adds a `filename` parameter to the transfer MCP `write_file_to_disk` tool so session-produced artifacts can be written to local disk — the backend rewrites the name to a tokenized download URL and the tool fetches the bytes (reusing the reader's size cap and `Content-Length` fast-fail), with directory-destination support and default access guards (home-root anchoring, hidden-path blocking, escape-hatch env vars). Also removes the dead `INCLUDE_FILE_CONTENT_BASE64` setting and corrects the file-IO docs.

## Verdict
Looks good.

## Evidence
No separate evidence assessment was produced for this review. The description reports the full suite passing (1647 passed, 17 skipped) and adds `test_mcp_transfer.py` cases covering backend fetch + URL resolution, directory naming, size-cap rejection, missing-source, unresolved-name, outside-root denial, the `ALLOW_ANY_PATH`/`ALLOWED_DIRS`/`ALLOW_HIDDEN` paths, and home-as-default-root — proportionate to a change of this size.

---

PR #685, opened 2026-06-29
Persist tool call input/output in saved conversations (#684)

### LINKED ISSUE #684: conversation history does not include tool calls and tool input and output
At least in the UI it is missing the tool input and output when reloading a conversation from the history. 

Support showing the tool i/o being saved and reloaded when restarting a conversation.

garland3:
Fixes #684.

## Diagnosis (answer to the triage question)

Tool input/output was **not being saved at all** — this is a storage/save-path fix, not a rendering/hydration fix. The investigation showed:

- **Schema already supports it**: `conversation_messages.metadata_json` + `message_type` can hold arbitrary per-message detail.
- **Reload already supports it**: `loadSavedConversation` spreads `msg.metadata` back onto the message, and `Message.jsx` already renders `tool_call` rows from `tool_name` / `arguments` / `result` / `status`.
- **The gaps were entirely on the save/export sides**:
  - *Server mode*: tool calls are emitted as transient WebSocket events and were never added to `session.history`, so `_save_conversation` never saw them.
  - *Local mode*: the IndexedDB autosave kept only `role`/`content`/`timestamp`/`message_type`, dropping the tool fields.
  - *Export*: the `.txt` export emitted only `m.content`.

## Changes

**Server mode (backend)**
- New `ToolCallRecorder` (`atlas/application/chat/utilities/tool_history.py`) wraps the turn's update callback in tools mode, capturing the already-UI-sanitized `tool_start` / `tool_complete` / `tool_error` payloads and flushing them as display-only `tool_call` messages into history just before the final assistant message (order: user → tool_call(s) → assistant). Covers the non-streaming and streaming (multi-round) tools paths.
- `ConversationHistory.get_messages_for_llm()` excludes `message_type=tool_call` rows, so they are persisted/rendered but **never replayed to the model** (avoids orphan-tool-message provider errors).
- `handle_restore_conversation` now preserves message `metadata` so restored tool rows keep their type (and re-save keeps their detail).
- The canvas tool is skipped (it renders into the canvas panel, not the transcript), matching live UI behavior.

**Local mode (frontend)**
- `buildPersistedMessage` tucks tool-call fields into message `metadata` for the IndexedDB autosave; the restore path already spreads metadata back.

**Export (frontend)**
- `.txt` export renders a tool block (name, input arguments, output result) with large base64 elided via the same display filters used in-app. `.json` export already carried the fields. Display-only rows are filtered out of the `restore_conversation` payload sent to the backend.

Scope note: in-chat **agent-loop** tool steps are a separate path and out of scope here, consistent with the product direction in `AGENTS.md`.

## Tests
- Backend: `test_tool_call_persistence.py` (recorder event→message mapping, callback pass-through, failed/canvas handling, ordering, LLM-context exclusion, repository save/load round-trip) + a restore metadata-preservation test in `test_chat_service_conversation_id_default.py`.
- Frontend: `buildPersistedMessage` and `formatToolCallForText` coverage in `chatExport.test.js`.
- Full suites green: backend 1693 passed / 17 skipped; frontend 474 passed; lint clean.

## Docs
- Updated `docs/admin/chat-history.md` (schema table + "Tool calls in saved conversations" section).


garland3:
## Summary
Persists tool-call input/output into saved conversations by recording tool events during a turn and flushing them into history before the final assistant message, with matching frontend serialization for save/restore and `.txt`/`.json` export. Restored `tool_call` rows are kept out of the LLM message list.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. The change is well-demonstrated for its size — a dedicated backend suite covers the recorder mapping, ordering, LLM-context exclusion, and a save/load round-trip, plus restore and frontend export tests, with full suites green. The one missing artifact is an end-to-end screenshot of a saved conversation reloaded with its tool calls re-rendering in the transcript (and the `.txt` export block); the round-trip is asserted only via unit tests.

## Notes
- Orphan `role=tool` rows can still reach the provider: `get_messages_for_llm` excludes only on `metadata.message_type`, so a tool row with null/corrupt metadata falls through — also drop by role (`atlas/domain/messages/models.py:173`).
- Auth-required (or any start-without-finish) tool calls persist as a permanent `status="calling"` row that reloads as a tool stuck mid-call; coerce non-terminal status at flush (`atlas/application/chat/utilities/tool_history.py:39`).
- Failed tool errors are printed under the `Output Result:` heading in the `.txt` export, reading like a normal result; branch the heading on `status === 'failed'` (`frontend/src/utils/chatExport.js:88`).

---

garland3:
## Summary
Persists tool-call input/output into conversation history (via a new `ToolCallRecorder`) so saved conversations re-render tool calls in the transcript and `.txt`/`.json` exports, while excluding those display-only rows from the LLM context. Touches the backend save path, frontend persistence, and export.

## Verdict
Request changes.

## Required changes
1. Guard the metadata deref at `atlas/domain/messages/models.py:176` — use `(msg.metadata or {}).get("message_type")`. `Message.from_dict` returns `metadata=None` for any serialized row carrying `"metadata": null`, so the new exclusion filter raises `AttributeError` and fails the entire LLM turn when an older saved conversation is replayed.
2. Also drop bare `role == MessageRole.TOOL` rows in `get_messages_for_llm` (`models.py:176`) — a tool row whose metadata lacks `message_type` (corrupted/partial metadata, the `None` case above) currently falls through and is replayed as an orphan tool message, which several providers reject. Tool results are never replayed standalone, so excluding by role is safe.
3. Coerce non-terminal status to a terminal state at flush in `atlas/application/chat/utilities/tool_history.py:74` — a tool that emits `tool_start` but no `tool_complete`/`tool_error` (auth gate, cancelled/interrupted turn) persists with `status="calling"` and reloads as a permanently animated `CALLING` row on static history.
4. Add an aggregate size guard in `_elide_for_storage` (`tool_history.py:39`) — the per-string `_MAX_STR_CHARS` cap does not bound element count, so a high-cardinality result (many small strings) writes a multi-MB `metadata_json` row unchecked. Cap total serialized bytes / list length, and apply the string cap at all depths (currently values past `_MAX_DEPTH = 6` are stored uncapped).

## Evidence
Evidence: 8/10. The change is well-demonstrated for its size — a dedicated 380-line backend suite, an end-to-end recorder-wiring test, restore coverage, and frontend export tests, with quantified pass counts. The one gap is that the user-facing outcome (a reloaded conversation re-rendering its tool calls, and the `.txt` export block) is asserted only via unit tests, never shown; a screenshot would close that loop but is not required.

## Notes
- `.txt` export labels a failed tool's error under `Output Result:`; branch the heading on `status === 'failed'` to `Error Details:` to match the in-app renderer (`frontend/src/utils/chatExport.js:88`).
- `message_type` is serialized two ways (inside `metadata` by the backend recorder, top-level by `buildPersistedMessage`), reconciled by a read-time fold; pick one canonical write location (`service.py:529`, `chatExport.js:43`).
- The recorder lifecycle (construct → reassign callback → flush before each assistant append) is hand-replicated across several exit paths in `tools.py`; a single `finalize_turn` helper would prevent a future early-return from silently dropping persistence.

---

PR #723, opened 2026-07-14
Enforce MCP tool group ACLs at execution time; make filter fail closed

garland3:
This PR closes the MCP tool ACL bypass described in the review notes.

## Changes
- Add a fail-closed group ACL check inside `MCPToolManager.execute_tool` at the single execution choke point, keyed on `context['user_email']`.
  - Missing user context, disabled servers, authorization exceptions, or unauthorized group membership now deny execution.
  - This fixes agent mode bypassing the request-time filter and blocks model-emitted (hallucinated/prompt-injected) tool names from executing against group-restricted servers.
- Convert `ToolAuthorizationService.filter_authorized_tools` to fail closed: return an empty list when the tool manager is unavailable or the ACL check raises, instead of returning the unfiltered `selected_tools`.
- Add new execution-time authorization tests and an exception fail-closed regression test for the request-time filter.
- Update `docs/admin/mcp-servers.md` and `CHANGELOG.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
-### PR #TBD - 2026-07-14
+### PR #723 - 2026-07-14
 - **Fail-closed MCP tool ACL enforcement at execution time**: Fixed agent mode bypassing the tool authorization filter by enforcing group ACL checks inside `MCPToolManager.execute_tool` at the single execution choke point, keyed on the trusted `context["user_email"]`. Missing user context, disabled servers, group-check exceptions, and unauthorized membership now deny execution rather than fall back to allowing the call. Also made `ToolAuthorizationService.filter_authorized_tools` fail closed: it returns an empty list whenever the ACL check cannot complete instead of returning the unfiltered selection.
 
 ### PR #717 - 2026-07-08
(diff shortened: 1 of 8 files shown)


garland3:
## Summary
Adds an execution-time authorization choke point in `MCPToolManager.execute_tool` so tool calls are re-checked against per-server group ACLs, closing an agent-mode gap where model-emitted tool names could bypass the request-time filter. Also flips the "no tool manager" and error paths in the request-time filter to fail closed.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. This backend-only change is well-demonstrated — two test files add behavior-level cases covering each fail-closed branch, and docs/CHANGELOG were updated. The only gap is that the PR description does not state the suite was run green; adding the test command and result would round it out.

## Notes
- Fail-open for unknown servers: `_is_server_authorized_for_user` treats a `server_name` absent from `servers_config` as public (returns authorized), the opposite of the request-time allow-list — safe under today's invariants but fragile (`atlas/modules/mcp_tools/mcp_execution.py:74`).
- Canvas/RAG pseudo-tools are silently dropped when `tool_manager` is None because the early `return []` precedes the pseudo-tool allow-list (`atlas/application/chat/policies/tool_authorization.py:55`).
- Group check runs sequentially without short-circuit, so an authorized user can incur N sequential auth-endpoint round-trips per call; short-circuit on first match (`atlas/modules/mcp_tools/mcp_execution.py:94`).

---

garland3:
## Summary
Closes an MCP tool ACL bypass by adding a fail-closed, execution-time group
authorization check at the `execute_tool` choke point, plus a new test file,
doc, and CHANGELOG updates.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. Appropriate for a non-visual backend security fix — eight
behavior-level tests cover each fail-closed branch, and docs/CHANGELOG describe
the new invariant. Only gap: the PR never states the suite was run green;
adding the `pytest` command and its result would close it.

## Notes
- Pseudo-tools (`canvas_canvas`, `atlas_rag`) are now silently dropped when
  `tool_manager` is None because the early `return []` sits ahead of the
  allow-list — behavioral regression, move the allow-list first at
  `atlas/application/chat/policies/tool_authorization.py:55`.
- Group ACL check awaits every group with no short-circuit at
  `atlas/modules/mcp_tools/mcp_execution.py:94` — loop and `return True` on the
  first match to bound the authorized case to one round-trip.
- Execution-time check duplicates the request-time authorization logic; extract
  a shared `is_server_authorized_for_user` predicate
  (`atlas/modules/mcp_tools/mcp_execution.py:68`) so the two paths cannot drift.

---

Issue #727, opened 2026-07-25
MCP tool inputSchema is never validated; one malformed schema breaks every request


modeofO:
## Summary

`MCPToolManager.get_tools_schema` forwards each MCP tool's advertised `inputSchema` to the LLM provider verbatim, with no validation. Providers require a function's `parameters` to be a JSON Schema of `type: "object"`. If any single selected tool advertises a schema that violates this, the provider rejects the **entire** request, so every turn fails until the user finds and deselects that one tool.

`atlas/modules/mcp_tools/mcp_discovery.py:599`

```python
"parameters": getattr(tool, 'inputSchema', {}) or {}
```

Nothing between discovery and `acompletion()` inspects this value.

## Why this is worse than a normal tool failure

The documented behaviour in `AGENTS.md` is that individual tool failures become error `ToolResult`s so other tools still succeed. That isolation lives downstream of the LLM call. An invalid schema fails the call itself, so no tool ever executes and the isolation never applies. One malformed server takes down all tool use.

## Reproduction

Minimal MCP server advertising one invalid and one valid tool:

```python
types.Tool(
    name="lookup",
    description="Look up a safety document by title.",
    inputSchema={"type": "string", "description": "the document title"},   # invalid
)
types.Tool(
    name="healthy",
    description="A correctly specified tool, for contrast.",
    inputSchema={"type": "object", "properties": {"query": {"type": "string"}}, "required": ["query"]},
)
```

Registered in `config/mcp.json`, both tools appear in `/api/config` with no warning:

```
badschema tools: ['lookup', 'healthy']
```

Driving a chat turn over `/ws`, the payload reaching the provider contains the invalid schema unchanged:

```json
badschema_lookup parameters: {"type": "string", "description": "the document title"}
```

Control turn with only valid schemas selected completes normally. Repro turn with the invalid tool added fails the whole turn.

Note: no provider credentials were available in the test environment, so the rejection was produced by a local OpenAI-compatible stub applying the documented `type: "object"` validation. Everything downstream of the rejection is unmodified Atlas code.

## Suggested fix

Validate `inputSchema` at discovery time. A tool whose schema is not a JSON Schema object should be dropped from the inventory and logged with its server name, rather than admitted and allowed to break every subsequent request. Surfacing the quarantined tool in the UI would also help operators notice a broken server.

## Related

The user-visible symptom of this bug is a generic error that does not identify the tool. Filed separately.

PR #733, opened 2026-07-25
Name the rejected tool when a provider rejects the request

### LINKED ISSUE #728: Tool-caused provider rejections surface as a generic LLM error that does not name the tool
## Summary

When a tool definition causes the provider to reject a request, the user sees:

> The LLM service encountered an error. Please try again or contact support if the issue persists.

The provider's error names the offending tool, and that name is present in the backend logs, but it never reaches the user. The message also implicitly points at the model, so the natural user response is to switch models, which does not help. In practice users recover by deselecting tools one at a time until the turn succeeds.

## Root cause: the detail is discarded before anything tries to classify it

Two layers each generalize, and the first destroys what the second would need.

**1.** `atlas/modules/llm/litellm_caller.py:161` — `_raise_llm_domain_error` has branches for rate limit, timeout, auth and context window. `BadRequestError` matches none of them and falls through:

```python
logger.error("LLM call failed (%s): %s", error_type, error_str)
raise LLMServiceError(
    "The LLM service encountered an error. Please try again or select a different model."
) from exc
```

The provider's diagnostic survives only in the log line. The raised error carries a fixed string.

**2.** `atlas/application/chat/modes/tools.py:242` then calls `error_handler.classify_llm_error` on that already-generic error. At `error_handler.py:80`, `error_str = str(error)` is now the generic sentence from step 1, not the provider's message. No branch matches, so it falls to `error_handler.py:110` and produces a second generic string, which is what the frontend receives.

Because step 1 runs first, fixing `classify_llm_error` alone cannot recover the tool name.

## Evidence

Backend logs for a single failing turn:

```
ERROR | litellm_streaming | litellm.BadRequestError: OpenAIException - Invalid schema for
        function 'badschema_lookup': schema must be a JSON Schema of 'type: "object"', got 'string'.
ERROR | litellm_caller   | LLM call failed (BadRequestError): ...same...
ERROR | chat.modes.tools | Streaming tools classified error: LLM error: The LLM service
        encountered an error. Please try again or select a different model.
```

Frames received by the client on the same turn:

```
[token_stream] is_last=True
[ERROR FRAME] error_type=None
[ERROR FRAME] message='The LLM service encountered an error. Please try again or contact support if the issue persists.'
```

Reproduced with an MCP server advertising an invalid tool schema (see #727) and a local OpenAI-compatible stub applying the documented provider validation. A control turn without the invalid tool completes normally.

## Suggested fix

Preserve the provider's diagnostic through `_raise_llm_domain_error` rather than replacing it with a fixed string, and add an explicit `BadRequestError` branch. When the failed request carried tools, the user-facing message should say the request was rejected because of a tool definition and name the tools that were in the request.

`error_handler.classify_llm_error` documents that `user_message` must not contain raw exception details. That constraint is reasonable, but the tool name is not sensitive data and is exactly what the user needs to act on.

## Related

- #727 (the schema that triggers it)

### LINKED ISSUE #729: Streaming tools error frame omits error_type
## Summary

The error frame sent from the streaming tools path omits `error_type`, unlike every other error frame the WebSocket emits.

`atlas/application/chat/modes/tools.py:246`

```python
await self.event_publisher.send_json({
    "type": "error",
    "message": user_msg,
})
```

Every typed handler in `atlas/main.py` sets `error_type` (`rate_limit`, `timeout`, `authentication`, `context_window_exceeded`, `validation`, `authorization`, `domain`, `unexpected`). This path does not, so clients receive `error_type: undefined`.

## Observed

Frames from a failing turn in tools mode:

```
[ERROR FRAME] error_type=None
[ERROR FRAME] message='The LLM service encountered an error. Please try again or contact support if the issue persists.'
```

## Impact

The frontend's `error` branch in `frontend/src/handlers/chat/websocketHandlers.js` still renders the message and ends the stream, so this is not user-visible today. It does mean any current or future handling keyed on `error_type` silently skips this path, and it makes the frame inconsistent with the documented server message contract.

`classify_llm_error` already returns an error class alongside the message, so the type is available at the call site and just is not being forwarded.

## Suggested fix

Map the returned error class to the same `error_type` strings used in `main.py` and include it in the frame.

modeofO:
Fixes #728. Fixes #729.

A malformed tool definition makes the provider reject the whole request, so no tool ever runs and the turn fails. The provider names the offending function, but that detail was replaced by a generic message before it could reach the user:

> The LLM service encountered an error. Please try again or contact support if the issue persists.

The message points at the model, so the natural response is to switch models, which does not help. Users recover by deselecting tools one at a time.

## Cause

Two layers each generalized, and the first destroyed what the second needed.

`_raise_llm_domain_error` (`litellm_caller.py`) had no `BadRequestError` branch, so it fell through to a fixed `LLMServiceError` string. `classify_llm_error` (`error_handler.py`) then ran its keyword matching against `str(error)`, which by then was that fixed string rather than the provider's message. Nothing matched, so it produced a second generic message.

Because the detail was gone before classification ran, fixing `classify_llm_error` alone could not recover the tool name.

## Changes

- `LLMBadRequestError` domain error carrying `tool_names`.
- `_raise_llm_domain_error` maps `litellm.BadRequestError` to it, and takes an optional `tools_schema` so it can report which tools were in the request. Placed after the context-window check, since `litellm.ContextWindowExceededError` subclasses `BadRequestError`.
- If the provider's text names a tool from the request, only that tool is reported; otherwise all request tools are listed so the user can bisect. The raw provider string is not exposed.
- `classify_llm_error` short-circuits on `LLMBadRequestError` and passes its message through.
- `error_type_for()` maps a domain error class to the `error_type` vocabulary already used in `main.py`; the tools-mode error frame now includes it (#729).

Existing classification behavior for rate limit, timeout, auth, context window and generic errors is unchanged.

## Test plan

- [x] `atlas/tests/test_tool_error_attribution.py` — 13 tests, written before the fix and confirmed failing for the expected reasons first
- [x] Provider names a tool: only that tool is reported
- [x] Provider names none: all request tools listed
- [x] No tools in request: message claims no tool
- [x] `ContextWindowExceededError` still classifies as context window, not bad request
- [x] A tool named like another error category (`vault_api_key_lookup`) is not rerouted to the auth branch
- [x] Error frame names the tool and carries `error_type: "bad_request"`
- [x] Full backend suite: same 16 pre-existing failures before and after, none introduced
- [x] `ruff check` clean on all touched files

Reproduced end to end before and after with an MCP server advertising `inputSchema: {"type": "string"}` and a local OpenAI-compatible stub applying the documented `type: "object"` validation.

## Not included

`main.py` routes `LLMBadRequestError` through its `DomainError` handler, so that path reports `error_type: "domain"` while carrying the same message. Left alone to keep this change scoped; worth aligning separately.

The underlying cause — MCP tool schemas are never validated at discovery — is #727 and needs its own fix. This change makes the failure legible, not impossible.


garland3:

## Summary
Adds an `LLMBadRequestError` domain error that carries the offending tool name(s), maps `litellm.BadRequestError` to it after the context-window check, and passes the resulting message through classification so tool-rejection failures name the tool instead of a generic service error.

## Verdict
Looks good.

## Evidence
No evidence score was produced for this change. The description documents 13 targeted tests covering the named-tool, unnamed-tool, no-tools, and context-window-precedence cases, plus an end-to-end reproduction with a malformed MCP schema, which is proportionate to the size of the change.

## Notes
- The `main.py` path still reports `error_type: "domain"` for `LLMBadRequestError`, as called out in the description — worth aligning in a follow-up.

---

garland3:
Pushed the review fixes to this branch (`0a39945`):

- **Tool attribution now needs positive evidence.** Treating any `BadRequestError` as tool-caused whenever tools were selected traded one misleading message for another — an out-of-range parameter or a malformed message history is a 400 too, and blaming the whole tool selection for those sends users to disable tools that were never at fault. Attribution is now: the tool the provider named (whole-token match, so `calc` is not blamed for text containing `calculate`), or every tool in the request when the text points at the tool payload (`tool`/`tools`/`function`/`tool_choice`) without naming one, or no tool at all.
- **RAG fallback paths.** `LLMBadRequestError` was missing from the passthrough tuples in `call_with_rag()` and `call_with_rag_and_tools()`, so a rejection was read as a RAG failure: the same refused request went to the provider a second time, with RAG blamed in the logs.
- **`error_type` consistency.** The tools-streaming path alone left the non-streaming path reporting `domain` for the same condition. `main.py` now has an `LLMBadRequestError` branch, so #729 holds across both paths rather than one.

Backend suite: 1842 passed, 17 skipped, 0 failures.

The red checks here are the fork-PR ceiling, not the change — `build-and-test` passed every test step and failed only on "Push production Docker image" (no ghcr write token for a fork), and "Security Summary" failed on a 403 posting its PR comment. Mirrored to a same-repo branch as #735 so CI can actually report green; that is the one to merge.

garland3:
## Summary
Narrows tool attribution on provider rejections so a tool is only named when the error text gives positive evidence for it, via a new `_tools_implicated_by` helper that feeds `LLMBadRequestError.tool_names`, plus documentation of the attribution rules.

## Verdict
Request changes.

## Required changes

1. Anchor the tool-name match in `_tools_implicated_by` at `atlas/modules/llm/litellm_caller.py:183` to whole tokens and require schema-specific evidence — the bare word `function` in a provider message implicates every selected tool, so "This deployment does not support function calling" returns all of them despite naming none.
2. Move the `isinstance(exc, litellm.BadRequestError)` branch (`atlas/modules/llm/litellm_caller.py:206`) above the substring heuristics at lines 179–196, or scope those heuristics to the provider message rather than the full `str(exc)` — a tool schema rejection for a tool whose name merely contains `timeout` or `authentication` anywhere is classified as `LLMTimeoutError` / `LLMAuthenticationError`, dropping `tool_names` and, for the timeout case, retrying a permanently invalid schema via `_is_retryable` at line 261.
3. Delete the unbalanced ` ```markdown ` fence at `docs/developer/error-flow-diagram.md:1` and its partner at line 8 — fence pairing across the file is shifted by one, so the new "Provider Rejections Caused by a Tool Definition" section and its attribution table (lines 147–165) and the new ordering-constraints prose (lines 182–207) render as literal monospaced text instead of Markdown.
4. Add regression tests for the two cases above: a tool named `calc` must not match "could not calculate length", and a tool-schema rejection for a tool named `network_timeout` must surface as a bad-request error carrying that tool name.

## Evidence
No evidence score was produced for this change. The change is small and self-contained, but the cases above are not covered by the added tests — the most useful additions would be a malformed tool-call-history case and a tool whose name contains a classification keyword.

## Notes
- The non-streaming wrapper at `atlas/application/chat/utilities/error_handler.py:182` rebuilds the bad-request error from its message alone, erasing the `tool_names` this PR adds; re-raise the classified error unchanged instead.

---

garland3:
See #735

PR #735, opened 2026-07-28
Name the rejected tool when a provider rejects the request

### LINKED ISSUE #728: Tool-caused provider rejections surface as a generic LLM error that does not name the tool

### LINKED ISSUE #729: Streaming tools error frame omits error_type

garland3:
Fixes #728. Fixes #729.

Same-repo mirror of #733 (fork PRs cannot get green CI here — the ghcr.io push and PR-comment steps fail on a missing token, not on the tests), with the review fixes applied on top.

A malformed tool definition makes the provider reject the whole request, so no tool ever runs and the turn fails. The provider names the offending function, but that detail was replaced by a generic message before it could reach the user:

> The LLM service encountered an error. Please try again or contact support if the issue persists.

The message points at the model, so the natural response is to switch models, which does not help. Users recover by deselecting tools one at a time.

## Cause

Two layers each generalized, and the first destroyed what the second needed.

`_raise_llm_domain_error` (`litellm_caller.py`) had no `BadRequestError` branch, so it fell through to a fixed `LLMServiceError` string. `classify_llm_error` (`error_handler.py`) then ran its keyword matching against `str(error)`, which by then was that fixed string rather than the provider's message. Nothing matched, so it produced a second generic message.

## Changes

- `LLMBadRequestError` domain error carrying `tool_names`.
- `_raise_llm_domain_error` maps `litellm.BadRequestError` to it, and takes an optional `tools_schema`. Placed after the context-window check, since `litellm.ContextWindowExceededError` subclasses `BadRequestError`.
- Attribution requires positive evidence, since most 400s have nothing to do with tools:

  | Provider text | Reported |
  | --- | --- |
  | names a tool from the request (whole-token match) | just that tool |
  | points at the tool payload (`tool`, `tools`, `function`, `tool_choice`, …) but names none | every tool in the request, to bisect |
  | anything else | no tool; a plain "the provider rejected this request" |

  Whole-token matching keeps a tool called `calc` from being blamed for provider text that merely contains `calculate`. The raw provider string is never exposed.
- `classify_llm_error` short-circuits on `LLMBadRequestError` and passes its message through.
- `LLMBadRequestError` added to the passthrough tuples in `call_with_rag()` and `call_with_rag_and_tools()`. Those fall back to a simpler call when RAG breaks; a provider rejection is not a RAG failure, and without the passthrough the same refused request went to the provider twice with RAG blamed in the logs.
- `error_type_for()` maps a domain error class to the `error_type` vocabulary used in `main.py`, and the tools-mode error frame now includes it. `main.py` gained an `LLMBadRequestError` branch so the non-streaming path reports `bad_request` too rather than falling through to `domain` — #729 now holds across both paths.

Existing classification behavior for rate limit, timeout, auth, context window and generic errors is unchanged.

## Test plan

- [x] `atlas/tests/test_tool_error_attribution.py` — 17 tests
- [x] Provider names a tool: only that tool is reported
- [x] Provider points at the tool payload but names none: all request tools listed
- [x] Provider rejects an out-of-range parameter or a malformed message with tools selected: no tool blamed
- [x] A tool name that is a prefix of an unrelated word in the provider text is not matched
- [x] `ContextWindowExceededError` still classifies as context window, not bad request
- [x] A tool named like another error category (`vault_api_key_lookup`) is not rerouted to the auth branch
- [x] `call_with_rag_and_tools` re-raises the rejection instead of retrying tools-only
- [x] Error frame names the tool and carries `error_type: "bad_request"`
- [x] Full backend suite: 1842 passed, 17 skipped, 0 failures
- [x] `ruff check` clean on all touched files (the pre-existing `I001` in `main.py` is unchanged)

Reproduced end to end with an MCP server advertising `inputSchema: {"type": "string"}` and a local OpenAI-compatible stub applying the documented `type: "object"` validation; scripted under `test/pr-validation/`.

## Not included

The underlying cause — MCP tool schemas are never validated at discovery — is #727 and needs its own fix. This change makes the failure legible, not impossible.


garland3:
## Summary
Adds an `LLMBadRequestError` domain error carrying the tool names implicated in a provider rejection, maps `litellm.BadRequestError` to it after the context-window check, and short-circuits classification so the specific message reaches the user instead of a generic service error. Also adds `error_type_for()` so the streaming tools path and the non-streaming path in `main.py` both report `bad_request`.

## Verdict
Looks good.

## Evidence
Evidence is proportionate to the change: `atlas/tests/test_tool_error_attribution.py` adds 17 tests covering named-tool, payload-marker, no-attribution, whole-token-matching, and context-window-precedence cases, all passing, plus a scripted end-to-end reproduction under `test/pr-validation/` using a malformed MCP schema against a strict provider stub. No missing artifact of note.

## Notes
- `_ERROR_TYPE_BY_CLASS` in `error_handler.py:73` is keyed by exact class, so any future subclass of the listed errors falls through to `"unexpected"` rather than inheriting its parent's `error_type`.
- Attribution ordering at `litellm_caller.py:200` depends on `ContextWindowExceededError` being checked before the `BadRequestError` branch; the comment records this, but it is a load-bearing ordering worth a regression test guard if the block is ever reshuffled.

---

garland3:
## Summary
Adds an `LLMBadRequestError` domain error carrying the tool names implicated in a provider rejection, maps `litellm.BadRequestError` to it after the context-window check, and short-circuits downstream classification so the specific message reaches the user instead of a generic service error. Attribution requires positive evidence — a whole-token tool-name match, or a tool-payload marker with no name — so unrelated 400s blame no tool.

## Verdict
Request changes.

## Required changes
1. Move the `isinstance(exc, litellm.BadRequestError)` branch at `atlas/modules/llm/litellm_caller.py:206` ahead of the message-keyword fallbacks at `:179-204` (keep the `ContextWindowExceededError` subtype check first) — a rejection whose provider text contains `timeout`, an authentication term, or a context-window keyword matches those `in error_str.lower()` tests first and is raised as the wrong domain error, so it never reaches the new attribution branch and the user is told to retry a request that will never succeed. A malformed schema for a tool named `timeout_checker` is exactly the case this PR exists to fix.
2. Add an explicit non-retryable branch for `litellm.BadRequestError` in `_is_retryable_error` at `atlas/modules/llm/litellm_caller.py:236` — it has no bad-request case, and the `"timeout"`/`"timed out"` substring test at `:259-263` marks such a permanent 400 as transient, so it is retried three times with backoff before failing.
3. Add a regression test covering a `BadRequestError` whose provider text contains a classification keyword (e.g. a tool or property named `*timeout*` or `*api_key*`) — the existing collision test at `atlas/tests/test_tool_error_attribution.py:157` covers `classify_llm_error`, but the ordering gap is one layer up in `_raise_llm_domain_error`, which is untested for this case.

## Evidence
The change is well demonstrated for its size: `atlas/tests/test_tool_error_attribution.py` adds 17 tests spanning named-tool, payload-marker, no-attribution, whole-token-matching and context-window-precedence cases, plus a scripted end-to-end reproduction under `test/pr-validation/` using a malformed MCP schema against a strict provider stub. The one gap is the keyword-collision case at the raise layer, listed above.

## Notes
- `safe_call_llm_with_tools` at `atlas/application/chat/utilities/error_handler.py:182` rebuilds the exception as `error_class(user_msg)`, dropping `tool_names` on the non-streaming path; the user-facing text survives and no current consumer reads the field there, but re-raising the already-classified error would preserve it.
- `docs/developer/error-flow-diagram.md` has unbalanced code fences, so prose and tables render as code blocks. This predates the change — the document-wide fence at line 1 is untouched here — but the new section at `:145-180` inherits it.

---

garland3:
## Summary
Adds a `LLMBadRequestError` domain error so provider 400s caused by a malformed tool schema are attributed to the offending tool instead of surfacing as a generic service fault, plus an `error_type` map for the error frame, RAG-fallback passthrough, unit tests, and an end-to-end reproduction harness.

## Verdict
Request changes.

## Required changes

1. Move the `isinstance(exc, litellm.BadRequestError)` branch at `atlas/modules/llm/litellm_caller.py:216` ahead of the message-substring fallbacks at `:189-211` (keeping the `ContextWindowExceededError` subtype check first) — today a rejection naming a tool such as `request_timeout_probe` matches `"timeout" in error_str.lower()` first and is raised as `LLMTimeoutError`, so `tool_names` is dropped and the user is told to retry a request that can never succeed.
2. Return `False` for `litellm.BadRequestError` in `_is_retryable_error` at `atlas/modules/llm/litellm_caller.py:246`, before the transient-keyword tests — a permanent 400 whose text contains `timeout`/`429`/`503` is currently retried three times with backoff, costing ~7s and three full-payload round-trips on a deterministic failure.
3. Narrow the tool-payload evidence in `TOOL_REJECTION_MARKERS` at `atlas/modules/llm/litellm_caller.py:78` so it does not fire on message-history rejections — a 400 reading `An assistant message with 'tool_calls' must be followed by tool messages...` matches `tool_calls` and blames every selected tool for a fault that disabling tools cannot fix.
4. Add a raise-layer regression test in `atlas/tests/test_tool_error_attribution.py` asserting `LLMBadRequestError` is raised (with `tool_names` populated) for provider text containing `timeout` / `api_key` / a context-window phrase, plus a case asserting `_is_retryable_error` is `False` for such a rejection — the existing `test_tool_name_containing_a_keyword_is_not_misclassified` at `:157` exercises `classify_llm_error` one layer below the gap and passes while the raise layer misclassifies.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size: 17 targeted unit tests, a runnable end-to-end reproduction under `test/pr-validation/` with a control turn, and a written causal account. The main gap is that the claimed full-suite and validation-script results are asserted in checkboxes rather than shown.

## Notes
- The message names the qualified tool id (`badschema_lookup`), which the tools panel neither displays nor matches in its search filter — `litellm_caller.py:221`, `ToolsPanel.jsx:305-328`.
- "Please try again" is wrong advice for a deterministic 400 and contradicts the comment three lines above it — `litellm_caller.py:231-234`.
- A stray document-wide ```` ```markdown ```` fence at `docs/developer/error-flow-diagram.md:1` makes this PR's new prose and attribution table render inside a code block; deleting line 1 and its close at `:208` repairs the pairing.

---

garland3:
## Summary
When a provider rejects a tools request with a 400, this change attributes the failure to the specific selected tool named in the rejection text (falling back to enumerating the selection when no name appears), raises a dedicated `LLMBadRequestError` carrying `tool_names`, and surfaces an `error_type: "bad_request"` on the error frame across the streaming and WebSocket paths.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — 438 lines of targeted unit tests plus a runnable end-to-end reproduction with a stub provider and a deliberately malformed MCP tool schema. The main missing artifact is pasted output: the test-plan claims ("1842 passed", "17 tests", script pass line) are checkboxes rather than shown runs, and no screenshot shows how the new message actually renders in the chat UI.

## Notes
- Message-history rejections that happen to contain a selected tool's name still blame that tool — the `MESSAGE_HISTORY_MARKERS` guard only covers the second attribution level, not the whole-token name match that returns first (`atlas/modules/llm/litellm_caller.py:176`). The user is told to disable a working tool and the next turn fails identically.
- `_ERROR_TYPE_BY_CLASS` omits `AuthorizationError` and `DomainError`, so `error_type_for` returns `"unexpected"` where `atlas/main.py:683`/`:706` send `"authorization"`/`"domain"` for the same failure (`atlas/application/chat/utilities/error_handler.py:72`); a test pinning the two vocabularies together would stop further drift.
- The message names the qualified id (`badschema_lookup`), which the Tools panel neither displays nor matches in its search box (`frontend/src/components/ToolsPanel.jsx:316`) — consider making the qualified key searchable, or splitting server and tool name in the copy. Also cap the untargeted enumeration at `litellm_caller.py:246`; large selections produce a multi-kilobyte wall of names.

---

Issue #182, opened 2025-12-17
MCP logging should be shown to the user in the chat area


PoldervaartS:
MCP Tool messages (context.log, context.info, context.error, context.debug) should also be shown to the user inside of the ChatArea respecting the env log level. Currently they are only viewable in the `admin/logs` page

Acceptance criteria:

- [ ] use the current `log_handler` to also forwards the MCP tool logs to the chatArea visible to the user. With a similar badge-styling as in the logs as a message from the tool. 
- [ ] MCP logs are added into the messages context for the LLM to use.
- [ ] Test to make sure that a tool (example the `logging_demo` tool) when it provides a `ctx.info` or otherwise is placed into the LLM context for messaging

PR #739, opened 2026-08-04
fix(setup): validate MCP token key at startup, document mcp-demos extra (#732)

### LINKED ISSUE #732: Documented dev setup on a fresh clone yields a 500 on /api/config and a broken pptx_generator
## Summary

Following the documented development setup on a fresh clone produces an app whose `/api/config` returns HTTP 500, and whose pptx_generator tool is missing. Two separate causes.

## 1. Placeholder encryption key fails inside a request handler, not at startup

`.env.example:235` ships:

```
MCP_TOKEN_ENCRYPTION_KEY=your-random-string-at-least-32-chars
```

Copying it to `.env` as documented and starting the backend gives a server that appears healthy — `/api/health` and `/api/config/shell` both return 200 — but `/api/config` returns 500:

```
RuntimeError: MCP_TOKEN_ENCRYPTION_KEY is not set (or is still the shipped placeholder).
  atlas/routes/config_routes.py:421 → atlas/modules/mcp_tools/token_storage.py:189
```

`get_token_storage()` is called lazily inside `get_config`, so the guard fires per-request. The frontend receives a bare 500 with no indication of the cause. The guard itself is correct; the timing and the lack of a startup signal are the problem.

Suggested fix: validate the key during startup so the server refuses to boot with a clear message, or generate a unique key in `atlas-init`. At minimum, note in `.env.example` that this value must be replaced before the app will work.

## 2. Documented dev install omits the extra that pptx_generator needs

`AGENTS.md` documents:

```
uv pip install -e ".[dev]"
```

With that install, pptx_generator crashes on startup:

```
File "atlas/mcp/pptx_generator/main.py", line 28, in <module>
    from pptx import Presentation
ModuleNotFoundError: No module named 'pptx'
```

Discovery reports only the downstream symptom:

```
ERROR | mcp_discovery | TOOL DISCOVERY FAILED for 'pptx_generator': McpError: Connection closed
```

`AGENTS.md` mentions elsewhere that MCP demo packages live in the `mcp-demos` extra, but the setup instructions do not include it. `uv pip install -e ".[dev,mcp-demos]"` resolves it and pptx_generator then discovers normally.

Suggested fix: update the documented dev install command, or note that the bundled demo MCP servers require the extra.

## Related

A server that fails discovery stays listed in `authorized_servers` in `/api/config` while being absent from `tools`, so the failure is invisible in the UI. That asymmetry overlaps #182.

garland3:
Fixes #732. Both reported problems reproduce; investigation notes below.

## 1. `MCP_TOKEN_ENCRYPTION_KEY` failed per-request instead of at startup

`get_token_storage()` is called lazily inside handlers (`atlas/routes/config_routes.py:420`, plus the MCP/LLM auth routes). Grepping every non-test caller confirmed there was **no** startup validation anywhere — `settings.py` defaults the key to `None`. So a server with no key, or with the `.env.example` placeholder (which is in the rejected `_PLACEHOLDER_ENCRYPTION_KEYS` set), booted looking healthy and then failed every `/api/config` with an unhandled `RuntimeError` → bare 500.

That also made the docs wrong: `docs/admin/mcp-server-authentication.md`, `docs/admin/mcp-servers.md`, `.env.example` and `CHANGELOG.md` all say "Atlas refuses to start without it." Now it does.

- Extracted `resolve_encryption_key()` in `token_storage.py`; `MCPTokenStorage.__init__` and the new lifespan check share it, so the placeholder list cannot drift between them.
- The lifespan check runs **before** MCP initialization, so the existing `try/except` around MCP startup cannot swallow it (pinned by a test).
- The error message now includes the `secrets.token_urlsafe(32)` one-liner.

**Additional finding not in the issue:** `atlas-init` wrote a minimal `.env` that omitted `MCP_TOKEN_ENCRYPTION_KEY` entirely, so the `atlas-init` onboarding path hit the same failure with `key_source = None`. It now generates a unique per-install key.

`.env.example` and `docs/getting-started/installation.md` now state that the shipped placeholder is public, must be replaced, and how to generate a replacement. (The placeholder value itself is unchanged — an inline `# REPLACE ME` comment would risk dotenv folding it into the value and defeating the guard.)

## 2. Documented dev install omits the `mcp-demos` extra

`atlas/mcp/pptx_generator/main.py` imports `PIL` and `pptx` at module scope. `python-pptx`/`Pillow` are only in the `pptx` and `mcp-demos` extras; `dev` has just alembic/podman-compose/pytest*/ruff. `pptx_generator` is enabled by default in both `config/mcp.json` and `atlas/config/mcp.json`, so the documented install always failed its discovery with `Connection closed`.

`AGENTS.md`, `README.md`, `docs/getting-started/installation.md` and `docs/getting-started/python-package.md` now document `uv pip install -e ".[dev,mcp-demos]"` and say why. CI (`security.yml`) and the archived docs are left on `.[dev]` deliberately — they don't run the demo servers.

## Testing

- New `atlas/tests/test_startup_encryption_key_validation.py`: `resolve_encryption_key()` accept/reject cases, lifespan aborts on missing and placeholder keys, and the check runs before any MCP manager access.
- New cases in `atlas/tests/test_init_cli.py`: the generated `.env` sets exactly one key, it passes the server's own guard, and it differs between installs.
- Full backend suite: 1836 passed, 17 skipped.

## Not addressed

The issue's "Related" note — a server that fails discovery stays in `authorized_servers` while absent from `tools` — is left to #182.


garland3:
## Summary
Makes `MCP_TOKEN_ENCRYPTION_KEY` a hard startup requirement — a new lifespan check rejects a missing key or the shipped `.env.example` placeholder via an extracted `resolve_encryption_key()` — and has `atlas-init --minimal` generate one. Also documents `.[dev,mcp-demos]` as the install extra needed for the bundled demo MCP servers.

## Verdict
Request changes.

## Required changes
1. Update every Docker snippet in `docs/getting-started/installation.md:12` (also :29, :41) and the Docker references in `README.md` to pass `MCP_TOKEN_ENCRYPTION_KEY` — the `Dockerfile` sets no default and `settings.py:217` defaults to `None`, so the documented `docker run -p 8000:8000 ...` quick start now exits at startup instead of serving the UI.
2. Generate a key on the default `atlas-init` path too, at `atlas/init_cli.py:222` — the non-`--minimal` branch copies `.env.example` verbatim, including the exact value `_PLACEHOLDER_ENCRYPTION_KEYS` rejects, so the documented `atlas-init && atlas-server` flow produces an install that refuses to boot. Extract a shared `_ensure_encryption_key(env_path)` used by both branches so they cannot drift.
3. Add `MCP_TOKEN_ENCRYPTION_KEY` to the Configuration Reference table and Step 1 of `docs/getting-started/python-package.md:406` — `pip install atlas-chat` users following that guide hit a startup abort the guide gives no way to anticipate.
4. Update the packaged `atlas/.env.example:232` to match the repo-root copy — `get_env_example_path()` prefers the packaged file, so installed-package users still get the pre-change guidance without the "you MUST replace this" warning; add a test asserting the two copies agree.
5. Create the generated-secret `.env` with mode `0600` at `atlas/init_cli.py:136` — `Path.write_text` yields `0644` under the common umask, exposing the key that decrypts every user's stored MCP tokens on the shared-install host the module docstring targets.
6. Reconcile `docker-compose.yml:120` with the new docs — it ships `dev-only-mcp-token-encryption-key-change-me`, an equally repo-public value that passes the guard, contradicting the newly documented rule that public placeholders are rejected. Either add it to `_PLACEHOLDER_ENCRYPTION_KEYS` and require the variable in compose, or narrow the docs wording.
7. Enforce the 32-character minimum in `resolve_encryption_key()` at `atlas/modules/mcp_tools/token_storage.py:77`, or soften the four surfaces that promise it — today `MCP_TOKEN_ENCRYPTION_KEY=x` passes the gate and is fed to PBKDF2 under a hardcoded constant salt. Check `test/e2e_tests.sh:60` and `mocks/wormhole-mcp-mock/run_e2e.sh:12` before choosing a hard floor.

## Evidence
Evidence: 8/10 — the encryption-key half is well covered by targeted tests (parametrized placeholder rejection, a lifespan ordering guard, and an assertion that the `atlas-init`-generated key passes the server's own check) and a full-suite result. The most useful missing artifact is captured output of a real `atlas-server` aborting on the placeholder, and a before/after MCP discovery log for `pptx_generator` under each install command — the `mcp-demos` half currently rests entirely on a code-reading argument.

## Notes
- Re-running `atlas-init --minimal --force` mints a new key and silently orphans every stored MCP token — carry the existing value forward or warn (`atlas/init_cli.py:95`).
- `atlas/init_cli.py` and `docs/getting-started/python-package.md` are whole-file CRLF→LF rewrites, so ~10 real lines are unreadable in the diff and blame is reset; land a `.gitattributes` so it cannot recur.
- No lifespan test runs with a *valid* key, which leaves `test_validation_runs_before_mcp_initialization` indistinguishable from the plain abort test (`atlas/tests/test_startup_encryption_key_validation.py:83`).

---

garland3:
Addressed the review. Merged `main` (the PR was `DIRTY`) and pushed 9d89d43.

**Implemented**

1. **Docker quick starts** (Codex P1, review item 1) — `docs/getting-started/installation.md` now generates a key up front and passes `-e MCP_TOKEN_ENCRYPTION_KEY` in all three `docker run` snippets, plus a new Compose section. `README.md` contains no `docker run` commands, so nothing to change there.
2. **Default `atlas-init` path** — the non-`--minimal` branch copied `.env.example` verbatim, placeholder included, so `atlas-init && atlas-server` produced an install that refused to boot. Both branches now share one key resolved in `run_init()` and applied via `_ensure_encryption_key()`.
3. **`python-package.md`** — the variable is documented in Step 1, in the "Option B: use a .env file" snippet, and in the Configuration Reference table.
5. **`0600`** — new `_write_env_securely()` opens with mode `0600` before the secret is written (and `chmod`s, since `O_CREAT` keeps an existing file's mode). Covered for both branches.
6. **`docker-compose.yml`** — took the stricter option: `dev-only-mcp-token-encryption-key-change-me` is now in `_PLACEHOLDER_ENCRYPTION_KEYS`, and compose requires the variable via `${MCP_TOKEN_ENCRYPTION_KEY:?...}`.
7. **32-character floor** — enforced in `resolve_encryption_key()`. Both e2e keys (`test/e2e_tests.sh:60`, `mocks/wormhole-mcp-mock/run_e2e.sh:12`) are already 43–46 chars, and `token_urlsafe(32)` yields 43, so the floor is safe. Three fixtures in `test_token_storage.py` used sub-32 values and were lengthened.

Notes: re-running `atlas-init` now carries a usable existing key forward and says so, rather than minting a new one and orphaning stored tokens. Added a valid-key lifespan test that asserts startup reaches `get_mcp_manager()`, so `test_validation_runs_before_mcp_initialization` has a positive control. CHANGELOG condensed to the two-line convention (the `PR #739` header was already correct — the `(#732)` in the title is the issue).

**Not implemented, with reasons**

- **Item 4 (`atlas/.env.example`)** — that file is not in the repo. It is gitignored (`.gitignore:7`) and generated at build time by `setup.py`, which copies the root `.env.example` into `atlas/`. There is no second committed copy that can drift, so there is nothing to update and no meaningful sync test to add. The path you read was a stale local build artifact.
- **`.gitattributes` for the CRLF rewrites** — agreed on the diagnosis, but `git ls-files --eol` reports 172 tracked files with CRLF in the index. Landing `.gitattributes` here would renormalize all of them and bury this change. Worth its own PR.

Full backend suite: 1929 passed, 17 skipped. `ruff` clean on every touched file.

garland3:
## Summary
Enforces `MCP_TOKEN_ENCRYPTION_KEY` at startup (non-empty, ≥32 characters, not a repo-committed placeholder), teaches `atlas-init` to generate/carry forward a valid key and write `.env` at mode 0600, removes the committed default from `docker-compose.yml` in favor of `${VAR:?}`, and updates the getting-started and Docker docs accordingly.

## Verdict
Request changes.

## Required changes
1. Set the mode on the descriptor before the first write in `_write_env_securely` at `atlas/init_cli.py:133` — call `os.fchmod(fd, 0o600)` right after `os.open` and keep the trailing `os.chmod` only as a fallback. The `0o600` argument to `os.open` applies only when the file is created, so on the default non-minimal path (`copy_with_prompt` → `shutil.copy2` reproduces `.env.example`'s 0644) and on every re-run the generated key is written into a world-readable file first. This is the guaranteed sequence, not a rare race.
2. Reconcile the key when the overwrite prompt is declined at `atlas/init_cli.py:373` (same gap via `create_minimal_env` at `:295`) — `_ensure_encryption_key` runs only inside `if copy_with_prompt(...)`, and `prompt_yes_no` defaults to no, so the plain interactive `atlas-init` over an existing `.env` leaves the rejected placeholder in place, prints "Setup complete!", and exits 0 while `atlas-server` still refuses to start. At minimum print the generated key and the variable name so the operator has a copy-paste fix.
3. Reject an explicitly-passed empty/whitespace key in `resolve_encryption_key` at `atlas/modules/mcp_tools/token_storage.py:85` instead of falling back to `AppSettings` — a `.env` line `MCP_TOKEN_ENCRYPTION_KEY=` makes `_read_encryption_key` resolve against the ambient process environment, so `atlas-init --target ./demo` run by an operator with the production key exported copies that secret into the target `.env` and reports it was already there.
4. Narrow the `except Exception: return None` at `atlas/init_cli.py:120` to the error `resolve_encryption_key` actually raises, and print the reason before regenerating — today an `ImportError` or any unrelated fault is indistinguishable from "key unusable", so a valid key is overwritten and every stored MCP token becomes undecryptable.
5. Distinguish "no key present" from "key present but rejected" in `run_init` (`atlas/init_cli.py:120`) — installs predating this PR with a sub-32-character key are silently replaced, permanently orphaning `mcp_tokens.enc`. Print the specific reason and state explicitly that stored MCP tokens will no longer be readable before replacing.
6. Replace `python -c` with `python3` in the key-generation one-liner across `docs/getting-started/installation.md:18`, `:67`, `:128`, `docker-compose.yml:122`, `.env.example:246` and `docs/getting-started/python-package.md:39` — `export KEY=$(python ...)` exits 0 and sets an empty string where only `python3` is on PATH, producing exactly the startup abort the section exists to prevent. Add a length check (`echo "${#MCP_TOKEN_ENCRYPTION_KEY}"`) after it.
7. Surface `_ensure_encryption_key`'s `OSError` path at `atlas/init_cli.py:151` — it returns silently, after which `run_init` prints "Created", "Setup complete!" and exits 0 over a `.env` that was never written. Print a warning and return non-zero.

## Evidence
Evidence: 8/10. The encryption-key half is well demonstrated — new startup-validation tests parametrized off the real placeholder set, boundary tests at the 32-character floor, subprocess-level `atlas-init` tests, and full-suite results at both revisions. Missing pieces worth adding: captured terminal output of a real `atlas-server` aborting on the placeholder (nothing exercises a real server process), `docker compose config` output with and without the variable set, and any runtime evidence for the `mcp-demos` half, which is argued from code reading only.

## Notes
- `_read_encryption_key` and `_ensure_encryption_key` stop at the *first* `MCP_TOKEN_ENCRYPTION_KEY=` line while the runtime loader honors the *last*, so a duplicated assignment lets `atlas-init` rewrite one value and the server boot on another (`atlas/init_cli.py:112`).
- `os.open` on `.env` follows symlinks — add `O_NOFOLLOW` so a pre-planted link cannot redirect the key write and the subsequent chmod (`atlas/init_cli.py:133`).
- Nothing pins `_PLACEHOLDER_ENCRYPTION_KEYS` to the values actually committed in `.env.example`; the tests parametrize over the frozenset, so a reworded placeholder silently stops being rejected (`atlas/modules/mcp_tools/token_storage.py:48`).

---

PR #740, opened 2026-08-04
Stop blocking tokenizer download from stalling the event loop; handle mid-turn websocket disconnects

garland3:
## Problem

Production logs showed repeated `Update callback failed: Cannot call "send" once a close message has been sent.` followed by an error during file ingest, around the same time a pod died and was restarted.

Tracing it back, the trigger is a **blocking network call on the async event loop**:

1. Atlas never sets `stream_options={"include_usage": True}`, so streamed responses carry no usage block.
2. At end-of-stream, LiteLLM reconstructs token counts locally — `CustomStreamWrapper.__anext__` calls `stream_chunk_builder` **synchronously, on the event loop** (`litellm_core_utils/streaming_handler.py:1980`).
3. That reaches `_select_tokenizer_helper`, and for any model name containing `llama-3` it calls `Tokenizer.from_pretrained("Xenova/llama-3-tokenizer")` (`litellm/utils.py:1985`) — a blocking `huggingface_hub` download.
4. In a network-restricted deployment this fails with `[Errno 101] Network is unreachable`, but only after retry/backoff (`HF_HUB_ETAG_TIMEOUT` defaults to 10s).

Atlas runs one process, one event loop (`workers=1`). While that call blocks, **nothing else runs** — not other users' streams, not the WebSocket keepalive timers (`ws_ping_timeout` defaults to 30s), and not `/api/health` or `/api/heartbeat`, which are ordinary `async def` handlers on the same loop. A liveness probe that can't get a response is a plausible explanation for the pod restart, so this is likely more than log noise.

Note this is the **LiteLLM Python package running in-process**, not a LiteLLM proxy server — the tokenizer call is client-side by construction and can't be fixed by upstream config.

## Fixes

**1. Don't make the call at all.** `litellm.disable_hf_tokenizer_download = True`. LiteLLM already catches the download failure and falls back to the bundled tiktoken tokenizer, so this deployment is *already* using that fallback — this just skips straight to it instead of blocking the loop to rediscover it every process start. Token counts are unchanged.

**2. Survive a disconnect cleanly.** Two latent defects turned the keepalive close into a worse incident. These will be exercised by any disconnect — tab close, proxy timeout, network blip — so they're worth fixing regardless of the trigger:

- **Both** send paths sent unconditionally: `websocket_update_callback` and `WebSocketConnectionAdapter.send_json` (the one the event publisher uses). Producers deep in the chat pipeline raised once the client was gone — logging spurious errors and aborting `handle_session_files` partway through, silently dropping attachments. Both now drop the message when the socket isn't connected, and tolerate the close racing the send.
- **The background chat task was never cancelled on disconnect.** It kept streaming tokens and holding MCP sessions against a session `end_session()` had already torn down. Cleanup now cancels the turn first. Extracted to `cleanup_disconnected_session()` so the ordering is directly testable. A failed error-frame send can no longer escape the task as an unretrieved exception either.

## Testing

- `atlas/tests/test_websocket_disconnect_resilience.py` — 15 tests covering both send paths, the disconnect cleanup and its ordering, and the LiteLLM setting. Each behavioral test was confirmed to **fail** with the corresponding fix reverted.
- `test/pr-validation/test_pr740_websocket_disconnect_no_eventloop_stall.sh` — real uvicorn server, real websocket client, real streaming turn against the mock gateway with the client hanging up mid-turn. Asserts the server stays healthy, the turn is cancelled, and no send-after-close or leaked task appears in the log. Also asserts the real `_select_tokenizer_helper` resolves a llama-3 name to the local tokenizer fast enough to prove no network attempt.
- Full backend suite: 1840 passed, 17 skipped.
- `ruff check` clean on all changed files (`atlas/main.py` has one pre-existing `I001` that also fails on `main`; left alone to keep this diff focused).

The e2e script earned its keep — it caught the `WebSocketConnectionAdapter` send path, which the unit tests alone had missed.

## Follow-up worth considering (not in this PR)

Passing `stream_options={"include_usage": True}` for providers that support it would let LiteLLM use provider-reported usage and skip local token counting entirely. Bigger change; needs per-provider capability handling.

## Caveat

I could not measure the actual stall duration against the 30s ping timeout in the production environment — `Errno 101` can fail fast when there's no route. The strongest evidence the stall crossed the threshold is that the first error is `Cannot call "send" once a close message has been sent`, meaning the **server** initiated the close (keepalive timeout), not the client. If you can get the wall-clock gap between the first HF retry log line and the first `event_notifier` warning, that would confirm it directly.


garland3:
## Summary
Disables LiteLLM's blocking HuggingFace tokenizer download so it can no longer stall the single event loop, and hardens mid-turn websocket disconnects by making both send paths close-aware and cancelling the background chat task during cleanup.

## Scope
This pass covered correctness only. Security, architecture, testing, performance, and UX were not assessed on this revision.

## Verdict
Request changes.

## Required changes
1. Route the remaining direct `websocket.send_json()` calls through the close-aware send helper and move teardown into a `finally` block at `atlas/main.py:883` — cleanup currently runs only for `WebSocketDisconnect`, so a send-after-close `RuntimeError` in the download, restore, reset, or attachment branches skips task cancellation, MCP session release, and `end_session()`.
2. Await the cancelled task before releasing resources in `cleanup_disconnected_session()` at `atlas/main.py:149` — `task.cancel()` only requests cancellation, so teardown can finish before the task observes `CancelledError`; suppress the expected cancellation and consider a bounded timeout.
3. Enforce one active turn per connection at `atlas/main.py:725` — each chat frame overwrites `active_chat_task["task"]` without cancelling the previous one, so if two frames arrive in quick succession the earlier task keeps running against a closed connection and a torn-down session.

## Evidence
Evidence was not scored on this revision. From the change itself, the demonstration is strong for its size: 15 targeted unit tests, each confirmed to fail with its fix reverted, plus an end-to-end script that exercises a real uvicorn server and a real mid-turn hangup. The gap is coverage for the failure modes above — an endpoint-level test where a direct response send raises the send-after-close error, an assertion that cancellation *completes* before teardown rather than merely being requested, and a two-concurrent-chat-task case.

---

Issue #438, opened 2026-03-18
Add ability to run Agent Skills


ktpedre:
ATLAS currently lacks the ability to add agent skills. Add capabilities similar to those available in Claude Code for adding agent skills to ATLAS and allowing ATLAS to run them when in agent mode. The Agent Skills specification is available at https://agentskills.io/specification

An example of an agent skill for querying a database is at: https://github.com/K-Dense-AI/claude-scientific-skills/tree/main/scientific-skills/pubmed-database

garland3:
What scope is in mind here — full compliance with the agentskills.io spec, or just enough to run skills shaped like the PubMed example? And how should skill execution be sandboxed, given skills can ship arbitrary code?

---
_Posted automatically by **AGENT-REVIEW-BOT-3**_

garland3:
Proposing we split this into two independently shippable halves, which also answers the sandboxing question above:

1. **Discovery + injection** — find skills on disk, inject their `name`/`description` into the system prompt, let the model read a skill body on demand. Executes nothing, so no sandbox needed.
2. **Execution of bundled scripts** — the part of the agentskills.io spec that ships arbitrary code. Needs a real sandboxing story and should be its own issue.

Everything below is scope for (1). It's useful on its own: a skill that is purely instructions (most of them) works with just this.

## Behavior

- **Feature flagged**, default off. New setting, e.g. `FEATURE_SKILLS_ENABLED`.
- **Auto-injection when the flag is on.** The skill index (name + one-line description per skill) is appended to the system prompt automatically. No user selection step — the model sees what's available and decides whether to load a body. This is the progressive-disclosure model from the spec: descriptions are always resident, bodies are not.
- **Index refreshed per session.** Rescan the directories when a new session starts, so dropping a new skill folder in doesn't require a server restart. Not a filesystem watcher — just no permanent process-lifetime cache.
- **Configurable directories, layered.** Global / user / project tiers, mirroring how `ConfigManager._search_paths` and `PromptProvider._resolve_base_paths` already do user-dir-beats-packaged-default. Difference: skills should **merge across tiers**, with override-by-name, not first-path-wins — a project skill should add to the global set, and shadow a global skill only when it reuses the name.

## Implementation notes

- Injection point is `atlas/application/chat/preprocessors/message_builder.py:126-139`, which builds the system message per turn.
[...]

PR #743, opened 2026-08-06
Add Agent Skills discovery and system prompt injection

garland3:
Implements the **discovery and injection** half of #438. Atlas can now find Agent Skills on disk and tell the model they exist; the model pulls a skill's full instructions on demand.

Executing skill-bundled scripts is deliberately **not** in this PR — see [Scope](#scope) below.

## What this does

With `FEATURE_SKILLS_ENABLED=true`, Atlas scans for skill directories, validates each `SKILL.md` against the [Agent Skills specification](https://agentskills.io/specification), and appends a compact index of skill names and descriptions to the system prompt:

```
# Available Skills
- **skill-authoring**: Write, validate, and install Agent Skills for Atlas. Use when...
```

That index is the only thing always in context (~100 tokens/skill). When the model decides a skill applies, it calls `read_skill` on the bundled skills MCP server to get the full body — the progressive-disclosure model from the spec.

## Where skills are found

Layered roots, lowest precedence first. Roots **merge**; a later root overrides only same-named skills.

| Precedence | Root |
| --- | --- |
| 1 | `atlas/config/skills/` (packaged) |
| 2 | `~/.atlas/skills/`, `~/.claude/skills/`, `~/.agents/skills/` |
| 3 | `.atlas/skills/`, `.claude/skills/`, `.agents/skills/` under the project root |

Reading the `.claude/` and `.agents/` locations is deliberate: it matches the conventions Claude Code and other agents-compatible tools already use, so a skill authored for one works in Atlas without being copied. Verified against a real `~/.claude/skills/` during development.

`SKILLS_PATHS` replaces the default list entirely.

## Design decisions worth reviewing

**Injection is additive, not a replacement.** `custom_system_prompt` *replaces* the system prompt (`message_builder.py:127`). The skill index is appended to whichever prompt wins, so selecting a custom prompt does not silently drop every skill.

**Appended after `str.format()`.** A skill description containing `{` would raise if it reached the template formatter. There is a regression test for this.

**Its own feature flag.** Not gated on `custom_prompts_effective`, which requires chat history — skills have no such dependency.

**Rescan per session, not per process.** `PromptProvider` caches templates for the process lifetime with no invalidation. Skills instead invalidate on `create_session`, so dropping a skill on disk takes effect in the next chat rather than needing a restart. No filesystem watcher.

**Invalid skills are skipped, not fatal.** A malformed `SKILL.md` logs `Skipping invalid skill` with the reason and the scan continues.

## Scope

In scope (this PR):
- Discovery, spec validation, layered precedence
- System prompt index injection behind a flag
- Read-only MCP server for skill bodies and bundled `references/` files

Out of scope (needs its own issue):
- **Executing skill `scripts/`.** This is the part of the spec that ships arbitrary code and needs a sandboxing story. Nothing here executes anything; `scripts/` files are discoverable and readable only.
- **Per-group authorization.** Every discovered skill is offered to every user. MCP tools here are group-aware and skills likely should be too. The spec's `metadata` field gives a forward-compatible place to put this later.
- Frontend UI for browsing or toggling skills.

## Security notes

The MCP server only reads files. `read_skill_resource` resolves the requested path and rejects anything landing outside the skill's own directory (covers `..` and symlinks), and refuses files over 256 KB. Both are tested.

## Enabling

Off by default. Two switches:

```bash
FEATURE_SKILLS_ENABLED=true          # index injection
```
plus merging `atlas/config/mcp-example-configs/mcp-skills.json` into `config/mcp.json` so the model can read skill bodies. Injection alone tells the model skills exist without letting it read them, which is not useful on its own — the docs call this out.

## Testing

33 new tests in `atlas/tests/test_agent_skills.py` covering frontmatter validation (including each rejection case from the spec), discovery precedence and merge, flag gating, cache invalidation on new session, index rendering, injection additivity, the brace regression, and the MCP server's path-traversal guard.

Full backend suite: **1873 passed, 17 skipped**.

## Docs

- `docs/user-guide/skills/README.md` — user-facing guide, including limitations
- `atlas/mcp/skills/README.md` — server reference
- `atlas/config/skills/skill-authoring/` — a packaged example skill that documents the format

Closes the discovery half of #438.


garland3:

## Summary
Adds an Agent Skills feature: a layered filesystem discovery/validation registry (`atlas/modules/skills`), injection of a rendered skill index into the chat system prompt, and a read-only stdio MCP server exposing `list_skills` / `read_skill` / `read_skill_resource`, behind `FEATURE_SKILLS_ENABLED`.

## Verdict
Request changes.

## Required changes
1. Normalize whitespace in `_validate_description` at `atlas/modules/skills/registry.py:96` (e.g. `" ".join(raw.split())`) — a YAML block-scalar description keeps its newlines, and `render_index` at `registry.py:303` emits it inside a `- **name**: ...` list item, so everything after the first newline lands in the system prompt as top-level content and can forge entries for skills that do not exist.
2. Move the skill scan off the event loop at `atlas/application/chat/preprocessors/message_builder.py:105` — `_render_skill_index` → `discover()` does synchronous `iterdir`, `read_text`, and `yaml.safe_load` over up to nine roots inside `async build_messages`, stalling all concurrent streams; combine with a TTL or background refresh so the process-wide `invalidate()` at `atlas/application/chat/service.py:246` does not force that blocking rescan on nearly every turn of a busy instance.
3. Bound the `SKILL.md` read at `atlas/mcp/skills/main.py:88` and `atlas/modules/skills/registry.py:134` with the same `MAX_RESOURCE_BYTES` cap `read_skill_resource` already applies — an oversized or generated `SKILL.md` is currently read whole on every scan and returned to the model in full.
4. Compare resolved paths instead of basenames at `atlas/mcp/skills/main.py:96` (skip only `path == skill.path`) — today every nested file named `SKILL.md` is dropped from `resources`, which silently hides the bundled examples in the packaged `skill-authoring` skill.
5. Guard the `Path.home()` call in `_default_roots` at `atlas/modules/skills/registry.py:194` — it raises `RuntimeError`, not `OSError`, when a container runs under a UID with no passwd entry and no `HOME`, which aborts the whole scan and escapes the MCP tools instead of returning the structured `{"error": ...}` the module otherwise maintains.
6. Add a symlink-escape test at `atlas/tests/test_agent_skills.py:432` — the containment check is documented as covering symlinks, but only `..` traversal and an absolute path are tested, and a symlink shipped inside a skill directory is the vector an author actually controls.
7. Add one test that exercises the real wiring `ChatService` → `get_orchestrator(skill_registry=...)` → `MessageBuilder` — `ChatOrchestrator` builds its own `MessageBuilder` at `atlas/application/chat/orchestrator.py:101`, so deleting the kwarg at `service.py:234` removes skills from every real chat with the suite still green.

## Evidence
Evidence: 8/10. The change is well demonstrated at the unit level — 33 targeted tests covering frontmatter rejection, root precedence, flag gating, cache invalidation, and three traversal vectors, with a full backend suite result quoted. The main gap is that the composed path is asserted rather than shown: nothing exercises the server the way the example `mcp-skills.json` launches it, so a captured startup registering the three tools would close the most valuable remaining hole.

## Notes
- Default roots include `~/.claude/skills` and `~/.agents/skills`, directories other tooling writes; their contents reach the system prompt for every user (`atlas/modules/skills/registry.py:41`). Consider making the non-Atlas roots opt-in on server deployments, and document each root as a privileged input.
- `FEATURE_SKILLS_ENABLED=false` does not disable the MCP server, which calls `discover()` rather than `get_skills()` (`atlas/mcp/skills/main.py:34`); if that split is intended, say plainly in the docs that the flag is not the kill switch.
- Docker mounts only `./logs` and `./config`, so every user/project root named in the skills guide is unreachable in a container (`docker-compose.yml:152`); a mount plus a "Running in Docker" note would prevent a silent empty scan.

---

garland3:

garland3:

## Summary
Adds an Agent Skills subsystem: a discovery/validation registry over layered skill roots, injection of a skill index into the system prompt behind `FEATURE_SKILLS_ENABLED`, and a read-only stdio MCP server exposing `list_skills`, `read_skill`, and `read_skill_resource`.

## Verdict
Request changes.

## Required changes
1. Normalize skill descriptions to a single line in `_validate_description` at `atlas/modules/skills/registry.py:89` — use `" ".join(raw.split())` before the length check. A YAML block scalar currently keeps its interior newlines, and `render_index` (`registry.py:303`) emits the value inside one list item, so the trailing lines land in the system prompt as top-level content and can forge index entries for skills that do not exist.
2. Compare paths instead of basenames when listing resources at `atlas/mcp/skills/main.py:96` — `if path == skill.path: continue`. The current `path.name == skill.path.name` drops every nested file named `SKILL.md`, including the example manifests a skill-authoring skill bundles under `references/`.
3. Guard the home-directory lookup at `atlas/modules/skills/registry.py:194` — `Path.home()` raises `RuntimeError`, which the `OSError` handler at `registry.py:234` does not catch, so a container running under a UID with no passwd entry and no `HOME` loses every skill including the packaged and project-root tiers.
4. Add a regression test asserting a block-scalar description renders as exactly one index line, and one covering the real `ChatService` → `get_orchestrator` → `MessageBuilder` wiring. Today all 33 tests stay green if `skill_registry=` is dropped at `atlas/application/chat/service.py:234`, because every injection test builds `MessageBuilder` directly.
5. Correct `docs/user-guide/skills/README.md:35` — the guide presents `FEATURE_SKILLS_ENABLED` as an off switch, but the MCP server calls `_registry.discover()` (`atlas/mcp/skills/main.py:34`) rather than `get_skills()` and keeps serving full skill bodies with the flag off. State that the server must also be removed from `config/mcp.json`.

## Evidence
Evidence: 7/10 — strongly evidenced at the unit level (33 targeted tests covering frontmatter rejection, root precedence, flag gating, cache invalidation, and three traversal vectors), but the composed runtime path is asserted rather than shown. The most important gap is that the quoted `1873 passed` suite result predates this head commit, which is a merge-conflict resolution touching exactly the wiring files (`orchestrator.py`, `message_builder.py`, `service.py`, `settings.py`) — re-run and quote the suite at this SHA.

## Notes
- `_render_skill_index` runs a blocking nine-root scan with per-skill YAML parsing inside `async def build_messages` (`atlas/application/chat/preprocessors/message_builder.py:163`), stalling every concurrent stream when the cache is cold.
- `SKILL.md` is read with no size cap in `parse_skill_file` (`registry.py:134`) and `read_skill` (`main.py:88`), while `read_skill_resource` enforces a 256 KB limit — the one file with no bound is the one read on every scan.
- A configured `SKILLS_PATHS` root that does not exist is skipped silently at `registry.py:231`; relative entries also resolve differently in the app and in the server, which is launched with `cwd: atlas`.

---

PR #744, opened 2026-08-06
Lifecycle hook / plugin system for the chat pipeline (#713)

### LINKED ISSUE #713: Proposal: config-driven hook system — run bash/python scripts at chat lifecycle events (Claude Code–style)
## Summary

Atlas has a clean, layered chat pipeline but no first-class way to **observe or intervene** at its lifecycle boundaries. Today, adding behavior like "audit every tool call", "redact PII from prompts", "force approval for a class of tools", "rewrite a RAG query", or "drop retrieved chunks above a risk threshold" means editing core files (`orchestrator.py`, `tool_executor.py`, `unified_rag_service.py`) directly.

This proposal is for a **hook system modeled on Claude Code and OpenCode**: an operator writes a **bash or Python script**, registers it in a **config file** against a lifecycle event, and Atlas runs it as a **subprocess** at that point in the turn. The event payload arrives as **JSON on stdin**; the script's **JSON on stdout** (plus its exit code) can allow, modify, block, or escalate the operation.

> This replaces the earlier framing of this issue (an in-process Python plugin bus). The change is deliberate: an in-process plugin API means writing Python against Atlas' internal types, shipping it inside the image, and inheriting whatever the plugin does to the event loop. A subprocess-and-JSON contract is what makes Claude Code hooks usable — an operator drops in a 10-line shell script and edits one config file, with no Atlas internals in the blast radius. Language-agnostic, restart-free, crash-isolated by construction.

This is a design/proposal issue. **No implementation is requested here** — the goal is to agree on the event set, the stdin/stdout contract, and the execution/security model.

> Fits the [product direction](../blob/main/AGENTS.md) toward **governed controls suitable for enterprise/government use** (authorization, auditing, approval gates, resource limits). A hook bus is the natural substrate for those controls and for the Agent Portal governance surface — rather than scattering one-off checks through the core loop.

## Goals

- A single, documented set of **lifecycle events**, each firing at an existing chokepoint.
- Hooks are **arbitrary executables** (bash, Python, anything) configured declaratively — no Atlas Python API to learn, no rebuild to add one.
- A **uniform JSON contract**: event payload on stdin, decision on stdout, exit code as the fast path.
- **Fail-closed, deterministic** execution: a hanging, crashing, or malformed hook must not silently weaken a security boundary or kill a chat turn.
- Reuse the existing seams (orchestrator, mode runners, `tool_executor`, RAG services, `EventPublisher`) rather than a parallel pipeline.

## Non-goals

- Not a client-side/browser plugin system — hooks run server-side, as the Atlas server process user, configured by an operator.
- Not a replacement for the outbound `EventPublisher` (that stays the UI event sink; hooks are the *inbound/interceptor* counterpart).
- Not a sandbox. A hook is operator-installed code running with server privileges, same trust level as `config/mcp.json` entries. Untrusted hooks are out of scope.
- Not a marketplace or remote distribution mechanism.

## Configuration

New two-layer config file, resolved by the existing `ConfigManager._search_paths` mechanism (user `config/` overrides packaged `atlas/config/`), alongside `mcp.json` and `rag-sources.json`:

```jsonc
// config/hooks.json
{
  "hooks": {
    "PreToolUse": [
      {
        "name": "block-destructive-fs",
        "matcher": "filesystem__.*",              // regex over tool name; omit or "*" = all
        "command": ["${ATLAS_CONFIG_DIR}/hooks/block_destructive.py"],
        "timeout_ms": 2000,
        "on_error": "deny"                        // deny | allow  (fail-closed default for this event)
      }
    ],
    "PostToolUse": [
      { "name": "audit-log", "command": ["/usr/local/bin/atlas-audit.sh"], "timeout_ms": 1000, "on_error": "allow" }
    ],
    "UserPromptSubmit": [
      { "name": "pii-redact", "command": ["python3", "${ATLAS_CONFIG_DIR}/hooks/redact.py"], "timeout_ms": 3000 }
    ]
  }
}
```

Notes on the shape:
- `command` is an **argv array**, spawned with **no shell** — no quoting bugs, no injection surface from the payload.
- `matcher` filters before spawn so a hook that only cares about one tool doesn't cost a process on every call.
- `timeout_ms` and `on_error` are per-hook, with per-event defaults (security events default `deny`, observability events default `allow`).
- Multiple hooks per event run **sequentially in config order**, piping modifications forward.

## Events

Each event fires at a concrete, already-existing chokepoint.

| Event | Fires at | Anchor | What the hook can do |
|-------|----------|--------|----------------------|
| **SessionStart** | Conversation/session created | `ChatService.create_session` (`application/chat/service.py:233`) | Attach session-scoped metadata; reject the session; seed policy state |
| **UserPromptSubmit** | Prompt received, before LLM/routing | after user msg added, before mode dispatch in `ChatOrchestrator.execute` (`application/chat/orchestrator.py:172`) | Rewrite/redact prompt text; inject context; **block** the turn with a message |
| **PreLlmCall** | Before each provider request | `LiteLLMCaller.call_plain` / `call_with_rag` / `call_with_tools` / `call_with_rag_and_tools` (`modules/llm/litellm_caller.py:935/984/1128/1174`) | Inspect or redact the outgoing message array; swap/veto the model; **block** the call. *(Requested in [this comment](https://github.com/sandialabs/atlas-ui-3/issues/713#issuecomment-5276133695).)* |
| **PreToolUse** ⭐ | Immediately before a tool executes | `execute_single_tool` before invoke (`application/chat/utilities/tool_executor.py:329`) | Mutate tool args, **deny** the call (synthesized error `ToolResult`), or **force approval** even when not normally required |
| **PostToolUse** | After a tool returns | `tool_executor.py`, post-invoke | Transform/redact the result before it reaches the model/UI; annotate; audit |
| **PermissionRequest** | When a tool needs consent | `requires_approval` (`tool_executor.py:184`) + `application/chat/approval_manager.py` | Auto-approve/auto-deny by policy; escalate a normally-auto call to a human; add audit reason |
| **RagCall** | Before RAG retrieval | classic: `UnifiedRAGService.query_rag` / `query_rag_batch` (`domain/unified_rag_service.py:400/581`); agentic: `_execute_atlas_rag_tool` (`modules/mcp_tools/mcp_execution.py:620`) | Rewrite the query; narrow sources; **block** retrieval |
| **RagResponse** | After chunks return, before prompt injection | the return path of the same `unified_rag_service` / `mcp_execution` calls, before injection in `litellm_caller` | Filter/redact/re-rank chunks; drop high-risk chunks; annotate provenance |
| **SessionEnd** *(optional)* | Turn/session completion | orchestrator teardown | Flush audit records; notify |

⭐ **PreToolUse is the most powerful control point**: every tool call in the system — tools mode *and* the agentic loop — passes through `execute_single_tool`, so one hook uniformly governs arguments, denial, and approval escalation across all tool usage.

### stdin

A JSON object with a stable envelope plus an event-specific `payload`:

```json
{
  "hook_event_name": "PreToolUse",
  "session_id": "…",
  "user_email": "user@example.gov",
  "compliance_level": 3,
  "payload": {
    "tool_name": "filesystem__write_file",
    "tool_args": { "path": "/etc/passwd", "content": "…" }
  }
}
```

`user_email` and `compliance_level` are **informational** — they are stamped server-side and re-asserted after the hook returns regardless of what the hook says (see Security).

### Exit codes (the fast path)

| Exit | Meaning |
|------|---------|
| `0` | Continue. If stdout is non-empty JSON, apply it; if empty, no change. |
| `2` | **Block.** stderr is surfaced as the reason (to the model for tool events, to the user for prompt events). |
| other | Error. Handled per `on_error` (`deny` or `allow`); always logged and audited. |

This mirrors Claude Code exactly, and it means the simplest useful hook is three lines of bash with no JSON emitted at all.

### stdout (the structured path)

```json
{ "decision": "modify", "payload": { "tool_args": { "path": "/tmp/safe.txt", "content": "…" } } }
{ "decision": "deny", "reason": "Writes outside /workspace are blocked by policy" }
{ "decision": "require_approval", "reason": "Cross-domain network call" }
{ "decision": "continue" }
```

- **`continue`** — no change (also the meaning of exit 0 with empty stdout).
- **`modify`** — replace the mutable part of the payload (prompt text, tool args, tool result, LLM messages, RAG query, RAG chunks). Core validates the replacement against the original schema and re-applies invariants.
- **`deny`** — short-circuit. For tools this becomes an error `ToolResult` so the loop continues gracefully rather than aborting the turn.
- **`require_approval`** — (PreToolUse / PermissionRequest) force the runtime approval gate even if `requires_approval` would have passed.

### Composition

Hooks for an event chain in config order; each sees the previous hook's `modify` output. **Most-restrictive-wins**: the first `deny` short-circuits the chain, and a `require_approval` cannot be downgraded by a later hook.

## Execution & resource model

- Spawned via `asyncio.create_subprocess_exec` (argv, **never** `shell=True`), inside the existing async turn.
- **Timeout** per hook; on expiry the process is killed and `on_error` decides. A hanging hook must never hang the turn.
- **Environment allow-list** — hooks get a minimal env plus `ATLAS_CONFIG_DIR`, `ATLAS_PROJECT_DIR`, not the server's full environment (which holds provider API keys).
- **Bounded stdout/stderr** (e.g. 1 MB) to keep a chatty hook from ballooning memory; overflow = error.
- Hooks are opt-in: no `config/hooks.json`, no subprocess, zero overhead on the hot path.
- Config is loaded through `ConfigManager`, so hook changes follow the same reload story as the rest of Atlas config.

## Security & governance

- A hook can **tighten but never widen** a boundary. Server-side identity (`_atlas_user`, `user_email`) and trusted `session.context["compliance_level"]` are **re-injected after** any `modify` — `tool_executor.py:495` already does exactly this for tool args ("ensures `_atlas_user` and other security-critical parameters cannot be tampered with"), and the same rule must apply to every hook payload.
- `modify` on `RagCall` may **narrow** the source list; it can never add a source outside the compliance-filtered allow-list computed in `mcp_execution.py` / `unified_rag_service.py`. Same for `PermissionRequest`: a hook may escalate to approval, but a `deny` from `ToolAuthorizationService` or `core/compliance.py` is not overridable by a hook returning `continue`.
- Every hook invocation is **auditable** — event, hook name, decision, reason, duration, exit code. Dovetails with the OpenTelemetry span work (#545/#697) and history/export work (#701).
- Because hooks are executables, `config/hooks.json` is a **privileged file**: write access to it is equivalent to code execution as the server user. Worth documenting alongside `mcp.json` and considering for the startup validation path.

## Open questions

1. **Payload schemas** — one versioned schema per event (`hook_event_name` + a `version` field?), and how much of the turn state to expose (full message history on `PreLlmCall`, or just the last turn?).
2. **`modify` granularity** — full replacement of the mutable object, or a structured patch? Full replacement is simpler to validate; patches compose better across chained hooks.
3. **Per-group / per-deployment enablement** — should hooks be scopeable to a user group or compliance level, or is a single global list enough for v1?
4. **Do we also want a fire-and-forget observability tier?** Blocking subprocesses on `PostToolUse` for pure logging costs latency for nothing; a non-blocking variant (`"blocking": false`) may be worth having.
5. **Can a hook emit outbound UI events** (via `EventPublisher`) — e.g. surfacing "your prompt was redacted" as a visible notice? Probably a reserved key in the stdout JSON.
6. **Relationship to existing extension points** — MCP tools, `PromptProvider`, elicitation/approval managers. Where does a hook stop and an MCP tool begin?
7. **Windows / container support** — hooks assume a POSIX-ish exec. Is that acceptable, or does the packaged container need a documented interpreter set?

## Out of scope

- Concrete implementation / PR.
- Frontend/client plugins.
- Sandboxing untrusted hook code.

## Prior art

- [Claude Code hooks](https://code.claude.com/docs/en/hooks) — event + matcher + argv command in `settings.json`; stdin JSON, exit 2 = block with stderr as reason, JSON stdout for structured decisions. The model this proposal follows most closely.
- [OpenCode plugins](https://opencode.ai/docs/plugins/) — `tool.execute.before` / `.after` as in-process TS functions; blocking by throwing, modification by mutating `output.args`. Also ships a small config-driven `experimental.hook` for `file_edited` / `session_completed`.

---

Related: #497 (safety checks), #438 (Agent Skills), #545/#697 (OpenTelemetry spans), #699 (server-side compliance enforcement at RAG query time), #701 (history export/audit).

garland3:
Implements the plugin / event-hook system proposed in #713.

Issue #713 is framed as a design proposal ("no implementation is requested here"). This PR takes the proposed hook surface, return-value contract, and execution/security model and implements them, resolving the open questions with the choices described below. Review the contract as much as the code -- the semantics are the part that is hard to change later.

## What this adds

A server-side hook bus (`atlas/core/hooks/`) that is the **inbound mirror of `EventPublisher`**: where the publisher fans events out to the UI with no return influence, the registry runs a blocking interceptor chain whose return values the core respects.

Seven hook points, each firing at an existing chokepoint:

| Hook | Fires at |
|------|----------|
| `SESSION_START` | `ChatService.create_session`, before the session is persisted |
| `USER_PROMPT_SUBMIT` | `ChatOrchestrator.execute`, before the message is written to history |
| `PRE_TOOL_USE` | `execute_single_tool`, before invocation |
| `POST_TOOL_USE` | `execute_single_tool`, after the tool returns |
| `PERMISSION_REQUEST` | `execute_single_tool`, once the approval requirement is computed |
| `RAG_CALL` | `UnifiedRAGService.query_rag` / `query_rag_batch`, before retrieval |
| `RAG_RESPONSE` | same methods, after retrieval |

`PRE_TOOL_USE` is the broadest control point: tools mode and the agentic loop both funnel through `execute_single_tool`, so one handler governs arguments, denial, and approval escalation across all tool usage.

## Return-value contract

`CONTINUE` / `MODIFY` / `DENY` / `REQUIRE_APPROVAL`. Returning `None` is `CONTINUE`.

`MODIFY` is a **structured patch**, not a full replacement (open question 2). Each event declares `MUTABLE_FIELDS`; anything else -- including the trusted `session_id`, `user_email`, `conversation_id`, `compliance_level` on every event -- is not patchable.

Composition is deterministic and most-restrictive-wins: handlers run in `(priority, registration order)`, `MODIFY` patches pipe through the chain, the first `DENY` short-circuits, and `REQUIRE_APPROVAL` is sticky but loses to a later `DENY`.

## Security model

Plugins are trusted operator code, but the core treats their **output** defensively -- a plugin may tighten a boundary, never widen one:

- **Identity cannot be forged.** After a `PRE_TOOL_USE` patch, `tool_executor` re-runs `inject_context_into_args()` and `_filter_args_to_schema()`, so `_atlas_user` is re-stamped server-side and undeclared parameters are dropped.
- **RAG sources are narrow-only.** `RagCallEvent` rejects any `data_sources` patch that adds to the already compliance-filtered list, and refuses to let it be emptied (`DENY` is the way to block retrieval).
- **Admin-mandated approval is absolute.** `PermissionRequestEvent` rejects `needs_approval=False` when `admin_required` is set by `FORCE_TOOL_APPROVAL_GLOBALLY` or a per-tool `require_approval`.
- **Prompt patches are narrow-only.** `UserPromptSubmitEvent` rejects patches that add tools/data sources or enable agent mode.
- **Compliance is untouched.** `compliance_level` remains read-only and continues to reach `execute_tool()` straight from the session.

A rejected patch is handled exactly like a handler failure.

## Failure semantics

Async handlers run under a timeout (`HOOK_TIMEOUT_SECONDS`, overridable per registration). A handler that raises, times out, returns a non-`HookResult`, or returns an invalid patch either becomes a `DENY` (fail closed) or is skipped (fail open). Defaults: `SESSION_START` and `POST_TOOL_USE` fail open; the other five fail closed. `asyncio.CancelledError` is never treated as a plugin fault -- it propagates so a disconnect unwinds normally.

## Registration / packaging (open question 1)

An explicit config allow-list, not autodiscovery:

```
FEATURE_HOOKS_ENABLED=true
HOOK_PLUGINS=my_org.atlas_plugins.audit,my_org.atlas_plugins.pii:register
HOOK_TIMEOUT_SECONDS=5.0
```

Loading happens in `AppFactory.__init__` before any service is wired, and is **fail-fast**: a plugin that cannot be imported or that raises while registering aborts startup, because a governance control that silently fails to load is indistinguishable from one that was never configured.

## Other open questions, as resolved here

3. **Interceptor, not fan-out bus.** One tier. Observation is `CONTINUE` plus metadata, which the chain result carries back to the call site for auditing.
4. **No direct `EventPublisher` access** for plugins. A plugin that wants to surface a warning denies with a `user_message`, which the orchestrator publishes as a normal assistant message.
5. **MCP tools add capability the model can invoke; hooks add policy over capability that already exists** and are not model-visible.

## Cost when unused

Off by default. Every call site short-circuits on `registry.has_hooks(point)`, so a deployment without plugins does no extra work.

## Testing

67 new tests (`atlas/tests/test_hook_registry.py`, `atlas/tests/test_hook_integration.py`) covering composition and ordering, error isolation and timeouts, every tighten-only invariant, plugin loading, and each hook point at its real call site -- including that a `PRE_TOOL_USE` patch cannot forge `_atlas_user` or smuggle undeclared parameters, and that a plugin cannot auto-approve an admin-mandated tool.

Full backend suite: 1907 passed, 17 skipped.

## Docs

New: `docs/developer/hook-plugin-system.md` (indexed in `docs/developer/README.md`). `.env.example` documents the three new settings.

Closes #713


garland3:

garland3:
## Summary
Adds a hook plugin system with seven governance points (session start, prompt submit, pre/post tool use, permission request, RAG call/response), a priority-ordered registry with per-handler timeouts and fail-open/fail-closed defaults, settings-driven plugin loading, and a developer guide.

## Verdict
Request changes.

## Required changes
1. Track "was modified" independently of the chain decision in `atlas/core/hooks/hook_registry.py:283` — when one handler returns `MODIFY` and another returns `REQUIRE_APPROVAL`, `chain.modified` is `False` even though the event was already patched, so the four call sites gated on it (`tool_executor.py:474`, `tool_executor.py:668`, `service.py:264`, `unified_rag_service.py:326`) silently discard the patch and the tool runs with the original arguments.
2. Clear `result.artifacts` and `result.display_config` on a `POST_TOOL_USE` denial in `atlas/application/chat/utilities/tool_executor.py:644` — today only `content` is rewritten, so `notify_tool_complete()` still ships the withheld tool's files to the UI and persists them with tokenized download URLs.
3. Make plugin loading idempotent in `atlas/infrastructure/app_factory.py:34` — `AppFactory` is constructed twice in a live process (module-level instance plus `AtlasClient.initialize()`), and `load_plugins_from_settings()` appends to the process-wide registry with no dedup, so every handler registers and runs twice.
4. Make `needs_approval` monotonic within a chain in `atlas/core/hooks/hook_types.py:283` — a later `PERMISSION_REQUEST` handler returning `modify({"needs_approval": False})` currently overwrites an earlier handler's escalation, making approval enforcement depend on plugin ordering.
5. Validate the `SessionStart` context patch in `atlas/application/chat/service.py:266` — the whole dict is merged into `session.context` unchecked, so a patch can seed `conversation_id`, `files`, or `agent_mode`, which every other writer validates first.
6. Populate `session_id`, `conversation_id`, and `compliance_level` on `RagCallEvent` / `RagResponseEvent` in `atlas/domain/unified_rag_service.py:299` — they are left at `None` while every other call site fills them, so a compliance-tiered retrieval policy silently degrades to the permissive branch.
7. Remove the unused `is_model_allowed` import at `atlas/application/chat/orchestrator.py:9` — merge residue that fails `ruff check` (F401).

## Evidence
Evidence: 7/10. 67 targeted new tests, a developer guide, and settings/compose entries make this well-documented for its size, but nothing exercises the feature end to end with hooks actually enabled, and the quoted full-suite result ("1907 passed") predates the merge commit under review. The most useful addition would be one reference plugin loaded through `HOOK_PLUGINS` against a running server, showing a real denial.

## Notes
- Sync handlers run un-timed on the event loop (`hook_registry.py:263`); `HOOK_TIMEOUT_SECONDS` only applies to coroutine handlers.
- `HookResult.deny()` defaults the user-facing message to the operator-facing `reason` (`hook_types.py:106`), leaking handler names and rule identifiers into chat.
- `HOOK_PLUGINS` is absent from the `docker-compose.yml` environment allow-list (`docker-compose.yml:135`), so enabling the feature there loads nothing.

---

garland3:
## Summary
Adds a lifecycle hook/plugin system: a process-wide registry with seven interception points (session start, user prompt, pre/post tool, permission request, RAG call/response), fail-open/fail-closed per point, a per-turn context var, startup plugin loading from `HOOK_PLUGINS`, plus 67 tests and a developer guide. The feature is off by default behind `FEATURE_HOOKS_ENABLED`.

## Verdict
Request changes.

## Required changes
1. Catch `AuthorizationError`/`DomainError` around the `reset_session` and `attach_file` branches at `atlas/main.py:790` and `atlas/main.py:798` and reply with the `{"type": "error", "error_type": "authorization"}` frame — `create_session` raises there too, and the unhandled exception unwinds the message loop, closing the socket with no explanation the moment a `SESSION_START` deny plugin fires.
2. Set the escalation flag whenever a handler asserts it, not only when it changes the value: at `atlas/core/hooks/hook_types.py:345`, use `if patch.get("needs_approval") is True: self.escalated_by_hook = True`. The tools-mode call site enters the chain with `needs_approval=True` (`requires_approval` returns `True` on every path), so today an escalating patch is a no-op flag-wise and a later `modify({"needs_approval": False})` validates — making the gate depend on plugin registration order, the exact invariant the docs promise. Add a registry test whose event starts at `needs_approval=True`.
3. Stop falling back to the operator-facing `reason` for user-visible text: use `DEFAULT_DENY_USER_MESSAGE` at `atlas/core/hooks/hook_registry.py:235` and drop `or chain.reason` at `atlas/application/chat/utilities/tool_executor.py:343`. `HookResult` is public and directly constructible, so `HookResult(decision=DENY, reason="rule pii-7 matched path=/etc/shadow")` currently puts rule internals into the chat and the model's context, contradicting the documented contract.
4. Close the in-place-mutation bypass at `atlas/application/chat/utilities/tool_executor.py:470` (and `:519`, `:657`, `service.py:253`, `unified_rag_service.py:317`): events are built from shallow copies, so a handler can edit a nested value (`event.arguments["options"]["path"]`) and return `CONTINUE`, and the edit reaches `execute_tool()` with `chain.modified` still `False` — so `inject_context_into_args()` and `_filter_args_to_schema()` never re-run. Deep-copy the mutable payloads, or re-run the compensating pass unconditionally after `PRE_TOOL_USE`.
5. Run sync handlers under the timeout — `await asyncio.wait_for(asyncio.to_thread(handler, event), ...)` at `atlas/core/hooks/hook_registry.py:282` — or reject non-coroutine handlers at `register()`. Today a sync handler runs unbounded on the event loop, so `HOOK_TIMEOUT_SECONDS` and the fail-closed guarantee are both unenforceable for the most likely first plugin shape (a blocking policy lookup).
6. Iterate a snapshot in `dispatch()`: `registrations = list(self._registrations[hook_point])` at `atlas/core/hooks/hook_registry.py:246` — `register()` sorts the bucket in place, so a handler that registers another during a chain can cause handlers to be skipped or run twice.
7. Make the compose block consistent at `docker-compose.yml:135`: `FEATURE_HOOKS_ENABLED=${FEATURE_HOOKS_ENABLED:-false}` and `HOOK_TIMEOUT_SECONDS=${HOOK_TIMEOUT_SECONDS:-5.0}` — `HOOK_PLUGINS` passes through from the host while the flag is pinned to a literal, so an operator who sets both in their environment gets the plugin list and a silently disabled feature.

## Evidence
Evidence: 7/10. Unusually thorough unit and integration coverage — every tighten-only invariant in the description has a matching negative test, and the prior review round's fixes each landed with a test — but nothing shows the system running with the feature actually enabled: every integration test registers handlers directly against the global registry, `test_app_factory_loads_configured_plugins` never constructs an `AppFactory`, and the quoted full-suite result (1907 passed) predates the four follow-up commits in this head. The most useful missing artifact is one end-to-end run with a plugin loaded through `HOOK_PLUGINS` showing a denial reaching the chat, plus a re-run suite result at `1d3890e`.

## Notes
- A RAG denial's `user_message` is concatenated into the prompt as retrieved context rather than shown to the user (`atlas/domain/unified_rag_service.py:339`) — same field name, different audience than the tool and prompt paths.
- `HookChainResult.metadata` is accumulated on every dispatch but read by no call site, so the documented observation-only plugin pattern (`continue_(risk_score=...)`) currently goes nowhere; wire it to the existing tool/RAG spans or adjust the doc.
- `_hook_denied_result` sets no `hook_denied` span attribute, unlike the RAG path (`unified_rag_service.py:409`), so policy blocks on the tool path are indistinguishable from tool crashes in telemetry.

---

garland3:
Add a pre LLM call hook as well

garland3:
## Summary
Adds a hook/plugin bus (`atlas/core/hooks/`) with eight hook points wired into the tool executor, chat orchestrator, chat service, RAG service, and both LLM call paths, plus settings-driven plugin loading (`FEATURE_HOOKS_ENABLED`, `HOOK_PLUGINS`, `HOOK_TIMEOUT_SECONDS`) and a developer guide.

## Verdict
Request changes.

## Required changes
1. Re-dispatch `PRE_TOOL_USE` over the edited arguments at `atlas/application/chat/utilities/tool_executor.py:588` (once, after the re-injection and re-filter) and honor a `DENY` — today a user who edits arguments in the approval dialog bypasses the argument-sanitizing hook entirely, on exactly the tools an operator chose to gate.
2. Let `AuthorizationError` pass through the LLM error funnels unchanged — add an `except AuthorizationError: raise` ahead of the blanket `except Exception` in `safe_call_llm_with_tools()` at `atlas/application/chat/utilities/error_handler.py:157` and re-raise instead of classifying at `atlas/application/chat/modes/tools.py:231` — otherwise a `PRE_LLM_CALL` denial renders as "The LLM service encountered an error…" and the plugin's `user_message` is discarded on both tools paths.
3. Stop returning a hook denial as retrieved context at `atlas/domain/unified_rag_service.py:334` — return a flagged/sentinel `RAGResponse` the caller drops (or publish the notice as an assistant message) — because the denial text is currently spliced into a system message labelled "Retrieved context from &lt;source&gt;. Use this context to inform your response", so the model reads the policy text and the user never sees it.
4. Skip the non-streaming fallback on a domain refusal in `stream_with_fallback()` at `atlas/application/chat/modes/streaming_helpers.py:90` — a denied or failed stream currently re-enters a leaf entry point and dispatches the whole `PRE_LLM_CALL` chain a second time for one turn, double-counting audit rows and quotas.
5. Resolve `needs_approval, allow_edit, admin_required` from `requires_approval()` whenever a `config_manager` is present and apply `skip_approval` only to `needs_approval` afterwards, at `atlas/application/chat/utilities/tool_executor.py:506` — in the agentic loop `PermissionRequestEvent.admin_required` is always `False`, even under `FORCE_TOOL_APPROVAL_GLOBALLY`.
6. Dispatch sync handlers on a dedicated bounded `ThreadPoolExecutor` owned by the registry rather than `asyncio.to_thread` at `atlas/core/hooks/hook_registry.py:301` — a timed-out handler's thread is abandoned but keeps a slot in the process-wide default pool, so a blocking plugin can starve every unrelated `to_thread` caller until restart.

## Evidence
Evidence: 7/10. Unit and call-site coverage is strong for a change this size, but nothing demonstrates the feature as an operator configures it — every integration test registers handlers directly against the registry, and `test_app_factory_loads_configured_plugins` never constructs an `AppFactory`; the single most useful missing artifact is one plugin loaded through `HOOK_PLUGINS` with its denial visible in a real turn. The quoted "1907 passed, 17 skipped" predates this head and the description still says seven hook points and omits `PRE_LLM_CALL`; both should be refreshed.

## Notes
- Default denial copy reads "Tool execution blocked by policy: This request was blocked by policy." — apply the prefix only when a plugin supplies its own text (`atlas/application/chat/utilities/tool_executor.py:359`).
- `notify_tool_complete` receives the pre-hook `parsed_args`, so a `PRE_TOOL_USE` rewrite of `canvas_canvas` arguments executes but pushes the original content to the canvas (`atlas/application/chat/utilities/tool_executor.py:713`).
- `FEATURE_HOOKS_ENABLED` gates loading only; dispatch keys off `has_hooks()`, so a handler registered by import runs with the feature nominally off — either make the flag authoritative on the registry or narrow the doc's "the bus is inert" claim (`atlas/core/hooks/hook_registry.py:190`).

---

garland3:
## Summary
Adds a pluggable hook/governance system (eight hook points, registry, loader, settings and docs) and wires dispatch into the chat orchestrator, chat service, tool executor, RAG service, and both LLM call paths. Off by default behind `FEATURE_HOOKS_ENABLED`.

## Verdict
Request changes.

## Required changes
1. Re-dispatch `PRE_TOOL_USE` over the post-edit arguments in `atlas/application/chat/utilities/tool_executor.py:588` — arguments edited in the approval dialog replace `filtered_args` with no second dispatch, so a sanitizing or denying handler is bypassed on exactly the tools an operator chose to gate. Honor `DENY` via `_hook_denied_result` and re-run the injection/filter pass on `MODIFY`; dispatch at most once to avoid recursion.
2. Let `AuthorizationError` escape the LLM error funnels — add `except AuthorizationError: raise` ahead of the blanket handler in `safe_call_llm_with_tools` (`atlas/application/chat/utilities/error_handler.py:184`) and in the streaming branch of `atlas/application/chat/modes/tools.py:241`. Today a `PRE_LLM_CALL` denial is reclassified as "The LLM service encountered an error… please try again or contact support", discarding the plugin's `user_message` and telling a blocked user to retry.
3. Skip the non-streaming fallback on a domain refusal in `atlas/application/chat/modes/streaming_helpers.py:90` — a denial re-enters a leaf entry point and runs the whole `PRE_LLM_CALL` chain a second time for one turn, so audit rows, quota decrements and external policy calls all double.
4. Give a denied retrieval a channel of its own in `atlas/domain/unified_rag_service.py:466` and `:511` — the denial is returned as an ordinary `RAGResponse`, so the policy text is spliced into the prompt as "Retrieved context from `<source>`… use this context to inform your response" and the user never sees it. Add an explicit blocked flag (or sentinel type), have the context assembly skip that source, and deliver `user_message` on the assistant-message channel.
5. Resolve `needs_approval`, `allow_edit` and `admin_required` from `requires_approval(...)` whenever a `config_manager` is present at `atlas/application/chat/utilities/tool_executor.py:506`, applying `skip_approval` to `needs_approval` alone — on the agent-portal path `admin_required` is always `False`, so the "cannot auto-approve an admin-mandated tool" check can never fire there.
6. Dispatch sync handlers on a registry-owned bounded `ThreadPoolExecutor` instead of `asyncio.to_thread` at `atlas/core/hooks/hook_registry.py:301` — a timed-out handler's thread is abandoned but keeps its slot in the process-wide default pool, so a plugin with a hung dependency starves every unrelated `to_thread` caller until restart.
7. Pass the executed `filtered_args` — not `parsed_args` — to `notify_tool_complete` at `atlas/application/chat/utilities/tool_executor.py:713`; for `canvas_canvas` the pre-hook arguments are published verbatim to the client, so a redacting hook governs what is stored while the unredacted text renders in the canvas.

## Evidence
Evidence: 6/10. Unit and call-site coverage is genuinely strong for a change of this size (~2,000 lines of new tests, a negative test for each tighten-only invariant), but nothing exercises the feature the way an operator enables it — every test registers handlers directly against the registry, and `test_app_factory_loads_configured_plugins` never constructs an `AppFactory`. The most valuable missing artifact is one end-to-end run with `FEATURE_HOOKS_ENABLED=true HOOK_PLUGINS=<module>` showing a plugin loaded at startup and a denial reaching the chat.

## Notes
- The PR description still says "seven hook points" and omits `PRE_LLM_CALL`; the quoted "1907 passed" predates this head commit — refresh both before merge.
- `FEATURE_HOOKS_ENABLED` gates loading only, not dispatch (`atlas/core/hooks/hook_registry.py:190`), so the documented "the bus is inert when disabled" is not what the code guarantees.
- The default denial reads "Tool execution blocked by policy: This request was blocked by policy." when no `user_message` is supplied (`atlas/application/chat/utilities/tool_executor.py:359`); apply the prefix only to plugin-supplied text.

---

garland3:
Closing in favor of a fresh PR. The branch has been force-pushed with all 18 commits squashed into a single commit (`ea8d5d9`) on top of current `main`, so review can start clean. A new PR will be opened from the same branch (`feat/issue-713-hook-plugin-system`).

PR #750, opened 2026-08-08
Fix horizontal overflow on mobile (closes #747)

### LINKED ISSUE #747: horizontal scrolling
### GitHub issue title

**Long text causes horizontal overflow and page scrolling on mobile**

### Description

On mobile, some content does not respect the width of the viewport. Long text, tool names, URLs, or other non-wrapping content can extend past the right edge of the screen, making the entire page wider than the device viewport.

This results in **horizontal scrolling of the whole page**. In the screenshot, the tool-call rows extend beyond the visible right boundary rather than wrapping, truncating, or remaining constrained to their parent container.

### Steps to reproduce

1. Open the chat UI on a mobile-sized viewport.
2. Open a conversation containing long text/tool names, such as `basic_memory_discover_topics`, long URLs, or similar content.
3. Scroll to the affected content.
4. Notice that the content extends beyond the right side of the viewport.
5. Swipe horizontally and observe that the entire page can be horizontally scrolled.

### Expected behavior

All chat content should remain constrained to the viewport width. Long strings should either wrap, truncate with an ellipsis where appropriate, or otherwise shrink within their parent container.

The main page should never become wider than `100vw` due to message/tool content.

### Actual behavior

Certain text elements maintain their intrinsic/minimum width and push their parent container beyond the viewport width. This creates horizontal overflow and allows the entire mobile page to scroll left/right.

This appears to be primarily a **mobile/responsive layout issue**.

### Possible cause

Likely a flex/grid child with the default `min-width: auto`, or a text element that does not allow long strings to break. Potential fixes to investigate include:

```css
min-width: 0;
max-width: 100%;
overflow-wrap: anywhere;
word-break: break-word;
```

For the outer application/chat container, `overflow-x: hidden` may be useful as a safeguard, but ideally the underlying overflowing child should be fixed rather than simply hiding the overflow.

garland3:
Closes #747.

## Symptom

At a phone viewport (375px and below) chat content escaped the right edge of the screen and made the whole page scroll sideways. Reproduced with Playwright at 375x700: the message list measured a scroll width of **859px inside a 375px viewport**, and the header alone measured **400px** even with an empty chat.

## Root cause

A long unbroken string -- a tool name like `basic_memory_discover_topics`, a raw URL, a base64 blob -- has a min-content width as wide as the string itself. Inside a flex row that width becomes the item's *automatic minimum* (`min-width: auto`), so the row refuses to shrink back to its column and pushes everything wider than the screen.

Three concrete offenders:

1. **Assistant bubble** (`Message.jsx`) used `w-full` as a flex sibling of the `w-8` avatar plus a `gap-3`, so it asked for 100% of the row *on top of* 44px of avatar. It only fit because flex shrink normally absorbs the difference -- and shrink is exactly what `min-width: auto` disables once the content cannot wrap.
2. **Message rows** (tool-call summary line, tool log, agent status, plain user/system text) had neither a cleared flex minimum nor a break opportunity, so a long tool name or URL widened the row.
3. **Header** -- the two button groups had no `min-w-0`, so the model dropdown, connection dot, save-mode button and hamburger together demanded 400px at a 375px viewport. This one overflowed with no messages at all.

## Fix

Underlying children first; `overflow-x` hiding is only a safeguard.

- `index.css`: a `.chat-messages` scope that clears the automatic flex minimum (`min-width: 0` via a zero-specificity `:where(...)` list, deliberately excluding `svg`/`img` so icons are not squashed) and sets `overflow-wrap: anywhere` so long strings have somewhere to break. Code blocks are excluded -- they keep their own horizontal scroller rather than being broken mid-token. Markdown tables are capped at `max-width: 100%`.
- `ChatArea.jsx`: the transcript container carries the `chat-messages` hook, plus `overflow-x-hidden` as a backstop. Uploaded-file chips and the tool/@file autocomplete rows truncate or wrap long names instead of stretching.
- `Message.jsx`: the bubble is `min-w-0` + `flex-1` instead of `w-full`; tool-call summary rows wrap (`flex-wrap`) with the status badge pinned (`flex-shrink-0`); plain-text bodies, tool logs, agent status lines and file-download buttons break long words.
- `Header.jsx`: both button groups are `min-w-0` so the model name truncates instead of pushing the header past the screen.
- `EnabledToolsIndicator.jsx`: tool chips truncate rather than widen the footer.

## Verification

Measured in a real browser at 320px, 375px and 1280px with long tool names, long URLs, wide tables and code blocks injected into the transcript:

| | before | after |
|---|---|---|
| transcript scroll width @375px | 859px | 375px (= viewport) |
| header scroll width @375px | 400px | 375px |
| elements past the right edge | 20 | 0 |

Desktop rendering is unchanged -- markdown tables still fill the message column (`display: table`, 972px in a 1004px bubble), and code blocks still scroll horizontally instead of wrapping.

- `frontend/src/test/message-viewport-containment.test.jsx` -- 7 new regression tests covering the structural contract (jsdom has no layout engine, so these assert the shrink/break classes and the CSS hook rather than pixel widths).
- `npx vitest run`: 489 tests passed (47 files).
- `bash test/atlas_tests.sh`: 1872 passed, 17 skipped.
- `npm run lint`: 0 errors (4 pre-existing warnings on unrelated files).


garland3:
## Summary
Contains the chat transcript, header, tool chips and autocomplete rows within the viewport on narrow screens, by clearing the automatic flex minimum via a scoped `.chat-messages :where(...) { min-width: 0 }` rule plus `overflow-wrap: anywhere`, adding `min-w-0`/`truncate`/`break-all` utilities across five components, and setting `overflow-x-hidden` on the transcript container. Seven jsdom regression tests are added.

## Verdict
Request changes.

## Required changes
1. Let the tool-approval action row shrink at `frontend/src/components/ToolApprovalMessage.jsx:347` and `:247` — add `min-w-0` to the `flex-1` rejection-reason input and `flex-wrap` to its parent row. Form controls are not in the `:where(...)` list at `frontend/src/index.css:636`, so the input keeps its intrinsic ~180px minimum and the row cannot fit the ~270px available at 375px; with the new `overflow-x-hidden` that overflow is now clipped instead of scrollable, leaving the control unreachable.
2. Give display math its own scroller — add `.chat-messages .katex-display { overflow-x: auto; overflow-y: hidden; }` next to the code-block carve-out in `frontend/src/index.css`. KaTeX sets `white-space: nowrap` on the display wrapper, so `overflow-wrap: anywhere` cannot break it and long equations are now silently cut off with no affordance.
3. Give wide markdown tables a scroller rather than clamping them — `.chat-messages .selectable-markdown table { display: block; overflow-x: auto; }` with `overflow-wrap: normal` on cells, at `frontend/src/index.css:655`. Under `max-width: 100%` plus `overflow-wrap: anywhere`, a 5–6 column table at 375px collapses to columns of stacked single characters.
4. Add `overflow-x-auto` to the markdown parse-failure fallback `<pre>` at `frontend/src/components/Message.jsx:591` — it is the only `<pre>` in the transcript without a local scroller, so an unbroken long token in an already-failing message is now clipped with no way to reveal it.
5. Restore a legible floor and an accessible name on the model selector at `frontend/src/components/Header.jsx:205` — `min-w-0` plus `truncate` lets the label collapse to a bare chevron at 320px, and the button carries no `title` or `aria-label`, so the active model becomes undiscoverable without opening the dropdown.
6. Make the container assertion in `frontend/src/test/message-viewport-containment.test.jsx:134` check the rendered element rather than substring-matching `ChatArea.jsx` source — `toContain('chat-messages')` also matches the comments that discuss it, so moving the class to the wrong element leaves the test green and the fix dead.
7. Pin the deliberate `svg`/`img` exclusion in the same test file — capture the `:where(...)` selector list and assert it does not contain `svg` or `img`. The current regex matches any list contents, so "simplifying" it to `*` would pass CI while squashing every icon in the transcript.

## Evidence
Evidence: 7/10. The description is unusually substantive for a layout fix — reproducible symptom with numbers, a correct root-cause explanation, a before/after measurement table and three quoted green runs — but every claim about rendered behavior rests on prose. The single most valuable missing artifact is a checked-in browser-level assertion that `document.documentElement.scrollWidth <= window.innerWidth` at 320/375px with adversarial content, since the jsdom tests only assert class names and would stay green if the overflow returned; a 375px before/after screenshot would be the cheap substitute.

## Notes
- The same shrink invariant is encoded twice for transcript descendants (the `:where(...)` rule and in-component `min-w-0` utilities) with neither site referencing the other — `frontend/src/index.css:635`.
- `.selectable-markdown` and `.message-content` are global selectors inside a transcript-scoped block, and the code carve-out below them is scoped, so the wrap policy and its exemption do not travel together — `frontend/src/index.css:640`. `.message-content` has no producer in `src/`.
- `break-all` on the download buttons (`frontend/src/components/Message.jsx:349`, `:372`, `:389`) breaks filenames at every character where the inherited `overflow-wrap: anywhere` would break only when needed.

---

garland3:
Thanks -- reviewed each point in a real browser (Chromium at 375x812 and 320x640, transcript seeded with adversarial content) rather than by reading. Results and what changed in 8adb3a6:

**1. Tool-approval row -- confirmed, and worse than described.** The input's intrinsic minimum is only half of it: `button` is in the `:where(...)` list, so `min-width: 0` applies to Approve/Reject, and in a nowrap row they absorbed all the shrinkage. Measured at 375px: **Approve 24px wide (needs 68), Reject 26px (needs 53)** -- they render as "Ap" and "Re". Forcing `min-width: auto` back onto just those two buttons restores 80px/67px, which pins the cause on the containment rule. Fixed at both call sites: `flex-wrap` on the row, `shrink-0` on the buttons, and `min-w-[12rem]` on the input so it drops to its own line instead of becoming a sliver. Verified: buttons full-width and the input 271px at 375px, 216px at 320px.

**2. Display math -- confirmed.** `.katex-display` measured `clientWidth 267` vs `scrollWidth 573`: 306px of the equation was clipped with no affordance. Added the scoped `overflow-x: auto` carve-out.

**3. Wide tables -- confirmed.** A six-column table at 375px gave 35-59px columns and 89px-tall cells ("brav/o", "fox/tro/t"). Now `display: block; overflow-x: auto` with `overflow-wrap: normal` on cells -- cell height drops to 41px and the table scrolls.

**4. Fallback `<pre>` -- valid, narrowly.** It carries `whitespace-pre-wrap`, so only a single unbroken token is affected, and only when markdown parsing throws. Added `overflow-x-auto` anyway.

**5. Model selector -- not reproduced.** At 320px the label measures 82px and reads `openrouter-...`, not a bare chevron. The missing `title`/`aria-label` is real but predates this branch (`main` has the same `truncate` span with no accessible name), so it is out of scope here -- worth its own issue.

**6 and 7 -- taken.** The transcript-hook assertion now matches the `<main>` element's className rather than the file at large, and the `:where(...)` check parses the selector list and asserts `svg`/`img`/`*` are absent.

**On evidence.** Fair hit. Whole-document measurements after the fixes, with adversarial content (150-char URLs, base64 blobs, 70-char tool names, display math, six-column table, 98-char download filenames):

| viewport | `documentElement.scrollWidth` | elements overflowing outside a scroller |
|---|---|---|
| 375 | 375 | 0 |
| 320 | 320 | 0 |

For comparison, `main` with the same content at 375px: transcript `scrollWidth 4080` vs `clientWidth 375`, and the header's right cluster at `right: 400` with the menu button off-screen. I did not check in a browser-level assertion: no workflow in `.github/workflows/` runs `test_e2e/`, so it would be a test nothing executes. The jsdom tests were extended instead (11 in this file, and the two new approval-row cases fail without the component fix -- verified by stashing it).

garland3:
## Summary
Contains the chat transcript within the viewport on narrow screens by adding a transcript-scoped `min-width: 0` / `overflow-wrap: anywhere` policy in `index.css`, making display math and markdown tables horizontally scrollable, and adding `min-w-0` / `truncate` / `shrink-0` classes across the header, composer and message components. Eleven jsdom tests are added.

## Verdict
Request changes.

## Required changes
1. Add `flex-wrap` to the edit-mode action row at `frontend/src/components/Message.jsx:681` and `shrink-0 whitespace-nowrap` to its Cancel and Send buttons — the row's cleared minimum lets both labels break mid-word at 320–375px, the same collapse already fixed in `ToolApprovalMessage.jsx`.
2. Apply the same `flex-wrap` + `shrink-0 whitespace-nowrap` treatment to the agent-question row at `frontend/src/components/ChatArea.jsx:851`, and give its input a legible floor (`min-w-[12rem]`) — the input keeps `min-width: auto` while the Send button absorbs the whole shortfall at 320px.
3. Move the table scroller onto a wrapper instead of the table at `frontend/src/index.css:666` (e.g. a `renderer.table` override in `frontend/src/utils/markdownRenderer.js:59` emitting `<div class="table-scroll">`), keeping `display: table; width: 100%` — `display: block` both drops implicit table semantics for screen readers and stops tables filling the message column on desktop, contradicting the description's stated 1280px invariant.
4. Give the new scroll containers at `frontend/src/index.css:660` and `:666` `tabindex="0"` plus `role="region"` and an accessible name — as written, content clipped past the right edge of a long equation or wide table is unreachable by keyboard (WCAG 2.1.1), and the transcript's `overflow-x-hidden` removes the outer fallback scroller.
5. Anchor the CSS assertions in `frontend/src/test/message-viewport-containment.test.jsx:180` to the rule block rather than matching across the file — the code-block carve-out test still passes when `overflow-wrap: normal` at `index.css:650` is flipped to `anywhere`, because the lazy bridge matches the new table-cell rule instead.
6. Add `shrink-0 whitespace-nowrap` to the code-block copy button at `frontend/src/utils/markdownRenderer.js:154` and `min-w-0 break-all` to the language span at `:152` — a long model-supplied fence info string squashes the Copy label by the same mechanism.
7. Restore a small mobile floor (`min-w-[5rem]`) on the model selector at `frontend/src/components/Header.jsx:205` and add `title` / `aria-label` — replacing `min-w-[80px]` with `min-w-0` leaves the active model unlabelled for sighted users and unnamed for assistive technology whenever the left header cluster grows.

## Evidence
Evidence: 7/10 — unusually well-measured for a layout change, with before/after numbers at 320/375/1280 against named adversarial content and a stashed-fix check confirming the tests bite. The gaps are that no visual artifact exists for a fix whose whole subject is what the screen looks like, the checked-in tests assert class names and CSS source text rather than layout, and the description plus the 1280px desktop measurement still describe the first round rather than the head commit.

## Notes
- Scope the `overflow-wrap: anywhere` policy at `frontend/src/index.css:640` the same way its carve-outs are scoped; today `.selectable-markdown` outside the transcript would inherit the wrap without the code/math/table exemptions, and `.message-content` has no producer in `src/`.
- Drop `break-all` from the download buttons at `frontend/src/components/Message.jsx:349`, `:372`, `:389` — it splits ordinary hyphenated filenames mid-token where the inherited wrap rule alone would not.
- The final hunk of `frontend/src/index.css:766` converts eleven unchanged lines from LF to CRLF; land that separately, ideally with a `.gitattributes` entry for `*.css`.

---

garland3:
## Summary
Fixes horizontal overflow of the chat transcript and header on phone-width viewports by adding a `.chat-messages`-scoped `min-width: 0` / `overflow-wrap: anywhere` policy in `index.css`, converting `w-full` bubbles to `flex-1 min-w-0`, pinning approval-row buttons with `shrink-0`, adding phone-only scrollers for tables and display math, and checking in 11 regression tests.

## Verdict
Request changes.

## Required changes
1. Make the agent-question answer row wrap at `frontend/src/components/ChatArea.jsx:851` — add `flex-wrap` to the row, `shrink-0 whitespace-nowrap` to the Send button (`:859`) and `min-w-[12rem]` to the input (`:853`). The new `min-width: 0` on `button` combined with `<input>` keeping its ~180px intrinsic minimum collapses Send to a sliver at 320–375px, and the transcript's new `overflow-x-hidden` clips it rather than letting it scroll.
2. Apply the same treatment to the edit-mode action row at `frontend/src/components/Message.jsx:681` — `flex-wrap` on the row and `shrink-0 whitespace-nowrap` on Cancel and Send — since flex shrinkage is proportional and both labels break mid-word at 375px.
3. Change the two remaining `w-full` bubbles at `frontend/src/components/ChatArea.jsx:848` and `:873` to `flex-1 min-w-0`, matching `AgentBusyIndicator.jsx:34`; they overhang the avatar-plus-gap width and the thinking bubble appears every turn.
4. Scope the code-block assertion at `frontend/src/test/message-viewport-containment.test.jsx:181` to the matched rule block instead of lazily bridging to the next `overflow-wrap: normal;` — the new phone table rule at `index.css:682` satisfies it, so flipping `index.css:650` to `anywhere` leaves the test green.
5. Also assert `display: block` in the phone-only table test at `frontend/src/test/message-viewport-containment.test.jsx:195` — `overflow-x: auto` alone on a `display: table` box gives no usable scroller, so removing the load-bearing half passes today.
6. Keep table semantics below 639px: move the scroller to a wrapper via a `renderer.table` override in `frontend/src/utils/markdownRenderer.js` and leave the `<table>` as `display: table; width: max-content; min-width: 100%`. `display: block` on the table itself drops the implicit table/row/cell roles for screen-reader users.
7. Make the new horizontal scrollers keyboard-reachable — give the KaTeX and table scroll containers (`frontend/src/index.css:657`, `:675`) `tabindex="0"`, `role="region"`, an accessible name and a visible focus ring; with `overflow-x-hidden` on the transcript they are now the only route to the clipped content.

## Evidence
Evidence: 7/10 — the prose measurements are unusually rigorous (before/after `scrollWidth` at 320/375/1280 against named adversarial content, three quoted green test runs, a stash-the-fix mutation check), but a purely visual change ships with no screenshot, and the two head commits are described only in the CHANGELOG. The single most useful missing artifact is a 375px before/after image of the transcript with the adversarial content already measured.

## Notes
- Drop `break-all` from the download buttons at `frontend/src/components/Message.jsx:349`, `:372`, `:389`; the inherited `overflow-wrap: anywhere` already contains them without splitting ordinary filenames mid-token.
- `@media (max-width: 639px)` at `frontend/src/index.css:674` leaves a gap against `sm` (`min-width: 640px`) at fractional widths — use `639.98px`.
- The PR description still describes round one (`Header.jsx` "both groups are `min-w-0`", tables capped at `max-width: 100%`); refresh it against head before merge.

---

PR #758, opened 2026-08-09
Add SKIP_AUTHORIZATION_CHECKS dev-only flag to reduce local setup friction

### LINKED ISSUE #757: Add DEBUG_MODE-gated skip_authorization_checks flag for local/dev setup
## Problem

New contributors running ATLAS locally currently have to manually add themselves to the admin group (by configuring `ADMIN_TEST_USER` to match their own email, or using the hardcoded `test@test.com` identity) before they can access admin-gated functionality, even with `DEBUG_MODE=true` set.

This is because `debug_mode` already provides a mock-authorization fallback in [`atlas/core/auth.py`](atlas-ui-3/atlas/core/auth.py) (`is_user_in_group`), but that fallback only recognizes a small hardcoded set of identities (`test_user`, `admin_test_user`, and two entries in a `mock_groups` dict) — not whatever real email/identity a new developer's environment produces. The result is unnecessary local setup friction for something that should be a one-line dev convenience.

## Proposed solution

Add a new setting, `skip_authorization_checks` (env var `SKIP_AUTHORIZATION_CHECKS`), defaulting to `False`:

- Only takes effect when `debug_mode` is also `True`. It has **no effect on its own** — it is not an independent bypass.
- Bypasses **authorization** (`is_user_in_group` returns `True` for any user/group) but **not authentication** — `AuthMiddleware`/`get_current_user` identity resolution is unchanged, so audit logs still show who did what.
- **Fails fast at startup** (via a pydantic `model_validator` on `AppSettings`, mirroring the existing `validate_agent_portal_dev_only` pattern) if `skip_authorization_checks=True` while `debug_mode=False`.
- Logs a loud startup warning when active, similar to the existing production-security-warning block in `atlas/main.py`.

### Implementation sketch

In `atlas/core/auth.py::is_user_in_group`, after the existing `if not app_settings.debug_mode: return False` guard:

```python
if app_settings.skip_authorization_checks:
    return True  # dev convenience: any group, any user — DEBUG_MODE only
```

## Guardrails (must never be usable in production)

1. Hard dependency on `debug_mode` enforced by a startup validator, not just documentation.
2. Defaults to `False` — strictly opt-in.
3. Only affects authorization, never authentication.
4. Loud startup warning when enabled.
5. `.env.example` should list it commented-out with an explicit "local dev only, never enable in production" warning, distinct from `DEBUG_MODE=true` which ships enabled by default in the example.
6. No UI toggle — env-var only, so it can't be flipped at runtime through the running app.

## Tests to update

- `atlas/tests/test_core_auth.py`: assert `skip_authorization_checks=True` grants an arbitrary user arbitrary groups when `debug_mode=True`; assert startup fails (or the flag is ignored, per final design decision) when `debug_mode=False`.
- `atlas/tests/test_security_admin_routes.py`: assert `require_admin` passes for a non-configured user when the flag is set.

## Docs to update

- `docs/admin/authentication.md`: add a subsection alongside the existing `DEBUG_MODE` documentation, with an explicit local-only warning, positioning this as the recommended way to skip the `ADMIN_TEST_USER`-matching step during local setup.
- `atlas-ui-3/.env.example`: add the new var (commented out, not enabled by default) near `DEBUG_MODE` with a warning comment.
- `atlas/init_cli.py`: optionally add the new var commented-out in the scaffolded `.env` generator.

## ⚠️ Production warning

This flag must never be enabled in production. It is gated entirely behind `DEBUG_MODE`, which is already documented and enforced as unsafe for production use. The startup validator is the primary guardrail — implementation should not proceed without it.

garland3:
## Summary
- New local users had to configure `ADMIN_TEST_USER` to match their own email before reaching admin-gated routes, even with `DEBUG_MODE=true` set — because the debug-mode mock authorization table only recognizes two hardcoded identities.
- Adds `skip_authorization_checks` (env `SKIP_AUTHORIZATION_CHECKS`), which makes `is_user_in_group` return `True` for any user/group. It is strictly opt-in (`False` by default) and only takes effect when `debug_mode=True`.
- A new `AppSettings` startup validator (`validate_skip_authorization_checks_dev_only`, mirroring the existing `validate_agent_portal_dev_only` pattern) makes the app **refuse to start** if `SKIP_AUTHORIZATION_CHECKS=true` while `DEBUG_MODE=false`, and logs a loud warning whenever the flag is active.
- Only authorization is bypassed — authentication (identity resolution in `AuthMiddleware`/`get_current_user`) is unchanged, so audit logs still show the real caller.
- Documented in `docs/admin/authentication.md`, `.env.example`, and the `atlas-init` scaffolded `.env` template (`atlas/init_cli.py`), all commented-out/off by default.

Closes #757

## Test plan
- [x] `atlas/tests/test_core_auth.py`: new tests cover the bypass granting an unconfigured user any group, and the startup failure when `SKIP_AUTHORIZATION_CHECKS=true` without `DEBUG_MODE=true`.
- [x] `atlas/tests/test_security_admin_routes.py`: new test confirms an admin route is reachable for a user with no admin configuration when the flag is set.
- [x] `python -m pytest atlas/tests/ -k "auth or admin or settings or config"` — 428 passed, 9 skipped.
- [x] `ruff check` on changed files — clean (one pre-existing unrelated import-order warning in `test_security_admin_routes.py`, not touched by this change).


garland3:
## Summary
Adds an opt-in `SKIP_AUTHORIZATION_CHECKS` flag that makes `is_user_in_group` return `True` for every group when `DEBUG_MODE=true`, with a startup validator that refuses to boot if the flag is set outside debug mode, plus tests and docs.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. Proportionate for a +113/-0 backend/config change — the bypass, the refusal-to-boot, and the admin-route effect are all asserted by the three added tests, and the description names the test command and its result. The one artifact worth adding is proof the guardrail surfaces as an actual startup abort rather than only as `AppSettings()` raising in isolation, since the boot path wraps settings construction in a retry-on-exception handler (`atlas/modules/config/config_loader.py:121-129`).

## Notes
- The bypass sits above the external-authorization branch, so a shared/staging instance running `DEBUG_MODE=true` with `AUTH_GROUP_CHECK_URL` configured has that service silently defeated — consider refusing the flag when an auth endpoint is set, or when the bind host is not loopback (`atlas/core/auth.py:43`; `docker-compose.yml:71` ships `ATLAS_HOST=0.0.0.0`).
- The new admin-route test's `monkeypatch.setenv("DEBUG_MODE", "true")` is a no-op — `AuthMiddleware` snapshots `debug_mode` at app construction, which happens at import — so the file passes only when `DEBUG_MODE` is already set in the shell (503 otherwise); set it before import in a fixture or `conftest.py` (`atlas/tests/test_security_admin_routes.py:64`).
- Untested branches worth pinning: the default-off value of the field, the `caplog` warning/error the docs advertise as the operator signal, and the `app_settings.debug_mode and` conjunct in the bypass condition — removing any of them leaves the suite green (`atlas/tests/test_core_auth.py:22`).

---

garland3:
## Summary
Adds an opt-in `SKIP_AUTHORIZATION_CHECKS` setting that makes `is_user_in_group` return `True` for every caller, gated behind `DEBUG_MODE=true` by a new startup validator, with docs and `.env` templates updated. The same diff also lands test-infrastructure changes: an autouse `_isolate_config_cache` fixture and a shared `mock_admin_authorization` fixture applied to five existing modules.

## Verdict
Request changes.

## Required changes
1. Move the bypass below the external group-check branch at `atlas/core/auth.py:43` (or reject the flag in the validator when `auth_group_check_url` is set) — as written it takes precedence over a configured authorization service, so an instance running `DEBUG_MODE=true` against a real group-check endpoint silently grants every group without ever calling it.
2. Neutralize the flag for the test session in `atlas/tests/conftest.py:47` (e.g. `os.environ.pop("SKIP_AUTHORIZATION_CHECKS", None)` before app import) — `atlas/main.py:92` loads `../.env` into `os.environ` at import time, so a developer who follows the new docs and sets the flag in `.env` gets 19 failures, including every "non-admin is denied" assertion and any test constructing `AppSettings()` without pinning `DEBUG_MODE=true`.
3. Add a production-mode negative test in `atlas/tests/test_core_auth.py` asserting `is_user_in_group(admin_test_user, admin_group)` is `False` with `DEBUG_MODE=false` and no external endpoint configured — the five new `pytestmark` call-sites now rest on that invariant, and deleting the `if not app_settings.debug_mode: return False` guard currently goes unnoticed in both CI legs.
4. Pin the boot-path refusal, not just the model: assert `ConfigManager().app_settings` raises with `SKIP_AUTHORIZATION_CHECKS=true` / `DEBUG_MODE=false` (`atlas/tests/test_core_auth.py:39`) — `config_loader.py:118-129` catches `Exception` and retries `AppSettings()` as a "fallback", so the documented refuse-to-start guarantee holds only incidentally today.

## Evidence
Evidence: 7/10. The feature itself is well pinned — a test per claimed behavior, plus the exact verification command and result in the description — but the test-infrastructure half is undescribed and unevidenced: the fixture docstring's claim that config leakage was the only reason ~39 admin-route tests passed in the `DEBUG_MODE=false` leg has no output behind it. The missing artifact is a full-suite run in both `DEBUG_MODE` legs, since the reported run was a `-k` subset that does not exercise the autouse fixture's blast radius.

## Notes
- `_CONFIG_CACHE_ATTRS` (`atlas/tests/conftest.py:89`) hand-mirrors the seven caches cleared by `reload_configs()`; snapshot `__dict__` generically so an eighth config can't silently drift.
- The refusal path surfaces as two stack traces plus a misleading "Create default settings as fallback" log rather than a clean abort (`atlas/modules/config/settings.py:610`); the error message also states the rule without the remedy.
- `test_security_admin_routes.py:71` mixes the new `pytestmark` idiom with a hand-rolled `monkeypatch.setenv` + `reload_configs()` + `try/finally` in the test right below it.

---

garland3:
## Summary
Adds a `SKIP_AUTHORIZATION_CHECKS` development flag that makes every group check
return `True` when `DEBUG_MODE=true`, guarded by a startup validator that refuses
to boot otherwise, plus docs, `.env` templates, and a set of test-isolation
fixtures (`_isolate_config_cache`, `mock_admin_authorization`,
`skip_auth_checks_env`) applied across five existing test modules.

## Verdict
Request changes.

## Required changes
1. Move the bypass at `atlas/core/auth.py:43` below the `if auth_url and api_key:`
   branch so it can only override the mock table — as written it also short-circuits
   a fully configured `AUTH_GROUP_CHECK_URL` authorizer, which is broader than the
   documented intent and defeats real authorization on any debug-mode instance that
   has one configured. Alternatively, reject the combination in
   `validate_skip_authorization_checks_dev_only` when `auth_group_check_url` is set.
2. Register teardown before mutating the environment in `skip_auth_checks_env`
   (`atlas/tests/conftest.py:176`) — use `monkeypatch.setenv`, or move the writes
   and the two asserts inside the `try`. Today a failure in the setup block skips
   the `finally`, leaving `DEBUG_MODE=true` and `SKIP_AUTHORIZATION_CHECKS=true`
   set for the rest of the session, which turns the suite's denial assertions into
   vacuous passes.
3. Add a production-mode negative test in `atlas/tests/test_core_auth.py`: with
   `DEBUG_MODE=false` and no `AUTH_GROUP_CHECK_URL`, assert
   `is_user_in_group(admin_test_user, admin_group)` is `False`. The
   `if not app_settings.debug_mode: return False` guard at `atlas/core/auth.py:70`
   is the invariant the five new `pytestmark` call-sites rely on, and deleting it
   currently leaves both CI legs green.
4. Pop `SKIP_AUTHORIZATION_CHECKS` from `os.environ` in `atlas/tests/conftest.py:47`
   before `main` is imported — `atlas/main.py:92` loads `../.env` at import time,
   and the new docs and `.env.example` instruct developers to set the flag in exactly
   that file, so a developer who follows them runs the whole suite with authorization
   off.
5. Update the `is_user_in_group` docstring (`atlas/core/auth.py:43`) and the
   guardrails list in `docs/admin/authentication.md:198-208` to state the actual
   precedence and the actual posture: with the flag on, an unauthenticated request
   is an administrator, since debug mode already assigns `test_user` to header-less
   requests (`atlas/core/middleware.py:134-145`).

## Evidence
Evidence: 6/10. The feature itself is proportionately proven — one test per claimed
behavior plus a quoted verification run — but the test-infrastructure half of the diff
is absent from the description and its central claim (that config leakage was the only
reason ~39 admin-route tests passed in the `DEBUG_MODE=false` leg) has no output behind
it. The most valuable missing artifact is a full-suite run under each `DEBUG_MODE` leg,
since the new autouse fixture affects every test, not the `-k`-filtered subset that was
run.

## Notes
- The "refuses to start" guarantee holds only incidentally: `ConfigManager.app_settings`
  (`atlas/modules/config/config_loader.py:121-129`) catches the `ValueError`, logs
  "Create default settings as fallback", and retries — two tracebacks and a misleading
  recovery message instead of one clean abort.
- `_CONFIG_CACHE_ATTRS` (`atlas/tests/conftest.py:88`) hand-mirrors the attributes
  cleared by `reload_configs()`; an eighth cached config silently stops being isolated.
- The validator keys only on `debug_mode` and ignores `environment`
  (`atlas/modules/config/settings.py:619`), so `ENVIRONMENT=production` with
  `DEBUG_MODE=true` boots with authorization disabled.

---

garland3:

garland3:
## Summary
Adds an opt-in `SKIP_AUTHORIZATION_CHECKS` flag that short-circuits `is_user_in_group` when `DEBUG_MODE=true`, guarded by a startup validator, plus documentation and a set of new pytest fixtures (`_isolate_config_cache`, `mock_admin_authorization`, `skip_auth_checks_env`) applied across five test modules.

## Verdict
Request changes.

## Required changes
1. Pop `SKIP_AUTHORIZATION_CHECKS` (and `DEBUG_MODE`) from `os.environ` in `atlas/tests/conftest.py:47`, next to the existing `env_file` guard and before `main` can be imported — `atlas/main.py:92` calls `load_dotenv("../.env")`, so a developer who follows the new instructions in `.env.example` and `docs/admin/authentication.md` runs the suite with authorization off (measured: 5 failures across `test_security_admin_routes.py`, `test_model_access_control.py`, and every 200-expecting assertion passing vacuously).
2. Move the environment writes, the `reload_configs()` call, and both asserts in `skip_auth_checks_env` inside the `try:` at `atlas/tests/conftest.py:176`, or switch to `monkeypatch.setenv` — today a failure during setup skips the `finally` and leaves the bypass enabled for the rest of the session.
3. Move the bypass at `atlas/core/auth.py:43` into the `else:` branch so it can only replace the mock table, or reject the combination in `validate_skip_authorization_checks_dev_only` when `auth_group_check_url` is set — as written it silently overrides a fully configured external authorizer, which contradicts both the docstring at `atlas/core/auth.py:22-33` and the documented motivation.
4. Extend the validator at `atlas/modules/config/settings.py:619` to also fail when `environment` is not a development value — `ENVIRONMENT=production DEBUG_MODE=true SKIP_AUTHORIZATION_CHECKS=true` currently boots with all authorization disabled and only a warning.
5. Correct `docs/admin/authentication.md:202` — "you still need a valid identity" is not a residual control, because `AuthMiddleware` (`atlas/core/middleware.py:132-145`) assigns `TEST_USER` to unauthenticated requests in debug mode, so with the flag on any request reaching the port is an administrator. State that, and that the blast radius covers restricted MCP servers and group-restricted models, not just admin routes.
6. Add a test in `atlas/tests/test_core_auth.py` asserting `is_user_in_group` returns `False` when `skip_authorization_checks` is true and `debug_mode` is false (settings injected, since `AppSettings()` refuses that combination) — deleting the `debug_mode` conjunct from the bypass currently leaves the suite green.
7. Add a test asserting `config_manager.app_settings` raises under `SKIP_AUTHORIZATION_CHECKS=true DEBUG_MODE=false` — the existing test pins `AppSettings()` directly, one layer below the boot path, which catches `Exception` and retries (`atlas/modules/config/config_loader.py:118-129`).

## Evidence
Evidence: 6/10. The flag, validator, and docs are proportionately demonstrated for a backend change, but the test-infrastructure half — 129 of 248 added lines, including an autouse fixture affecting every test — is absent from the description, and the only run reported is a `-k`-filtered subset. The most important missing artifact is a full-suite run in both `DEBUG_MODE` legs against the merged head commit.

## Notes
- Startup error at `atlas/modules/config/settings.py:610-621` states the rule without a remedy or doc link, unlike the sibling validator it mirrors.
- Boot failure surfaces as two tracebacks plus a misleading "Create default settings as fallback" line (`atlas/modules/config/config_loader.py:126`).
- `_CONFIG_CACHE_ATTRS` at `atlas/tests/conftest.py:88` hand-mirrors `reload_configs()` with no drift detection; deriving it from `__dict__` would keep the two in sync.

---

garland3:
## Summary
Adds a dev-only `SKIP_AUTHORIZATION_CHECKS` flag that makes `is_user_in_group` return `True` for everyone, gated by three startup guardrails (`DEBUG_MODE=true`, `ENVIRONMENT != production`, no `AUTH_GROUP_CHECK_URL`), plus documentation and test-isolation fixtures.

## Verdict
Request changes.

## Required changes

1. Add `ENVIRONMENT=development` to the scaffolded `.env` template at `atlas/init_cli.py:110` — `AppSettings.environment` defaults to `"production"` (`atlas/modules/config/settings.py:514`), so a user who uncomments the `SKIP_AUTHORIZATION_CHECKS=true` line the template itself adds gets a hard boot refusal instead of the documented shortcut. Update the same comment to name all three guardrails, not just `DEBUG_MODE`.
2. Pop the bypass variables from `os.environ` in `atlas/tests/conftest.py:47`, before any test can import `main` — e.g. `for _k in ("SKIP_AUTHORIZATION_CHECKS", "AUTH_GROUP_CHECK_URL", "AUTH_GROUP_CHECK_API_KEY"): os.environ.pop(_k, None)`. The existing `env_file = None` guard does not cover `load_dotenv(dotenv_path="../.env")` at `atlas/main.py:92`, so a developer who follows the new docs runs the whole suite with authorization disabled and the denial assertions in `test_security_admin_routes.py` and `test_model_access_control.py` pass vacuously.
3. Make the environment guardrail at `atlas/modules/config/settings.py:631` an allowlist rather than an exact-match denylist — require `self.environment.strip().lower()` to be in `{"dev", "development", "local", "test"}`. Today `ENVIRONMENT=prod`, `staging`, or a typo all satisfy the check and boot with authorization off, and the rest of the app (`otel_config.py:219`) already treats the field as an allowlist.
4. Add a test covering the boot-path refusal at `atlas/modules/config/config_loader.py:125` — assert `config_manager.app_settings` raises under `SKIP_AUTHORIZATION_CHECKS=true` / `DEBUG_MODE=false`. The three existing tests construct `AppSettings()` directly, one layer below the new clause, so deleting it leaves the suite green while the documented "refuses to start" guarantee degrades.
5. Log the bypass at request time in `atlas/core/auth.py:54` — a single startup warning is the only current signal, so an audit-trail admin action granted by the bypass is indistinguishable from one that passed a real group check.
6. State the real blast radius in `docs/admin/authentication.md:198` and `docs/getting-started/installation.md:100` — the flag unlocks every group-restricted model (`atlas/core/model_access.py:58`), every MCP server gated by `required_groups` (`mcp_execution.py:101`), and the feedback/capture routes, not just admin pages.
7. Fix `atlas/tests/test_core_auth.py:39` so it pins the guardrail it names — set `ENVIRONMENT=development`, delete `AUTH_GROUP_CHECK_URL`, and match on `DEBUG_MODE=true`; the current `match="SKIP_AUTHORIZATION_CHECKS"` substring appears in all three refusal messages, so the test stays green if the debug-mode branch is deleted.

## Evidence
Evidence: 6/10. The flag and its three guardrails each have a dedicated test, but the PR description and its "428 passed" run predate the head commit, which also adds two guardrails, a boot-path change, and an autouse fixture touching every test in the suite; the most important missing artifact is an unfiltered `pytest atlas/tests/` run at this commit, ideally one per `DEBUG_MODE` leg.

## Notes
- `except ValueError: raise` at `config_loader.py:125` also intercepts ordinary `ValidationError`s (bad `PORT`, bad `LOG_LEVEL`), which now propagate with no error log at all.
- Module-scope `pytestmark` in `test_model_access_control.py:29` forces debug mode on pure-helper tests that never touch `core.auth`; scope it to the `TestClient` tests.
- `_isolate_config_cache` (`conftest.py:112`) restores snapshotted attributes but never deletes ones created during a test, so the docstring's "either direction" claim is stronger than the code.

---

garland3:
## Summary
Adds an opt-in `SKIP_AUTHORIZATION_CHECKS` flag that makes `is_user_in_group` grant every group, gated by three startup guardrails (debug mode, non-production `ENVIRONMENT`, no `AUTH_GROUP_CHECK_URL`), plus docs, `.env` template updates, and new pytest fixtures applied across five existing test modules.

## Verdict
Request changes.

## Required changes
1. Invert the environment guardrail at `atlas/modules/config/settings.py:631` to an allowlist — `self.environment.strip().lower()` must be in an explicit development set (e.g. `{"dev", "development", "local", "test"}`) — because the current `== "production"` equality lets `prod`, `staging`, `qa`, `Production ` (trailing space) and any typo boot with authorization fully bypassed, and `atlas/core/otel_config.py:220` already reads the same field as an allowlist. Add a test asserting `ENVIRONMENT=staging` is refused.
2. Clear the authorization-relevant variables from the process environment in `atlas/tests/conftest.py:47`, before any test can import `main` — `for _k in ("SKIP_AUTHORIZATION_CHECKS", "AUTH_GROUP_CHECK_URL", "AUTH_GROUP_CHECK_API_KEY"): os.environ.pop(_k, None)` — because `atlas/main.py:92` calls `load_dotenv("../.env")` at import time, so a developer following the new docs runs the suite with the bypass on. Measured at this commit: 6 denial tests fail (`test_admin_routes_require_admin`, `test_system_status_requires_admin`, three in `test_model_access_control.py`, `test_is_user_in_group_denied_in_production_mode`) and the positive assertions pass vacuously.
3. Fix `test_admin_routes_allow_any_user_with_skip_authorization_checks` at `atlas/tests/test_security_admin_routes.py:73` so it builds its app after debug mode is set — `AuthMiddleware` snapshots `debug_mode` at construction (`atlas/core/middleware.py:81`), so with `DEBUG_MODE` unset the request 503s before the bypass is reached; this is the one net-new suite failure the PR introduces in that leg.
4. Add `ENVIRONMENT=development` to the scaffolded template at `atlas/init_cli.py:110` and expand the comment to name all three guardrails, because `AppSettings.environment` defaults to `"production"` and uncommenting the template's own `SKIP_AUTHORIZATION_CHECKS=true` line therefore aborts startup.
5. Correct or make true the claim in `mock_admin_authorization` (`atlas/tests/conftest.py:139`) and the five copied `pytestmark` comments that the marked tests "pass in both CI legs" — of the 163 pre-existing failures with `DEBUG_MODE` unset, the fixture fixes zero, for the same import-time snapshot reason as item 3.
6. Log before re-raising in the `except ValueError` branch at `atlas/modules/config/config_loader.py:125` (and narrow it to the guardrail refusals, e.g. a dedicated exception type), because `ValidationError` subclasses `ValueError`, so a mistyped `PORT` or `LOG_LEVEL` now aborts with no `Failed to load application settings` record at all.
7. Add a `caplog` assertion covering the bypass `logger.warning` at `atlas/core/auth.py:58` — `docs/admin/authentication.md` promises this per-request audit line, and deleting the call leaves the suite green.

## Evidence
Evidence: 6/10. The feature's validators are reasonably pinned, but the reported test plan is a filtered `-k` run made against an earlier revision and does not cover the test-infrastructure half of the diff; the most important missing artifact is an unfiltered `pytest atlas/tests/` run at this commit in both the `DEBUG_MODE=true` and `DEBUG_MODE=false` legs.

## Notes
- The per-check WARNING logs a user email on every membership check and fans out to one line per (model × group) and (server × group) — `atlas/core/auth.py:54`.
- `_isolate_config_cache` restores snapshotted attributes but never deletes ones created during a test, so the docstring's "cannot leak in either direction" is stronger than the code — `atlas/tests/conftest.py:112`.
- Module-scope `pytestmark` forces a config reload onto pure-helper tests that never reach the auth layer — `atlas/tests/test_model_access_control.py:26`.

---

PR #763, opened 2026-08-09
Collapse tool calls to one row and one line (#762)

### LINKED ISSUE #762: Tool calls take two rows each: drop the redundant auto-approved row and collapse to one line
## Problem

On mobile, a single tool call consumes roughly two rows of vertical space and, in practice, about half the screen for a two-call turn. Each tool call renders as **two separate messages**:

1. `tool_approval_request` -> `ToolApprovalMessage.jsx:285` — `▶ [AUTO-APPROVED] basic_fns_bash · 2 params`, plus an `Auto-approve ON` pill that wraps onto its own line at phone widths
2. `tool_call` -> `Message.jsx:210` — `▶ [SUCCESS] basic_fns_bash (basic_fns) · 2 params`, which also wraps to two lines

With `space-y-4` between transcript children, two `basic_fns_bash` calls take ~10 lines and four inter-message gaps, pushing the actual assistant answer off screen.

The first message is nearly information-free once the second arrives: `SUCCESS` supersedes `AUTO-APPROVED` within about a second, and the two rows name the same tool.

## Why the redundant row exists

`Message.jsx:183` states the reason:

> Note: Tool auto-approval handled inside ToolApprovalMessage; we keep message visible so inline toggle remains accessible.

The `Auto-approve ON/OFF` pill (`ToolApprovalMessage.jsx:140`) is the only in-transcript way to turn auto-approve back off, so an entire message is being kept alive to host one control — and it is repeated for every call in the turn.

## Proposal

**1. Do not render the approval message when the call was auto-approved.**
The `APPROVAL REQUIRED` path stays exactly as it is today — the review/approve/reject/edit-args UI is valuable and should not change. This only removes the row that says "we approved a thing we are about to report anyway."

**2. Rehome the auto-approve toggle.**
Preferred: put it on the `Active Tools:` strip above the composer. That strip is already the persistent per-conversation status bar, it stays visible instead of scrolling away, and one indicator replaces one per call. Fallback: render the inline pill only on the first auto-approved call of a turn.

**3. Slim the surviving `tool_call` row so it fits one line at 390px.**
It currently cannot: badge + tool name + `(server_name)` + `· N params`. In the collapsed state:
- drop `(server_name)` (`Message.jsx:235`) — it is usually a prefix of the tool name (`basic_fns` beside `basic_fns_bash`)
- drop `· N params` (`Message.jsx:237`) — the disclosure triangle already signals expandability
- replace the `SUCCESS`/`FAILED` text pill with a colored glyph

Target:

```
▶ ✓ basic_fns_bash
▶ ✓ basic_fns_bash
```

Two calls in three lines instead of ten. All detail stays one tap away — `Message.jsx:191` already expands input arguments and output together in the same row, so no new plumbing is needed to "put the output in the same box"; that is already the behavior once the redundant sibling row is gone.

## Related defect to fix in the same pass

`autoApproved` is derived from **live** settings, not from what happened at approval time:

```js
// ToolApprovalMessage.jsx:27
const autoApproved = Boolean(settings?.autoApproveTools && !message.admin_required)
```

So collapsing rows based on this value makes old transcripts re-render differently when the setting is later toggled — a historical auto-approved call would reappear as a separate `APPROVED` row once auto-approve is switched off. The approval decision is history and should be persisted on the message, with rendering keyed off that field.

## Out of scope for now

Collapsing consecutive identical calls into `✓ basic_fns_bash ×2` is a larger win when an agent makes many calls, but it hides per-call status. Worth revisiting after the above lands.

## Notes

Behavior confirmed on `8286331` by reading the components; the screenshot that prompted this shows two `basic_fns_bash` calls occupying the space between a one-line user message and a one-line assistant reply.

garland3:
Fixes #762

## Problem

An auto-approved tool call rendered **two** transcript rows:

1. `tool_approval_request` — `▶ [AUTO-APPROVED] basic_fns_bash · 2 params`, plus an `Auto-approve ON` pill that wrapped onto its own line at phone widths
2. `tool_call` — `▶ [SUCCESS] basic_fns_bash (basic_fns) · 2 params`, which also wrapped

With `space-y-4` between children, two `basic_fns_bash` calls took ~10 lines and pushed the assistant's answer off screen. The first row is nearly information-free once the second arrives.

## Changes

**1. No approval row for auto-approved calls.** `ToolApprovalMessage` returns `null` when the call was auto-approved, and `Message` renders it inside a `display:none` wrapper — the component must stay mounted because it owns the effect that sends the approval, but a hidden child contributes no height and no `space-y-4` gap. The `APPROVAL REQUIRED` review/approve/reject/edit-args path is unchanged.

**2. Auto-approve toggle rehomed** to the `Active Tools:` strip above the composer (the preferred option in the issue). One persistent indicator replaces one pill per call, and it no longer scrolls away with the transcript.

**3. The compact `tool_call` row now fits one line at 390px**: the server name is dropped (usually a prefix of the tool name — `basic_fns` beside `basic_fns_bash`) and moves into the expanded detail; the param count is dropped (the disclosure triangle already signals expandability); and the `SUCCESS`/`FAILED` text pill becomes a colored glyph (`✓` / `✗`) carrying the old label as `aria-label`/`title`. Result: `▶ ✓ basic_fns_bash`. Active calls keep the spinner and elapsed-time readout. The classic (compact-off) row is unchanged.

**4. The related defect.** `autoApproved` was derived from live settings rather than from what happened at approval time, so collapsing rows based on it would make old transcripts re-render differently once the setting was toggled. The decision is now persisted on the message as `auto_approved` when it is made (`true` from the auto-approve effect, `false` on a manual approve/reject) and rendering keys off that field through a shared `resolveAutoApproved` helper in `src/utils/toolApproval.js`. The live setting is consulted only for a still-pending row that has no recorded decision yet; a resolved row without the field predates it and renders as a manual approval.

Out of scope, as the issue notes: collapsing consecutive identical calls into `✓ basic_fns_bash ×2`.

## Notes

Removing the auto-approved branch made the `toolApprovalArgsCollapsed` persisted preference dead — the only rows the component now renders are review-required ones, which always open expanded — so it was dropped along with its `usePersistentState` import.

## Verification

- `npx vitest run` — 505 passed (48 files), including the `#747` viewport-containment assertions on the rows touched here
- `npx eslint src/` — no errors (4 pre-existing warnings, none in changed files)
- `npm run build` — clean
- New tests: auto-approved request renders nothing but still sends the approval and records `auto_approved: true`; a persisted auto-approved call stays hidden after the setting is toggled off; an admin-required call still shows `APPROVAL REQUIRED` with auto-approve on; the approval row no longer hosts the toggle; the compact row has no status pill / server name / param count and exposes a labelled `✓`/`✗`; `Message` wraps a hidden auto-approved request in `.hidden` with no text


garland3:
## Summary
Collapses the tool-approval transcript to a single line per call: auto-approved approval rows are hidden (while staying mounted so the approval still sends), the compact tool-call row drops its status pill, server name and parameter count in favour of a labelled glyph, and the per-row auto-approve toggle is rehomed onto the `Active Tools:` strip. A new `resolveAutoApproved` helper keys rendering off a persisted `auto_approved` field instead of the live setting.

## Verdict
Approve with nits.

## Evidence
Evidence: 6/10. The behavioral claims are well covered by seven new or rewritten tests and the description is unusually specific, but the headline justification is a rendered-layout claim ("fits one line at 390px") and no screenshot at that viewport is attached — a before/after pair of a two-call transcript at phone width is the missing artifact.

## Notes
- `auto_approved` is not persisted — `buildPersistedMessage` writes metadata only for `tool_call` rows (`frontend/src/utils/chatExport.js:47`), so after save/reload an auto-approved call renders a stray `APPROVED` pill with a blank tool name, restoring the two-row layout this PR removes.
- The rehomed toggle is gated on a non-empty tool selection — `EnabledToolsIndicator` returns `null` at `frontend/src/components/EnabledToolsIndicator.jsx:16`, so the "persistent" auto-approve indicator disappears whenever no tools are selected, and it has no test.
- `resolveAutoApproved` returns a persisted `auto_approved: true` before checking `admin_required` (`frontend/src/utils/toolApproval.js:8`); an admin-required pending row carrying that field renders no approval prompt at all. It fails closed (the call is not executed) but stalls with no visible row.

---

garland3:
## Summary
Collapses the tool-approval transcript rows into a single compact line: auto-approved approval requests now render nothing, the `tool_call` row drops its status pill / server name / param count, the auto-approve toggle moves to the `Active Tools:` strip, and the approval decision is persisted so reloaded transcripts render consistently.

## Verdict
Request changes.

## Required changes
1. Stop persisting a live `status: 'pending'` for `tool_approval_request` rows in `frontend/src/utils/chatExport.js:55` (or normalize it on restore) — auto-save fires ~1s after the row appears, so a conversation saved mid-approval reloads as pending and, with auto-approve on, emits a real `tool_approval_response {approved: true}` 100ms later for a `tool_call_id` the backend no longer has.
2. Mark restored messages in `frontend/src/contexts/ChatContext.jsx:564` (e.g. `_restored: true`) and have the auto-approval effect at `frontend/src/components/ToolApprovalMessage.jsx:52` refuse to act on them — a restored pending row currently also renders live Approve/Reject controls bound to a dead `tool_call_id`, and clicking Approve rewrites archived history to `APPROVED` with nothing executed.
3. Handle approval rows restored from conversations saved before this change in `frontend/src/components/ToolApprovalMessage.jsx:248` — they carry no `metadata`, so they render a green `APPROVED` pill with a blank tool name; skip the row when it has no `tool_call_id`/`tool_name`, or fall back to a neutral label when `status` is undefined.
4. Make the admin gate unconditional and first in `frontend/src/utils/toolApproval.js:9` — `if (message?.admin_required) return false` ahead of the persisted-value branch, so a row carrying `auto_approved: true` can never suppress an admin-required prompt.
5. Surface a failed auto-approval send at `frontend/src/components/ToolApprovalMessage.jsx:68` — record `auto_approved: false` (or an equivalent marker) on the `!sent` path so the row falls back to the visible `APPROVAL REQUIRED` prompt instead of staying hidden with no approval in flight.
6. Render the auto-approve control unconditionally in `frontend/src/components/EnabledToolsIndicator.jsx:16` — gate only the chip list on `allTools.length`, so deselecting all tools no longer removes the only always-visible indicator that auto-approve is ON.
7. Add tests for the two untested new surfaces: a unit table over `resolveAutoApproved`'s branches and precedence (`frontend/src/utils/toolApproval.js`), and a render/click test for the rehomed toggle including the empty-selection case (`frontend/src/components/EnabledToolsIndicator.jsx`).

## Evidence
Evidence: 6/10. The written verification is unusually thorough for the size — named commands with outcomes, and seven tests mapping onto the behavioral claims — but the headline claim is a rendered-layout one ("the compact row now fits one line at 390px") with no screenshot; a before/after pair at phone width showing two consecutive auto-approved calls is the missing artifact, and a save→restore→render round-trip test would close the persistence loop the current tests approach from both ends.

## Notes
- Persisting `arguments` and `allow_edit` duplicates the paired `tool_call` row's payload on every save; only the pending path reads them (`frontend/src/utils/chatExport.js:48`).
- `title` on an `<svg>` produces no tooltip — use a `<title>` child so the spinner's status is readable by pointer (`frontend/src/components/Message.jsx:254`, `:295`).
- Dropping `server_name` unconditionally makes like-named tools from different servers indistinguishable; drop it only when it prefixes the tool name (`frontend/src/components/Message.jsx:272`).

---

frontend/src/components/ToolApprovalMessage.jsx
```diff
@@ -33,50 +39,44 @@ const ToolApprovalMessage = ({ message, compact = true }) => {
...
   const [reviewCollapsed, setReviewCollapsed] = useState(false)
-  const isExpanded = needsReview ? !reviewCollapsed : !argsCollapsed
-  const toggleCollapsed = () => {
-    if (needsReview) setReviewCollapsed(c => !c)
-    else setArgsCollapsed(!argsCollapsed)
-  }
+  const isExpanded = !reviewCollapsed
+  const toggleCollapsed = () => setReviewCollapsed(c => !c)
 
   useEffect(() => {
```
garland3, line 64: **[suggestion]** The auto-approval effect still keys only on `settings?.autoApproveTools && !message.admin_required && message.status === 'pending'`. After a successful send the PR persists `auto_approved: true` but leaves `status` as `pending` until the tool lifecycle overwrites it. `sendApprovalResponse` is `sendMessage` from `WSContext`, which is not memoized, so any parent re-render changes the dependency identity, re-runs the effect, and can send a second (or Nth) `tool_approval_response` for the same `tool_call_id` during the window before `tool_start`/`tool_complete`.

**Suggestion:** Gate the effect with `message.auto_approved !== true` (and/or a local "alreadySent" ref) so a successful auto-approve is not re-sent when deps churn. Example: `if (settings?.autoApproveTools && !message.admin_required && message.status === 'pending' && message.auto_approved !== true)`.

frontend/src/components/ToolApprovalMessage.jsx
```diff
@@ -33,50 +39,44 @@ const ToolApprovalMessage = ({ message, compact = true }) => {
...
+        }
       }, 100)
       return () => clearTimeout(timer)
     }
-  }, [settings?.autoApproveTools, message.admin_required, message.status, message.tool_call_id, message.arguments, sendApprovalResponse])
+  }, [settings?.autoApproveTools, message.admin_required, message.status, message.tool_call_id, message.arguments, sendApprovalResponse, updateToolResult])
 
   const handleApprove = () => {
     if (resolvedStatus !== 'pending') return
     setDecision('approved')
```
garland3: **[suggestion]** Manual `handleApprove` (and the matching reject path at line 92) still calls `updateToolResult` with a terminal `status: 'approved'|'rejected'` before confirming `sendApprovalResponse` returned true. The auto-approval path in this same PR deliberately fixed that failure mode (persist only after a successful send) because a stuck terminal state with no in-flight approval is hard to recover from. On a dropped socket, the UI will show APPROVED/REJECTED and drop the controls while the backend never received the response.

**Suggestion:** Mirror the auto path: `const sent = sendApprovalResponse(...); if (sent) { setDecision(...); updateToolResult?.(...); }` (or revert local decision if `sent` is false and surface a toast). At minimum, do not patch `status`/`auto_approved` when the send fails.

frontend/src/components/Message.jsx
```diff
@@ -598,6 +646,14 @@ const Message = ({ message, userIndex = null, onRewind = null, onCorrect = null
     }
   }
 
+  if (isHiddenApprovalRow) {
```
garland3, line 650: **[nit]** The early return uses `<div className="hidden">` and the surrounding comment claims this contributes "no `space-y-4` gap". Tailwind 3's `space-y-*` utilities use the selector `> :not([hidden]) ~ :not([hidden])`, which looks for the HTML `hidden` attribute — not the `hidden` utility class (`display: none`). In the common case (hidden row between two visible messages) spacing still looks fine because `display: none` removes box contribution, but the comment is inaccurate and a leading hidden child can still force margin-top onto the following sibling.

**Suggestion:** Prefer `<div hidden>` or `<div className="hidden" hidden>` so the element is excluded from `space-y` the way the comment describes, or soften the comment to only claim no height.

frontend/src/components/ToolApprovalMessage.jsx
```diff
@@ -33,50 +39,44 @@ const ToolApprovalMessage = ({ message, compact = true }) => {
...
       const timer = setTimeout(() => {
-        sendApprovalResponse({
+        // Only record the decision once the server has actually been told.
+        // `sendApprovalResponse` returns false when the WebSocket isn't open;
+        // persisting `auto_approved: true` before the send completes would
+        // leave the row stuck hidden if the socket dropped during the 100ms
+        // delay — the approval never reached the backend, but rendering keys
+        // off the persisted flag and so the row never comes back to retry.
+        // The live `autoApproveTools` setting still hides the row while the
+        // socket is down, so there is no visible flicker on the success path.
```
garland3: **[suggestion]** When `sendApprovalResponse` returns false, the row stays hidden via the live `autoApproveTools` setting (by design, to avoid flicker) with no user-visible signal that approval never left the client. Recovery depends on the effect re-running after reconnect/re-render; if nothing re-triggers deps, the transcript shows neither an approval prompt nor a tool call and the turn looks stuck until the user toggles Auto-approve off.

**Suggestion:** On `sent === false`, either keep the row visible for manual retry, or show a non-blocking toast/banner ("Auto-approve failed — reconnecting or toggle Auto-approve off to retry"). A local "sendFailed" flag that forces `resolveAutoApproved` false until the next successful send would also work.

garland3 (commented):
## Summary

PR #763 successfully collapses auto-approved tool calls to a single compact `tool_call` row by hiding the redundant approval transcript entry, moving the Auto-approve toggle to the Active Tools strip, and shortening the compact tool-call chrome (glyph + name only). The shared `resolveAutoApproved` helper plus `auto_approved` persistence (including `buildPersistedMessage` for approval rows) correctly addresses the live-settings history re-render bug, and the new Vitest coverage is strong for the main paths. Dominant residual risks are the auto-approval effect still re-firing after a successful send while status stays `pending`, and the manual approve/reject path still recording terminal state before a successful WebSocket send — the same class of failure the auto path was carefully hardened against.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 1

garland3:
## Summary
Compacts the tool-call transcript row (drops the inline server name, param count and per-row auto-approve pill; relocates the server to the expanded detail), rehomes the auto-approve toggle onto the Active Tools strip, and extends conversation persistence so `tool_approval_request` rows round-trip `tool_call_id`, `arguments`, `status`, `admin_required` and `auto_approved` through `metadata`.

## Verdict
Request changes.

## Required changes
1. Stop persisting a live `pending` state for approval rows in `frontend/src/utils/chatExport.js:55` — normalize `status` to a terminal/neutral value (e.g. `unresolved`) on write, or drop `status`/`tool_call_id` for rows that never reached a decision. The restore path spreads `metadata` back onto the message (`frontend/src/contexts/ChatContext.jsx:569`), so opening an archived conversation with auto-approve on emits a real `tool_approval_response {approved: true, tool_call_id, arguments}` 100 ms later for a decision the user never made this session.
2. Mark restored messages (e.g. `_restored: true`) in `loadSavedConversation` at `frontend/src/contexts/ChatContext.jsx:564` and have both the auto-approval effect and the Approve/Reject handlers in `frontend/src/components/ToolApprovalMessage.jsx:104` refuse to send for them — today a restored row renders live Approve/Reject/edit-arguments controls whose click rewrites archived history to `APPROVED` while nothing executed, with fully user-controlled arguments on the wire.
3. Make the admin gate unconditional and first in `frontend/src/utils/toolApproval.js:9` — add `if (message?.admin_required) return false` ahead of the persisted-`auto_approved` branch, so a stored flag cannot suppress an admin-required prompt (today that pairing hides the row entirely and the call stalls with no prompt and nothing in flight).
4. Drop `arguments` and `allow_edit` from `TOOL_APPROVAL_PERSISTED_FIELDS` (`frontend/src/utils/chatExport.js:48`) — the paired `tool_call` row already persists the same payload under the same id, so every autosave duplicates an unbounded, frequently sensitive blob to IndexedDB and server history.
5. Add a save → restore → render round-trip test that runs a live pending approval message through `buildPersistedMessage`, spreads it the way `loadSavedConversation` does, renders with `settings.autoApproveTools: true` and fake timers, and asserts `sendApprovalResponse` is never called; plus a unit table for `resolveAutoApproved` covering the `{auto_approved: true, admin_required: true}` precedence case.
6. Include the relocated server line in the expandability test at `frontend/src/components/Message.jsx:205` (`argCount > 0 || message.result != null || (compactMessages && message.server_name)`) — a zero-argument call with no result renders no triangle, making `server_name` unreachable now that it left the compact row.
7. Move the `allTools.length === 0` early return in `frontend/src/components/EnabledToolsIndicator.jsx:16` so it gates only the chip list — the toggle is now the sole always-visible readout of whether auto-approve is on, and it disappears whenever the user deselects every tool.

## Evidence
Evidence: 6/10. The description names the commands run with their outcomes and the new tests map cleanly onto the behavioral claims, but the PR's stated purpose is a rendered-layout outcome ("fits one line at 390px") and nothing demonstrates it — the most important missing artifact is a before/after screenshot pair at 390px width, with a shot of the rehomed toggle in ON/OFF states a close second.

## Notes
- `title` on `<svg>` produces no tooltip; use a `<title>` child or drop the hover claim — `frontend/src/components/Message.jsx:254`, `:295`.
- `aria-pressed` alongside dynamic `Auto-approve ON`/`OFF` text announces the state twice — `frontend/src/components/EnabledToolsIndicator.jsx:76`.
- Rows saved before this change restore with no `metadata` and render a green `APPROVED` pill with a blank tool name — `frontend/src/components/ToolApprovalMessage.jsx:271`.

---

PR #769, opened 2026-08-10
security(deps): apply in-range npm audit fixes to frontend

garland3:
## What changed

Lockfile-only update from `npm audit fix` (no `--force`) in `frontend/`. `package.json` is untouched — every fix lands inside an existing `^` range, so no breaking bumps.

Resolved 6 of 11 flagged packages (all non-major, in-range):

| Package | Severity | Advisory |
|---|---|---|
| `@babel/core` | low | Arbitrary File Read via sourceMappingURL comment |
| `brace-expansion` | high | DoS via unbounded / zero-step expansion |
| `flatted` | high | Unbounded recursion DoS in `parse()` revive phase |
| `form-data` | high | CRLF injection via unescaped multipart field names |
| `picomatch` | high | Method injection in POSIX character classes (incorrect glob matching) |
| `ws` | high | Uninitialized memory disclosure |

## What remains

5 packages are unchanged because all five require the **same semver-major bump** (`vitest` 2 -> 4.1.10) and `npm audit fix` correctly refuses to cross a major without `--force`. These are tracked as a separate task:

- `vitest` (critical), `vite` (high), `@vitest/mocker` (moderate), `esbuild` (moderate), `vite-node` (moderate)

`vitest` is a **dev-only** dependency and ATLAS never runs `vite dev` / the dev server in production (per `AGENTS.md`), so real production exposure for the cluster is low; the critical is a dev-tooling risk. The `vitest` 2->4 upgrade will be its own PR with a migration pass and test review — `npm audit fix --force` was intentionally **not** run to avoid pulling that breaking major into this PR.

## Verification

- `npm run lint` — 0 errors (4 pre-existing warnings, unrelated to this change)
- `npm test -- --run` — **521 / 521 passed** (49 test files)
- `npm run build` — success
- `npm audit` — 11 packages -> 5 (the `vitest` cluster only)

No source changes; `frontend/package-lock.json` only.


garland3:
## Summary
Bumps six transitive frontend dependencies in `frontend/package-lock.json` to clear npm advisories (11 flagged packages down to 5), deferring the remaining five behind a `vitest` 2 → 4 major upgrade, with a CHANGELOG entry describing the split.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. Adequate for a lockfile-only change — the description names each resolved advisory, explains the deferral, and lists lint/test/build/audit results. Independent comparison of the base and head lockfiles confirms the claim: zero packages added or removed, root dependency block unchanged, and all bumps patch/minor within existing ranges. The main missing artifact is captured output (or a CI link) for the `npm audit` 11 → 5 delta, which is the one number this change exists to move and the only one the pipeline does not re-establish on its own.

## Notes
- The committed lockfile now uses CRLF on all 6,376 lines while the rest of the repo uses LF, which is why a six-version bump shows as a ~12,700-line diff; the next `npm install` will silently rewrite it back to LF and produce a second whole-file diff. Convert with `sed -i 's/\r$//' frontend/package-lock.json` and consider `* text=auto` in a top-level `.gitattributes` (`frontend/package-lock.json:1`).
- The audit step at `.github/workflows/security.yml:63` ends in `|| true` and nothing parses the report, so the improvement this PR makes has no regression guard — a future re-pin would reintroduce the advisories with a green pipeline. Adding a threshold check with the deferred `vitest` cluster allowlisted would make the CI run itself the evidence for the audit delta.
- The dependency-scan job runs `npm ci` on Node 18 (`.github/workflows/security.yml:36`) while the locked tree declares floors of Node 20+ (`vite@7.3.6`, `marked@16`, `react-router@7.18.2`) and the test image uses Node 20; `npm ci` only warns, so the scan audits a tree it cannot run.

---

PR #770, opened 2026-08-10
fix: keep RAG context out of assistant tool_calls blocks

garland3:
## Problem

A RAG + tools turn that took a second tool round failed with a provider 400 from Azure/OpenAI:

> An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: call_slT5qS0Q05qPdumdHfhDZU4D

and the recovery path then crashed:

```
TypeError: 'NoneType' object is not iterable
  atlas/application/chat/modes/tools.py:422
  canvas_calls = [tc for tc in llm_response.tool_calls if tc.function.name == "canvas_canvas"]
```

## Root cause

Two separate defects, the second masking the first.

**1. The RAG context message split the tool-call block.** It was injected with `messages.insert(-1, ...)` — "before the last message". That is the user turn only on the first round. On a tool-calling continuation round the conversation ends with `assistant(tool_calls=[...]), tool, tool, ...`, so the context landed *between* the assistant message and its replies:

```
assistant(tool_calls=[call_slT5...])
system("Retrieved context from ...")   <- breaks adjacency
tool(call_slT5...)
```

The provider requires the tool replies to immediately follow the assistant message, so it reported the id as unanswered. Only reachable when data sources are selected and the loop reaches an extra round.

**2. The rejection crashed instead of degrading.** A failed continuation round yields no `LLMResponse`, so the loop substitutes `LLMResponse(content="")` whose `tool_calls` defaults to `None`, then hands it to the synthesis path, whose first statement iterates `tool_calls` unguarded. That made every mid-continuation provider error a hard failure instead of the intended graceful message.

## Fix

- New `LiteLLMCaller._rag_insert_index()` returns the index of the last `user` message (or `len(messages)` when there is none), and all four RAG injection sites use it: `litellm_streaming.py` (`stream_with_rag`, `stream_with_rag_and_tools`) and `litellm_caller.py` (`call_with_rag`, `call_with_rag_and_tools`). Behavior on a first-round turn is unchanged — the last message *is* the last user message there.
- Both canvas-only shortcuts (`modes/tools.py`, `utilities/tool_executor.py`) treat a response with no tool calls as not-canvas-only and fall through to synthesis. A `None` no longer raises, and an empty list no longer returns the misleading "Content displayed in canvas."

## Tests

New `atlas/tests/test_rag_context_insert_position.py` covers `_rag_insert_index` directly and asserts, for both the streaming and non-streaming RAG+tools paths, that every assistant `tool_calls` block is immediately followed by a reply for each id. New case in `test_tools_mode_iteration.py` drives a provider rejection on the continuation round and asserts the turn degrades to synthesis rather than raising.

All six new assertions fail on `main` and pass with the fix. Full backend suite: 1900 passed, 17 skipped.

## End-to-end validation

`test/pr-validation/test_pr770_rag_context_tool_call_ordering.sh` (8 checks) runs a stub provider that enforces the real ordering rule, the mock ATLAS RAG API, and an MCP server with one tool, then drives a RAG + tools turn over the WebSocket and inspects the message list the provider received. On the parent commit it reproduces both symptoms — the continuation request arrives as `[system, user, assistant, system:rag, tool]`, the stub returns the 400 naming `call_pr770StubToolCallId`, and the turn then dies with `'NoneType' object is not iterable`, sending the client nothing at all. With the fix the request is `[system, system:rag, user, assistant, tool]` and the turn completes.

```
bash test/pr-validation/test_pr770_rag_context_tool_call_ordering.sh
Passed: 8 | Failed: 0
```


garland3:
## Summary
Fixes a provider 400 on the RAG + tools continuation path by inserting the retrieved-context system message before the last user turn (`_rag_insert_index`) instead of at `insert(-1, ...)`, which split an assistant `tool_calls` block from its tool replies. Also guards the canvas-only synthesis shortcut against `None`/empty `tool_calls` in both `modes/tools.py` and `tool_executor.py`, and records the ordering invariant in a design note.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — strong for a non-visual fix: a verbatim provider error, a dedicated `test_rag_context_insert_position.py`, a continuation-failure regression test, and a full-suite pass count. The most useful missing artifact is the actual pytest output for the fail-on-`main` / pass-with-fix claim; two of the four changed insertion sites (`call_with_rag`, `stream_with_rag`) are also edited but never exercised.

## Notes
- The RAG context message dict is built and inserted identically at four sites (`litellm_caller.py:993,1180`, `litellm_streaming.py:331,409`) — collapsing the tail into one `_insert_rag_context` helper would give the new invariant a single enforcement point.
- The two copies of the canvas-only shortcut now diverge: `modes/tools.py:427` reads the name shape-tolerantly via `_tool_call_signature(tc)[0]`, while `tool_executor.py:816` still uses `tc.function.name` and would raise on a dict-shaped call.
- The design note at `docs/developer/design-notes/multi-tool-calling-2026-02-22.md:115` cites PR #769; this change is #770.

---

garland3:
## Summary
Fixes RAG context injection so the retrieved-context system message is placed before the last user turn (`_rag_insert_index`) instead of `insert(-1, ...)`, which split an assistant `tool_calls` block from its `tool` replies and caused provider rejections on continuation rounds. Also hardens the canvas-only synthesis shortcut against `None`/empty `tool_calls`, and adds unit tests plus an end-to-end validation script.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size: verbatim provider error and traceback, direct unit tests for the index helper, a regression test for the graceful-degrade path, and an end-to-end script with a stub provider that self-verifies its own ordering rule before asserting on the product. The one missing artifact is the actual on-parent-commit run output; the "fails on main" claim is stated in prose rather than shown.

## Notes
- The canvas-only shortcut in `atlas/application/chat/utilities/tool_executor.py:816` still reads `tc.function.name` while the sibling copy at `atlas/application/chat/modes/tools.py:427` uses the shape-tolerant accessor — a dict-shaped tool call would raise `AttributeError` there. Worth collapsing the two copies onto one shared helper.
- The RAG context message is still built and inserted at four copy-pasted sites (`litellm_caller.py:993,1180`, `litellm_streaming.py:331,409`); a single `_insert_rag_context(...)` helper would give the new ordering invariant one enforcement point instead of a prose rule.
- Test gaps: no fixture has more than one `user` message, so a forward scan in `_rag_insert_index` would pass all tests; `call_with_rag` / `stream_with_rag` are edited but unasserted; `handle_synthesis_decision` is not reached by any test.

---

PR #771, opened 2026-08-12
security: fix calculator eval RCE and add chat WebSocket origin check

garland3:
Fixes the two high-severity findings from a whole-codebase security review of `main` (3eb7b8e). Both are reachable in a normal deployment.

## 1. Calculator MCP server: remote code execution

`evaluate` ran `eval(expr, {"__builtins__": {}}, allowed_names)` with only a 200-character length cap. Emptying `__builtins__` does not contain Python: attribute access reaches the class hierarchy, and any module the interpreter has already imported — `os` among them — is a couple of lookups away from an expression that never names it.

A 122-character payload, comfortably inside the cap, reached `os.system`:

```
[c for c in ().__class__.__mro__[1].__subclasses__()
   if c.__name__=='_wrap_close'][0].__init__.__globals__['system']('id')
```

I confirmed this executing against the shipped tool before the fix, and confirmed it refused afterward with no side effect on disk.

**Exposure was broad.** The calculator is a stdio server, so this ran as the backend's own OS user. `atlas/config/mcp.json` ships it with `groups: ["users"]`, and `is_user_in_group` returns `True` for `users` unconditionally — every authenticated user had it. And since the expression arrives through an LLM tool call, so did any prompt-injection vector that reached the model: a poisoned RAG document, an uploaded file, web content pulled in by another tool.

**Fix:** new `atlas/mcp_shared/safe_math_eval.py`. It parses with `ast.parse(mode="eval")` and walks the tree against an allowlist of arithmetic nodes. `Attribute`, `Subscript`, comprehensions, `Lambda`, `IfExp`, f-strings, and starred/keyword call arguments are simply not in the allowed set — the escape becomes unrepresentable rather than merely filtered. A call's callee must be a bare `Name` present in the caller's table, so evaluation cannot reach a callable the caller did not supply. Literal exponents are bounded, so `9**999999` can no longer hang the process.

The allowed-names table moved to module scope (it was rebuilt on every call). No change to the tool's contract or to any expression a user would legitimately write.

## 2. Chat WebSocket: cross-site WebSocket hijacking

`/ws` validated the proxy secret and resolved identity from the auth header, then accepted — without ever inspecting `Origin`. WebSocket upgrades are not preflighted, so the same-origin policy does not protect this endpoint; the check has to be explicit.

The consequence: any page a logged-in user visits could open a socket to Atlas. The browser attaches their cookies, the reverse proxy authenticates the upgrade through `auth_request` and injects `X-User-Email` plus the proxy secret, and the backend sees a legitimate session. The attacker's page then reads conversation history, calls MCP tools, and reaches stored files as that user. CSP `connect-src 'self'` does not help — it constrains Atlas's own pages, not the attacker's.

**Fix:** `/ws` now validates `Origin` before authentication runs. An upgrade is allowed when the origin is loopback, names the same host the request was addressed to (compared against `Host`, ignoring port), or appears in `WEBSOCKET_ALLOWED_ORIGINS`. Anything else closes with 1008.

A **missing** `Origin` is allowed, deliberately. Browsers always send the header on an upgrade, so absence means a non-browser client — a CLI, a test harness, a service integration — and those carry no ambient cookies for an attacker page to borrow. CSWSH is a browser attack; rejecting header-less clients would break legitimate integrations without closing anything.

**No configuration is needed for a normal deployment** — the same-origin rule covers the case where browser and backend agree on the hostname. `WEBSOCKET_ALLOWED_ORIGINS` is only for proxies that rewrite `Host` to an internal name. `FEATURE_WEBSOCKET_ORIGIN_CHECK_ENABLED` (default `true`) is the escape hatch.

### Shared with the Agent Portal

The portal's stream socket already did this correctly. Its helper moved to `atlas/core/websocket_origin.py` so both sockets share one implementation instead of drifting. **Portal behaviour is unchanged** — it stays stricter, consulting neither `Host` nor admitting a missing `Origin`, because it is a dev-only preview that grants command execution. Its existing tests pass untouched.

## Testing

- `atlas/tests/test_safe_math_eval.py` — 61 tests. Every escape shape (the confirmed payload, attribute access on literals and on allowed functions, subscripting, comprehensions, lambdas, `__import__`, `open`, nested `eval`, f-strings, walrus, conditional expressions) must raise; the full range of legitimate arithmetic must still evaluate.
- `atlas/tests/test_websocket_origin_check.py` — 33 tests covering the primitives (host extraction incl. IPv6 and ports, allowlist normalization, suffix-confusion rejection, non-HTTP schemes), the policy wrapper, and the hijacking case end to end through `TestClient`.
- Full backend suite: **2023 passed, 17 skipped**.
- `ruff check atlas/` introduces no new findings (44 pre-existing, unchanged).

## Docs

`docs/admin/authentication.md` gains a "WebSocket Origin Validation" section explaining the attack, the three allow rules, why a missing `Origin` is permitted, and when the allowlist is actually needed. `.env.example` documents both new variables.

## Not addressed here

The review turned up further items, deliberately left for separate PRs: MCP-controlled iframe `url`/`sandbox` in the canvas (medium, currently blunted by CSP), the Agent Portal's unresolved per-user ownership TODOs, and rate limiting keyed on the proxy's IP.


garland3:
## Summary
Replaces the calculator MCP tool's `eval` with an AST-walking evaluator (`atlas/mcp_shared/safe_math_eval.py`) and adds a WebSocket `Origin` check to the chat upgrade path, backed by a shared `atlas/core/websocket_origin.py` also now used by the Agent Portal. Adds `FEATURE_WEBSOCKET_ORIGIN_CHECK_ENABLED` and `WEBSOCKET_ALLOWED_ORIGINS`, plus docs and `.env.example` entries.

## Verdict
Request changes.

## Required changes
1. Bound exponentiation by evaluated operand magnitude in the `ast.Pow` branch of `_evaluate` (`atlas/mcp_shared/safe_math_eval.py:79`) instead of the syntactic `_check_pow_bounds`. Because `**` is right-associative, `9**9**9` has a `BinOp` exponent, skips the check entirely, and never returns — the exact expression the module docstring cites as the motivation for the guard. `2 ** (500 + 500)` and `((9**999)**999)**999` evade it the same way.
2. Bound the arguments of the superlinear entries in `ALLOWED_NAMES` (`atlas/mcp/calculator/main.py:14`) — `pow`, `factorial`, `comb`, `perm` reach the same unbounded computation through `ast.Call` and never touch `_check_pow_bounds`; `pow(9, 9999999)` takes ~6 s and returns a 4 MB int, `factorial(3000000)` does not terminate.
3. Bound the shift count for `ast.LShift` (`atlas/mcp_shared/safe_math_eval.py`, `_BIN_OPS`) — `1 << 10000000000` is 16 characters, returns instantly, and allocates a 1.33 GB integer.
4. Reject oversized integer results inside `evaluate` (`atlas/mcp/calculator/main.py:130`) with the documented `is_error: true` payload. An `int` above 4300 digits cannot be serialized, so the response fails in the MCP layer after `evaluate` returns, breaking the error contract the docstring promises on every path.
5. Add rejection tests for the shapes that currently defeat the guard (`atlas/tests/test_safe_math_eval.py:170`) — `9**9**9`, `((9**999)**999)**999`, `pow(9, 99999999)`, `factorial(3000000)`, `1 << 10000000000`. `test_nested_huge_exponent_is_rejected` uses `2 * (9 ** 999999)`, whose exponent is still a literal, so it does not cover the nesting that matters and the suite currently reads as evidence this class of input is handled.
6. Normalize allowlist entries through the module's existing `extract_host` in `parse_allowed_hosts` (`atlas/core/websocket_origin.py:35`) — `WEBSOCKET_ALLOWED_ORIGINS=https://atlas.example.com`, the value the setting's name invites, is stored verbatim and compared against a bare hostname, so it never matches and the operator sees only continued 1008 rejections.
7. Document the evaluator's new refusals in the calculator tool docstring (`atlas/mcp/calculator/main.py:60-104`) — keyword arguments (`round(3.14159, ndigits=2)`), boolean/membership operators, and string literals were accepted by `eval` and are now rejected, and the docstring is the contract the caller plans against.

## Evidence
Evidence: 8/10 — strong for a backend security change: 94 new tests, a regression case for the reported payload, an end-to-end hijacking test through `TestClient`, and both settings documented. The most important missing artifact is any demonstration of the pre-fix exploit actually executing; the RCE severity claim currently rests on prose assertion, and a two-block terminal transcript (payload before, refusal after) would close it.

## Notes
- Read the new settings as direct attributes rather than `getattr(..., default)` — `atlas/main.py:461,470` — so a later rename fails loudly instead of silently emptying the allowlist.
- Loopback origins are admitted unconditionally, including in production — `atlas/core/websocket_origin.py:104`; consider gating that branch on debug mode.
- `atlas/main.py:76-83` splits one `files_routes` import into two; recombine to keep the diff to the security change.

---

garland3:
Thanks — the resource-exhaustion finding was correct and I'd gotten it wrong. Pushed 52c99b9.

## Fixed

**Resource exhaustion (Copilot + Codex, P1).** Confirmed before fixing: `pow(9, 999999)` ran past a 25s timeout. My guard was parse-time and operator-specific, so five shapes walked past it — `9**9**9` (outer exponent is a `BinOp`, not a `Constant`), `pow(9, 999999)` and `factorial(999999)` (calls are invisible to an operator check), `1 << 99999999` (different operator), and `(10**1000)**1000` (every exponent under the cap, result unbounded).

Both bounds now apply to *evaluated* operands rather than source literals. `MAX_EXPONENT` catches the direct case; `MAX_RESULT_BITS` bounds the size of the number an operation would have to build, which catches chaining. Left shifts get the same check.

Codex's point that the MCP timeout doesn't help is right and worth stating: calculator calls run synchronously in the stdio server, so nothing external interrupts the computation. That's why this had to be a pre-computation bound rather than a timeout.

Functions in the name table are opaque to the evaluator — it can't know what a caller-supplied callable costs — so `guard_exponent` / `guard_operand_size` are exported and the calculator wraps `pow`, `factorial`, `comb`, `perm`. Three-argument `pow` stays unbounded since modular exponentiation is cheap at any exponent.

All six reject in under a millisecond now; `pow(9,999999,7)`, `factorial(20)`, `2**32` unaffected.

**Loopback origins on non-loopback targets (Codex, P1).** Also correct, and a real hole I'd missed — in production a malicious page from another app on the user's own machine could open the production socket. Loopback is now trusted only when the target is loopback too, or when no `request_host` was supplied. The latter preserves the Agent Portal's behaviour exactly (it binds loopback and passes no host).

**Docstring contract (Copilot).** Right — `divmod` yields a tuple and comparisons yield a bool. Documented the real shape.

## Not changing, with reasons

**Compare full origins including scheme and port (Codex, P1).** The threat is real but this fix isn't available to us, and a naive version would break every deployment. Behind a TLS-terminating proxy the backend sees plain HTTP on an internal port and has no way to reconstruct the browser-facing origin, so a strict comparison rejects every legitimate upgrade. Port-strictness fails too: with `proxy_set_header Host $host` nginx strips the port, so `Origin: https://atlas.example.com:8443` vs `Host: atlas.example.com` would reject a correct request.

Hostname is also the granularity the credentials this defends actually live at — cookies are not isolated by port, and without `Secure` not by scheme either. The residual risk is a *different* app on another port of the same hostname; I've documented that in `websocket_origin.py` and pointed such deployments at `WEBSOCKET_ALLOWED_ORIGINS` plus not sharing a hostname with untrusted apps. Happy to revisit if you'd rather key off `X-Forwarded-Proto`, but that only closes the scheme half and adds a header-trust dependency.

**Remove tool-branded commit attribution (Codex, P1).** False positive. Author and committer on all commits are `anthony garland <garland3@gmail.com>`; `git log --format='%(trailers)'` is empty and there are no AI tool names anywhere in the messages. Nothing to recreate. (The comment was also filed against `.env.example:61`, which is unrelated to commit metadata.)

**CodeQL log-injection at `main.py:526`.** Known false-positive class for this repo — the value goes through `sanitize_for_logging`, which strips CR/LF, C0/C1 controls, and U+2028/2029; CodeQL doesn't recognise the custom sanitizer. Verified: an origin containing `\r\n` plus a forged log line comes out as a single line. Same pattern as the Agent Portal's existing origin-rejection log.

Full suite: **2056 passed, 17 skipped**. `ruff check atlas/` unchanged at 44 pre-existing.

garland3:
Thanks — this was a good review. Items 1, 2, 3 and 5 had landed in 52c99b9 just before it posted; 4, 6, 7 and the notes are addressed in 0d82f47.

## The exploit transcript you asked for

Fair challenge — the RCE claim was resting on prose. Same script against both trees:

```
########## BEFORE (origin/main, 3eb7b8e) ##########
payload length: 145 chars (tool cap is 200)
tool reported is_error = False
!! COMMAND EXECUTED. /tmp/ATLAS_RCE_PROOF contains:
  uid=1000(garlan) gid=1000(garlan) groups=1000(garlan),4(adm),20(dialout),...

########## AFTER (this branch) ##########
payload length: 145 chars (tool cap is 200)
tool reported is_error = True
refused: Evaluation error: Only direct calls to allowed functions
no command executed; marker file absent
```

The payload runs `id > /tmp/ATLAS_RCE_PROOF`; the pre-fix run reports success *and* leaves the file, so it is the shell's output, not the tool's.

## Required changes

**1, 2, 3, 5 — already in 52c99b9.** Bounds moved to evaluated operands in the `ast.Pow` branch, `pow`/`factorial`/`comb`/`perm` wrapped at the `ALLOWED_NAMES` boundary, `LShift` bounded, and rejection tests added for `9**9**9`, `((9**999)**999)**999`, `pow(9, 99999999)`, `factorial(3000000)`, `1 << 10000000000`. `test_nested_huge_exponent_is_rejected` is still there but is no longer load-bearing.

One correction on your item 1's examples: **`2 ** (500 + 500)` should not be rejected.** It evades the *syntactic* check exactly as you say, but under the evaluated-operand check it resolves to `2**1000` — within the cap, 302 digits, a legitimate result. I initially wrote it as a rejection test, it failed, and the test was wrong rather than the code. Tests now pin both sides: `500 + 501` rejected, `500 + 500` allowed. That pair is what actually demonstrates the guard reads the computed value.

**4 — oversized results. Confirmed and fixed.** `(2**1000)**15` and `factorial(1000)*factorial(1000)` both returned `is_error: false` and then raised `ValueError` in `json.dumps`. The evaluator's bounds don't cover this because they're sized for compute cost, and both of those are cheap; multiplication has no bound at all, deliberately, since it's linear. `evaluate()` now checks serializability before returning, sizing ints via `bit_length` rather than `str()` (the operation under test) and walking tuples since `divmod`/`modf` return pairs. There's a test asserting every payload this tool can produce is JSON-serializable — success, refusal, arithmetic error, over-length.

**6 — allowlist normalization. Fixed.** Entries now go through `extract_host`, so `https://atlas.example.com` and a bare hostname both work. This fixes `AGENT_PORTAL_ALLOWED_ORIGINS` too, since it shares the function.

**7 — docstring. Fixed.** Added a "Not Supported" section covering keyword arguments, string literals, `and`/`or`/`in`/`is`, attribute access and indexing, and the size refusals.

## Notes

- **Direct attribute reads** — done, for both settings. Agreed on the reasoning: silently disabling a security check on a rename is the worst failure mode available. Replaced the "absent setting" test with one asserting both fields exist on `AppSettings`, which is the invariant that makes direct access safe.
- **Loopback in production** — fixed in 52c99b9, though gated on the *target* rather than on debug mode. Loopback is trusted only when the request's own host is also loopback. That covers the production case you're pointing at while keeping the Agent Portal working unchanged (it passes no `request_host`) and without tying a network-boundary decision to `DEBUG_MODE`.
- **`files_routes` import** — recombined. Worth flagging: an earlier `ruff --fix` of mine had split it, which incidentally cleared a *pre-existing* I001 in `main.py`. So my earlier "44 pre-existing" was wrong — the real baseline on `origin/main` is **45**, and this branch is now 45. Lint-neutral, not lint-improving.

Full suite: **2073 passed, 17 skipped**.

garland3:
## Summary
Replaces the calculator MCP tool's `eval` with a restricted AST evaluator (`atlas/mcp_shared/safe_math_eval.py`) that bounds exponents, shifts and growth functions, adds a result-serializability guard, and adds an `Origin`/`Host` check to the `/ws` upgrade path with a shared `websocket_origin` helper, settings, docs and tests.

## Verdict
Request changes.

## Required changes
1. Bound tuple repetition in `atlas/mcp_shared/safe_math_eval.py:69` — `ast.Mult` with a tuple operand is unguarded, so `(1,)*(10**8)` allocates a 100M-element tuple (~778 MB measured) from a 12-character expression; reject when `len(seq) * abs(n)` exceeds a small element cap, or drop tuple operands from `_BIN_OPS`.
2. Wrap `round` in `ALLOWED_NAMES` at `atlas/mcp/calculator/main.py:98` with `guard_operand_size` on `ndigits` — a large negative `ndigits` is superlinear (`round(5, -32000000)` measured at 35 s; `round(5, -999999999)` does not return), the same class the `pow`/`factorial` wrappers close.
3. Reject complex results at `atlas/mcp/calculator/main.py:31` — drop `complex` from `_ALLOWED_CONSTANTS` or have `_result_is_returnable` refuse any non-`int`/`float`/`bool` leaf; today `evaluate("2j")` returns `is_error: false` and then raises `TypeError` in the MCP encoding layer, escaping the documented error contract.
4. Make `guard_operand_size` (`atlas/mcp_shared/safe_math_eval.py:129`) bound floats as well as ints — on Python 3.11, which `pyproject.toml` still supports, `math.factorial` accepts integral floats, so `factorial(1e6)` passes the guard untouched.
5. Add complex, `nan`/`inf` and nested-tuple cases to `test_every_payload_this_tool_returns_is_json_serializable` (`atlas/tests/test_calculator_tool.py:126`) — it currently enumerates expressions while its name asserts a universal property, so the class above passes CI.
6. Correct the loopback rule in `docs/admin/authentication.md:78` and `.env.example:63` — both state loopback origins are always allowed, but `origin_is_allowed` trusts them only when the target host is also loopback; add that `WEBSOCKET_ALLOWED_ORIGINS` matches by hostname only (scheme and port are ignored) and accepts full origins as well as bare hostnames.
7. Document the working spelling for aggregates at `atlas/mcp/calculator/main.py:169` and in the CHANGELOG entry — `sum([1, 2, 3])` and `round(x, ndigits=2)` worked before and now error, and nothing tells a caller that `sum((1, 2, 3))` is the replacement.

## Evidence
Evidence: 9/10. The change is unusually well demonstrated — the RCE payload is run against both `origin/main` and this branch with a filesystem side effect as proof, the CSWSH rejection is driven end to end through a test client, and the size bounds are pinned on both the accept and reject side. The one missing artifact is a measured timing transcript (or a duration assertion) for the resource-exhaustion bounds, which are currently asserted only as `is_error is True`.

## Notes
- Two disagreeing normalization rules for allowlist entries: `parse_allowed_hosts` strips scheme/port via `extract_host`, `origin_is_allowed` re-normalizes with a weaker rule (`atlas/core/websocket_origin.py:151`).
- A refused expression leaves no server-side log, and every evaluator rejection reports the same `meta_data.reason` (`atlas/mcp/calculator/main.py:235`), so a sandbox-escape attempt is indistinguishable from a typo.
- Dead `_LOOPBACK_HOSTS` constant left behind in `atlas/routes/agent_portal_routes.py:1221`.

---

garland3:
All seven addressed in 501571e, plus the WS JWT bypass from a separate review. Verified each finding before fixing.

## Required changes

**1. Tuple repetition — confirmed, fixed.** `(1,)*(10**7)` measured 238 MB RSS in 2.2s. `ast.Mult` with a sequence operand isn't arithmetic, so the numeric guards never saw it. Added a repetition guard capped at 1000 elements (tuples exist only so `divmod` results and `sum((1,2))` read naturally). Kept tuples in `_BIN_OPS` rather than dropping them, so `(1,)*5` still works.

**2. `round` ndigits — confirmed, fixed.** `round(5, -3000000)` took 0.69s. Wrapped like `pow`/`factorial`/`comb`/`perm`.

**3. Complex results — confirmed, fixed.** `evaluate("2j")` returned `is_error: false` then raised `TypeError` in encoding. Dropped `complex` from `_ALLOWED_CONSTANTS` *and* rewrote `_result_is_returnable` to refuse any non-`int`/`float`/`bool` leaf. Extended to **`inf`/`nan`** while there — `json.dumps` emits bare `Infinity`/`NaN`, which isn't valid JSON, so those had the same escape.

**4. Float bounds — fixed, but the specific claim didn't reproduce.** `math.factorial(5.0)` raises `TypeError` on 3.11.14 (float support was removed, not just deprecated), so `factorial(1e6)` was already an error rather than a hang. The bound is still right and now covers `round`'s `ndigits`, which *is* reachable with a float. Flagging it so the reasoning isn't taken on trust.

**5. Property test — fixed, and this was the right call.** The test now parametrizes over every shape a caller can reach (tuples, nested tuples, non-finite floats, complex, oversized ints, and each refusal class) and encodes with `allow_nan=False`. A new refusal class is caught by the existing property rather than needing its own test.

**6. Docs — fixed.** Both said loopback is always allowed, which stopped being true when I gated it on the target. Corrected, plus documented hostname-only matching and that entries accept full origins.

**7. Working spellings — fixed.** Tool docstring now lists each refused form *with its replacement* (`sum((1,2,3))` not `sum([1,2,3])`, `round(x,2)` not `round(x,ndigits=2)`), and the CHANGELOG has a dedicated entry, since this is a behaviour change callers plan against.

## Evidence: timing transcript

Fair — `is_error is True` would pass even if the guard ran after a 30-second computation. Added `test_resource_bounds_reject_quickly`, which asserts elapsed < 1.0s per expression. Measured:

```
(1,)*(10**8)           err=True    0.10ms
round(5, -999999999)   err=True    0.05ms
2j                     err=True    0.03ms
factorial(1e6)         err=True    0.04ms
inf / nan              err=True    0.01ms
```

Still working: `round(3.14159,2)`, `round(2.5)`, `sum((1,2,3))`, `divmod(7,2)`, `2**1000`, `(1,)*5`.

## Notes

- **Disagreeing normalization rules** — fixed; `origin_is_allowed` now uses `extract_host` too.
- **Indistinguishable rejections** — fixed. Refused constructs report `reason: "rejected_expression"` and log a warning; ordinary maths errors (`ZeroDivisionError`, `sqrt(-1)`) stay unlogged with their exception name. A sandbox-escape attempt is now visible to an operator.
- **Dead `_LOOPBACK_HOSTS`** — already removed in ddb5599, before this review; grep confirms it's gone.

## Also in this push: WebSocket JWT bypass

A separate review caught something worse than anything above. `AuthMiddleware` verifies `aws-alb-jwt` properly, but **both** WebSocket endpoints called `get_user_from_header()` unconditionally — so in an ALB-JWT deployment, `X-User-Email: admin@company.com` opened an authenticated socket that the same value could not open over HTTP. Full auth bypass on the socket.

All three call sites now resolve through one `resolve_user_from_auth_header()`. New tests cover forged, unsigned, malformed, and wrong-signer tokens on both sockets, plus that plain `email-string` deployments are unaffected.

Full suite: **2121 passed, 17 skipped**. Lint unchanged at 45 (matches `origin/main`).

Two follow-up PRs are in progress: authorization/resource limits (tools flag, sampling ACL, turn concurrency, dev-only MCP prod guard) and cleanup/supply-chain (domain whitelist removal, lockfile sync + CI gating).

garland3:
## Summary
Hardens the calculator MCP tool by replacing `eval` with a bounded AST evaluator (`mcp_shared/safe_math_eval.py`) plus operand/exponent/result guards, adds WebSocket origin validation, and routes both WebSocket endpoints through `resolve_user_from_auth_header` so a forged `X-User-Email` no longer bypasses ALB-JWT verification.

## Verdict
Request changes.

## Required changes
1. Bound the *total* element count in `_guard_repetition` at `atlas/mcp_shared/safe_math_eval.py:162` — `len(sequence) * abs(count)` only counts top-level elements, so `(((1,)*1000,)*1000,)*1000` (25 chars, inside the 200-char cap) passes the guard and produces ~10^9 leaves. Size the operand recursively (early-exit past the cap), or reject repetition whose operand contains a tuple; `divmod` pairs and `sum((1, 2, 3))` are unaffected.
2. Give `_result_is_returnable` a hard node budget at `atlas/mcp/calculator/main.py:33` and refuse once it is exceeded — it currently walks every leaf of the structure above, and measurements on this branch show the call not returning within 40–120 s while `pending` grows unboundedly. The calculator is a synchronous stdio server, so this wedges it for every user.
3. Add rejection tests for two- and three-level nesting (`((1,)*1000,)*1000`, `(((1,)*1000,)*1000,)*1000`) to `test_resource_bounds_reject_quickly` at `atlas/tests/test_calculator_tool.py:169` — it currently pins only the flat shape `(1,)*(10**8)`, so the suite reads as if this class is closed.
4. Move the ALB key fetch off the event loop at the two new async call sites (`atlas/main.py:558`, `atlas/routes/agent_portal_routes.py:1246`) — `_get_alb_public_key` does a synchronous `httpx.get(..., timeout=5.0)`, and a cache miss stalls every in-flight connection for up to 5 s. Negatively cache failed fetches so varying `kid` values cannot force one outbound request per upgrade.
5. Fail closed on an unparseable `Host` at `atlas/core/websocket_origin.py:120` — a supplied-but-unparseable value (e.g. `::1`, empty, `host/x//y`) yields `target is None` and takes the same unconditional-loopback-trust branch as "no host supplied". Make the permissive case explicit (a `trust_loopback` argument the Agent Portal sets) rather than the default of an omitted parameter.
6. Add positive-path tests for the new bounded wrappers in `atlas/mcp/calculator/main.py` — `round(3.14159, 2)`, `round(2.5)`, `round(1234, -2)`, `perm(5)`, `sum((1, 2, 3))`. Only the rejection branches are currently exercised, so a tightening that breaks legitimate `round(x, -2)` would land green.

## Evidence
Evidence: 9/10 — unusually strong for a change this size: the RCE is demonstrated executing against `origin/main` and refusing here with a filesystem side effect as the discriminator, resource bounds are pinned on both sides, and the CSWSH fix asserts ordering (origin checked before authentication), not just outcome. The one gap worth closing is the WebSocket JWT bypass, which arrived in the final push with regression coverage of the fix but no before/after transcript showing the original exploitability.

## Notes
- Remove `inf`/`nan` from the advertised constants list, or annotate them as usable as arguments but not as a final result — `atlas/mcp/calculator/main.py:179` contradicts the "Not Supported" section at `:206`.
- Error text names AST internals (`Unsupported expression element: List` for `sum([1,2,3])`) — mapping the reachable node types to caller-facing wording turns a failed call into a successful retry (`atlas/mcp_shared/safe_math_eval.py:267`).
- Origin rejections happen pre-`accept()`, so the browser sees 1006 with no reason and the client only surfaces reasons for 1008 — worth a troubleshooting line pointing operators at the backend log (`atlas/main.py:532`, `docs/admin/authentication.md:62`).

---

garland3:
All six plus the three notes addressed in 063d4dd. Findings 1 and 2 were the same root cause and both reproduced.

## Required changes

**1 + 2. Nested repetition — confirmed, and worse than the flat case.** `_guard_repetition` sized the operand with `len(sequence)`, counting only the top level. Repetition composes, so each `* n` multiplies what the operand already holds while its own top-level length stays at 1. Measured on this branch before the fix:

```
((1,)*1000,)*1000            is_error=False   243ms   (1e6 leaves)
((1,)*1000,)*100             is_error=False    24ms
```

Three levels is 1e9 leaves in 25 characters. Operands are now sized with `total_elements()`, which walks nested tuples and early-exits past the cap, so an already-enormous structure costs no more to reject than a small one. After: all three refuse in **under 0.2ms**; `(1,)*5`, `divmod` pairs and `((1,2),(3,4))` unaffected.

`_result_is_returnable` got the hard node budget you asked for. Your framing is the right one — the evaluator guard *should* mean nothing that large ever arrives, but this function must not be the thing that wedges a synchronous stdio server if that assumption is wrong.

**3. Nesting tests — added**, both two- and three-level, alongside the flat shape in `test_resource_bounds_reject_quickly`.

**4. ALB fetch on the event loop — fixed, and it applied more widely than the two new sites.** `AuthMiddleware.dispatch` is also a coroutine and had the same defect *before* this branch; leaving HTTP blocking while fixing WS would be arbitrary, so all three now use a new `resolve_user_from_auth_header_async` that runs verification via `asyncio.to_thread`. The plain-header path stays inline — no I/O, so a thread hop would cost more than it saves. There's a test asserting the verifier actually runs on a different thread, since that's the observable consequence.

Negative caching added (5min TTL) — `kid` comes from an *unverified* header, so a fresh value per request forced one outbound HTTPS call per upgrade. The cache is also bounded at 256 entries, since that same attacker-influenced key would otherwise make the negative cache its own growth lever.

**5. Unparseable `Host` — fixed.** Trust is now an explicit `trust_loopback` argument set only by the Agent Portal, rather than inferred from a missing `request_host`. You were right that inference was the bug: `""`, `::1` and `host/x//y` all produced `target=None` and took the same branch as a deliberate opt-in. Tests cover each malformed value.

**6. Positive-path tests — added**, and worth doing: `round(x, -2)` is exactly the case a tightening would silently break. Covers `round(3.14159, 2)`, `round(2.5)`, `round(1234, -2)`, `round(-1234, -2)`, `perm(5)`, `perm(5,2)`, `comb(5,2)`, `factorial(0)`, `sum((1,2,3))`, `(1,)*5`.

## Notes

- **`inf`/`nan` contradiction** — fixed; now listed as usable as arguments but not as a final result, with the reason.
- **AST internals in error text** — fixed, and this was a good catch. `Unsupported expression element: List` is a dead end for an LLM caller. Now:
  ```
  sum([1,2,3])  -> list syntax is not supported; use parentheses, e.g. sum((1, 2, 3))
  a[0]          -> indexing (x[0]) is not supported
  1 and 2       -> and/or are not supported; this evaluates numbers, not logic
  ```
- **1006 vs 1008** — documented. The close happens pre-`accept()`, so the reason never reaches the browser; added a troubleshooting block pointing at the backend log line and noting that a rewritten `Host` is the usual cause.

## On the evidence gap

Fair: the JWT bypass arrived with regression coverage but no before/after transcript. It's harder to demonstrate honestly than the RCE was — it needs a real ALB-signed token to show the positive side, and I'm not going to fabricate one. What I can state precisely is the code path: `middleware.py:140/153` branched on `auth_header_type` and called `get_user_from_aws_alb_jwt`; `main.py:560` and `agent_portal_routes.py:1244` called `get_user_from_header` unconditionally, which is `header.strip()`. In `aws-alb-jwt` mode those two lines accepted any non-empty value the HTTP path rejected. If a transcript against a stub ALB would close it, say so and I'll build one.

Full suite: **2144 passed, 17 skipped**. Lint unchanged at 45. check-docs passes.

Follow-ups are open: **#772** (tools flag, sampling ACL, turn/step limits, dev-only MCP prod guard) and **#773** (domain whitelist removal, lockfile sync, enforceable CI gates), both stacked.

PR #772, opened 2026-08-12
security: enforce tools flag, MCP sampling model ACL, and per-turn/step limits

garland3:
Second of three security PRs from the review. **Stacked on #771** — targets `fix/mcp-eval-rce-and-ws-origin`, so review that one first; this diff shows only its own changes. Retarget to `main` once #771 lands.

## 1. `FEATURE_TOOLS_ENABLED` was cosmetic

The flag only filtered what `/api/config` advertised. MCP clients initialized unconditionally, and the WebSocket handler passed client-supplied `selected_tools`, `selected_prompts` and `agent_mode` straight through — so turning tools off in the UI mitigated nothing. A crafted frame still executed them, which matters because the flag is the natural first reach for anyone responding to a tool vulnerability.

Enforced in the **orchestrator** rather than the WebSocket handler, so programmatic callers (`AtlasClient`) go through the same rule; discarded requests are logged. MCP initialization, discovery, auto-reconnect, and the cache sweeper are now skipped entirely when the flag is off — starting subprocesses and connecting to remote servers is real attack surface with nothing to serve.

## 2. MCP sampling bypassed the model ACL

`_SamplingRoutingContext` carried no user identity, so the handler took the MCP server's preferred model — or `next(iter(models))` — and invoked it with **no group check**. An MCP server available to a user could consume a model whose `groups` deny that user; the listing-layer filtering that governs the chat UI never touches this path.

`user_email` is now threaded through the context. The server's preference is honoured only if `check_model_access` permits it; otherwise the first authorized model is used; if none is authorized the request is **denied rather than silently downgraded**, since succeeding with an unauthorized model is the bug. A policy-resolution error fails closed.

`max_tokens` is clamped to the new `MCP_SAMPLING_MAX_TOKENS` (default 4096) — it arrived from the MCP server and was an unbounded lever on a metered API.

## 3. Unbounded concurrent turns

Every `chat` frame created a task and overwrote the single active-task reference **without cancelling the previous one**. Earlier turns kept running with no handle on them: still making metered model calls, still calling tools, racing each other writing session history — while stop/reset could only reach the most recent. HTTP rate limiting never applied here; it's HTTP middleware and doesn't see WebSocket frames.

Additional frames are now rejected with a `turn_in_progress` error while a turn is active.

## 4. `agent_max_steps` had no server-side ceiling

It arrived off the frame and each step is a metered model call plus its tool calls. Now clamped to the configured `agent_max_steps`; non-integer, `bool` (an `int` subclass — would otherwise have become 1 step) and out-of-range values fall back to the default rather than failing the turn.

## 5. Dev-only MCP servers refuse to start in production

`file_viewer`, `filesystem`, `code-executor`, `code_executor_v2` and `transfer` give any authenticated user arbitrary host file access or outbound fetches. They're fine on a laptop — the failure mode is an operator copying an example config into a shared deployment, where nothing previously objected.

Startup now **aborts** if one is enabled while `ENVIRONMENT=production` or `DEBUG_MODE=false`, mirroring `validate_agent_portal_dev_only`. A refusal rather than a warning, because a line in a startup log isn't a control.

Detection is **by server name**, so copying `mcp-file_viewer.json` is caught without the operator having to know a marker exists. `"dev_only": true` marks a custom server; `"dev_only": false` opts out one that has been reviewed and hardened — an explicit statement, not a way to silence the message.

## Testing

- `test_tools_flag_and_step_limits.py` — flag reading and the clamp across valid, over-limit, negative, string, `bool` and float inputs.
- `test_sampling_model_authorization.py` — preference honoured when authorized, fallback when not, filtered default pick, denial when nothing is authorized, fail-closed on policy error, and the token clamp.
- `test_mcp_dev_only_guard.py` — name-based detection incl. hyphen/underscore equivalence, explicit marker in both directions, disabled servers ignored, and each production signal (`ENVIRONMENT` or `DEBUG_MODE`) sufficing alone.

Full suite **2170 passed, 17 skipped**. Lint unchanged at 45 (matches `origin/main`). `check-docs.sh` passes.

## Note on scope

The review also recommended sandboxing MCP subprocesses and minimizing their inherited environment (they currently inherit the backend's, so a compromised tool sees its credentials). That's a larger change with real blast radius and isn't in this PR — the startup guard here reduces the exposure that motivated it, but does not replace it.

PR 3 (domain whitelist removal, lockfile sync, CI security gating) follows.


garland3:
## Summary
Hardens several server-side boundaries: `FEATURE_TOOLS_ENABLED` is now enforced in `ChatOrchestrator.execute` (and gates MCP startup), client-supplied `agent_max_steps` and MCP sampling `maxTokens` are clamped to configured ceilings, sampling model selection is authorization-checked, one chat turn per WebSocket is enforced, and a new startup guard aborts production boot when development-only MCP servers are enabled.

## Verdict
Request changes.

## Required changes
1. Add `dev_only: Optional[bool] = None` to `MCPServerConfig` at `atlas/modules/config/models.py:154` — the guard reads `config.get("dev_only")` (`atlas/core/mcp_dev_only_guard.py:73`) but is fed `{name: server.model_dump()}` (`atlas/modules/mcp_tools/client.py:103`), and pydantic's default `extra="ignore"` drops the key, so `"dev_only": true` on a custom host-access server is ignored in production and the documented `"dev_only": false` opt-out cannot clear the abort. Add a test that round-trips through `MCPServerConfig(...).model_dump()` before calling `find_enabled_dev_only_servers`, since the current tests pass hand-written dicts and cannot catch this.
2. Handle `error_type === 'turn_in_progress'` as a non-terminal case in `frontend/src/handlers/chat/websocketHandlers.js:485` — the generic `case 'error'` branch calls `setIsThinking(false)`, `clearAgentRunning()` and `endTokenStream()`, so sending a follow-up while an answer streams (Enter is not gated in `ChatArea.handleSubmit`) tears down the UI for the turn that is still running and streaming. Also gate Send/Enter on in-flight state so the rejection is rare rather than routine.
3. Call `enforce_dev_only_mcp_servers` after every `mcp_manager.reload_config()` in `atlas/routes/admin_routes.py:314` and `:871`, before `initialize_clients()`, returning a 400/409 naming the offending servers and leaving the previous config applied — today an admin can add `file_viewer` and hit reload to get it running in production without a restart, which is the scenario the guard exists for.
4. Move `start_user_client_cache_sweeper()` back outside the `if tools_enabled:` block at `atlas/main.py:297`, and call `mcp_manager.cleanup()` unconditionally on shutdown (`atlas/main.py:305`) — `RAGMCPService` creates MCP user clients lazily at request time regardless of the flag (`atlas/domain/rag_mcp_service.py:74`), so with tools disabled `_user_clients` grows unbounded and its subprocesses are never torn down.
5. Return a conservative default rather than the ceiling for unusable input in `_clamp_sampling_tokens` (`atlas/modules/mcp_tools/mcp_routing.py:313`) — `None`/`0`/`"9999"` currently resolve to 4096, so an explicit `maxTokens: null` now costs 8× the previous 512 default and malformed input is the cheapest route to the maximum allowance. Apply the same split in `_clamp_agent_steps` (`atlas/application/chat/orchestrator.py:193`).
6. Clamp `agent_max_steps` unconditionally at `atlas/application/chat/orchestrator.py:268` (drop the `if "agent_max_steps" in kwargs` guard) and read `kwargs["agent_max_steps"]` at `:402` instead of `kwargs.get("agent_max_steps", 10)` — the ceiling is currently enforced only against callers that supply the key, and the hardcoded 10 can drift from `AGENT_MAX_STEPS`.
7. Clear `active_chat_task["task"]` on the cancel paths (`atlas/main.py:963`, `:973`, `:141`) rather than relying on `done()` — `task.cancel()` only requests cancellation, so a stop followed promptly by a new `chat` frame is rejected, and a turn parked on an unanswered tool-approval dialog wedges the socket until the client reconnects.

## Evidence
Evidence: 7/10. The description, the three new test files, the full-suite and lint numbers, and the docs/CHANGELOG updates are proportionate for a backend change of this size, but the two behaviours that actually changed at runtime — the WebSocket single-turn rejection and the lifespan skip/abort — are asserted only in prose. The most important missing artifact is a WebSocket test that sends a second `chat` frame *after* the first turn completes, which is what proves the new gate does not wedge the socket.

## Notes
- Tests exercise `_tools_are_enabled` / `_clamp_agent_steps` in isolation via `__new__`; deleting the enforcement block from `execute()` leaves the new suite green (`atlas/tests/test_tools_flag_and_step_limits.py:36`).
- `_clamp_sampling_tokens` tests read the real global config, so the configured ceiling is indistinguishable from the hardcoded 4096 fallback (`atlas/tests/test_sampling_model_authorization.py:100`).
- `annotate_dev_only` has no caller outside its tests despite a docstring claiming admin UI/config API use (`atlas/core/mcp_dev_only_guard.py:135`).

---

garland3:
**Heads-up on CI: this PR will show no checks until it targets `main`.**

All three workflows (`ci.yml`, `security.yml`, `build-artifacts.yml`) trigger on `pull_request: branches: [main]`, so a PR based on another branch gets nothing — not a failure, just no runs. Checks will appear when #771 merges and this retargets to `main`.

Rebased onto #771's latest (which picked up review fixes). Verified locally with the same commands CI runs:

- Backend suite: **2193 passed, 17 skipped**
- `ruff check atlas/`: 45, unchanged from `origin/main`
- `scripts/check-docs.sh`: OK

One thing I checked specifically, because CI runs `test/run_tests.sh all` in **both** debug and production mode: the new dev-only MCP guard raises when `DEBUG_MODE=false`, so it could in principle break the production-mode run. It doesn't — the shipped `atlas/config/mcp.json` contains no dev-only servers:

```
shipped servers: calculator, elicitation_demo, pptx_generator, progress_demo, prompts,
                 session_state_demo, structured_output_demo, task_demo, tool_planner,
                 username-override-demo
flagged dev-only: []
RESULT: production startup OK with shipped mcp.json
```

`config/` is untracked, so CI never sees a local override that might enable `file_viewer`. That's the case the guard exists for, and it's exactly why it can't be verified by CI here.

PR #773, opened 2026-08-12
security: remove domain whitelist, sync lockfile, make CI security gates enforceable

garland3:
Third of three security PRs. **Stacked on #772** (which is stacked on #771), so this diff shows only its own changes. Retarget down the stack as each lands.

## 1. Remove the email domain whitelist

Requested outright, and the review independently found it **fails open**. The loader searched obsolete paths that did not include the documented project override, the package default, or `APP_CONFIG_DIR`, so when enabled it typically found no file — and `is_domain_allowed()` returns `True` on that path.

A restriction that silently admits every domain is worse than no restriction: it reads as a control that is present. Rather than fix path resolution for a feature nobody uses, it's removed.

Gone: `domain_whitelist.py`, `domain_whitelist_middleware.py`, the settings flag and its `FEATURE_DOE_LAB_CHECK_ENABLED` alias, `atlas/config/domain-whitelist.json`, `docs/admin/domain-whitelist.md` and its index entry, the `docker-compose.yml` variable, the `.env.example` entry, and `test_domain_whitelist.py`. The mention in `docs/archive/` stays — it's a dated record of how things were.

## 2. `uv.lock` was resolving below its own declared floors

`uv export --locked` had been failing. The consequence wasn't just staleness: the lock pinned **`reportlab` 4.4.10 under a `>=5.0.0,<6.0.0` constraint**. So the bump to 5.0.0 here *fixes a constraint violation* rather than introducing a major upgrade — worth knowing before reading the diff. 38 packages moved; the full list is in the commit message.

Worth flagging: CI installs with `pip install -e ".[dev]"` and does not use the lock, so the lock has been decorative for CI. The new consistency check is what stops it drifting again.

## 3. Two CI checks now actually block

Every dependency scan used `continue-on-error` or `|| true`. A green "Security Checks" only meant a scan had *run* — which is why it stayed green through everything above.

Two checks now fail the build:

| Check | Enforces |
|---|---|
| Lockfile is in sync | `uv export --locked` succeeds |
| Production frontend advisories | No high/critical in `npm audit --omit=dev` |

The rest stay advisory **on purpose**. They scan the whole dev environment, where most findings are build and test tooling that never reaches a user; a job that fails on those gets ignored, which is worse than not running it. `--omit=dev` is what makes the frontend gate meaningful — it covers only what is actually served.

## 4. Shipped frontend advisories cleared

An in-range `npm audit fix` takes `npm audit --omit=dev` to **zero**, so the new gate passes on merit rather than being tuned around the current state. I dry-ran the gate's own logic against the tree to confirm (`COUNT=0` → passes).

`package.json` is untouched — lockfile-only. Frontend: **521 tests pass**, lint unchanged, `npm run build` succeeds.

The 5 advisories left in the full tree (1 critical, 1 high, 3 moderate) are all `vitest`/`vite`/`esbuild`, need a `vitest` major, and don't ship to a browser. New [`docs/developer/dependency-advisories.md`](docs/developer/dependency-advisories.md) records what blocks, why the production/dev split exists, and how to accept an advisory with an expiry rather than weakening the gate.

## Testing

Backend **2158 passed, 17 skipped** — 12 fewer than #772, which is exactly the whitelist tests leaving with the feature. Lint unchanged at 45. `check-docs.sh` passes. `security.yml` validated as parseable YAML.

## Not done here

The review also wanted the Python side gated (Safety on the resolved production set rather than the installed dev environment). That needs a production-only resolution step that doesn't exist yet, and gating on the current dev-environment scan would block on test tooling. Left out rather than added as a gate that would need immediate exceptions.


garland3:
**Heads-up on CI: this PR will show no checks until it targets `main`.**

All three workflows trigger on `pull_request: branches: [main]`, so a PR based on another branch gets no runs. Checks appear once #771 and #772 merge and this retargets.

That's worth naming explicitly here, because **this PR adds two CI gates that cannot run on this PR**. I dry-ran the frontend gate's own logic against the tree instead:

```
gate would see COUNT=0
RESULT: gate PASSES
```

And the lockfile gate directly — `uv export --locked` now succeeds, where it failed before this change.

Rebased onto #772's latest. One conflict in `atlas/main.py`, resolved: #771 renamed the auth import to `resolve_user_from_auth_header_async` on the line adjacent to the `DomainWhitelistMiddleware` import this PR deletes. Kept both intents.

Verified locally:

- Backend suite: **2181 passed, 17 skipped** (12 fewer than #772 — exactly the whitelist tests leaving with the feature)
- Frontend: **521 passed**, lint unchanged, `npm run build` succeeds
- `ruff check atlas/`: 45, unchanged from `origin/main`
- `scripts/check-docs.sh`: OK
- `security.yml` validated as parseable YAML

Issue #760, opened 2026-08-09
Client disconnect cancels and discards the in-flight turn; no session resume across reconnect or device


garland3:
## Problem

Closing the browser tab, switching devices, or any transport interruption cancels the in-flight agent turn and discards everything it had already done. Reconnecting does not resume — it starts a new session with no way to reattach.

Reproduce:

1. Start an agent-mode turn that will run a batch of tool calls.
2. Close the tab (or put the laptop to sleep, or lose the network for ~30s).
3. Reopen the app and load the same conversation.

The turn is gone. Not partially saved — absent. Nothing indicates it ever ran, and there is no error. A follow-up message starts from scratch and re-does work.

This is not a local-setup quirk. On a clustered deployment it is routine: a rolling update, a pod eviction, an HPA scale-down, or a mobile network handover each produce the same silent loss of an in-flight turn for every affected user.

## Mechanism

A client disconnect runs the same cancel path as the Stop button. `atlas/main.py:892`:

```python
except WebSocketDisconnect:
    await cleanup_disconnected_session(
        chat_service, session_id, user_email, active_chat_task
    )
```

`cleanup_disconnected_session` (`atlas/main.py:139-162`) cancels first, then tears down:

```python
task = active_chat_task.get("task")
if task and not task.done():
    logger.info("Cancelling active chat task (client disconnected)")
    task.cancel()
...
await mcp_tool_manager.release_sessions(conv_id, user_email=user_email)
await chat_service.end_session(session_id)
```

The `task.cancel()` raises `asyncio.CancelledError` out of `orchestrator.execute(...)`, which — as documented in detail in #755 — bypasses the conversation save block at `atlas/application/chat/service.py:506-538`, because `CancelledError` is a `BaseException` and is caught by neither `except DomainError` nor `except Exception`. So nothing from the turn is persisted.

The existing docstring is explicit that cancelling is intentional:

> Cancels the in-flight turn *before* releasing resources: without this the background chat task keeps streaming tokens, calling tools and holding MCP sessions against a session that end_session() is about to tear down -- work nobody can receive.

That reasoning is sound for token streaming. It does not hold for tool calls that **already completed** — those are not "work nobody can receive", they are work already done, and they are thrown away rather than committed.

## Why reconnecting cannot recover it

Session identity is bound to the socket, not to the user or the conversation. `atlas/main.py:547`, inside `websocket_endpoint`:

```python
session_id = uuid4()
```

[...]

PR #776, opened 2026-08-13
fix: persist stopped agent turns and carry tool work into the next turn (#755)

### LINKED ISSUE #755: Agent mode: stopping the agent discards the turn - prior tool calls are not persisted and not in the next turn's context
## Problem

Run agent mode, let it work through a batch of tool calls (~20 is enough to make it obvious), then click the red "Stop agent" button. Send a follow-up message such as "continue" or a steering instruction.

The agent behaves as if it is starting from scratch. It has no idea the 20 tool calls happened, does not know what it found, and re-does work. Reloading the page afterwards is worse: the whole interrupted turn is gone from the saved conversation.

Two separate defects combine here.

### 1. A stopped turn is never persisted

"Stop agent" is a plain task cancel. `atlas/main.py:851-861`:

```python
elif message_type == "agent_control":
    action = data.get("action")
    if action == "stop":
        task = active_chat_task.get("task")
        if task and not task.done():
            logger.info("Cancelling active chat task (agent_control stop)")
            task.cancel()
```

The cancel raises `asyncio.CancelledError` out of `orchestrator.execute(...)` inside `ChatService.handle_chat_message` (`atlas/application/chat/service.py:463` / `:482`). The conversation save lives *after* that await, at `atlas/application/chat/service.py:506-538`:

```python
result = await orchestrator.execute(...)   # <-- CancelledError raised here
...
# Persist conversation (if not incognito and feature enabled)
if (self.conversation_repository is not None and ...):
    saved = self._save_conversation(session, user_email, model, ...)
    ... await update_callback({"type": "conversation_saved", ...})
```

`CancelledError` is a `BaseException`, so neither `except DomainError` (`service.py:541`) nor `except Exception` (`service.py:545`) catches it, and the only `finally` block just resets the compliance context token. The save block is skipped entirely — no `_save_conversation`, no `conversation_saved` event.

The cancel is then absorbed at the transport layer (`atlas/main.py:685-699`), which sends a terminal `token_stream` and `response_complete` and returns. Nothing on that path writes history either.

`AgentModeRunner.run` has the same shape: the final assistant message is appended at `atlas/application/chat/modes/agent.py:179-184`, after `agent_loop.run(...)` returns. Its `except Exception` at line 159 does not catch `CancelledError` either, so a stopped turn also ends with no assistant message in history at all.

### 2. Even in memory, prior tool calls are not part of the next LLM context

The agentic loop keeps the real tool-call transcript in a **local** list. `atlas/application/chat/agent/agentic_loop.py:168-221`:

```python
messages.append({
    "role": "assistant",
    "content": llm_response.content,
    "tool_calls": [_to_tool_call_dict(tc) for tc in tool_calls],
})
...
for result in results:
    messages.append({
        "role": "tool",
        "content": result.content,
        "tool_call_id": result.tool_call_id,
    })
```

That `messages` list is the parameter passed into `run()` and is discarded when the turn ends. What *does* go into `session.history` is display-only: the narration row (`agentic_loop.py:180-189`, `message_type="agent_intermediate"`) and the recorder rows (`agentic_loop.py:234` → `ToolCallRecorder.flush`, `message_type="tool_call"`).

Both types are explicitly excluded from the LLM context (`atlas/domain/messages/models.py:11-13` and `:165-179`):

```python
DISPLAY_ONLY_MESSAGE_TYPES = frozenset({"tool_call", "agent_intermediate"})
...
return [
    {"role": msg.role.value, "content": msg.content}
    for msg in self.messages
    if msg.metadata.get("message_type") not in DISPLAY_ONLY_MESSAGE_TYPES
]
```

The next turn builds its context from exactly that call — `MessageBuilder.build_messages` at `atlas/application/chat/preprocessors/message_builder.py:143`:

```python
history_messages = session.history.get_messages_for_llm()
```

So the follow-up request contains the system prompt, the original user message, and the new user message. Twenty tool calls, their arguments, and their results contribute nothing.

The exclusion itself is correct as written (issue #684 added these rows for re-render; replaying orphaned `tool` rows would get rejected by providers). The gap is that there is no *model-visible* record of the agent's work to replace them. On a normal completion the final assistant answer at least carries a summary forward; after a stop there is not even that.

## Expected behavior

Stop should cancel only the in-flight LLM call / tool execution. Everything already completed should be committed:

- The user message, narration rows, and every completed `tool_call` row are saved to the conversation, and `conversation_saved` is emitted, so the turn survives a page reload.
- A follow-up message continues the same thread with the agent's prior work in the model's context — it should know which tools it already ran and what came back, and pick up from there.

## Actual behavior

- **Nothing from the interrupted turn is persisted.** `CancelledError` bypasses the save block at `service.py:506-538`. The transcript is visible in the browser only; reload and it is gone, and no `conversation_saved` arrives.
- **No assistant message closes the turn** — `modes/agent.py:179-184` is never reached.
- **The next turn's LLM context has no tool history.** The working `messages` list dies with the loop, and the surviving `tool_call` / `agent_intermediate` history rows are filtered out by `get_messages_for_llm()`. The model sees the original prompt plus the new one, and starts over.

## Suggested fix

Two independent pieces; the first is the smaller one.

**Commit the turn on cancel.** Wrap the `orchestrator.execute(...)` await in `ChatService.handle_chat_message` so `asyncio.CancelledError` runs the same persistence path before re-raising — factor `service.py:496-538` into a helper and call it from both the success path and a `except asyncio.CancelledError:` handler. `AgentModeRunner.run` should likewise catch `CancelledError` and append a terminal assistant message (e.g. metadata `interrupted: True`) so the turn is well-formed. Note `_save_conversation`'s `start_index` / incognito floor logic must be preserved in both paths.

**Give the follow-up turn something to reason over.** Options, roughly in increasing cost:

1. On stop, append a single assistant summary row (normal, non-display-only) recording the steps taken and tools called, so the next turn at least sees the trajectory. Cheapest, no format risk.
2. Keep the agent loop's working `messages` list on the session (e.g. `session.context["agent_transcript"]`) and prepend it for a follow-up turn in the same conversation. Higher fidelity but needs sanitizing so no `tool` message is orphaned from its `assistant` `tool_calls` message, or providers will reject the request.
3. Reconstruct assistant/`tool` pairs from the persisted `tool_call` rows at `get_messages_for_llm()` time. Most faithful, most invasive, and highest risk of the strict-alternation problems that motivated the display-only exclusion in the first place.

Option 1 combined with the cancel-path save fixes the reported symptom with the least surface area.

## Files involved

- `atlas/main.py:851-861` (`agent_control` stop), `:685-699` (`CancelledError` transport handling)
- `atlas/application/chat/service.py:463-540` (turn execution + persistence block)
- `atlas/application/chat/modes/agent.py:150-186` (loop invocation, final assistant append)
- `atlas/application/chat/agent/agentic_loop.py:134-234` (local `messages` transcript, history flush)
- `atlas/application/chat/utilities/tool_history.py` (`ToolCallRecorder`, display-only rows)
- `atlas/domain/messages/models.py:11-13`, `:165-179` (`DISPLAY_ONLY_MESSAGE_TYPES`, `get_messages_for_llm`)
- `atlas/application/chat/preprocessors/message_builder.py:143`
- `frontend/src/contexts/ChatContext.jsx:521`, `:605` (stop send sites)

garland3:
Closes #755.

## Defect 1 — a stopped turn was never persisted

"Stop agent" is a plain `task.cancel()`. `asyncio.CancelledError` is a `BaseException`, so it slipped past both `except DomainError` and `except Exception` in `ChatService.handle_chat_message` and skipped the persistence block that sits after the `orchestrator.execute(...)` await. No `_save_conversation`, no `conversation_saved` — the user message, the agent's narration, and every completed tool call were browser-only and gone on reload. The same path runs on client disconnect (#760) and `reset_session`, so the fix is written for any cancel, not for the Stop button specifically.

- `service.py`: the save block is factored into `_commit_turn()`, called from the completion path and from a new `except asyncio.CancelledError:` handler that re-raises afterwards. The incognito save floor is applied identically in both, so a cancelled turn cannot persist messages produced while the session was incognito (covered by a test).
- `agentic_loop.py`: the step loop moved into `_run_steps()`; on any unwind the recorder's calls for the in-flight step are flushed into history, so tool calls that completed just before the stop survive. Calls that never reported a result are closed out (`status="failed"`, explanatory result) instead of persisting as `status="calling"` and reloading as forever-in-progress.
- `modes/agent.py`: `AgentModeRunner` catches `CancelledError`, appends a terminal assistant message (`interrupted: True`, "[Stopped by the user before this turn finished…]") so the turn is well-formed, emits `agent_completion` so the frontend clears its agent UI state, then re-raises.

## Defect 2 — prior tool calls were not in the next turn's context

The loop's real `assistant`/`tool` transcript is a local list discarded when the turn ends; what survives in history (`tool_call`, `agent_intermediate`) is display-only and excluded from `get_messages_for_llm()`. So a follow-up message reached the model with nothing but the original prompt. Per the measurements in the issue thread this degrades ordinary completed turns too — ~20% of tool calls in that corpus were re-establishing context from an earlier turn — so the digest runs on every agent turn, not only on the cancel path.

Implemented as **option 1** from the issue: a compact digest of the turn's calls (name, arguments, result) is attached to that turn's closing assistant message as `agent_tool_digest` metadata, and `ConversationHistory.get_messages_for_llm()` folds it into that message's content. Because it rides on an existing message, the role sequence is byte-identical to today's — no consecutive assistant turns, no orphaned `tool` row, nothing for a strict-alternation provider to reject. Options 2 and 3 were not pursued for exactly those risks.

Bounded on purpose (`utilities/agent_digest.py`): 400 chars of result and 300 of arguments per call, 30 calls per digest with head+tail kept and the elided middle announced. The issue's corpus showed ~14% of tool-result tokens were box-drawing characters from TUI capture; a digest with a hard cap keeps the identifiers the next turn actually needs without replaying the payload.

## Tests

`atlas/tests/test_interrupted_turn_persistence.py` (12 tests): digest construction (format, per-call caps, call-count elision, failed-call marking, turn scoping), the LLM-context fold (digest present; role sequence unchanged; messages without a digest untouched), the agent runner on both a completed turn and a mid-tool cancel (history reads `user -> agent_intermediate -> tool_call -> assistant(interrupted)`, digest carries the completed call, `agent_completion` still published, `CancelledError` still propagates), and `ChatService` on cancel (turn saved with its `tool_call` rows, `conversation_saved` emitted, cancel re-raised; incognito turn still not saved).

Full backend suite: **2156 passed, 17 skipped**. `ruff check` clean on all changed and added files. No frontend changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,7 @@ wheels/
 MANIFEST
 
 # Virtual environments
-.venv/
+.venv
 venv/
 ENV/
 env/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### PR #776 - 2026-08-12
+- **Stopping a turn no longer discards it** (closes #755): `asyncio.CancelledError` is a `BaseException`, so a stop / disconnect / `reset_session` skipped `ChatService`'s persistence block entirely and the interrupted turn was lost on reload. Both paths now go through one `_commit_turn()`; agent and tools mode flush the in-flight tool calls while unwinding (a call stopped mid-flight persists as `interrupted`, rendered neutrally rather than as an error), plain/RAG keep the text that already streamed, and every stopped turn is closed by an assistant message marked `interrupted`.
+- **A follow-up turn can see what the agent already ran**: each agent turn's closing assistant message carries a capped digest of its tool calls in `agent_tool_digest` metadata, folded into that message's content by `get_messages_for_llm()` — no new message, no role-sequence change — so the model stops re-deriving work it did in an earlier turn.
+
 ### PR #771 - 2026-08-11
 - **WebSocket authentication honours the configured header type**: `AuthMiddleware` branched on `AUTH_USER_HEADER_TYPE` and cryptographically verified `aws-alb-jwt` tokens, but both WebSocket endpoints called `get_user_from_header()` unconditionally — which only strips whitespace. In an ALB-JWT deployment any non-empty `X-User-Email` value therefore authenticated a socket that the same value could not authenticate over HTTP. All three call sites now resolve identity through one `resolve_user_from_auth_header()` helper, so the header type cannot be interpreted differently in different transports.
 - **Calculator expression syntax is stricter than `eval` was**: `sum([1, 2, 3])` and `round(x, ndigits=2)` no longer work — use `sum((1, 2, 3))` with parentheses and `round(x, 2)` positionally. String literals, complex numbers such as `2j`, and non-finite results (`inf`, `nan`) are now refused as well, the last two because they cannot be encoded in the tool's JSON response. The tool docstring lists every refused form with its replacement.
--- a/atlas/application/chat/agent/agentic_loop.py
+++ b/atlas/application/chat/agent/agentic_loop.py
@@ -17,7 +17,7 @@ to manage its own control flow.
 from __future__ import annotations
 
 import logging
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 from atlas.domain.messages.models import Message, MessageRole
 from atlas.interfaces.llm import LLMProtocol, LLMResponse
@@ -120,8 +120,6 @@ class AgenticLoop(AgentLoopProtocol):
                 self.tool_manager, selected_tools,
             )
 
-        steps = 0
-        final_answer: Optional[str] = None
         use_streaming = streaming and event_publisher
 
         # Record tool input/output as they stream to the UI so agent-mode tool
@@ -130,6 +128,80 @@ class AgenticLoop(AgentLoopProtocol):
         # mode executes tools through this loop's own callback, so it needs the
         # same wrapper here.
         recorder = ToolCallRecorder(self.connection.send_json if self.connection else None)
+
+        try:
+            steps, final_answer = await self._run_steps(
+                model=model,
+                messages=messages,
+                context=context,
+                tools_schema=tools_schema,
+                data_sources=data_sources,
+                max_steps=max_steps,
+                temperature=temperature,
+                event_handler=event_handler,
+                use_streaming=use_streaming,
+                event_publisher=event_publisher,
+                recorder=recorder,
+            )
+        except BaseException:
+            # A stop, a client disconnect, or a mid-step failure leaves this
+            # step's already-executed tool calls only in the recorder. Flush
+            # them so the interrupted turn still persists what actually ran
+            # (issue #755); anything that never reported a result is closed out
+            # rather than left rendering as in-progress forever.
+            try:
+                recorder.flush(context.history, mark_incomplete=True)
+            except Exception:  # pragma: no cover - never mask the real failure
+                logger.warning("Failed to flush tool calls while unwinding", exc_info=True)
+            raise
+
+        # Max steps exhausted without a text-only response
+        if final_answer is None:
+            if use_streaming:
+                final_answer = await stream_final_answer(
+                    self.llm, event_publisher, model, messages,
+                    temperature, context.user_email,
+                )
+            else:
+                final_answer = await self.llm.call_plain(
+                    model, messages, temperature=temperature,
+                    user_email=context.user_email,
+                )
+
+        await event_handler(AgentEvent(
+            type="agent_completion", payload={"steps": steps},
+        ))
+        return AgentResult(
+            final_answer=final_answer,
+            steps=steps,
+            metadata={
+                "agent_mode": True,
+                "strategy": "agentic",
+            },
+        )
+
+    async def _run_steps(
+        self,
+        *,
+        model: str,
+        messages: List[Dict[str, Any]],
+        context: AgentContext,
+        tools_schema: List[Dict[str, Any]],
+        data_sources: Optional[List[str]],
+        max_steps: int,
+        temperature: float,
+        event_handler: AgentEventHandler,
+        use_streaming: bool,
+        event_publisher,
+        recorder: ToolCallRecorder,
+    ) -> Tuple[int, Optional[str]]:
+        """Run the tool-calling steps, returning ``(steps, final_answer)``.
+
+        ``final_answer`` is ``None`` when the step budget ran out before the
+        model produced a text-only response.
+        """
+        steps = 0
+        final_answer: Optional[str] = None
 
         while steps < max_steps:
             steps += 1
@@ -233,30 +305,7 @@ class AgenticLoop(AgentLoopProtocol):
             # final assistant message the caller appends after run() returns.
             recorder.flush(context.history)
 
-        # Max steps exhausted without a text-only response
-        if final_answer is None:
-            if use_streaming:
-                final_answer = await stream_final_answer(
-                    self.llm, event_publisher, model, messages,
-                    temperature, context.user_email,
-                )
-            else:
-                final_answer = await self.llm.call_plain(
-                    model, messages, temperature=temperature,
-                    user_email=context.user_email,
-                )
-
-        await event_handler(AgentEvent(
-            type="agent_completion", payload={"steps": steps},
-        ))
-        return AgentResult(
-            final_answer=final_answer,
-            steps=steps,
-            metadata={
-                "agent_mode": True,
-                "strategy": "agentic",
-            },
-        )
+        return steps, final_answer
 
     async def _call_llm(
         self,
--- a/atlas/application/chat/modes/agent.py
+++ b/atlas/application/chat/modes/agent.py
@@ -1,9 +1,15 @@
 """Agent mode runner - handles LLM calls with agent loop execution."""
 
+import asyncio
 import logging
 from typing import Any, Awaitable, Callable, Dict, List, Optional
 
-from atlas.domain.messages.models import Message, MessageRole, ToolResult
+from atlas.domain.messages.models import (
+    AGENT_TOOL_DIGEST_KEY,
+    Message,
+    MessageRole,
+    ToolResult,
+)
 from atlas.domain.sessions.models import Session
 from atlas.interfaces.events import EventPublisher
 
@@ -11,6 +17,11 @@ from ..agent import AgentLoopFactory
 from ..agent.protocols import AgentContext
 from ..events.agent_event_relay import AgentEventRelay
 from ..utilities import event_notifier
+from ..utilities.agent_digest import build_tool_digest
+from ..utilities.interrupted_turn import (
+    INTERRUPTED_TURN_CONTENT,
+    INTERRUPTED_TURN_CONTENT_WITH_DIGEST,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -142,6 +153,10 @@ class AgentModeRunner:
             artifact_processor=process_artifacts,
         )
 
+        # Everything the loop appends to history from here on belongs to this
+        # turn; remember where it starts so the tool digest covers only it.
+        turn_start_index = len(session.history.messages)
+
         # Run the loop (always streaming final answer)
         try:
             result = await agent_loop.run(
@@ -156,17 +171,28 @@ class AgentModeRunner:
                 streaming=True,
                 event_publisher=self.event_publisher,
             )
+        except asyncio.CancelledError:
+            # Stop button, client disconnect, or reset_session (issue #755).
+            # The loop has already flushed every completed step's narration and
+            # tool_call rows into history, but nothing closes the turn, so the
+            # saved conversation would end mid-trajectory and the next turn
+            # would see no trace of the work. Append a terminal assistant
+            # message carrying the same tool digest a completed turn gets, so
+            # the turn is well-formed and the follow-up can pick up from there.
+            self._close_turn(
+                session,
+                turn_start_index,
+                content=INTERRUPTED_TURN_CONTENT,
+                metadata={"agent_mode": True, "interrupted": True},
+                content_with_digest=INTERRUPTED_TURN_CONTENT_WITH_DIGEST,
+            )
+            await self._publish_completion(steps=0)
+            raise
         except Exception:
             # Send agent completion event so the frontend clears agent UI state
             # (currentAgentStep, thinking indicator, etc.) before the error
             # message arrives via the WebSocket error handler.
-            try:
-                await self.event_publisher.publish_agent_update(
-                    update_type="agent_completion",
-                    steps=0,
-                )
-            except Exception as cleanup_exc:
-                logger.warning("Failed to send agent_completion cleanup event: %s", cleanup_exc)
+            await self._publish_completion(steps=0)
             raise
 
         # Append final message. Ordering contract with AgenticLoop: the loop
@@ -176,12 +202,12 @@ class AgentModeRunner:
         # returns — reloaded history reads user -> intermediate assistant
         # -> tool_call(s) -> assistant. Guarded by
         # TestAgentModeRunnerPersistedOrder in test_tool_call_persistence.py.
-        assistant_message = Message(
-            role=MessageRole.ASSISTANT,
+        self._close_turn(
+            session,
+            turn_start_index,
             content=result.final_answer,
             metadata={"agent_mode": True, "steps": result.steps},
         )
-        session.history.add_message(assistant_message)
 
         # Completion update
         await self.event_publisher.publish_agent_update(
@@ -190,3 +216,48 @@ class AgentModeRunner:
         )
 
         return event_notifier.create_chat_response(result.final_answer)
+
+    def _close_turn(
+        self,
+        session: Session,
+        turn_start_index: int,
+        content: str,
+        metadata: Dict[str, Any],
+        content_with_digest: Optional[str] = None,
+    ) -> Message:
+        """Append the turn's closing assistant message, carrying a tool digest.
+
+        The digest (issue #755) is the only model-visible record of what the
+        agent ran: the loop's working transcript dies with the turn and the
+        persisted ``tool_call`` rows are display-only. Without it a follow-up
+        turn re-derives everything the agent already established — measured at
+        roughly a fifth of tool calls in ordinary, uninterrupted use.
+        """
+        try:
+            digest = build_tool_digest(session.history.messages, turn_start_index)
+        except Exception:  # pragma: no cover - defensive
+            logger.warning("Failed to build agent tool digest", exc_info=True)
+            digest = None
+        if digest:
+            metadata = {**metadata, AGENT_TOOL_DIGEST_KEY: digest}
+            if content_with_digest:
+                # Only claim a record exists when one is actually attached.
+                content = content_with_digest
+
+        message = Message(
+            role=MessageRole.ASSISTANT,
+            content=content,
+            metadata=metadata,
+        )
+        session.history.add_message(message)
+        return message
+
+    async def _publish_completion(self, steps: int) -> None:
+        """Emit agent_completion, tolerating a dead transport."""
+        try:
+            await self.event_publisher.publish_agent_update(
+                update_type="agent_completion",
+                steps=steps,
+            )
+        except Exception as cleanup_exc:
+            logger.warning("Failed to send agent_completion cleanup event: %s", cleanup_exc)
--- a/atlas/application/chat/modes/plain.py
+++ b/atlas/application/chat/modes/plain.py
@@ -1,5 +1,6 @@
 """Plain mode runner - handles simple LLM calls without tools or RAG."""
 
+import asyncio
 import logging
 from typing import Any, Dict, List, Optional
 
@@ -85,16 +86,29 @@ class PlainModeRunner:
         user_email: Optional[str] = None,
     ) -> Dict[str, Any]:
         """Execute plain LLM mode with token streaming."""
-        accumulated = await stream_and_accumulate(
-            token_generator=self.llm.stream_plain(
-                model, messages, temperature=temperature, user_email=user_email,
-            ),
-            event_publisher=self.event_publisher,
-            fallback_fn=lambda: self.llm.call_plain(
-                model, messages, temperature=temperature, user_email=user_email,
-            ),
-            context_label="plain",
-        )
+        # Text the user already watched stream in must survive a Stop
+        # (issue #755); the helper drops it here before re-raising.
+        partial: list = []
+        try:
+            accumulated = await stream_and_accumulate(
+                token_generator=self.llm.stream_plain(
+                    model, messages, temperature=temperature, user_email=user_email,
+                ),
+                event_publisher=self.event_publisher,
+                fallback_fn=lambda: self.llm.call_plain(
+                    model, messages, temperature=temperature, user_email=user_email,
+                ),
+                context_label="plain",
+                partial_sink=partial,
+            )
+        except asyncio.CancelledError:
+            if partial:
+                session.history.add_message(Message(
+                    role=MessageRole.ASSISTANT,
+                    content=partial[0],
+                    metadata={"interrupted": True},
+                ))
+            raise
 
         assistant_message = Message(
             role=MessageRole.ASSISTANT,
--- a/atlas/application/chat/modes/rag.py
+++ b/atlas/application/chat/modes/rag.py
@@ -1,5 +1,6 @@
 """RAG mode runner - handles LLM calls with RAG integration."""
 
+import asyncio
 import logging
 from typing import Any, Dict, List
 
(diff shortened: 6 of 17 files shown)


garland3:
## Summary
Persists an interrupted chat turn instead of discarding it: `CancelledError` is now caught in `ChatService` and in `AgentModeRunner`, unfinished tool calls are flushed as incomplete, and a per-turn tool digest is stored on the closing assistant message and folded into LLM context on later turns.

## Verdict
Request changes.

## Required changes
1. Snapshot the incognito persistence decision before the turn's await points at `atlas/application/chat/service.py:547` — the cancel sites (`atlas/main.py:151`, `atlas/main.py:866`) call `task.cancel()` and proceed straight to `end_session()`, which discards `_incognito_sessions` and pops `_incognito_save_floor`; if teardown wins the race, `_commit_turn` sees a non-incognito session with `start_index` defaulting to `0` and writes the entire incognito transcript. Capture `is_incognito` and the effective `start_index` in `handle_chat_message` and pass them in, or make a missing save floor fail closed.
2. Close the turn on cancellation for non-agent modes too — `PlainModeRunner.run`/`run_streaming` (`atlas/application/chat/modes/plain.py:68`, `:103`) append the assistant message only on success, so a stopped plain/RAG turn persists history ending on a `user` message; after reload the next request is `user -> user` and strict-alternation providers reject it. Either append the accumulated/interrupted assistant message from the runners, or have `_commit_turn` skip a slice ending on a `user` role.
3. Remove the tracked `.venv` symlink (`.venv:1`, git mode `120000`, pointing at an absolute path outside the checkout) with `git rm --cached .venv`, and change `.gitignore:35` from `.venv/` to `.venv` so a symlink of that name is also ignored — it is unrelated to this change and breaks `uv run`/`uv sync` on every other machine.
4. Frame the digest's tool results as untrusted data at `atlas/application/chat/utilities/agent_digest.py:47` — results (fetched pages, external server responses, command output) are currently merged into **assistant**-role content under an imperative header and replayed on every later turn, so instruction-shaped text in a result reads as the assistant's own words. Wrap each result in an explicit data delimiter and state in the header that it is verbatim tool output, not instruction; also update the now-false invariant comment at `atlas/application/chat/utilities/tool_history.py:16-19` ("never replayed to the model").
5. Bound digest accumulation in `ConversationHistory.get_messages_for_llm()` (`atlas/domain/messages/models.py:186`) — each turn's digest is kept and re-folded on every subsequent request, so prompt size grows linearly with agent turns with no cap. Fold only the most recent N digests, or stop at a character budget walking newest-first.
6. Suppress or scope the client notification on the cancel path at `atlas/main.py:866` — after `reset_session`, `_commit_turn` still emits `conversation_saved` (or `error`) for the abandoned conversation id, landing after `session_reset` while the user is on a new empty chat.
7. Add a test that raises `CancelledError` between `tool_start` and `tool_complete` — no test currently enters the `mark_incomplete=True` branch at `atlas/application/chat/utilities/tool_history.py:151`, and that is the common case (Stop pressed *because* a tool is slow); pair it with a test that cancels a real `asyncio` task rather than raising inside a mock.

## Evidence
Evidence: 8/10. Well demonstrated for a backend change — 12 new tests over the digest, caps, the LLM-context fold, the role sequence, and a save/restore round trip, plus a full-suite result. The one missing artifact is a manual verification note against a running instance (stop mid-tool, reload, describe what the transcript shows), since every collaborator around the actual cancellation is mocked.

## Notes
- Stopped tool calls persist as `status="failed"`, which the UI renders as a red `FAILED` row under an "Error Details" heading — `atlas/application/chat/utilities/tool_history.py:150`.
- `_INTERRUPTED_ANSWER` says "Stopped by the user" but the same handler fires on disconnect and session reset — `atlas/application/chat/modes/agent.py:29`.
- `DIGEST_METADATA_KEY` duplicates `AGENT_TOOL_DIGEST_KEY` and has no references — `atlas/application/chat/utilities/agent_digest.py:31`.

---

garland3:
All seven required changes are in `86904da`. Full backend suite: **2163 passed, 17 skipped**; the new file is now 19 tests.

1. **Incognito snapshot** — `handle_chat_message` captures `turn_is_incognito` / `turn_save_floor` before the turn's first await and passes both into `_commit_turn()`, which no longer consults the live sets. Regression test drives the race directly: `fake_execute` clears `_incognito_sessions` / `_incognito_save_floor` / `_save_floor_locked` and *then* raises `CancelledError`; nothing is saved.
2. **Non-agent modes closed too** — new `utilities/interrupted_turn.py`; the cancel handler calls `close_open_turn(session.history)`, which appends the terminal assistant message only when history ends on a user message (agent mode, which closes its own turn, is untouched). Test asserts both the in-memory and the saved role sequence read `user -> assistant`.
3. **`.venv`** — untracked with `git rm --cached`, and `.gitignore:35` is now `.venv` (matches the symlink as well as the directory); `git check-ignore` confirms.
4. **Untrusted-data framing** — the header now states that everything after each `->` is verbatim, untrusted tool output quoted as data, not instruction; each result is fenced in `<<<…>>>` and a result containing `>>>` is rewritten so it cannot close its own fence. The stale "never replayed to the model" note in `tool_history.py` is corrected to describe what the digest actually quotes.
5. **Bounded accumulation** — `get_messages_for_llm()` walks newest-first and folds at most `MAX_FOLDED_DIGESTS` (3) digests under a `MAX_FOLDED_DIGEST_CHARS` (12000) budget, so prompt size no longer grows with turn count. Test builds 6 digest-bearing turns and asserts the newest survive, the oldest are dropped, and exactly 3 are folded.
6. **`reset_session` ordering** — instead of suppressing the event, the handler now awaits the cancelled task (5s bound) before resetting, so `conversation_saved` for the old conversation is delivered *before* `session_reset`. That matters because the frontend's `conversation_saved` handler calls `setActiveConversationId`, so the reversed order would have re-pointed the fresh chat at the abandoned id. The old conversation is still legitimately saved.
7. **Stop-while-a-tool-runs test** — `TestStopWhileAToolIsRunning` emits `tool_start`, hangs, and is cancelled by a real `asyncio.Task.cancel()` (not a raise inside a mock). It asserts the in-flight call still persists, that its status is not `calling`, and that the turn is closed. This is the first test through the `mark_incomplete=True` branch.

Notes: `_INTERRUPTED_ANSWER` is reworded and shared (it no longer blames the user, since the same path runs on disconnect and reset), and the duplicate `DIGEST_METADATA_KEY` is gone — `AGENT_TOOL_DIGEST_KEY` is the only definition.

On the secret-scan failure: two test method names matched TruffleHog's Lob detector (`test_` + 35 chars). Reproduced locally with trufflehog 3.96.0 against the same commit range, renamed both, and the scan is clean.

Still outstanding from your evidence note: manual verification against a running instance (stop mid-tool, reload, describe the transcript). Every collaborator around the cancellation is mocked, so that gap is real; flagging rather than claiming it.

garland3:
Force-pushed the branch as a single commit (`441ce3c`, same tree as `241efef`).

TruffleHog scans every commit in the range, not just the final tree, so the two test names that matched its Lob detector kept failing the check even after they were renamed — the earlier commits still contained them. Squashing removes them from the branch's history; a local run of trufflehog 3.96.0 over `3d1b5e4..HEAD` now reports zero findings.

Also added since the last review round: `TestStopThenContinueEndToEnd`, which drives the reported scenario through the production path — `ChatService.handle_chat_message(agent_mode=True)` → orchestrator → `AgentModeRunner` → `AgenticLoop` → persistence — with a tool that hangs on the second call, a real `asyncio.Task.cancel()`, and then a "continue" turn on the same session. It asserts the stopped turn is saved with both tool rows (one completed, one closed out), that it ends on an `interrupted` assistant message, and that the follow-up request the LLM receives actually contains `tmux ls` and its output. Only the LLM, tool manager, and transport are stand-ins.

Full backend suite: **2164 passed, 17 skipped**.

garland3:
Cleared the three CodeQL alerts on the test file in `1b5c7c7` (amended, same scope):

- `py/ineffectual-statement` ×2 — the bare `await task` inside `pytest.raises`. Both cancel sites now read `outcome = await asyncio.gather(task, return_exceptions=True)` followed by an `isinstance(outcome[0], asyncio.CancelledError)` assertion, which asserts the same propagation without a value-less expression statement.
- Mixed implicit/explicit returns in the fake tool — the hanging branch now ends in an explicit `raise AssertionError` after the never-completing await, so there is no implicit `None` path.

Suite still 20/20 on the new file, ruff clean.

garland3:
## Summary
Persists a turn that was stopped mid-flight instead of discarding it: the agentic loop flushes its tool-call recorder while unwinding from `CancelledError`, an interrupted-turn helper closes the open turn, and a compact tool digest is attached to the closing message and folded into later LLM requests. Also reorders `reset_session` to await the cancelled task before resetting.

## Verdict
Request changes.

## Required changes

1. Truncate an over-budget digest instead of dropping it at `atlas/domain/messages/models.py:211` — a single digest can reach ~21 KB against the 12,000-char fold budget, so `if len(digest) > budget: continue` skips the newest, largest digest while still folding a smaller older one, defeating the re-derivation fix precisely on the longest turns. Cut on a line boundary with an elision marker (or bound `build_tool_digest` output so it can never exceed `MAX_FOLDED_DIGEST_CHARS`), and add a test whose newest digest exceeds the budget asserting its head still reaches the model.
2. Make the data fence unforgeable at `atlas/application/chat/utilities/agent_digest.py:89` — `result.replace(">>>", ">>")` is not idempotent, so `">>>>"` becomes `">>>"` and a tool result (fetched page, third-party MCP response) can close its own delimiter inside an assistant-role message. Escape every `>` (or loop until fence-free) and assert `_RESULT_CLOSE not in body` over `">>>>"`, `">>>>>>"`, `"><>>>"`.
3. Normalize `tool_name` before interpolation at `atlas/application/chat/utilities/agent_digest.py:82` — it is the only field bypassing `_stringify`, so an MCP-advertised name containing a newline injects arbitrary extra digest lines, including a forged header or fabricated call record. Run it through `_stringify` or reject names outside `[A-Za-z0-9_.:-]+`, and test that digest line count equals tool-call count.
4. Flush the recorder on the cancel path in tools mode at `atlas/application/chat/modes/tools.py:166` — `ToolsModeRunner` flushes only on success paths, so a Stop during `execute_tools_workflow` discards every completed tool call, mirroring the exact defect fixed for agent mode. Wrap the tool-execution section in `try/except BaseException` and call `recorder.flush(session.history, mark_incomplete=True)` before re-raising, as `agentic_loop.py:151` does.
5. Persist the partial streamed text at `atlas/application/chat/modes/plain.py:88` (and the RAG runner) — `accumulated` dies with the frame on cancel, so paragraphs the user already watched stream in are replaced on reload by the interrupted placeholder. Append an assistant message with the accumulated content and `metadata["interrupted"] = True` before re-raising, leaving `close_open_turn` as the fallback when nothing streamed.
6. Stop rendering a deliberately stopped call as an error at `atlas/application/chat/utilities/tool_history.py:151` — `status="failed"` hits the red "Error Details" branch at `frontend/src/components/Message.jsx:470`, so a user's own Stop reads as a failure and is indistinguishable from a real tool error. Carry a distinct marker (`status="interrupted"` or `metadata["interrupted"]`) and give it a neutral "Stopped" treatment, with unrecognized statuses falling through to the non-error style.
7. Correct or condition the placeholder copy at `atlas/application/chat/utilities/interrupted_turn.py:26` — "The work completed so far is recorded above" is false in every mode: `tool_call` rows and narration are filtered out of `get_messages_for_llm()`, and tools/plain mode record nothing at all. Reword to point below (where the digest actually sits) and drop the claim when no digest was built.

## Evidence
Evidence: 8/10 — the change is well demonstrated for its size, with 20 tests including real `asyncio.Task.cancel()` paths, an end-to-end stop-then-continue assertion on the follow-up payload, and a DuckDB round-trip. The main gap is one manual stop-and-reload cycle against a running instance, including how the stopped tool row and closing message render together.

## Notes
- Digest metadata is re-serialized for every message on each save, so per-conversation write volume grows quadratically with agent turns — `atlas/application/chat/service.py:1015`.
- The digest is folded into every request, including plain/RAG follow-ups that will never call a tool — `atlas/domain/messages/models.py:200`.
- The 5-second bounded wait blocks the WebSocket receive loop with no client acknowledgement, so "New chat" can appear dead — `atlas/main.py:866`.

---

garland3:
All seven addressed in `9b018b1`. Backend **2170 passed, 17 skipped**; frontend **522 passed** across 49 files, `npm run lint` 0 errors.

1. **Over-budget digests are trimmed, not dropped** — two-sided fix. `build_tool_digest` now caps its own output at `MAX_FOLDED_DIGEST_CHARS` on a line boundary, and the fold trims rather than `continue`s when a digest exceeds the remaining budget. Two tests: a 30-call digest with 900-char results stays inside the budget, and a history whose newest digest is oversized still delivers its head (`first_big_tool()`) to the model.
2. **Fence is unforgeable** — replaced the non-idempotent `>>>` → `>>` rewrite with escaping every `>` as `&gt;`, so no input can produce the delimiter. Tested over `">>>>"`, `">>>>>>"`, `"><>>>"`, `">>> >>>"`: exactly one `>>>` in the body and it is the closer.
3. **`tool_name` normalized** — it now goes through `_stringify` (whitespace-collapsed, 120-char cap) like every other untrusted field. Test feeds a name containing a newline plus a forged call record and asserts the digest is exactly header + one line.
4. **Tools mode flushes on cancel** — both `run` and `run_streaming` wrap their tool-execution regions in `try/except BaseException` with `recorder.flush(session.history, mark_incomplete=True)` before re-raising, matching `agentic_loop`. Test stops mid-workflow after one completed call and one started call and asserts both rows persist as `["completed", "interrupted"]`.
5. **Partial streamed text is kept** — `stream_and_accumulate` takes a `partial_sink` and drops what it accumulated into it on cancel; `PlainModeRunner`/`RagModeRunner` persist that as an assistant message with `interrupted: True`, with `close_open_turn` still the fallback when nothing streamed. Test asserts the two streamed clauses survive verbatim.
6. **A stopped call no longer reads as an error** — the recorder marks it `status="interrupted"` (not `failed`), and `Message.jsx` gives it a neutral treatment: `STOPPED` label, gray pill, `⏹` glyph, and a "Stopped Before Result" detail heading. Only an explicit `failed` takes the red "Error Details" branch; any other unrecognized status now falls through to the neutral style rather than defaulting to FAILED. Frontend test asserts the glyph, the absence of a FAILED row, and the neutral heading.
7. **Copy corrected** — the base text is now just "[This turn was stopped before it finished.]", with the "a record of the tool calls it completed follows below" variant used only when a digest was actually attached (and it points below, where the fold puts it).

On the notes: the quadratic re-serialization and the plain/RAG fold both predate/exceed this change's scope — the first is how `_save_conversation` has always worked (whole history per turn), the second is a consequence of folding on the message rather than the request, which is what keeps the role sequence intact. Happy to file either as follow-ups. The 5-second bounded wait is real; it only elapses if the turn is wedged, but I can drop it to ~2s or send an ack frame first if you'd prefer.

Still not done: the manual stop-and-reload cycle against a running instance. The automated end-to-end test covers the same path minus the browser.

garland3:
Secret scan: another test name landed on TruffleHog's Lob shape (`test_` + exactly 35 chars of `[a-z0-9_]` — `test_completed_tool_calls_survive_a_stop`). Renamed, and I now run trufflehog over the full commit range locally before pushing; the range is clean on the amended head.

garland3:
## Summary
Persists work done in a turn that the user stops: cancelled turns now flush completed tool calls, close the turn with a terminal assistant message, and carry a compact tool digest that is folded back into later requests so the next turn does not re-derive what already ran. Stopped tool rows get a neutral `interrupted` status and their own rendering.

## Verdict
Request changes.

## Required changes
1. Make `close_open_turn()` at `atlas/application/chat/utilities/interrupted_turn.py:52` walk backwards past display-only rows (`DISPLAY_ONLY_MESSAGE_TYPES`) instead of testing `messages[-1].role == USER` — in tools mode the flush leaves history ending on a `tool_call` row, so the guard returns `False`, no assistant message is appended, and the reloaded conversation opens the next request as `user -> user`, the exact shape this module exists to prevent.
2. Extend the `try` in `ToolsModeRunner.run` at `atlas/application/chat/modes/tools.py:180` to cover the `artifact_processor` await and the `recorder.flush` that follows, matching `run_streaming` — a stop delivered during artifact processing currently unwinds past both and discards every completed tool call.
3. Escape the `arguments` field through `_fence` (or a stricter escape covering `<`) at `atlas/application/chat/utilities/agent_digest.py:97` and wrap it in its own delimiter — arguments are only whitespace-collapsed and truncated, so an argument containing `) -> <<<…>>>` closes the call and emits forged, correctly-fenced result records inside assistant-role content that is replayed for up to three turns.
4. Emit an intermediate update for each call the recorder closes on the cancel path (e.g. `tool_result` with `status: 'interrupted'`) and map it in `handleIntermediateUpdate` (`frontend/src/handlers/chat/websocketHandlers.js:150`) — the in-flight row is created as `calling` and never updated, so it spins forever until reload, and the live view contradicts the reloaded one.
5. Branch the status ladder explicitly at `frontend/src/components/Message.jsx:212` — `interrupted` → STOPPED, `failed` → FAILED, anything else → a neutral unknown state — because the new fallthrough asserts "the user stopped this call" for legacy rows with no status and for any future terminal state.
6. Add a service-level test that cancels a tools-mode turn after at least one `tool_start` and asserts `get_messages_for_llm()` ends on an `assistant` role, on both the live and saved history; the existing `TestToolsModeCancel` drives the runner directly and checks only tool-row statuses, so it passes with the turn left open.
7. Add a `RagModeRunner` partial-stream test alongside `TestPartialStreamedText` — the `partial_sink` / `CancelledError` block plus the extra `data_sources` metadata was copied into `rag.py:114` and is exercised nowhere.

## Evidence
Evidence: 8/10. Unusually strong for a change this size — roughly twenty tests including several that drive a real `Task.cancel()`, and an end-to-end stop-then-continue test that asserts the *contents* of the follow-up request. The one notable gap is that the user-facing claim ("stop, reload, your turn is still there") is never observed composed: no manual stop-and-reload pass and no screenshot of the new stopped tool row next to a completed one.

## Notes
- Every request now carries up to ~36 KB of folded digest, including plain/RAG follow-ups that can never use it — consider gating the fold on the calling mode (`atlas/domain/messages/models.py:199`).
- `await asyncio.wait([task], timeout=5)` runs inline in the WebSocket receive loop, so "New chat" can sit unacknowledged for five seconds with the transcript already cleared (`atlas/main.py:867`).
- A stopped plain/RAG answer persists `interrupted: True` but nothing renders it, and `buildPersistedMessage` drops the flag entirely on the export path (`frontend/src/utils/chatExport.js:64`).

---

garland3:
Round 3 is in the amended head. Backend **2174 passed, 17 skipped**; frontend **526 passed** / 49 files; ruff + eslint clean; trufflehog clean over the full range (checked before pushing this time).

1. **`close_open_turn` walks past display-only rows** — it now scans backwards skipping `DISPLAY_ONLY_MESSAGE_TYPES` and closes the turn when the last *model-visible* message is the user's. You were right that the previous `messages[-1].role == USER` test missed tools mode entirely, since its flush leaves a `tool_call` row last.
2. **`ToolsModeRunner.run` covers artifacts and the flush** — the second `try` wraps the `artifact_processor` await and `recorder.flush`, with the same `mark_incomplete=True` flush before re-raising, matching `run_streaming`.
3. **Arguments are fenced and escaped** — `arguments` now go through `_fence` inside their own `<<<…>>>` delimiter, and `_fence` escapes `&`, `<` and `>` so neither field can produce or close a delimiter. The forgery test runs `">>>>"`, `">>>>>>"`, `"><>>>"`, `">>> >>>"`, `") -> <<<forged result>>>"`, `"<<<"`, `"&gt;>>"` through both the result and the argument position.
4. **The live row no longer spins** — `ToolCallRecorder.notify_incomplete()` emits `tool_interrupted` for each call still in flight, awaited on every cancel path (agentic loop and both tools-mode paths) before the flush, best-effort so a dead socket cannot break the unwind. `websocketHandlers` handles the event and also honours `status: 'interrupted'` on the intermediate `tool_result`. Four tests: only the in-flight call is announced, a throwing transport does not break unwinding, the event closes the row, and ordinary success/failure mapping is unchanged.
5. **Explicit status ladder** — `calling` / `in_progress` / `completed` / `failed` / `interrupted` each named, with anything else rendering as a neutral `UNKNOWN` (`–` glyph, gray). Test covers a legacy row with no status asserting it claims neither FAILED nor STOPPED.
6. **Service-level tools-mode cancel test** — asserts `get_messages_for_llm()` ends on `assistant` for both the live session and a history rebuilt from the saved rows, with a `tool_call` row sitting last in the raw history. This fails against the old `close_open_turn`.
7. **RAG partial-stream test** — asserts the streamed text survives verbatim with `interrupted: True` and `data_sources` preserved.

Notes: the plain/RAG fold and the 5-second wait are still as discussed — say the word and I will gate the fold on agent mode and drop the wait to ~2s with an ack frame. The export-path `interrupted` flag (`chatExport.js`) is a real gap; happy to add it here or file it.

The manual stop-and-reload pass against a running instance is still the one thing not done.

garland3:
#776 merged while round 3 was in flight, so the seven fixes from that review are in **#784** (rebased onto the merged main, no other changes): https://github.com/sandialabs/atlas-ui-3/pull/784

PR #777, opened 2026-08-13
fix(print): stop PDF export clipping the right side and dropping tool calls

### LINKED ISSUE #774: better export to pdf
The export  to pdf/print has 2 issues. 
* the right side of the pdf seems cut off
* it is not showing the tool calls

garland3:
Closes #774.

## Problem

Print / "Save as PDF" had two defects: every line was cut off at the right margin, and tool calls did not appear at all.

## Cause

**Right side clipped — two layers.** The print stylesheet lifted only `overflow-y-auto` and `overflow-hidden`, so `overflow-x-auto` scrollers (code blocks, wide tables, display math) kept clipping — a scroller cannot be panned on paper. Lifting that clip then exposed the deeper cause: the chat column is a flex item whose default `min-width: auto` was being held at zero *by that same horizontal clip*. With the clip gone, one long code line stretched the whole layout to roughly three page-widths and everything past the first page-width was lost.

Wrapping `<pre>` was not enough either: the mobile-containment rule from #747 sets `overflow-wrap: normal` on `.chat-messages pre code`, and every highlight.js token span inherits it, so long tokens stayed unbreakable until the inner `<code>` and its spans were named explicitly.

**Tool calls missing.** The compact/classic tool summary renders inside a `<button>`, and the print hide rule is `button:not(.no-print-hide)` — so the whole row vanished. Its details (input arguments + output) were additionally rendered only while expanded, so even a printed row would have shown no input/output unless the user had opened every row by hand.

## Fix

- Print block also overrides `overflow-x-auto`, and pins `body *` to `min-width: 0` / `max-width: 100%` so nothing can outgrow the page.
- `pre`, `pre code`, `pre code *` get `overflow: visible` + `overflow-wrap: anywhere` + `white-space: pre-wrap` in print.
- `bg-gray-900` boxes are whitened alongside the existing `bg-gray-800` / `bg-blue-600` rules, so the now-printable detail boxes are not black-on-dark.
- Both tool summary buttons carry `no-print-hide`.
- Tool details are always in the DOM, `hidden print:block` — hidden on screen when collapsed, present in the PDF.

## Testing

- `frontend/src/test/print-export.test.jsx` (new, 12 tests) covers the print-stylesheet contract and the DOM contract, including a guard that copy/edit/rewind buttons stay hidden in print.
- Full frontend suite: **533 passed / 50 files**. `npm run lint`: 0 errors.
- Browser-verified against the running app with a real `calculator_evaluate` tool call and a 280-character single-line code block, print media emulated:
  - before this branch's layout fix the PDF cut the line at `...verylongtoken_veryl`; now it wraps across four lines and **0 elements** extend past the page width (`document.body.scrollWidth === clientWidth`).
  - with the tool row **collapsed** on screen (`display: none`), the generated PDF contains the tool name, `Input Arguments`, and `Output Result` with the full result value.


garland3:
Review feedback addressed in f23a34c.

**Codex (P2) — keep collapsed tool details lazy outside print.** Agreed, and fixed. Always-mounting meant every tool call serialized its arguments and result into hidden `<pre>` elements on every render, and MCP results are not size-bounded. A `useIsPrinting()` hook now mounts them only for the duration of the print job: `beforeprint`/`afterprint`, with the state change wrapped in `flushSync` so React has committed the extra DOM before the browser snapshots the page (a batched update would land after the snapshot and the details would be missing from the PDF), plus a `matchMedia('print')` fallback for renderers that only switch media. The window listeners are registered once at module scope rather than per component — a long transcript mounts hundreds of `Message`s, and one `flushSync` each would mean hundreds of synchronous renders.

Verified against the running app: with both tool rows collapsed the on-screen DOM contains zero detail wrappers before and after printing, while the PDF generated through the browser's own print path contains both rows' `Input Arguments` and `Output Result`.

**Copilot — empty details wrapper in print.** Fixed. A classic-mode row with no arguments and no result no longer mounts the wrapper at all, so it cannot print as a band of indented whitespace. (Compact rows still mount it when they carry only a server name, since that is where the server name lives.) Covered by a new test.

**Copilot — greedy `@media print` extraction in the test.** Fixed. `[\s\S]*` ran to the last `}` in `index.css`, so a rule from a later at-rule such as `@media (hover: none)` could have satisfied assertions meant to be print-scoped. The block is now extracted by counting braces, with a test asserting the extraction stops at the print block's own closing brace.

Frontend suite: 536 passed / 50 files. `npm run lint`: 0 errors.

garland3:
## Summary
Fixes print/PDF export of the chat transcript: a print stylesheet block that lifts horizontal clipping and wraps long code lines, plus a change making tool-call details always-rendered (`hidden print:block`) so collapsed rows still appear in the PDF. Adds a 12-case regression suite and a changelog entry.

## Verdict
Request changes.

## Required changes
1. Memoize the tool-detail serialization in `frontend/src/components/Message.jsx:452` — moving the block out from behind `showDetails &&` makes `filterArgumentsForDisplay` + `processToolResult` + two `JSON.stringify(..., 2)` calls run on every render of every collapsed tool call, so a session with large tool results re-serializes megabytes per streamed token. Wrap both expressions in `useMemo` keyed on `message.arguments` / `message.result` / `debugMode` (or extract a `memo`-wrapped child).
2. Reset overflow on display math and tables in the print block at `frontend/src/index.css:783` — the attribute selector `[class*="overflow-x-auto"]` cannot match `.chat-messages .katex-display`, whose `overflow-x: auto` comes from the stylesheet at `frontend/src/index.css:657`, so long equations are still clipped in the PDF despite being listed as fixed. Add `.chat-messages .katex-display` (and the markdown `table`) to the `overflow: visible !important` group, and extend the test to assert those selectors.
3. Scope the print-block extraction in `frontend/src/test/print-export.test.jsx:58` — the greedy `[\s\S]*` captures through the last `\n}` in the file, pulling in the trailing `@media (hover: none)` block, so the "scoped to print" guarantee stated in the file's own header comment does not hold. Use a lazy match or brace-balance from `@media print {`, and assert the captured text does not contain `@media (hover: none)`.
4. Add assertions for the two print rules the DOM changes depend on (`frontend/src/test/print-export.test.jsx:132`) — nothing pins `button:not(.no-print-hide)` (`index.css:704`) or `.print\:block { display: block !important }` (`index.css:801`), so a one-token edit to either re-hides tool rows or their details in the PDF with the suite still green.

## Evidence
Evidence: 6/10. The root-cause narrative and browser verification are unusually concrete (named tool call, 280-character line, `scrollWidth === clientWidth` post-fix check), but a purely visual print fix ships with no rendered artifact — the single most useful addition is a before/after screenshot of the print preview showing the clipped line and its wrapped replacement, ideally including a page with a collapsed tool row's input/output.

## Notes
- Printed PDFs now include every tool call's full arguments and result with no opt-out; call that out in user-facing terms in the changelog entry (`CHANGELOG.md:9`), and consider a key-name denylist in `filterArgumentsForDisplay` (`frontend/src/utils/toolResultUtils.js:5`), which masks only `file_data_base64`.
- `body * { min-width: 0; max-width: 100% }` (`frontend/src/index.css:797`) is a universal `!important` override for one flex item's `min-width: auto`; scoping it to the chat layout chain would keep the print block's targeted-selector convention.
- Download-link labels print as "N file(s) available for download:" with nothing under them (`frontend/src/components/Message.jsx:383`), since the links are `<button>` elements the print rule hides.

---

garland3:
Thanks — this reviewed `674de71`; two of the four were already fixed in `f23a34c`, and the other two are addressed in `d890dee`.

**1. Memoize the tool-detail serialization — resolved by a different fix.** Same defect Codex flagged, and I agree it was real. Rather than memoizing, `f23a34c` stopped rendering collapsed details on screen at all: a `useIsPrinting()` hook mounts them only for the duration of the print job (`beforeprint`/`afterprint`, `flushSync`ed so the DOM is committed before the page snapshot, with a `matchMedia('print')` fallback). A collapsed row now serializes nothing, which is strictly better than serializing once and caching it. Verified in the browser: with both tool rows collapsed the DOM contains zero detail wrappers before and after printing, while the PDF contains both rows' input and output.

**2. Display math / tables still clipped — correct, and fixed with a caveat.** `.chat-messages .katex-display` gets `overflow-x: auto` from the stylesheet, so the attribute selector could never match it; same for the phone-width table rule. Both are now named explicitly, with tests.

But I checked what this actually buys, and it is less than "long equations print in full". I rendered an 18-term display equation (intrinsic width 1182px against a ~668px printable width): removing the box-level clip lets it escape its container, and it is then truncated at the *page* edge instead. KaTeX lays the formula out at a fixed width and cannot reflow, so no CSS override makes an over-wide equation fit. The rule is a strict improvement, not a cure — the comment and changelog now say that rather than listing display math as fixed. Code blocks are genuinely fixed because they also wrap.

**3. Greedy print-block extraction — fixed in `f23a34c`,** by brace-balancing, plus the assertion you suggested that the captured text excludes `@media (hover: none)`.

**4. Pin `button:not(.no-print-hide)` and `.print\:block` — fixed in `d890dee`.** Good catch: those two rules are load-bearing for the DOM changes and nothing tied them down.

**On the notes:**
- *Download-link labels print with nothing under them* — real, and fixed: the three download buttons now carry `no-print-hide`, so the file names print under their label.
- *Printed PDFs include full tool arguments/results with no opt-out* — called out in the changelog in user-facing terms. I did not extend the `filterArgumentsForDisplay` denylist: that function governs on-screen display too, so broadening it changes what expanded rows show, which is a separate decision from this print fix and shouldn't ride along in it.
- *`body * { min-width: 0; max-width: 100% }` is broad* — deliberate, and I'd keep it. Scoping it to the chat layout chain hard-codes a specific ancestor chain into the print block, which silently stops working the next time that markup is restructured; the constraint we actually want is "in print, nothing may be wider than the page", and the universal selector states exactly that. It is confined to `@media print`.

Frontend suite: 540 passed / 50 files. `npm run lint`: 0 errors.

garland3:
## Summary
Fixes print/PDF export of chat transcripts: a new `useIsPrinting` hook mounts collapsed tool-call details during a print job, and the `@media print` block is reworked so horizontally-scrolling content (code blocks, tables, math) wraps instead of clipping at the right margin. Download-file names and tool rows now survive into the PDF, covered by a new 12-case regression suite.

## Verdict
Approve with nits.

## Evidence
Evidence: 7/10. The change is well-described and well-tested — a new regression file pins both the stylesheet and DOM contracts, the full suite and lint are clean, and browser verification is stated with concrete numbers rather than assertions. For a fix whose entire deliverable is visual, the one missing artifact is a before/after screenshot of the print preview showing the long line clipped and then wrapped; naming the browser(s) the `beforeprint` ordering was confirmed against would also close a real gap.

## Notes
- Exported PDFs now include every tool call's full arguments and output with no opt-out, and the only redaction is a single-key mask (`file_data_base64`) in `filterArgumentsForDisplay` (`frontend/src/utils/toolResultUtils.js:5`); consider a key-name denylist for credential-shaped arguments, or a setting to exclude tool I/O from export. Unbounded results also mean a few large tool calls can turn a 2-page transcript into a 30-page dump (`frontend/src/index.css:783` lifts `max-height`).
- `print-export.test.jsx:263` dispatches `beforeprint` without a matching `afterprint`, and `printingViaEvent` is module-scoped, so every later test runs stuck in the printing state — add an `afterEach` that dispatches `afterprint` (or a test-only reset) so coverage stops depending on declaration order.
- The download buttons use `rounded`, not `rounded-lg`, so the whitening rule `[class*="rounded-lg"][class*="bg-blue-600"]` (`frontend/src/index.css:739`) misses them while `* { color: black !important }` applies — file names print black-on-blue when "Background graphics" is enabled.

---

garland3:
Both code nits were real and are fixed in `ff7ac2c`.

**Test state leaked across tests.** Correct — `beforeprint` was dispatched without a matching `afterprint`, and the hook's flag is module-scoped, so every later test in the file ran stuck in the printing state. `afterEach` now dispatches `afterprint`, and the file passes under `--sequence.shuffle` (20/20), which it would not have before.

**Download buttons printed black-on-blue.** Also correct: they carry `rounded`, not `rounded-lg`, so the whitening rule never matched them while `* { color: black !important }` did. Rather than widening the radius selector, buttons that opt into printing now drop their screen fill outright (`button.no-print-hide { background: transparent !important }`) — that covers any future printable button regardless of which corner-radius utility it happens to use. Verified in Chromium 150.0.7871.114: under print media a button with the download-button classes computes to `display: flex`, `background: rgba(0,0,0,0)`, `color: rgb(0,0,0)`.

**On the evidence gap** — fair. The browser is Chromium 150.0.7871.114, and the `beforeprint` ordering was confirmed through `page.pdf()` with no media override, which is the same path a Ctrl+P / "Print / Save as PDF" click takes (`beforeprint` → snapshot → `afterprint`); the `flushSync` is what makes the mount land before the snapshot, and dropping it is what would break silently. Concretely, on a transcript with two collapsed tool rows the extracted PDF text contains both rows' `Input Arguments` and `Output Result` while the live DOM contains zero detail wrappers before and after, and the 280-character line comes back as five wrapped lines rather than one truncated at `...verylongtoken_veryl`. I can't attach images to a PR comment from the CLI, so the artifact stays a text extraction rather than a screenshot.

**On tool I/O in exported PDFs** (page count and redaction): including tool input/output is the substance of #774 — "it is not showing the tool calls" — so unbounded length is the requested behavior rather than a regression, and I'd rather not pair the fix with a switch that partially undoes it. The redaction point is worth pursuing, but `filterArgumentsForDisplay` governs the on-screen expanded view too, so a credential-shaped key denylist changes what users see in the app, not just in exports. That deserves its own issue and its own decision about what gets masked where.

Frontend suite: 541 passed / 50 files. `npm run lint`: 0 errors.

garland3:

garland3:
Note on the last two commits, so the history reads clearly:

The conflicts with `main` (from #776) were already resolved in `00b1e6f` before the automated merge landed. Both resolutions agree on the substance, and the important half is worth stating explicitly: #776 added the `failed` / `wasInterrupted` status labels ("Stopped Before Result") *inside* the tool-detail block that this branch moved behind `renderDetails` so it can mount for printing. The merged code keeps this branch's structure and #776's labels, so a tool call stopped mid-flight renders neutrally on screen **and** still appears in the PDF.

`3edb2a1` restores `frontend/package-lock.json` to main's LF copy. The merge commit `a4c0934` had rewritten the whole file with CRLF endings — byte-identical once carriage returns are stripped, so its only effects were 12,644 lines of noise in the PR diff and a future conflict magnet. This branch changes no dependencies, and the diff against `main` is now back to the six files it should be:

```
CHANGELOG.md                                     |   2 +
frontend/src/components/Message.jsx              | 113 +++---
frontend/src/hooks/useIsPrinting.js              |  74 +++++
frontend/src/index.css                           |  58 +++-
frontend/src/test/print-export.test.jsx          | 330 ++++++++++++++++++
frontend/src/test/tool-approval-message.test.jsx |   4 +-
```

Full frontend suite on the merged branch: 542 passed / 50 files. `npm run lint`: 0 errors.

PR #781, opened 2026-08-13
feat: reasoning content streaming for reasoning-capable LLMs

garland3:
## Summary

Rebuilds #655 as a single commit on current `main`. That branch was ~230 commits behind and no longer applied; this reproduces its intent against today's code, which had meanwhile refactored several of the touched call sites.

Reasoning-capable models (o-series, Qwen3, GPT-OSS via vLLM) emit chain-of-thought before their answer, and Atlas discarded all of it. This streams it to the frontend in real time and shows it in a collapsible per-message section that persists across reloads.

### Backend

- `ReasoningToken` / `ReasoningBlock` markers in `atlas/modules/llm/models.py`, plus `LLMResponse.reasoning_content`.
- `litellm_streaming.py` yields a `ReasoningToken` per reasoning delta and a `ReasoningBlock` with the full text once content begins — or at end of stream, so a reasoning-only turn still reports its reasoning.
- The markers travel in the same stream as content strings, so every consumer type-checks items rather than assuming `str`. `stream_and_accumulate`, `tools.py`, `agentic_loop.py`, and `stream_final_answer` forward them as `reasoning_token` / `reasoning_content` events instead of concatenating them onto the answer — which would raise.
- All four mode runners persist reasoning into assistant message metadata; `ChatResult` exposes it to API/CLI callers.

### Frontend

- `reasoning_token` / `reasoning_content` handlers reuse the existing 30ms buffered flush; the buffer is cleared on stream end so a late timer can't create an orphan message.
- `STREAM_REASONING_TOKEN` / `STREAM_REASONING_END` reducer actions in `useMessages`. The final event carries the backend's authoritative text, letting the reducer reconcile tokens coalesced or dropped while buffering.
- Collapsible reasoning section in `Message.jsx` — auto-expands while streaming with a blinking cursor, auto-scrolls only when the user is already near the bottom, and is keyboard/screen-reader accessible via `aria-expanded` / `aria-controls`.

### Two ordering details worth review

- A turn that reasons and then goes straight to tool calls must still close its token stream. Otherwise the reasoning bubble stays flagged `_streaming` and the post-tool synthesis appends to it instead of starting a fresh message.
- The local-save effect now depends on `messages` rather than `messages.length`. Content and reasoning stream in without changing the array length, so keying on length alone fired the 1s debounce mid-stream and persisted a truncated message — reasoning would not have survived a reload. Re-running per token only resets a timeout. This also made the existing `eslint-disable` on that effect redundant, so it's removed.

### LiteLLM patch

vLLM sends `delta.reasoning`, which LiteLLM drops during chunk conversion because its `Delta` model only recognizes `reasoning_content` ([BerriAI/litellm#20246](https://github.com/BerriAI/litellm/issues/20246)). A self-delimited monkey-patch in `litellm_caller.py` remaps it per chunk. It wraps async streams only (wrapping a sync `completion_stream` would break sync streaming), avoids a per-token `model_dump()`, and no-ops for providers that already populate `reasoning_content`. Removal instructions are in `docs/developer/reasoning-content-streaming.md`.

## Deliberately not carried over from #655

Two drive-by changes in that branch are unrelated to reasoning and are left out, to keep this reviewable and avoid regressing code main has since changed:

- A rework of the `streaming_helpers.py` error path. Main has since added `on_error_message` there; that rework should be argued on its own merits.
- Skipping `OPENAI_API_KEY` fallback for localhost model URLs, and the `test/e2e_tests_live.sh` rewrite.

## Testing

- `bash test/atlas_tests.sh` — 2159 passed, 17 skipped.
- `npx vitest run` — 538 passed across 50 files.
- `npx eslint src/` — 0 errors; `npm run build` succeeds.
- New: `atlas/tests/test_reasoning_content.py` (18 cases) covers marker emission, the reasoning-only stream, non-streaming extraction, that reasoning never leaks into answer text, metadata persistence, and that the reasoning-only case skips the fallback LLM call. `frontend/src/test/useMessages-reasoning.test.js` (10) and 8 new handler cases cover buffering, reconciliation, orphan-flush prevention, and the separate-message-after-tools case.

Not verified: live end-to-end against an actual reasoning model — I have no vLLM/o-series endpoint here. The LiteLLM patch in particular is exercised only by unit tests that simulate the chunk shape, so it's worth a manual check against a real reasoning model before merge.


garland3:
## Screenshots: reasoning streaming running end-to-end

Captured against a real build of this branch — `bash agent_start.sh` (frontend rebuilt, backend restarted), driven through Chromium via Playwright. I verified the served bundle actually contained the new code before trusting anything (`STREAM_REASONING_TOKEN`, `STREAM_REASONING_END`, `_reasoningStreaming` all present in `/assets/index-BHpEXmLr.js`), since `atlas/static` can shadow `frontend/dist`.

**How reasoning was produced:** I have no o-series/vLLM endpoint here, so I pointed a model config at a local mock that streams OpenAI-compatible SSE emitting **`delta.reasoning`** — the vLLM field name, *not* `reasoning_content`. That is deliberate: it exercises the LiteLLM monkey-patch in `litellm_caller.py` on the real path. Without the patch these chunks are dropped by LiteLLM's `Delta` model and no reasoning would appear at all. The content and reasoning text are canned; **every layer above the socket is the real code.**

### 1. Streaming — auto-expands, live cursor

The panel opens by itself while reasoning arrives, header reads `Reasoning...`, blinking `▌` cursor at the tail.

streaming

### 2. Complete — auto-collapses, answer renders below

Once content starts, the header drops to `Reasoning`, the panel collapses, and the markdown answer renders.

complete

### 3. Toggle works

Clicking the disclosure re-expands it. Playwright asserted the accessible state flips: `aria-expanded: false -> true`.

expanded

### 4. Survives a reload

With `saveMode: local`, after a full page reload and reopening the conversation, the reasoning is restored from persisted metadata. Read back from IndexedDB (`atlas-chat-local`), `messages[].metadata.reasoning_content` is present and intact.

after reload

Worth noting what this specific shot proves: the restored text includes the **last** reasoning chunks ("So the honest answer depends on the arrival pattern... Let me write that up concisely."). Those arrive well after the assistant message is created. On `messages?.length` the 1s debounce fired mid-stream and would have persisted a truncated message — so this screenshot is the evidence for the `messages` dependency change described in the PR body.

### Protocol check

Reasoning frames observed on the client WebSocket for the single turn:

```
{"reasoning_token": 8, "reasoning_content": 1}
```

8 incremental tokens (one per mock chunk) followed by exactly one authoritative final block — the shape the reducer's reconcile step expects.

### Caveats, stated plainly

- **Not a real reasoning model.** Provider-specific quirks beyond the `reasoning`/`reasoning_content` naming are still unverified. A pass against actual vLLM or an o-series model before merge is still worth doing.
- **Only plain mode is shown.** The mock advertises `supports_tools: false` (hence the tools warning in the shots), so the tools-mode and agent-mode reasoning paths are covered by unit tests, not by these screenshots. The reasoning-only-then-tool-calls ordering fix in particular is test-covered only.
- **"Error: Conversation not found"** in shot 4 is pre-existing and unrelated — it is the backend session-restore path rejecting a *locally*-saved conversation id that has no server-side row (`atlas/application/chat/service.py`, untouched by this branch).

Images live on the `assets/pr-781-screenshots` branch to keep binaries out of this PR's diff; delete that branch whenever it has served its purpose. The local mock config used for the run was temporary and has been reverted.

PR #786, opened 2026-08-13
feat(hooks): config-driven lifecycle hook system

### LINKED ISSUE #713: Proposal: config-driven hook system — run bash/python scripts at chat lifecycle events (Claude Code–style)

garland3:
## Summary

Implements the config-driven hook system proposed in #713. An operator writes a bash/python script, registers it in `config/hooks.json` against a lifecycle event, and Atlas runs it as a subprocess at that chokepoint. The event payload arrives as **JSON on stdin**; the script's **JSON on stdout** (plus exit code) can allow, modify, block, or escalate the operation. Modeled on Claude Code / OpenCode hooks.

This is the substrate for governed controls (audit, redaction, approval gates, policy enforcement) without scattering one-off checks through the core loop.

### Scope: all 8 events from the proposal

| Event | Anchor | Hook can |
| --- | --- | --- |
| `SessionStart` | `ChatService.create_session` | attach metadata, reject session |
| `UserPromptSubmit` | `ChatOrchestrator.execute` | rewrite/redact prompt, narrow tools/sources, block turn |
| `PreLlmCall` | `LiteLLMCaller` call_plain/call_with_tools (+ streaming leaves) | inspect/redact messages, swap model, block call |
| `PreToolUse` ⭐ | `tool_executor.execute_single_tool` | mutate args (re-injected), deny, force approval |
| `PostToolUse` | `tool_executor.execute_single_tool` | transform/redact result, deny |
| `PermissionRequest` | `tool_executor.execute_single_tool` approval gate | auto-approve / auto-deny / escalate |
| `RagCall` / `RagResponse` | `UnifiedRAGService.query_rag(_batch)` + agentic path | rewrite query, narrow sources (cannot widen), filter chunks, block |
| `SessionEnd` | `ChatService.end_session` | audit/notify |

## Engine (`atlas/hooks/`)
- Pydantic `HookConfig`/`HooksConfig` + `HookEvent` enum with **per-event `on_error` defaults** (security events fail-closed = `deny`; observability/lifecycle fail-open = `allow`).
- `HookManager`: **zero-overhead when no `hooks.json`** (no subprocess; the test suite asserts this invariant per event), `asyncio.create_subprocess_exec` (**never `shell=True`**), per-hook timeout (kills on expiry), minimal env allow-list (hooks do **not** inherit provider API keys), bounded 1 MB stdout/stderr.
- Exit-code contract: `0` = continue (apply stdout JSON if present), `2` = deny (stderr = reason), other = error → `on_error`.
- Composition: sequential in config order, **first `deny` short-circuits**, **`require_approval` is sticky** (cannot be downgraded by a later hook).
- **Security model**: trusted envelope fields (`session_id`, `user_email`, `compliance_level`) are stamped server-side and re-asserted after every `modify`; for tool args, `_atlas_user` and other security-critical parameters are re-injected (same rule already used for user-edited args). A hook can **tighten but never widen** a boundary.

## Config integration
- `ConfigManager.hooks_config` loads `config/hooks.json` via the existing `_search_paths` two-layer lookup, cached and cleared by `reload_configs()`.
- New `AppSettings.hooks_config_file` (default `hooks.json`, env `HOOKS_CONFIG_FILE`).

## Tests
- **64 new tests** (`test_hooks_engine.py` + `test_hooks_integration.py`): config models, subprocess contract, composition rules, security (trusted-field re-injection, env allow-list, no-shell), failure modes, matcher filtering, per-event wiring, plus a **no-config no-subprocess invariant** for every wired event.
- Updated 4 mock signatures in `test_atlas_rag_agent_tools.py` for the new `query_rag(_batch)` `_skip_hooks` kwarg.
- Full suite: **2242 passed, 17 skipped**. `ruff` clean on all changed files.

## Docs & examples
- `docs/hooks.md`: full reference.
- `atlas/config/hooks-example/`: runnable example hooks + illustrative `hooks.json` (not auto-loaded).

## Decisions on the proposal's open questions
1. Payload schemas — one per event, full replacement of the mutable part.
2. `modify` granularity — full replacement; trusted fields stripped and re-asserted.
3. Per-group enablement — single global list for v1; operators filter on envelope `user_email`/`compliance_level`.
4. Observability tier — PostToolUse/RagResponse/SessionEnd default `on_error: allow`; non-blocking variant is a follow-up.
5. Outbound UI events — not in v1 (reserved stdout key is a follow-up).
6. vs MCP tools — hooks govern invocations at lifecycle boundaries; MCP tools are how the model invokes capabilities.

Out of scope for v1: sandboxing untrusted hooks, non-blocking observability variant, outbound UI-event emission, per-group enablement.

Closes #713.

### DIFF
--- a/atlas/application/chat/agent/agentic_loop.py
+++ b/atlas/application/chat/agent/agentic_loop.py
@@ -263,7 +263,12 @@ class AgenticLoop(AgentLoopProtocol):
                     "session_id": context.session_id,
                     "user_email": context.user_email,
                     "files": context.files,
-                    "selected_data_sources": data_sources or [],
+                    # None and [] mean different things downstream: None is "the
+                    # user made no selection" (atlas_rag_query falls back to every
+                    # authorized source), [] is "explicitly no sources" (query
+                    # nothing). Collapsing None to [] would break RAG for every
+                    # agent turn where the user picked no sources.
+                    "selected_data_sources": data_sources,
                     # Trusted compliance level so RAG tools enforce the boundary;
                     # the model cannot set or change this.
                     "compliance_level": context.compliance_level,
--- a/atlas/application/chat/modes/tools.py
+++ b/atlas/application/chat/modes/tools.py
@@ -126,7 +126,11 @@ class ToolsModeRunner:
         # they do in agent mode, instead of falling back to all authorized
         # sources. build_session_context() only reflects session state, not this
         # per-request selection.
-        if selected_data_sources:
+        # ``is not None``, not truthiness: an explicitly empty list means "no
+        # sources" (e.g. a UserPromptSubmit hook narrowed the turn to nothing).
+        # Dropping it would leave mcp_execution reading None and widening the
+        # query back to every source the user is authorized for.
+        if selected_data_sources is not None:
             session_context["selected_data_sources"] = selected_data_sources
 
         # Ensure update_callback is never None (critical for elicitation)
@@ -289,8 +293,9 @@ class ToolsModeRunner:
 
         session_context = build_session_context(session)
         # See note above: propagate the per-request RAG selection so atlas_rag
-        # tools behave consistently with agent mode in the streaming path too.
-        if selected_data_sources:
+        # tools behave consistently with agent mode in the streaming path too,
+        # preserving an explicit empty selection.
+        if selected_data_sources is not None:
             session_context["selected_data_sources"] = selected_data_sources
         effective_callback = update_callback
         if effective_callback is None:
--- a/atlas/application/chat/orchestrator.py
+++ b/atlas/application/chat/orchestrator.py
@@ -290,6 +290,10 @@ class ChatOrchestrator:
                 reason = outcome.reason or "Prompt blocked by policy hook."
                 block_msg = Message(role=MessageRole.ASSISTANT, content=reason, metadata={"blocked_by_hook": True})
                 session.history.add_message(block_msg)
+                # Publish the reason before completing the turn: a streaming
+                # client renders what it receives, so completing without a
+                # message would end the turn silently and look like a hang.
+                await self.event_publisher.publish_chat_response(reason)
                 await self.event_publisher.publish_response_complete()
                 return event_notifier.create_chat_response(reason)
             if outcome.verdict == "modify":
@@ -298,12 +302,20 @@ class ChatOrchestrator:
                     content = new_prompt
                     # Reflect the (possibly redacted) prompt in the stored user message
                     user_message.content = content
+                # Tools/sources may only be *narrowed*: intersect the hook's list
+                # with what the user actually selected so a hook cannot grant
+                # access to a tool or source the user never chose. An explicitly
+                # empty list is preserved (it means "none"), not treated as
+                # "unset" -- collapsing [] to None would widen the turn back to
+                # the caller's full selection.
                 new_tools = outcome.payload.get("selected_tools")
                 if isinstance(new_tools, list):
-                    selected_tools = new_tools or None
+                    allowed_tools = set(selected_tools or [])
+                    selected_tools = [t for t in new_tools if t in allowed_tools]
                 new_sources = outcome.payload.get("selected_data_sources")
                 if isinstance(new_sources, list):
-                    selected_data_sources = new_sources or None
+                    allowed_sources = set(selected_data_sources or [])
+                    selected_data_sources = [s for s in new_sources if s in allowed_sources]
                 new_agent = outcome.payload.get("agent_mode")
                 if isinstance(new_agent, bool):
                     agent_mode = new_agent
--- a/atlas/application/chat/orchestrator.py
+++ b/atlas/application/chat/orchestrator.py
@@ -7,6 +7,7 @@ from uuid import UUID
 from atlas.core.model_access import ModelAccessDecision, check_model_access
 from atlas.domain.errors import AuthorizationError, SessionNotFoundError
 from atlas.domain.messages.models import Message, MessageRole
+from atlas.hooks import HookEvent, get_hook_manager
 from atlas.interfaces.events import EventPublisher
 from atlas.interfaces.llm import LLMProtocol
 from atlas.interfaces.sessions import SessionRepository
@@ -20,7 +21,7 @@ from .modes.tools import ToolsModeRunner
 from .policies.tool_authorization import ToolAuthorizationService
 from .preprocessors.message_builder import MessageBuilder
 from .preprocessors.prompt_override_service import PromptOverrideService
-from .utilities import file_processor
+from .utilities import event_notifier, file_processor
 
 logger = logging.getLogger(__name__)
 
@@ -262,6 +263,50 @@ class ChatOrchestrator:
         )
         session.history.add_message(user_message)
         session.update_timestamp()
+
+        # UserPromptSubmit hook (GH #713): fires after the user message is added,
+        # before file ingestion and mode dispatch. A hook may rewrite/redact the
+        # prompt text, narrow selected_tools/data_sources, disable agent_mode,
+        # or block the turn with a message. Opt-in; zero overhead without
+        # config/hooks.json.
+        hook_mgr = get_hook_manager()
+        if hook_mgr is not None and hook_mgr.has_hooks(HookEvent.USER_PROMPT_SUBMIT):
+            payload = {
+                "prompt": content,
+                "selected_tools": list(selected_tools) if selected_tools else [],
+                "selected_data_sources": list(selected_data_sources) if selected_data_sources else [],
+                "agent_mode": bool(agent_mode),
+            }
+            outcome = await hook_mgr.run_event(
+                HookEvent.USER_PROMPT_SUBMIT,
+                payload,
+                session_context={
+                    "session_id": str(session_id),
+                    "user_email": user_email,
+                    "compliance_level": session.context.get("compliance_level"),
+                },
+            )
+            if outcome.verdict == "deny":
+                reason = outcome.reason or "Prompt blocked by policy hook."
+                block_msg = Message(role=MessageRole.ASSISTANT, content=reason, metadata={"blocked_by_hook": True})
+                session.history.add_message(block_msg)
+                await self.event_publisher.publish_response_complete()
+                return event_notifier.create_chat_response(reason)
+            if outcome.verdict == "modify":
+                new_prompt = outcome.payload.get("prompt")
+                if isinstance(new_prompt, str) and new_prompt:
+                    content = new_prompt
+                    # Reflect the (possibly redacted) prompt in the stored user message
+                    user_message.content = content
+                new_tools = outcome.payload.get("selected_tools")
+                if isinstance(new_tools, list):
+                    selected_tools = new_tools or None
+                new_sources = outcome.payload.get("selected_data_sources")
+                if isinstance(new_sources, list):
+                    selected_data_sources = new_sources or None
+                new_agent = outcome.payload.get("agent_mode")
+                if isinstance(new_agent, bool):
+                    agent_mode = new_agent
 
         # Handle file ingestion
         update_callback = kwargs.get("update_callback")
--- a/atlas/application/chat/orchestrator.py
+++ b/atlas/application/chat/orchestrator.py
@@ -296,7 +296,7 @@ class ChatOrchestrator:
                 await self.event_publisher.publish_chat_response(reason)
                 await self.event_publisher.publish_response_complete()
                 return event_notifier.create_chat_response(reason)
-            if outcome.verdict == "modify":
+            if outcome.modified:
                 new_prompt = outcome.payload.get("prompt")
                 if isinstance(new_prompt, str) and new_prompt:
                     content = new_prompt
--- a/atlas/application/chat/service.py
+++ b/atlas/application/chat/service.py
@@ -20,6 +20,7 @@ from atlas.core.user_identity import normalize_user_email
 from atlas.domain.errors import AuthorizationError, DomainError
 from atlas.domain.messages.models import Message, MessageRole, MessageType, ToolResult
 from atlas.domain.sessions.models import Session
+from atlas.hooks import HookEvent, get_hook_manager
 from atlas.interfaces.events import EventPublisher
 from atlas.interfaces.llm import LLMProtocol
 from atlas.interfaces.sessions import SessionRepository
@@ -242,6 +243,21 @@ class ChatService:
         await self.session_repository.create(session)
 
         logger.info(f"Created session {sanitize_for_logging(str(session_id))} for user {sanitize_for_logging(user_email)}")
+
+        # SessionStart hook (GH #713): opt-in, zero overhead when no hooks.json.
+        # Fires on every session creation including restore; a hook can reject
+        # the session (deny) or attach metadata to session.context (modify).
+        mgr = get_hook_manager()
+        if mgr is not None and mgr.has_hooks(HookEvent.SESSION_START):
+            outcome = await mgr.run_event(
+                HookEvent.SESSION_START,
+                {"session_id": str(session_id)},
+                session_context={"session_id": str(session_id), "user_email": user_email},
+            )
+            if outcome.verdict == "deny":
+                raise DomainError(f"Session creation blocked by hook: {outcome.reason}")
+            if outcome.verdict == "modify" and isinstance(outcome.payload, dict):
+                session.context.update(outcome.payload)
         return session
 
     async def handle_chat_message(
@@ -1068,3 +1084,21 @@ class ChatService:
         self._incognito_save_floor.pop(session_id, None)
         self._save_floor_locked.discard(session_id)
         logger.info(f"Ended session {sanitize_for_logging(str(session_id))}")
+
+        # SessionEnd hook (GH #713): observability/lifecycle. Cannot block (a
+        # session already ended); deny is treated as continue with a log so a
+        # misconfigured audit hook cannot wedge the teardown path.
+        mgr = get_hook_manager()
+        if mgr is not None and mgr.has_hooks(HookEvent.SESSION_END):
+            try:
+                await mgr.run_event(
+                    HookEvent.SESSION_END,
+                    {"session_id": str(session_id), "user_email": session.user_email},
+                    session_context={
+                        "session_id": str(session_id),
+                        "user_email": session.user_email,
+                        "compliance_level": session.context.get("compliance_level"),
+                    },
+                )
+            except Exception as e:
+                logger.warning("SessionEnd hook failed for %s: %s", session_id, e)
--- a/atlas/application/chat/service.py
+++ b/atlas/application/chat/service.py
@@ -263,7 +263,7 @@ class ChatService:
                     sanitize_for_logging(user_email),
                 )
                 raise DomainError(f"Session creation blocked by hook: {outcome.reason}")
-            if outcome.verdict == "modify" and isinstance(outcome.payload, dict):
+            if outcome.modified and isinstance(outcome.payload, dict):
                 session.context.update(outcome.payload)
 
         await self.session_repository.create(session)
--- a/atlas/application/chat/service.py
+++ b/atlas/application/chat/service.py
@@ -240,13 +240,15 @@ class ChatService:
     ) -> Session:
         """Create a new chat session."""
         session = Session(id=session_id, user_email=user_email)
-        await self.session_repository.create(session)
-
-        logger.info(f"Created session {sanitize_for_logging(str(session_id))} for user {sanitize_for_logging(user_email)}")
 
         # SessionStart hook (GH #713): opt-in, zero overhead when no hooks.json.
         # Fires on every session creation including restore; a hook can reject
         # the session (deny) or attach metadata to session.context (modify).
+        #
+        # This runs *before* session_repository.create(). Persisting first would
+        # make deny a one-message speed bump: the row would already exist, so the
+        # caller's next message would find it via session_repository.get() and be
+        # answered normally. Nothing is written until the hook allows the session.
         mgr = get_hook_manager()
         if mgr is not None and mgr.has_hooks(HookEvent.SESSION_START):
             outcome = await mgr.run_event(
@@ -255,9 +257,18 @@ class ChatService:
                 session_context={"session_id": str(session_id), "user_email": user_email},
             )
             if outcome.verdict == "deny":
+                logger.warning(
+                    "Session %s for user %s blocked by SessionStart hook",
+                    sanitize_for_logging(str(session_id)),
+                    sanitize_for_logging(user_email),
+                )
                 raise DomainError(f"Session creation blocked by hook: {outcome.reason}")
             if outcome.verdict == "modify" and isinstance(outcome.payload, dict):
                 session.context.update(outcome.payload)
+
+        await self.session_repository.create(session)
+
+        logger.info(f"Created session {sanitize_for_logging(str(session_id))} for user {sanitize_for_logging(user_email)}")
         return session
 
     async def handle_chat_message(
--- a/atlas/application/chat/utilities/tool_executor.py
+++ b/atlas/application/chat/utilities/tool_executor.py
@@ -12,6 +12,7 @@ from typing import Any, Awaitable, Callable, Dict, List, Optional
 
 from atlas.core.capabilities import create_download_url
 from atlas.domain.messages.models import ToolCall, ToolResult
+from atlas.hooks import HookEvent, get_hook_manager
 from atlas.interfaces.llm import LLMResponse
 from atlas.modules.mcp_tools.token_storage import AuthenticationRequiredException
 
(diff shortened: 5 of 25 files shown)


garland3:
## CI status note

The actual code/test gates pass in CI:
- **build-and-test** (both pytest legs in Docker): `2231 passed, 28 skipped` (debug mode) and `2224 passed, 35 skipped` (production mode) — plus E2E: `All E2E tests passed!`
- **build-artifacts** (incl. `check-docs.sh`), **SAST**, **CodeQL**, **Trivy**, **grype**, **Dependency Vulnerability Scan**, **Container Security Scan**: all ✅

The three red checks are infrastructural, not from this diff:
1. **Secrets Scanning** — a pre-existing flaky TruffleHog false positive ("verified Lob result"). The same check failed on `main` in the scheduled run ~1h before this PR ([run 31665353379](https://github.com/sandialabs/atlas-ui-3/actions/runs/31665353379)). There is no `test_[a-f0-9]{35}` / `live_[a-f0-9]{35}` string in this PR.
2. **build-and-test** job shows red only on the final `Push production Docker image` step: `denied: installation not allowed to Write organization package` — a fork→org GHCR permission (this fork can't push to `ghcr.io/sandialabs/...`); both test steps completed successfully before it.
3. **Security Summary** — cascades from the secrets-scan result.

Happy to address any review feedback.

garland3:
## Summary
Adds a lifecycle hook subsystem (`atlas/hooks/`) that runs operator-configured subprocesses at nine chokepoints — session start/end, prompt submit, tool permission/pre/post, pre-LLM call, and RAG call/response — with a JSON stdin envelope, exit-code and stdout decision contract, per-event fail-open/fail-closed defaults, operator docs, and example hooks.

## Verdict
Request changes.

## Required changes
1. Run the SessionStart hook before `await self.session_repository.create(session)` at `atlas/application/chat/service.py:243` (or delete the session on deny before raising) — the session is persisted first, so a denied session is answered normally on the caller's second message once `session_repository.get()` starts succeeding, making `deny` a one-message speed bump rather than a control.
2. Preserve an explicitly empty hook-supplied source list at `atlas/application/chat/orchestrator.py:306` instead of `new_sources or None`, and make `_execute_atlas_rag_tool` (`atlas/modules/mcp_tools/mcp_execution.py:730`) fall back to all sources only when the key is absent — today a hook narrowing to `[]` widens the query to every authorized source.
3. Enforce the 1 MB output cap at read time in `atlas/hooks/manager.py:311` — replace `proc.communicate()` with bounded reads that kill the child on overflow (routed through `on_error`); `_truncate` currently runs after the entire stream is already buffered, so a hook emitting hundreds of MB can OOM the server despite the documented bound.
4. Fix the example command paths in `atlas/config/hooks-example/hooks.json:2` (and align `docs/admin/hooks.md`'s `config/hooks/` quick start) — `${ATLAS_CONFIG_DIR}/hooks-example/<script>` does not exist after the documented copy step, and because `PreToolUse` defaults to `on_error: deny`, following the instructions blocks every `filesystem__*` tool call.
5. Publish the block reason before completing the turn on the UserPromptSubmit deny path at `atlas/application/chat/orchestrator.py:293` — only `publish_response_complete()` is called, so a streaming client sees the turn end with no message.
6. Correct the `matcher` contract at `atlas/hooks/models.py:118` and the docs table, and warn at config load when a matcher is set on `SessionStart`/`SessionEnd`/`UserPromptSubmit` — the docstring says the matcher is ignored for those events, but `matches()` returns `False`, so the hook silently never fires.
7. Build the query coroutines after the RagCall hook settles the source set in `atlas/modules/mcp_tools/mcp_execution.py:806` — the deny and narrow paths discard the eagerly built `coros`, leaking un-awaited coroutines (`RuntimeWarning`) on every retrieval, and force a verbatim copy of the HTTP/MCP grouping block.

## Evidence
Evidence: 8/10. Well demonstrated for a backend subsystem — 64 tests spawning real subprocesses, negative and failure-mode coverage, a per-event assertion that nothing spawns without `hooks.json`, and a complete operator reference. The most useful missing artifact is any evidence of a hook running in an assembled deployment (e.g. the `tool-audit.jsonl` lines produced by the shipped example against a live instance), which would also have caught the broken example paths above.

Coverage gaps worth closing alongside the fixes: the `_atlas_user` re-injection test is a provable no-op (the fixture schema does not declare the field), nothing pins the rule that `PreToolUse` cannot auto-approve while `PermissionRequest` can, and the ~100 new lines of RagCall/RagResponse wiring in the agentic path are untested. `test_permission_request_require_approval_forces_gate` also rebinds `te.get_approval_manager` without teardown — use `monkeypatch`.

## Notes
- `require_approval` from a hook also sets `admin_required = True` (`atlas/application/chat/utilities/tool_executor.py:496`); the docs describe only "force the approval gate", so an "ask the user" hook can produce a request the user cannot satisfy.
- `docs/admin/hooks.md:112` advertises chunk filtering on `RagResponse`, but only `payload["content"]` is read back — returned metadata is silently discarded.
- A malformed `hooks.json` is swallowed into an empty config (`atlas/modules/config/config_loader.py:422`), disabling every fail-closed hook with only a log line; hooks are not covered by `validate_config`.

---

garland3:
## Summary
Adds a subprocess-based hook system (`atlas/hooks/`) that fires operator-configured scripts at eight lifecycle points — session start/end, prompt submit, permission request, pre/post tool use, pre-LLM call, and RAG call/response — with a JSON stdin envelope, exit-code verdicts, per-event fail-closed defaults, an operator reference in `docs/admin/hooks.md`, example hooks, and 64 tests.

## Verdict
Request changes.

## Required changes
1. Fire the `SessionStart` hook before `await self.session_repository.create(session)` at `atlas/application/chat/service.py:243` (or delete the session in the `deny` branch before raising) — today the row is persisted first, so the second message on the same `session_id` takes the `session_repository.get()` path, never re-runs the hook, and is answered normally. Add a regression test that a second `handle_chat_message` after a denied create is still refused.
2. Stop collapsing an explicitly empty list to `None` at `atlas/application/chat/orchestrator.py:306` (`selected_data_sources = new_sources or None`) and make the fallback at `atlas/modules/mcp_tools/mcp_execution.py:730` trigger on *absence* rather than emptiness — a hook narrowing to zero sources currently widens retrieval to the user's entire authorized set, inverting the stated tighten-never-widen rule.
3. Replace `proc.communicate(stdin_bytes)` at `atlas/hooks/manager.py:311` with bounded concurrent reads of stdout/stderr, killing the child and returning `_error_decision(...)` on overflow — `_truncate` runs only after the whole stream is already buffered, so the documented 1 MB cap does not bound server memory.
4. Align the example hook paths in `atlas/config/hooks-example/hooks.json:5`, its `_comment`, and the quick start at `docs/admin/hooks.md:32` on one layout — the three disagree, and because `PreToolUse` defaults to `on_error: deny`, following the documented copy step blocks every `filesystem__*` and `http_*` tool call with "executable not found".
5. Publish the block reason before completing the turn on the `UserPromptSubmit` deny path at `atlas/application/chat/orchestrator.py:293` — only `publish_response_complete()` is emitted, so a streaming client sees the turn end with an empty assistant bubble and no explanation.
6. Re-run `check_model_access` and re-derive the model compliance level after a hook-supplied model swap in `_apply_pre_llm_modify` (`atlas/modules/llm/litellm_caller.py:206`), or restrict swaps to an operator allow-list — authorization is validated upstream against the originally selected model only. Preserve `max_tokens` on the `call_with_tools` re-resolution path while there.
7. Make a present-but-unparseable `hooks.json` loud at `atlas/modules/config/config_loader.py:422` — distinguish it from "file absent", surface it through `validate_config`, and fix the `atlas/hooks/models.py:120` comment claiming this is already done. A typo currently disables every fail-closed hook with one log line while `config validate` reports healthy.

## Evidence
Evidence: 8/10 — strong for a backend change of this size: 64 tests spawning real subprocesses exercise the stdin envelope, exit-code contract, timeout kill, and env allow-list end to end. The main gap is that nothing loads a real `hooks.json` off disk (every test assigns `_hooks_config` directly), so the config-loading path and the shipped example hooks are never demonstrated running; the ~100 lines of agentic `RagCall`/`RagResponse` wiring in `mcp_execution.py` and the two streaming `PreLlmCall` leaves are likewise untested.

## Notes
- `matcher` set on `SessionStart`/`SessionEnd`/`UserPromptSubmit` makes the hook never fire, contrary to its docstring — warn at config load (`atlas/hooks/models.py:118`).
- `coros` is built before the `RagCall` hook runs at `atlas/modules/mcp_tools/mcp_execution.py:807`, so deny/narrow drop un-awaited coroutines and duplicate the grouping block.
- `te.get_approval_manager` is rebound without teardown at `atlas/tests/test_hooks_integration.py:305`, leaking an auto-approving stub into later tests.

---

garland3:
## Summary
Adds a subprocess-based policy hook system (`atlas/hooks/`) wired into nine chokepoints — session lifecycle, prompt submit, tool permission/pre/post, LLM call, and RAG call/response — with a `hooks.json` config, operator docs, and runnable example hooks. Also introduces a `None` vs `[]` distinction for `selected_data_sources` so a hook can narrow a turn to zero sources.

## Verdict
Request changes.

## Required changes
1. Fix the agent-mode RAG regression: `atlas/application/chat/agent/agentic_loop.py:266` sends `"selected_data_sources": data_sources or []`, so every agent turn where the user picked no sources now hits the new "explicit selection of nothing" branch at `atlas/modules/mcp_tools/mcp_execution.py:735` and `atlas_rag_query` returns `success=False` / "No authorized RAG data sources are available" instead of falling back to all authorized sources. Pass `None` (or carry an explicit `data_sources_explicit` flag) when the user selected nothing, and add tests pinning both directions.
2. Change the guards at `atlas/application/chat/modes/tools.py:129` and `:293` to `if selected_data_sources is not None:` — today an explicit `[]` from a `UserPromptSubmit` hook is dropped from `session_context`, so `mcp_execution` reads `None` and queries every source the user is authorized for: a narrowing hook produces the widest possible query. Add a tools-mode regression test asserting zero sources are queried.
3. Make a present-but-unparseable `hooks.json` a detected failure at `atlas/modules/config/config_loader.py:406` — `_load_file_with_error_handling` swallows `json.JSONDecodeError` and returns `None`, so a syntax typo takes the "no file" branch, leaves `_hooks_config_load_failed` False, and `validate_config` reports `hooks_config: true` while every fail-closed control is off. Check whether any candidate path exists, set the flag, log at ERROR, and test that `validate_config()["hooks_config"] is False`.
4. Apply a hook's `modify` payload independently of the verdict rank in `atlas/hooks/manager.py:216` — when any hook raises the outcome to `require_approval`, call sites that compare `verdict == "modify"` (`tool_executor.py:487`/`:522`, `orchestrator.py:296`, `litellm_caller.py:_fire_pre_llm_hook`, `unified_rag_service.py:528`/`:745`, `mcp_execution.py:837`) discard the merged payload, so redacted messages and rewritten tool args are silently dropped. Add a `modified` flag on `HookOutcome` and gate on that.
5. Reconcile the example hook paths with the quick start — `atlas/config/hooks-example/hooks.json:2` references `${ATLAS_CONFIG_DIR}/hooks/...` while `docs/admin/hooks.md:21` tells operators to copy scripts to `config/hooks/`; because `PreToolUse` defaults to `on_error: deny`, following the documented steps blocks every filesystem and HTTP tool.
6. Redact the shipped audit example at `atlas/config/hooks-example/audit_tool.sh` — `PostToolUse` payloads carry `dict(filtered_args)` (`tool_executor.py:686`), which by then contains tokenized download URLs with HMAC capability tokens, and the example appends the raw envelope to a plaintext `tool-audit.jsonl`. Log a projection (tool name, arg keys, result size) and state in `docs/admin/hooks.md` that payloads are secret-bearing.
7. Log the failure inside `has_hooks` at `atlas/hooks/manager.py:120` instead of `except Exception: return False` silently — any transient config read/validation error makes all four fail-closed events behave as if no hooks existed, with no diagnostic. Same for the `get_hook_manager()` init-failure path.

## Evidence
Evidence: 8/10. Strong for a backend change of this size — 64 tests spawning real subprocesses, per-event no-config invariants, full-suite and CI numbers, and a complete operator reference. The one notable gap is that the production config-loading path (`ConfigManager.hooks_config` reading a real `hooks.json` off disk) is never exercised: every test assigns `_hooks_config` directly, which is why item 3 above went unnoticed.

## Notes
- `require_approval` is silently dropped on the seven events with no approval gate — log a warning and mark approval-capable events in the docs table (`atlas/application/chat/orchestrator.py:293`).
- RAG hook matchers are tested against a comma-joined source list, so an anchored per-source pattern like `^atlas_rag:classified$` never fires on batch/agentic paths (`atlas/domain/unified_rag_service.py:175`).
- Per-event payload keys are documented only for `PreToolUse`; a misspelled key in a `modify` is dropped without error (`docs/admin/hooks.md:91`).

---

garland3:
Pushed `342544f`. Verified each item against current `HEAD` and fixed the six that reproduce; skipping one that no longer applies.

**Fixed**

1. **Agent-mode RAG regression** — `agentic_loop.py` passed `data_sources or []`, so any agent turn with no source selection hit the "explicitly no sources" branch and `atlas_rag_query` failed instead of falling back to all authorized sources. Passes the value through unchanged now.
2. **Narrowing hook widened the query** — tools mode gated `session_context["selected_data_sources"]` on truthiness, dropping an explicit `[]`; `mcp_execution` then read `None` and queried everything the user is authorized for. Now `is not None` in both the buffered and streaming paths.
3. **Malformed `hooks.json` was silent** — an existing-but-unparseable file took the "no file" branch, so `_hooks_config_load_failed` stayed `False` and `validate_config` reported healthy with every fail-closed control off. Now distinguished by checking whether a candidate path exists, logged at ERROR.
4. **`modify` dropped on escalation** — verdicts are ranked, so redact-then-escalate reported `require_approval` while call sites applied the payload only on `verdict == "modify"`. Added `HookOutcome.modified` and gated all twelve call sites on it.
5. **Audit example leaked secrets** — `PostToolUse` payloads carry tokenized download URLs and retrieved document text; the example appended the raw envelope to a plaintext log. Now logs a projection (tool name, arg *keys*, sizes), with a note in the docs that payloads are secret-bearing.
6. **`has_hooks` swallowed config errors** — now logged, so a broken config isn't indistinguishable from "no hooks configured".

**Not changed**

- **Example hook paths** — already consistent at `HEAD`: `hooks.json`, its `_comment`, and the quick start all resolve to `${ATLAS_CONFIG_DIR}/hooks/`, which is the same directory the docs call `config/hooks/`.
- **`get_approval_manager` teardown** — the rebind at `test_hooks_integration.py:414` is already restored in a `finally`.

10 regression tests added covering both directions of the `None` vs `[]` contract, modify-survives-escalation, and malformed-config detection. Full suite: **2263 passed, 17 skipped**; `ruff` clean on all changed files.

Note on the two red checks: both are fork-permission artifacts, not code. `build-and-test` passes both pytest legs and fails only on the final `Push production Docker image` step (`denied: installation not allowed to Write organization package`), and `Security Summary` fails posting its PR comment (`Resource not accessible by integration`). Neither step can succeed from a fork. Happy to mirror this to an origin branch for a fully green same-repo run if that's preferred before merge.

garland3:
## Summary
Adds an operator-configurable hooks subsystem: nine lifecycle events (`SessionStart`, `UserPromptSubmit`, `PermissionRequest`, `PreToolUse`, `PostToolUse`, `PreLlmCall`, `RagCall`, `RagResponse`, `SessionEnd`) dispatched to subprocess hooks via a stdin JSON envelope with a deny/modify/require_approval contract, wired through the orchestrator, tool executor, LLM caller, and RAG paths, plus `hooks.json` config, operator docs, example scripts, and 74 tests.

## Verdict
Request changes.

## Required changes
1. Fix the `None` vs `[]` data-source contract at the entry point, not downstream — `atlas/modules/mcp_tools/mcp_execution.py:735` treats `[]` as "explicitly zero sources", but `frontend/src/contexts/ChatContext.jsx:371` always sends `[]` when the RAG toggle is off, so `atlas_rag_query` now returns "No authorized RAG data sources are available" in the default UI state where it previously queried all authorized sources. Normalize an absent key or empty list from the client to `None` in `atlas/main.py:683`, or carry an explicit `data_sources_explicit` flag through `session_context`.
2. Preserve the three-state selection across the `UserPromptSubmit` envelope at `atlas/application/chat/orchestrator.py:277` — send JSON `null` when the caller made no selection, and on write-back treat a missing/`null` value as "unchanged" rather than a narrowing to zero. As written, the payload-echo idiom taught in `docs/admin/hooks.md:126` and shipped in `atlas/config/hooks-example/redact_pii.py` disables RAG and tools on every turn.
3. Read `payload["metadata"]` back on the `RagResponse` modify path at `atlas/domain/unified_rag_service.py:210` and rebuild `response.metadata` (narrow-only: drop refs not in the original set) — metadata is sent to the hook but discarded, so a redaction hook that strips text from `content` still ships the unredacted snippets, which `atlas/modules/llm/litellm_caller.py:1502` renders into the References block. If honoring edits is out of scope, stop sending `metadata` and correct the "filter chunks" claim in the docs.
4. Refresh `original_display_args` after each hook-driven recomputation of `display_args` in `atlas/application/chat/utilities/tool_executor.py:444` — a hook `modify` makes a plain user Approve look like a manual edit, so `filtered_args` is replaced with the UI-sanitized dict and the tool runs with a bare basename instead of the tokenized download URL.
5. Apply source narrowing on the single-source RAG path at `atlas/domain/unified_rag_service.py:535` (return an empty `RAGResponse` when the hook drops the source, as `query_rag_batch` does at :746) — today the same policy hook enforces "never retrieve from this corpus" only when the query happens to be batched.
6. Constrain `result_success` to the tightening direction at `atlas/application/chat/utilities/tool_executor.py:711` and clear `result.error` when success changes — a `PostToolUse` hook can currently relabel a failed or denied call as `success=True`, leaving `success=True` alongside a populated `error` and telling the model the tool worked.
7. Correct the exit-code table at `docs/admin/hooks.md:117` — it says non-JSON stdout on exit 0 is ignored, but `_parse_decision` routes it through `on_error`, so one `echo` or `set -x` line in a bash hook denies every matching tool call. State that stdout must be empty or a single JSON object and that diagnostics go to stderr.

## Evidence
Evidence: 8/10. Well-demonstrated for a backend subsystem of this size — 74 tests spawn real subprocesses and cover the envelope, exit codes, timeout kill, output cap, and env allow-list end to end. The most valuable missing artifact is coverage of the newly wired agentic `RagCall`/`RagResponse` paths in `mcp_execution.py` and the streaming `PreLlmCall` leaves, none of which any test touches.

## Notes
- Refuse a hook model swap on `UNKNOWN` as well as `DENIED`, and re-check compliance level — `atlas/modules/llm/litellm_caller.py:264`.
- RAG hook matchers are evaluated against a comma-joined source string, so anchored per-source patterns stop matching on batch queries — `atlas/domain/unified_rag_service.py:175`.
- Rename `_skip_hooks` to a public keyword (`fire_hooks`) or move it into ambient state; an underscore-prefixed flag that an out-of-module caller must set for correctness will be missed — `atlas/domain/unified_rag_service.py:488`.

---

garland3:

garland3:
Superseded by #803, which carries the same commits plus a merge of `main`, opened from a branch on this repository so CI can push images and comment on the PR. CI here could never go green: as a fork PR the Docker push and PR-comment steps are permission-denied, though all backend and E2E tests passed.

Issue #791, opened 2026-08-13
RAG API v2: tool-oriented query interface (explicit query + raw/synthesized modes)


garland3:
## Problem

ATLAS-UI's RAG integration (v1) ships the **entire conversation** to ATLAS-RAG on every query. The HTTP endpoint `POST /api/v1/rag/completions` takes the full `messages` array, but the server only uses the last user message as the query (the mock does `for m in reversed(request.messages): if m.role == "user": ...`). The rest is wasted payload and an OPSEC/privacy exposure (full conversation history sent to an external service that only needs the query).

Three concrete pain points:

1. **Message-dumping, not query-targeted.** No explicit `query` field on the HTTP path; the query is *derived* from the last user message. If that message is "thanks", that becomes the RAG query. The MCP RAG tools (`rag_get_raw_results`, `rag_get_synthesized_results`) already take an explicit `query` string — so the two RAG subsystems are asymmetric (same product, two contracts).
2. **Retrieval and generation are coupled.** `/rag/completions` does retrieval + synthesis in one call and returns a completion. There is no HTTP equivalent of `rag_get_raw_results`, so ATLAS-UI cannot fetch raw retrieved chunks as context for its own LLM. The only retrieval shape on HTTP is `is_completion=True`, which **short-circuits the LLM call entirely**.
3. **RAG cannot act as a first-class tool.** Because completion short-circuits the LLM, RAG can't return evidence the LLM reasons over alongside other tools. This is why `call_with_rag` / `call_with_rag_and_tools` / `_build_rag_completion_response` and the `atlas_rag_query` pseudo-tool exist as special-case bridges rather than a normal tool call.

Full design proposal with request/response examples and a phased migration plan: [`docs/planning/rag-api-v2-tool-interface.md`](docs/planning/rag-api-v2-tool-interface.md) (also on a [private gist](https://gist.github.com/garland3/d4ca04e5c11081f0594d5dd74dcf9e1c)).

## Proposed interface

New endpoint, versioned alongside v1 (v1 stays alive during migration):

```
POST /api/v2/rag/query?as_user={user}
Authorization: Bearer {token}
```

| Field | Type | Required | Default | Description |
|-------|------|----------|---------|-------------|
| `query` | string | yes | - | The specific question to ask. Non-empty. |
| `corpora` | string \| string[] | yes | - | Corpus id(s) to search. |
[...]

PR #793, opened 2026-08-13
feat(rag): v2 tool-oriented query interface (Phase 1) (#791)

garland3:
Closes part of #791 (Phase 1; keeps the issue open for Phases 2-4).

## Why

v1 posts the **entire conversation** to `POST /api/v1/rag/completions` on every turn, even though the server only reads the last user message. That is wasted payload and conversation history leaving the process for no reason, and it makes the query implicit — if the last message is "thanks", that is the RAG query. The MCP RAG tools have always taken an explicit `query`, so the two RAG subsystems disagreed on the contract.

## What this does

New `POST /api/v2/rag/query`: an explicit `query` string, and a `mode` that says whether the caller wants evidence or an answer.

- **`AtlasRAGClient.query_v2()`** posts `{query, corpora, mode, top_k, filters, synthesis_params}` and parses both response shapes. `mode: "raw"` renders retrieved sections as an evidence block carrying `[N]` document markers — the same citation form the UI already parses — with `is_completion=False`, so Atlas' own LLM writes the answer. `mode: "synthesized"` returns the backend's answer with `is_completion=True`, i.e. v1's user-visible behaviour.
- **`RAGSourceConfig.api_version`** (`"v1"` default / `"v2"`) picks the contract per configured source; `default_mode` picks the shape for callers that do not ask for one. Endpoint paths default per version, and an explicit `query_endpoint`/`discovery_endpoint` still wins.
- **`UnifiedRAGService.query_rag` / `query_rag_batch`** take optional `query` and `mode`. An omitted `query` still derives from the last user message, so **every existing caller behaves exactly as before** and v1 sources ignore both arguments.
- **The `atlas_rag_query` agent tool** passes its already-explicit query straight through and defaults to `mode: "raw"`, so on a v2 source the model reasons over retrieved evidence alongside other tool results instead of relaying a backend-written answer. The tool schema gains an optional `mode`; an unrecognized value falls back to `"raw"` rather than failing the call, since it is model-supplied.
- **The mock** serves `GET /api/v2/discover/datasources` (each source declaring `api_version`) and `POST /api/v2/rag/query`. v1 stays in place.

## Security / authorization

Unchanged, deliberately. `_ensure_source_query_allowed`, group and compliance filtering, `as_user` impersonation and `strip_domain` all apply identically on both paths — the contract version decides the request shape, not who may read what. The v2 tool schema stays model-facing only: no identity or compliance level is a tool argument.

Data minimization improves: `messages` is not part of a v2 request, so conversation history never reaches a v2 backend. The `rag.query` span keeps `query_hash`/`query_chars` and gains `mode` and `explicit_query`; the query text itself is still never logged.

## Not in this PR

Phases 2-4 from the design note: discovery-driven version negotiation (config declares the version today; `api_version` in the discovery response is parsed but not yet acted on), removing the `call_with_rag` / `call_with_rag_and_tools` / `_build_rag_completion_response` special-casing, a v2 MCP transport shape, and retiring v1. Open questions #1 and #4 are answered for Phase 1 (`synthesized` default per source, `raw` for the agent tool); the rest remain open in [the design note](docs/planning/rag-api-v2-tool-interface.md).

## Testing

- `atlas/tests/test_atlas_rag_v2.py` — 33 new tests: request body (including that `messages` is absent), both response shapes, malformed-entry tolerance, caller-error vs backend-error handling, status mapping, and the v1/v2 routing decision in `UnifiedRAGService`.
- `test/pr-validation/test_issue791_rag_api_v2.sh` — 11 checks, including the mock end-to-end over both modes plus 400/403/401 handling. All pass locally.
- Full backend suite: **2211 passed, 17 skipped**. Three existing tests were updated for the new optional kwargs (test fakes and one `assert_called_once_with`).

## Docs

`docs/admin/external-rag-api.md` gains an "API v2" section (config fields, both request/response shapes, what changes in the UI) and labels the existing contract as v1. The design note from the issue is added under `docs/planning/`, with an implementation-status section recording what landed.


garland3:
Review triage — pushed `f1caffe`.

**Fixed**

- **Codex P2, v2 citations dropped from the tool result** — real gap, and it applied to v1 sources too. `_tool_references()` now folds a capped `{document_ref, filename, citation, url}` list into the `atlas_rag_query` payload for both contracts. Raw evidence already carried its `[N]` markers and snippet text inline, so the snippets are not duplicated — only the identity that lived solely on `RAGResponse.metadata`.
- **Copilot, error body logged on v2 failures** — correct, and it undercut the point of v2. The body is no longer logged; the status code and endpoint are what diagnose a backend failure. (v1 is left alone: it already posts the whole conversation, so its logging is a separate question.)
- **Copilot, ambiguous `mode` span attribute** — renamed to `requested_mode`, so empty reads as "the caller let the source decide". The mode that actually ran is already on the response attributes as `retrieval_method` (`v2_raw` / `v2_synthesized`).
- **CodeQL log-injection ×6 on the new mock handlers** — the mock now strips control characters and caps `as_user`/`corpora` before logging.
- **Copilot, grammar in the validation script comment** — fixed.

**Not changing**

- **Codex P1, missing changelog entry** — stale review; it ran against `da24257`. The `### PR #793` entry landed in the follow-up commit and is in the diff.
- **Copilot, "changelog says #793 but this PR is #791"** — #791 is the issue; this PR is #793, and the changelog convention keys on the PR number. Entry header is correct as-is.

Verification after the fixes: 36 v2 unit tests pass, full backend suite 2214 passed / 17 skipped, `test/pr-validation/test_issue791_rag_api_v2.sh` 11/11, ruff clean on the touched files.

garland3:
## Summary
Adds a v2 external RAG contract: a query-only request shape (no conversation history), `raw`/`synthesized` response modes, a v2 parser and metadata builder in `AtlasRAGClient`, version routing in `UnifiedRAGService`, and a `references` payload plus `mode` argument on the `atlas_rag_query` agent tool. Docs, a mock backend, and 33 new unit tests come with it.

## Verdict
Request changes.

## Required changes
1. Renumber the evidence markers in `_format_raw_evidence` (`atlas/modules/rag/atlas_rag_client.py:645`) from `enumerate(documents, start=1)` instead of `doc.document_ref` — the citation instructions and references footer number positionally (`litellm_caller.py:1265`, `:1325`), so a backend returning non-contiguous refs, a hit with no `document_ref`, or a second source in `_combine_rag_contexts` (`litellm_caller.py:570`) emits `[N]` markers that point at the wrong document or at none.
2. Guard the blank derived query in `_query_http_client` (`atlas/domain/unified_rag_service.py:443`) before calling `query_v2` — `_extract_query_text` returns `""` for an attachment-only or whitespace-only turn, and `query_v2`'s `ValueError` is not an `HTTPException`, so a request that degraded gracefully on v1 becomes an unhandled 500 on v2.
3. Stop rendering "0% relevance" for v2 `synthesized` citations (`atlas/modules/rag/atlas_rag_client.py:595`, `litellm_caller.py:1347`) — citations carry neither `relevance` nor `sections`, so `confidence_score` is always `0.0` and every reference under the shipped default mode is shown to the user as 0% relevant. Omit the clause when no score was reported.
4. Normalize or validate `mode` at the service boundary in `query_rag`/`query_rag_batch` (`atlas/domain/unified_rag_service.py:448`) — the public API documents `"raw"`/`"synthesized"` but forwards anything else into a `ValueError`, so a typo like `"Raw"` from any non-tool caller is a 500.
5. Add `isinstance(requested_mode, str)` to the membership test at `atlas/modules/mcp_tools/mcp_execution.py:766` — a model-supplied list or object raises `TypeError: unhashable type`, failing the entire tool call including every other source, which contradicts the fallback the comment above it describes.
6. Bound the rendered evidence in `_format_raw_evidence` (`atlas/modules/rag/atlas_rag_client.py:623`) — cap documents, sections per document, and characters per section, and note the truncation — since `_MAX_TOOL_REFERENCES` caps the compact citation list while the far larger snippet text goes through uncapped on every turn.
7. Fence and label the retrieved text in `_format_raw_evidence` (`atlas/modules/rag/atlas_rag_client.py:623`) the way tool-digest results are already escaped and fenced — verbatim corpus text now enters the model's context with tools live and no delimiter or "this is data, not instructions" preamble, so a document added to an indexed corpus can steer another user's agent.

## Evidence
Evidence: 8/10 — strong for a change of this size (33 tests pinning both response shapes and the routing decision, an 11-check mock script, concrete payloads in the docs). The most important gap is that nothing runs a real `AtlasRAGClient(api_version="v2")` against the real mock, so field-name drift between mock and parser would not be caught, and the `[N]`-citation guarantee is asserted only as a substring rather than through the citation pipeline.

## Notes
- `_tool_references` emits a key named `filename` holding `doc.title or doc.source` — rename to `title` while the field has no consumer (`atlas/modules/mcp_tools/mcp_execution.py:54`).
- The span attribute is `requested_mode` but the description, CHANGELOG and design note all call it `mode` (`atlas/domain/unified_rag_service.py:501`).
- Endpoint path defaults are declared in both `config/models.py:240` and `atlas_rag_client.py:74`, and the client copy is unreachable in production — pick one owner.

---

garland3:
**CI: all 10 checks green** on `f1caffe` (build-and-test, build-artifacts, CodeQL, SAST, Trivy, grype, Container/Dependency/Secrets scans, Security Summary). Mergeable, no conflicts with `main`.

On the remaining `py/log-injection` annotations in `mocks/atlas-rag-api-mock/main.py`: the flagged calls now go through `safe_log()`, which strips control characters and caps length, so a newline cannot forge a log line — CodeQL just does not model the custom sanitizer. `mocks/*/main.py` is also already listed in `.codeqlignore`, and these are dev-only mock handlers that mirror what the existing v1 handlers log. The CodeQL check itself passes.

garland3:

PR #794, opened 2026-08-13
fix(tests): address review nits from #789 (status check, clone_from, fixture adoption)

garland3:
Follow-up to #789 (merged as 4a74584). AGENT-REVIEW-BOT-3 posted "Approve with nits" after the merge; this PR addresses the actionable nits so the merged test-identity cleanup is fully consistent.

## Changes

**Status check on mock registration (real bug)**
`test_atlas_rag_integration.py` `mock_service` only caught `httpx.HTTPError` (connection errors), not 4xx/5xx. An older mock without `/admin/users` returns 404, the fixture silently proceeds, and the live integration tests later fail with unrelated assertions. Now calls `resp.raise_for_status()` so any 4xx/5xx skips with a clear reason.

**No more hand-copied group data**
Replaces `MOCK_ALL_CORPORA_GROUPS = [...]` (which hand-copied the `test@test.com` entry from `mock_data.json`) with `clone_from: "test@test.com"` on the new `POST /admin/users` endpoint, so the test references the mock's source-of-truth user instead of duplicating its groups.

**Adopt the shared identity fixtures**
`test_capture_routes.py` / `test_feedback_routes.py` snapshot `config_manager.app_settings.test_user/admin_test_user` into module-level header constants at import time, invisible to `reload_configs()`, and the new `conftest.py` fixtures had no consumers. Both files now use the `test_user_headers` / `admin_test_user_headers` fixtures (resolved at test time). The AtlasRAGClient unit tests adopt the `test_user` fixture too. The fixtures now have real consumers.

**Docs**
Documents the `POST /admin/users` endpoint in `mocks/atlas-rag-api-mock/README.md` (Endpoints section + Test Users note), which the merged PR omitted.

## Verification
- Full backend suite: 2178 passed, 17 skipped (the 17 skips are the `RUN_RAG_INTEGRATION`-gated cases and unrelated env-gated tests).
- `ruff` clean on all changed files.
- Verified the mock endpoint end-to-end against a running mock: `clone_from` copies groups correctly (registered user sees all 3 corpora and can query `technical-docs`); `clone_from` a nonexistent user returns 404; unauthenticated registration returns 401.

Follows the OPSEC rules (no AI attribution).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### PR #794 - 2026-08-13
+- **Test-identity cleanup follow-up (review nits on #789)**: the RAG integration fixture's mock registration now calls `raise_for_status()` and verifies the returned groups are non-empty, so a 4xx/5xx or an older mock without `clone_from` (which would register an empty group list) skips with a clear reason instead of silently proceeding; the configured test user is registered via `clone_from` (cloning the mock's all-corpora identity) rather than hand-copying group data from `mock_data.json`; the mock's `POST /admin/users` enforces an either/or contract for `groups`/`clone_from` via a `model_validator`; the shared `test_user_headers` / `admin_test_user_headers` / `test_user` fixtures are now adopted by the capture/feedback route tests and the AtlasRAGClient unit tests, replacing module-level header constants that snapshotted config at import time; the `POST /admin/users` endpoint is documented in the RAG mock README and covered by unit tests.
+
 ### PR #789 - 2026-08-13
 - **Tests use configured development identities**: Python tests now reference the configured `test_user` and `admin_test_user` values instead of hardcoded default email strings (including the remaining `admin@test.com` admin-route mocks), with shared pytest fixtures available for future auth-focused tests. The RAG API mock gains an authenticated `POST /admin/users` endpoint so live integration tests can register the configured identity and stay robust to `TEST_USER` overrides.
 
--- a/atlas/tests/conftest.py
+++ b/atlas/tests/conftest.py
@@ -158,6 +158,12 @@ def admin_test_user():
 
 
 @pytest.fixture
+def admin_group():
+    """Return the configured admin group name (the group admin-route checks test against)."""
+    return _config_manager.app_settings.admin_group
+
+
+@pytest.fixture
 def test_user_headers(test_user):
     """Return auth headers for the configured non-admin test identity."""
     return {"X-User-Email": test_user}
--- a/atlas/tests/test_atlas_rag_integration.py
+++ b/atlas/tests/test_atlas_rag_integration.py
@@ -25,12 +25,12 @@ MOCK_URL = "http://localhost:8002"
 MOCK_TOKEN = "test-atlas-rag-token"
 MOCK_STARTUP_TIMEOUT = 10
 
-# Groups that grant access to every corpus in the mock (matches the
-# ``test@test.com`` entry in mock_data.json). The integration test fixture
-# registers the configured ATLAS test identity with these groups so the live
-# mock recognizes it regardless of TEST_USER overrides -- the mock otherwise
-# ships a fixed user database.
-MOCK_ALL_CORPORA_GROUPS = ["employee", "engineering", "devops", "admin"]
+# The mock ships a fixed user database (see mocks/atlas-rag-api-mock/mock_data.json);
+# ``test@test.com`` is its all-corpora identity. The integration test fixture
+# clones this identity into the configured ATLAS test user so live integration
+# tests are robust to TEST_USER overrides -- the mock otherwise wouldn't
+# recognize an overridden TEST_USER and would only return the public corpus.
+MOCK_ALL_CORPORA_USER = "test@test.com"
 
 
 def is_mock_running() -> bool:
@@ -40,6 +40,24 @@ def is_mock_running() -> bool:
         return response.status_code == 200
     except Exception:
         return False
+
+
+def _abort_registration(process, reason):
+    """Handle a test-user registration failure/incompatibility.
+
+    A failure against a freshly spawned mock (``process is not None``) is a
+    regression in this repo's own mock and fails the suite so it is not
+    silently hidden; against a pre-existing external mock it is treated as
+    an incompatibility and the live integration tests skip.
+    """
+    if process is not None:
+        process.terminate()
+        process.wait(timeout=5)
+        pytest.fail(
+            f"Freshly spawned RAG mock rejected test-user registration: {reason}",
+            pytrace=True,
+        )
+    pytest.skip(reason)
 
 
 @pytest.fixture(scope="module")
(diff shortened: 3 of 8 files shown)


garland3:
## Summary
Tightens the RAG integration test fixture — the mock-registration call now checks its response status and clones groups from the mock's all-corpora user instead of hand-copying a group list — and adds a `clone_from` field to the mock's `POST /admin/users`, plus conftest identity fixtures adopted by the capture and feedback route tests.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. Adequately demonstrated for a test/mock change: the full backend suite result is reported, and the new endpoint behavior was verified by hand. The one missing artifact is the actual output of that manual verification (or a `RUN_RAG_INTEGRATION=1` run) — the successful `clone_from` response showing the copied groups is the claim the whole change rests on.

## Notes
- `raise_for_status()` still passes against a mock that predates `clone_from`: pydantic drops the unknown field, `groups` defaults to `[]`, and registration returns 200 with no memberships — validate `resp.json()["groups"]` is non-empty at `atlas/tests/test_atlas_rag_integration.py:96` rather than just the status.
- `RegisterUserRequest` accepts both `groups` and `clone_from` (clone silently wins) or neither (silently registers an empty group list, revoking access for a known user) — a `model_validator` requiring exactly one would make the README's "either/or" contract structural, at `mocks/atlas-rag-api-mock/main.py:213`.
- The new clone branch has no automated test; a `TestClient` case for clone-succeeds and unknown-source-404 at `mocks/atlas-rag-api-mock/main.py:524` would pin the behavior the fixture now depends on.

---

garland3:
## Summary
Adds a `clone_from` field to the RAG mock's `/admin/users` endpoint so the integration fixture copies an existing user's groups instead of hardcoding a group list, adds `raise_for_status()` to the registration call, and converts three test modules to the shared `test_user` identity fixtures.

## Verdict
Approve with nits.

## Evidence
Evidence: 7/10. The suite result, lint run, and written account of manual verification are proportionate for a test/mock change, but the reported run skips the `RUN_RAG_INTEGRATION`-gated tests that actually exercise the modified fixture — pasting the `RUN_RAG_INTEGRATION=1 pytest atlas/tests/test_atlas_rag_integration.py` output (or the three curl responses) would move this from asserted-working to shown-working.

## Notes
- `raise_for_status()` misses the closest skew case at `atlas/tests/test_atlas_rag_integration.py:96`: a mock that has `/admin/users` but predates `clone_from` drops the unknown field (pydantic `extra="ignore"`), registers with `groups=[]`, and returns 200 — asserting the response body has non-empty groups would catch it.
- Enforce the either/or contract structurally at `mocks/atlas-rag-api-mock/main.py:212` with a `model_validator`: both-set silently prefers `clone_from`, neither-set registers with no groups and overwrites an existing user's memberships.
- The new clone branch and its 404 path have no automated test — a `TestClient` case over the mock app would pin the contract the fixture depends on.

---

garland3:
## Summary
Follow-up to #789: hardens the RAG integration fixture's mock-registration step, adds a `clone_from` option and an either/or validator to the mock's `POST /admin/users`, converts capture/feedback tests onto the shared identity fixtures, and documents/tests the endpoint.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. Adequately demonstrated for a test/mock change — the new `TestClient` module is the strongest artifact — but the modified integration fixture itself is never exercised in the reported run; a `RUN_RAG_INTEGRATION=1` pytest tail pasted into the description would close that gap.

## Notes
- `_require_exactly_one_source` keys on truthiness, not presence (`mocks/atlas-rag-api-mock/main.py:213`): `{"user": x, "groups": []}` now 422s, so a groupless user is unrepresentable, while `{"groups": [], "clone_from": ...}` passes despite the README saying both-set is rejected. Declaring `groups: Optional[List[str]] = None` and testing `is None` fixes both. Note `clone_from` on a groupless source still yields the empty registration the validator exists to prevent.
- `test_rag_mock_admin_users_endpoint.py` mutates `mock_main.USERS_GROUPS_DB` with no restore and asserts the clone against live state, so it would pass even if the copy aliased the source list — add an autouse snapshot/restore fixture and assert against a literal plus an `is not` check.
- `atlas/tests/test_atlas_rag_integration.py:96`: `resp.json().get("groups")` raises `AttributeError` on a non-dict 200 body, erroring the fixture instead of skipping; the teardown pair is also repeated across four exit paths with no `kill()` fallback.

---

garland3:
## Summary
Adds a `clone_from` option to the RAG mock's `POST /admin/users` endpoint with a mutual-exclusion validator, hardens the `mock_service` integration fixture so a failed or stale registration skips with a clear reason instead of failing later, and adds a `TestClient` module pinning the endpoint contract.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. Well demonstrated for a test/mock change — the new endpoint tests plus the reported manual end-to-end check cover the server side. The one gap is the fixture itself: it is gated behind `RUN_RAG_INTEGRATION` and excluded from the reported run, so a short `RUN_RAG_INTEGRATION=1` transcript (or the raw curl output) would convert its behavior from asserted to shown.

## Notes
- `_require_exactly_one_source` keys on truthiness, not presence, so `groups: []` is indistinguishable from absent — `{"groups": [], "clone_from": "..."}` returns 200 despite the README saying 422, and `{"groups": []}` is rejected so a groupless user can't be registered. Declaring `groups: Optional[List[str]] = None` and branching on `is None` fixes both directions (`mocks/atlas-rag-api-mock/main.py:213`).
- `assert body["groups"] == list(mock_main.USERS_GROUPS_DB["test@test.com"])` compares the response against the live global the handler read from, so an aliased (non-copying) clone would still pass; assert against a literal plus an `is not` identity check, and add an autouse fixture restoring `USERS_GROUPS_DB` so the module isn't order-coupled (`atlas/tests/test_rag_mock_admin_users_endpoint.py:45`).
- The fixture now has four shutdown sites with no `kill()` fallback, undrained `PIPE`s, and a `resp.json().get(...)` that raises `AttributeError` on a non-dict 200 body — routing all exits through one `_stop(process)` helper and using `DEVNULL` would keep a hung mock from orphaning port 8002 (`atlas/tests/test_atlas_rag_integration.py:66`).

---

garland3:
Update on the latest commit (52e8a01) — all CI is green (build-and-test, CodeQL, Secrets, SAST, container scans). This round addressed the further review feedback:

**CodeQL / Log Injection (2 medium)** — the mock's `register_test_user` logged `req.user` (user-controlled) via `logger.info("...%s", req.user, groups)`. Now logs only the group count (`"Registered test user with %d group(s)"`), so no user input flows into the log.

**CodeQL / Potentially uninitialized local variable (error, was blocking)** — `registered_groups` was read after a try/except that could skip its assignment. Moved the empty-groups check into a try/except `else` block, so `registered_groups` is only read when the try succeeded.

**Codex P2 — fail registration errors from a freshly spawned mock** — `raise_for_status()` previously routed any 4xx/5xx to `pytest.skip`, which would silently skip every live RAG integration test on a regression in this repo's own mock. Added `_abort_registration(process, reason)`: a failure against a freshly spawned mock (`process is not None`) now `pytest.fail`s (it's our code), while incompatibility with a pre-existing external mock still `skip`s.

**Copilot — `clone_from` precedence** — already structural via the `RegisterUserRequest.model_validator` added earlier (providing both or neither is rejected with 422), so clone no longer silently wins.

**Copilot — admin mocks didn't verify the group name** — `mock_admin` / `mock_admin_check` returned True for the admin user regardless of the `group` argument, so a route checking the wrong group name would still pass. Added an `admin_group` conftest fixture and the mocks now match on both identity and group (`user == admin_test_user and group == admin_group`).

**TruffleHog false-positive** — the Lob detector matched the test function name `test_clone_from_unknown_user_returns_404` (the `_404` suffix) as a Lob key and Lob's test API verified it, failing the secrets scan. Renamed to `test_clone_from_unknown_user_not_found`; verified the diff scans clean (0 verified secrets) with the same TruffleHog image CI uses.

Verification: full backend suite 2184 passed / 17 skipped; `ruff` clean on all changed files; TruffleHog + CodeQL both pass in CI.

garland3:
## Summary
Hardens the RAG integration test fixture so a failed test-user registration fails or skips loudly instead of silently proceeding, adds a `clone_from` option to the mock's `POST /admin/users` so tests can copy an existing user's groups instead of hand-copying them, and hoists shared identity/header fixtures into `conftest.py` with new unit coverage for the mock endpoint.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — well demonstrated for a test/mock change, with the full backend suite reported green. The one missing artifact is a run of the very fixture this PR hardens: the `RUN_RAG_INTEGRATION`-gated cases are among the 17 skips, so none of the new fail-vs-skip logic actually executed in the reported run.

## Notes
- `_require_exactly_one_source` branches on truthiness rather than presence, so `{"groups": [], "clone_from": ...}` is accepted despite the README saying both-supplied is a 422, while `{"groups": []}` is rejected — declare `groups: Optional[List[str]] = None` and test `is None` (`mocks/atlas-rag-api-mock/main.py:228`, `README.md:126`).
- The fixture's `terminate()`/`wait(timeout=5)` pair is repeated at four exit sites with no `kill()` fallback, and the mock is spawned with undrained `PIPE`s — route every exit through one `_stop(process)` helper and use `subprocess.DEVNULL` (`atlas/tests/test_atlas_rag_integration.py:54`, `:83`).
- `test_clone_from` asserts the response against the same live global the handler wrote, so an aliasing (non-copying) clone would still pass; assert against a literal and add an identity check, plus an autouse snapshot/restore of `USERS_GROUPS_DB` (`atlas/tests/test_rag_mock_admin_users_endpoint.py:45`).

---

PR #802, opened 2026-08-16
fix: carry the cross-turn tool digest into the default tools-mode path (#798)

### LINKED ISSUE #798: Cross-turn tool digest (#755) is agent-mode only, so it misses the default path
### Summary

The cross-turn tool digest added in #755 is attached only in **agent mode**. A normal turn
runs in **tools mode**, which never attaches a digest — so in the default configuration the
fix does not reach most conversations, and tool work is still invisible to every later turn.

### Evidence

**Agent mode is off by default at every layer**

- Frontend toggle default: `frontend/src/hooks/chat/useAgentMode.js:5` —
  `usePersistentState('chatui-agent-mode-enabled', false)`
- Sent per request: `frontend/src/contexts/ChatContext.jsx:400` — `agent_mode: agent.agentModeEnabled`
- WS entry point: `atlas/main.py:686` — `agent_mode=data.get("agent_mode", False)`
- Service signature default: `atlas/application/chat/service.py:257`
- Other clients also default off: `atlas/atlas_client.py:76`, `atlas/atlas_chat_cli.py:380`
- Gated behind `feature_agent_mode_available` (`atlas/modules/config/settings.py:102`)

It is additionally **force-downgraded** to tools/chat mode when the model lacks tool calling
(`atlas/application/chat/orchestrator.py:311-317`) or when no tools are selected
(`:334-342`). Dispatch is `orchestrator.py:345` (agent) vs `:347` (tools).

**The digest is agent-only**

- Only call site: `atlas/application/chat/modes/agent.py:20,237` (`_close_turn`)
- `grep digest atlas/application/chat/modes/tools.py` → no hits

**So tools mode has no cross-turn carry-over at all**

- Its three closing assistant messages carry only `tools`/`data_sources` metadata:
  `tools.py:181-191`, `tools.py:413-422`, `tools.py:667-675`
- Persisted `tool_call` rows are dropped from the model view
  (`atlas/domain/messages/models.py:13,206,226`), so the digest fold at `:208` finds nothing
  for a tools-mode turn
- Interrupted tools-mode turns close with a generic message, also digest-less:
  `atlas/application/chat/utilities/interrupted_turn.py:53-78`
- Within a turn the loop keeps a local working transcript (`tools.py:318-372`), but it dies
  with the turn — the only survivor is whatever the final prose happens to restate

The practical effect is the one #755 describes: ask a follow-up question about work a tool
did in the previous turn and the assistant has no record of it, so it either re-runs the
tool or answers from the prose summary. This is most visible on tools that produce large
results the model summarized briefly.

### Suggested change

Wire `build_tool_digest` into `modes/tools.py` rather than changing the `agent_mode`
default. Concretely: capture `turn_start_index` before the loop and add an agent-style
`_close_turn` that attaches `AGENT_TOOL_DIGEST_KEY` at the three closing sites, and ideally
on the interrupted path too.

The regression surface looks small: `get_messages_for_llm` already caps folds at 3 digests /
12000 chars (`atlas/domain/messages/models.py:24-25`) and folds into an existing message, so
role alternation is unchanged. Costs are prompt growth on tool-heavy chats, a shared budget
with agent digests, and inheriting the digest's untrusted-data quoting as-is.

Flipping the `agent_mode` default would be the wrong fix — it changes latency, cost, step
budgets, approval UX, and the emitted event stream for every user, and it still would not
help the turns that get downgraded back to tools mode by `orchestrator.py:311-342`.

### Question for maintainers

Was the agent-only scoping deliberate? The commit bodies for #755 (`d31bc07`, `0ea1754`)
describe the digest purely in agent-loop terms ("an agent turn's tool work was invisible to
every later turn"), and tools mode was touched in the same work but only for unwind/flush and
turn-closing — no stated reason to withhold the digest from it. If it was intentional, the
reasoning is worth a comment in `tools.py`, since the omission reads as incidental. Happy to
send a PR once the intent is confirmed.

One thing I did not verify: whether the persisted `tool_call` rows in tools mode carry every
field `agent_digest` expects. `tool_history.py` suggests they do, but I did not execute it.

garland3:
## Summary
Closes #798. The cross-turn tool digest added in #755 was attached only in
agent mode, so a normal turn -- which runs in tools mode by default -- never
carried it and tool work was invisible to every later turn.

## Changes
- `ToolsModeRunner` now attaches `agent_tool_digest` at all three closing
  sites (`run`, `run_streaming` synthesis, `_finalize_text_response`) via a
  new `_close_turn` helper that mirrors `AgentModeRunner._close_turn`.
- `close_open_turn` in `interrupted_turn.py` attaches a digest from the
  flushed `tool_call` rows on the interrupted path, switching the terminal
  message to `INTERRUPTED_TURN_CONTENT_WITH_DIGEST` so it only claims a
  record follows when one is actually attached.
- Adds tools-mode digest tests and a PR validation script.

## Regression surface
Small: `get_messages_for_llm` already caps folds at 3 digests / 12000 chars
and folds into an existing message, so role alternation is unchanged. Costs
are prompt growth on tool-heavy chats (shared budget with agent digests) and
inheriting the digest's untrusted-data quoting as-is.

## Test plan
- `pytest atlas/tests/test_interrupted_turn_persistence.py` (42 passed)
- `ruff check` on changed files (clean)
- `test/pr-validation/test_pr798_tools_mode_tool_digest.sh`


garland3:
## Summary
Attaches the tool-call digest at the three tools-mode turn-closing sites (`run`, the `run_streaming` synthesis, and `_finalize_text_response`) and on the interrupted-turn path, so a follow-up turn no longer re-derives what the previous turn's tools already established. Adds tests, a validation script, and docs/CHANGELOG updates.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — proportionate for a backend prompt-assembly change with no visual surface: all three closing sites, the interrupted path, an end-to-end stop, and the save/reload round-trip are covered, plus a negative case. The most useful missing artifact is a short sample of the folded digest text as the model actually receives it, so a reader can judge shape and readability rather than just substring presence.

## Notes
- The stopped-turn copy at `atlas/application/chat/utilities/interrupted_turn.py:36` promises "a record of the tool calls it completed follows below", but the digest is metadata-only — the user sees nothing below it (the `tool_call` rows render above), and for the model the claim goes stale once the 3-digest fold budget evicts it. Consider dropping the content switch and relying on the digest's own header.
- `_finalize_text_response(..., turn_start_index: int = 0)` at `atlas/application/chat/modes/tools.py:679` defaults to a wrong value rather than an inert one; a future caller that omits it would silently digest the entire conversation. Make it required and keyword-only — the single call site already passes it by keyword.
- No test pins the digest to the turn boundary (`atlas/tests/test_interrupted_turn_persistence.py:1170`): every case starts from a single-user-message session, so a regression hardcoding index 0 would produce a byte-identical digest. A case seeding one completed prior turn and asserting the old tool name is absent would close that gap.

---

garland3:
## Summary
Extends the per-turn tool digest — previously attached only in agent mode — to the default tools-mode path, wiring it through `ToolsModeRunner`'s three closing sites and the shared cancel-path closer `close_open_turn`, with tests, a validation script, and docs.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is adequately demonstrated for its size: every closing site it touches is exercised through production code paths, including a save/reload round-trip, a negative case, and a turn-boundary pin. The most useful missing artifact is a short sample of the rendered digest as `get_messages_for_llm()` actually emits it — current assertions only check that a tool name appears somewhere in the folded text.

## Notes
- Rename `test/pr-validation/test_pr798_tools_mode_tool_digest.sh` and its header to reference PR #802 — the documented pattern is the PR number, so `grep 802` finds nothing today.
- `./test/run_tests.sh backend > /dev/null 2>&1` at `test/pr-validation/test_pr798_tools_mode_tool_digest.sh:259` discards all output, so a failing gate reports no test name or traceback; capture to a file and tail it on failure.
- The digest/metadata-merge policy is now duplicated three times (`atlas/application/chat/modes/tools.py:713`, `agent.py:216`, `utilities/interrupted_turn.py:90`); folding it into one helper next to `build_tool_digest` would keep future changes (size caps, redaction) from needing three edits.

---

PR #815, opened 2026-08-18
fix(chat-history): stop relying on DuckDB secondary indexes

garland3:
## Problem

DuckDB's ART indexes can silently stop matching rows that are present in the file. An equality lookup such as `WHERE user_email = ?` returns nothing, while a full scan of the same table returns the row.

Two properties make this worse than a normal bug:

- **It is durable.** The index does not recover on reopen. Once a file is affected it stays affected, and every subsequent write lands in a table whose index no longer resolves it.
- **It is silent.** No error, no warning. The application receives an empty result set and renders it as "no data".

The user-visible symptom reads as data loss: conversations open with **no messages**, and custom prompts **disappear** from the prompt list. The conversation list still looks complete, which makes it look like a rendering bug rather than a lookup one — the index backing that list is separate from the ones that broke.

No data is actually lost. The rows are intact and a full scan finds all of them; only the indexes are wrong.

Upstream has an open report of the same shape: [duckdb/duckdb#19190](https://github.com/duckdb/duckdb/issues/19190) ("Missing results for index lookup"). That report describes a non-deterministic variant resolved by `SET threads=1`; the durable variant here is not affected by the thread count, so it appears to be the same family with a persistent outcome.

## Scope

**Development only.** DuckDB is the local/dev backend; PostgreSQL is the production one. PostgreSQL's indexes are unaffected and remain necessary there, so production behavior is unchanged.

## Change

Drop the declared secondary indexes on the DuckDB backend during `init_database()`, for both the chat-history and agent-portal stores. Dropping them also repairs a database that is already affected, so existing local databases heal on the next startup.

**Why drop rather than rebuild.** A rebuild repairs the rows that exist today but leaves the same failure mode armed for tomorrow. Local datasets are small enough that a full scan is cheap, so these indexes buy little and are the only surface the bug needs.

One implementation note worth flagging for review: `Index.drop(checkfirst=True)` **cannot** be used here. It probes for the index via reflection, which duckdb-engine does not implement, so it skips every drop while reporting success. The DDL is emitted directly instead. This was caught by a test that asserted through `inspect().get_indexes()` — which always returns `[]` on duckdb-engine and therefore passed vacuously. The tests now read `duckdb_indexes()` directly.

## Repair script

`scripts/repair_duckdb_indexes.py` repairs a file out of band, for inspecting or fixing a database without starting the application:

```bash
python scripts/repair_duckdb_indexes.py data/chat_history.db          # drop
python scripts/repair_duckdb_indexes.py data/chat_history.db --check  # inspect only
python scripts/repair_duckdb_indexes.py data/chat_history.db --rebuild
```

It copies the database before making changes. The application must be stopped first — DuckDB holds an exclusive lock on the file.

## Testing

- New `atlas/tests/test_duckdb_index_dropping.py`: indexes are dropped on DuckDB, rows stay queryable afterwards, the helper is a no-op on other backends, and the chat-history startup path leaves no indexes behind.
- Full suite: **2334 passed, 17 skipped**.
- Verified against a database exhibiting the fault: affected conversations returned 0 messages and one of two stored prompts before the change, and all conversations plus both prompts after it.

### DIFF
--- a/atlas/core/duckdb_indexes.py
+++ b/atlas/core/duckdb_indexes.py
@@ -1,0 +1,66 @@
+"""Work around DuckDB's unreliable secondary (ART) indexes.
+
+DuckDB's ART indexes can silently stop matching rows that are plainly present:
+an equality lookup such as ``WHERE user_email = ?`` returns nothing while a full
+scan of the same table returns the row. Upstream has an open report of the same
+shape (duckdb/duckdb#19190). Once a file is affected the damage is durable --
+it survives reopening the database, and it is invisible to the application,
+which simply sees empty result sets and renders them as "no data".
+
+The blast radius is development only: DuckDB is the local/dev backend and
+PostgreSQL is the production one (see ``chat_history.database``). PostgreSQL's
+indexes are unaffected and remain necessary there, so this workaround is scoped
+to the DuckDB dialect and leaves every other backend untouched.
+
+Dropping the indexes rather than rebuilding them is deliberate. A rebuild
+repairs the rows that exist today but leaves the same failure mode armed for
+tomorrow; local datasets are small enough that a full scan is cheap, so the
+indexes buy nothing and are the only surface the bug needs. Dropping them on
+startup also repairs a database that is already affected.
+"""
+
+import logging
+
+from sqlalchemy import MetaData
+from sqlalchemy.engine import Engine
+
+logger = logging.getLogger(__name__)
+
+
+def drop_duckdb_secondary_indexes(engine: Engine, metadata: MetaData) -> int:
+    """Drop the declared secondary indexes when ``engine`` is backed by DuckDB.
+
+    Only indexes declared on ``metadata`` are touched; the indexes DuckDB
+    maintains internally for primary-key and unique constraints are left alone,
+    since those enforce correctness rather than accelerate lookups.
+
+    Returns the number of indexes dropped (0 on every non-DuckDB backend).
+    """
+    if engine.dialect.name != "duckdb":
+        return 0
+
+    # ``Index.drop(checkfirst=True)`` cannot be used here: it probes the index
+    # via reflection, which duckdb-engine does not implement, so it silently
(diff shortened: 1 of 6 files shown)


garland3:
## Summary
Works around a DuckDB defect where secondary indexes return empty result sets by dropping declared non-constraint indexes after `create_all` on the DuckDB backend, wired into both the chat-history and agent-portal startup paths, plus a standalone operator repair script and admin docs.

## Verdict
Request changes.

## Required changes
1. Name the pre-repair backup so existing ignore rules cover it, or add a `data/*.prerepair-*` entry to `.gitignore` — `scripts/repair_duckdb_indexes.py:88` writes `data/chat_history.db.prerepair-<stamp>`, which matches neither `data/*.db` nor `data/*.wal`, so `git add -A` stages a file containing full conversation transcripts and user email addresses.
2. Open the DuckDB connection before taking the backup at `scripts/repair_duckdb_indexes.py:88` (or delete the copy in the `IOException` branch) — running the script with the application still up copies a file another process is writing, prints "backed up to ...", then exits 1, leaving an untrustworthy stray copy behind on every retry.

## Evidence
Evidence: 8/10. The change is well demonstrated for a backend fix — the failure mode is described precisely, four tests cover the drop and the chat-history startup path, and the vacuous-test trap is called out explicitly. The one load-bearing gap is that the field verification ("affected conversations returned 0 messages before, all of them after") is asserted in prose rather than shown as before/after query output.

## Notes
- `dropped` counts `DROP INDEX IF EXISTS` calls rather than indexes that existed, so every restart after the first logs the full count while dropping nothing (`atlas/core/duckdb_indexes.py:56`).
- `alembic/versions/001_initial_chat_history_schema.py` and `002_add_user_prompts.py` still `create_index`, so `alembic upgrade head` against a local DuckDB file re-arms the failure mode until the next startup.
- The agent-portal store (`data/agent_portal.db`) is affected identically but is documented only under chat history, and its startup path has no test (`atlas/modules/agent_portal/database.py:109`).

---

PR #816, opened 2026-08-18
fix(chat-history): stop relying on DuckDB secondary indexes

garland3:
## Problem

DuckDB's ART indexes can silently stop matching rows that are present in the file. An equality lookup such as `WHERE user_email = ?` returns nothing, while a full scan of the same table returns the row.

Two properties make this worse than a normal bug:

- **It is durable.** The index does not recover on reopen. Once a file is affected it stays affected, and every subsequent write lands in a table whose index no longer resolves it.
- **It is silent.** No error, no warning. The application receives an empty result set and renders it as "no data".

The user-visible symptom reads as data loss: conversations open with **no messages**, and custom prompts **disappear** from the prompt list. The conversation list still looks complete, which makes it look like a rendering bug rather than a lookup one -- the index backing that list is separate from the ones that broke.

No data is actually lost. The rows are intact and a full scan finds all of them; only the indexes are wrong.

Upstream has an open report of the same shape: [duckdb/duckdb#19190](https://github.com/duckdb/duckdb/issues/19190) ("Missing results for index lookup"). That report describes a non-deterministic variant resolved by `SET threads=1`; the durable variant here is not affected by the thread count, so it appears to be the same family with a persistent outcome.

## Scope

**Development only.** DuckDB is the local/dev backend; PostgreSQL is the production one. PostgreSQL's indexes are unaffected and remain necessary there, so production behavior is unchanged.

## Change

Drop the declared secondary indexes on the DuckDB backend during `init_database()`, for both the chat-history and agent-portal stores. Dropping them also repairs a database that is already affected, so existing local databases heal on the next startup.

**Why drop rather than rebuild.** A rebuild repairs the rows that exist today but leaves the same failure mode armed for tomorrow. Local datasets are small enough that a full scan is cheap, so these indexes buy little and are the only surface the bug needs.

One implementation note worth flagging for review: `Index.drop(checkfirst=True)` **cannot** be used here. It probes for the index via reflection, which duckdb-engine does not implement, so it skips every drop while reporting success. The DDL is emitted directly instead. This was caught by a test that asserted through `inspect().get_indexes()` -- which always returns `[]` on duckdb-engine and therefore passed vacuously. The tests now read `duckdb_indexes()` directly.

## Repair script

`scripts/repair_duckdb_indexes.py` repairs a file out of band, for inspecting or fixing a database without starting the application:

```bash
python scripts/repair_duckdb_indexes.py data/chat_history.db          # drop
python scripts/repair_duckdb_indexes.py data/chat_history.db --check  # inspect only
python scripts/repair_duckdb_indexes.py data/chat_history.db --rebuild
```

It copies the database before making changes. The application must be stopped first -- DuckDB holds an exclusive lock on the file.

## Testing

- New `atlas/tests/test_duckdb_index_dropping.py`: indexes are dropped on DuckDB, rows stay queryable afterwards, the helper is a no-op on other backends, and the chat-history startup path leaves no indexes behind.
- Full suite: **2334 passed, 17 skipped**.
- Verified against a database exhibiting the fault: affected conversations returned 0 messages and one of two stored prompts before the change, and all conversations plus both prompts after it.


garland3:
## Summary
Works around a DuckDB index fault (silently empty result sets) by dropping all declared secondary indexes at startup for the DuckDB dialect in both the chat-history and agent-portal databases, and adds `scripts/repair_duckdb_indexes.py` plus admin docs for repairing existing files. The core mechanism is sound and covered by tests; the issues below are all in the operator-facing repair script.

## Verdict
Request changes.

## Required changes
1. Write the pre-repair backup to an already-ignored path at `scripts/repair_duckdb_indexes.py:75` — `data/chat_history.db.prerepair-<stamp>` is not matched by the `data/*.db` glob (confirmed with `git check-ignore`), so following the documented procedure leaves a full copy of user emails and conversation text untracked in the working tree where `git add -A` will stage it. Either keep `.db` last (`chat_history.prerepair-<stamp>.db`) or use `data/backups/`, and add the matching `.gitignore` entry in this PR.
2. Take the backup only after `duckdb.connect` succeeds at `scripts/repair_duckdb_indexes.py:83` — the most likely failure is the app holding the exclusive lock, so today a locked run prints `backed up to ...`, then fails, and every retry accretes another full copy taken from underneath a live writer.
3. Put `--check` and `--rebuild` in a mutually exclusive argparse group at `scripts/repair_duckdb_indexes.py:60` — `--check` returns at line 92 before any DDL, so `--check --rebuild` silently performs no rebuild while producing output indistinguishable from a plain check.
4. Validate the `sql` column before re-executing it in the `--rebuild` path at `scripts/repair_duckdb_indexes.py:104` — that string comes from the database file being repaired, not from this codebase, so require it to match a `CREATE INDEX` shape for the index and table named in the same row.

## Evidence
Evidence: 8/10 — proportionate for a ~325-line non-visual change: written failure model, four new tests, and a full-suite result. The most important missing artifact is the raw before/after query output against the real affected database; the central corruption claim currently rests on one prose sentence, and no test exercises the fault itself.

## Notes
- `dropped` counts `DROP INDEX IF EXISTS` statements, not actual drops, so the startup log reports the same count on every boot forever — `atlas/core/duckdb_indexes.py:52`.
- The drop is gated on dialect alone, and DuckDB is the default backend, so a deployment that simply never sets `CHAT_HISTORY_DB_URL` runs unindexed — `atlas/modules/chat_history/database.py:94`.
- `--rebuild` prints `rows are now visible to lookups` without noting it re-arms the fault and is undone at next startup — `scripts/repair_duckdb_indexes.py:111` and `docs/admin/chat-history.md:88`.

---

garland3:
Addressed the review feedback in 7e449c1.

**Required changes**
1. **Backup path is ignored** — the copy is now `<stem>.prerepair-<stamp><suffix>` (e.g. `data/chat_history.prerepair-20260817-222454.db`), so the `.db`/`.wal` suffix stays last and `data/*.db`/`data/*.wal` keep matching it. Added `data/*.prerepair-*` to `.gitignore` to cover a database path with no suffix; verified with `git check-ignore -v` on the `.db`, `.wal`, and suffixless names.
2. **Backup after the connection succeeds** — `duckdb.connect` runs first; the copy is taken inside the open connection, before any DDL. A locked run now prints only the "stop the application first" error and leaves no copy. Covered by `test_no_backup_when_the_database_cannot_be_opened`.
3. **`--check`/`--rebuild` mutually exclusive** — argparse now rejects the combination with exit code 2 instead of silently skipping the rebuild.
4. **`sql` validated before re-execution** — `_validated_create_index()` requires a single `CREATE [UNIQUE] INDEX [IF NOT EXISTS]` statement naming the same index and table as the row it came from; anything else aborts the rebuild before any index is dropped (exit 1, file untouched). Unit tests cover the accepted shapes and the rejected ones (`DROP TABLE`, a smuggled second statement, mismatched index/table names).

**Notes**
- **`dropped` count** (also raised by Copilot) — the helper now reads `duckdb_indexes()` first and drops only indexes confirmed present, so the count and log line describe drops rather than statements issued; a second startup reports 0. If that introspection query fails it falls back to dropping every declared index, since the repair matters more than the count.
- **`--rebuild` messaging** — the success line and the admin doc now state that a rebuild re-arms the same failure mode and is undone at the next startup: a diagnostic option, not a repair.
- **Dialect gating** — documented in `docs/admin/chat-history.md`: any deployment left on the default DuckDB backend runs without these indexes, which is one more reason production should point at PostgreSQL. Left as-is rather than adding a second gate, since running unindexed is the intended behavior on DuckDB regardless of how it was selected.

Also added the missing `## [Unreleased]` CHANGELOG entry flagged by Codex.

Not addressed: the evidence gap on the fault itself. The corruption is not reproducible on demand, so there is still no test that exercises it and the before/after observation remains prose. Raw query output from the affected database can be attached if that is wanted before merge.

Verification: `bash test/atlas_tests.sh` — 2348 passed, 17 skipped. `bash scripts/check-docs.sh` — OK. Manual run of the script against a scratch database exercising `--check`, `--rebuild`, plain drop, and the rejected `--check --rebuild`.

garland3:
## Summary
Works around a DuckDB index corruption fault by dropping declared secondary indexes at startup for both the chat-history and agent-portal stores, and adds an operator script (`scripts/repair_duckdb_indexes.py`) with `--check` / `--rebuild` modes to repair an existing database file out of band.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — the change is well demonstrated for a backend workaround: precise failure model in the description, two new test files covering the mechanism and the script's failure paths, and full-suite results quoted for both revisions. The one gap worth closing is the original fault itself, which is asserted in prose rather than shown; pasting the raw before/after `SELECT` row counts and the `duckdb_indexes()` listing from the affected file would turn the central claim into an observation.

## Notes
- A refused `--rebuild` returns at `scripts/repair_duckdb_indexes.py:156` after the backup at `:143` has already been written, so each retry leaves another full copy of conversation text and user emails on disk; move the backup below the validation loop.
- Identifiers read back from `duckdb_indexes()` are interpolated into `DROP INDEX "{name}"` and `select count(*) from "{table}"` (`:159`, `:91`) without doubling embedded quotes — low risk since it needs a tampered database file, but worth a `_quote_ident` helper at both sites.
- When the introspection query raises, `atlas/core/duckdb_indexes.py:70` falls back to the declared set and counts `DROP INDEX IF EXISTS` statements issued, so an already-clean database logs "Dropped N secondary index(es)" on every boot; the `--rebuild` success path and this fallback branch are both untested.

---

PR #818, opened 2026-08-18
fix(tests): remove shared state and order dependence from the suite

garland3:
## Why

The Python suite was green in its default order only. Auditing it for isolation turned up four leaks; each is reproduced below before its fix.

### 1. `sys.modules` fake for the LiteLLM caller broke 86 tests, order-dependently

`test_capability_tokens_and_injection.py` replaced `sys.modules["atlas.modules.llm.litellm_caller"]` with a fake **at import time**, restoring it only from a module-scoped fixture. Anything imported while the fake was installed bound `_FakeLLM` for the rest of the session:

```
pytest tests/test_capability_tokens_and_injection.py tests/test_llm_retry.py
  -> ERROR collecting: ImportError: cannot import name 'MAX_LLM_RETRIES'

pytest tests/test_capability_tokens_and_injection.py tests/test_rag_citations.py
  -> 33 failed: AttributeError: type object '_FakeLLM' has no attribute '_build_citation_instructions'
```

Same for `test_tool_error_attribution` (21), `test_strict_role_ordering` (13), `test_llm_user_api_key` (7), `test_rag_context_insert_position` (5) -- 86 tests total. Control (`test_health_route.py` first): 57 passed.

The full suite survived only because `test_llm_env_expansion.py` sorts alphabetically earlier and un-faked the module on import. That is one of **four** copies of the same workaround the pattern had spawned (`test_llm_customer_id_header`, `test_log_level_sensitive_data`, `test_hooks_integration`, whose docstring already documented the leak as "a pre-existing test-isolation quirk").

The stub was never needed -- importing `main` does not call an LLM. It is removed, and so are all four workarounds.

### 2. ProcessManager singleton leaked across tests (a failing test today)

The agent-portal e2e fixture cleared `pm_mod._singleton_manager`; the real global is `_singleton`, so the assignment silently created a stray attribute and the process manager -- with every process it tracked -- survived from test to test.

```
pytest tests (reverse item order)   -> 1 failed: test_launch_list_get_cancel_round_trip
pytest tests (random item order)    -> same failure
  test_agent_portal_e2e.py:101: AssertionError: assert [{'id': '8266...}] == []
```

`test_agent_portal_groups.py` already did this correctly, which is what made the typo visible.

### 3. The suite wrote into real developer state

`AppFactory()` is constructed at import time in `atlas/infrastructure/app_factory.py` and calls `get_session_factory()`, so importing `main` (about 30 test modules do) bound the process-wide engine to the repository's `data/chat_history.db`. Measured on this checkout:

| Store | Damage |
| --- | --- |
| `data/chat_history.db` | `test_security_header_injection` inserted a conversation row per run; 318 had accumulated, all `test@test.com` / `attacker@evil.com` |
| `data/agent_portal.db`, `data/agent_portal_audit.jsonl` | 1204 audit rows, 1181 of them `e2e@test.com`; +2850 bytes per run |
| `runtime/feedback/` | +1 real feedback JSON per run, from a route-registration test that skipped the temp-dir fixture its neighbours use |
| `logs/security_high_risk.jsonl` | `test_prompt_risk_and_acl` deleted and rewrote it unconditionally |
| `config/secure/` | created by the MCP token store |

`conftest.py` now redirects `CHAT_HISTORY_DB_URL`, `AGENT_PORTAL_DB_URL`, `AGENT_PORTAL_AUDIT_PATH`, `RUNTIME_FEEDBACK_DIR`, `RUNTIME_CAPTURE_DIR` and `MCP_TOKEN_STORAGE_DIR` into a per-session temp dir, before any test module is imported. They are unconditional assignments, not `setdefault`, so an exported shell value cannot point the suite back at a real database.

The one product change in this PR is in that table's last row but one: `prompt_risk.log_high_risk_event` now resolves its path through a new `high_risk_log_path()` that honors `APP_LOG_DIR`, matching `telemetry._tool_output_dir`, `otel_config`, `admin_routes` and `telemetry_routes`, which already do. Deployments that do not set `APP_LOG_DIR` are unaffected.

### 4. Leaked env vars, sys.path and mocks

- `GLOBUS_AUTH_BASE_URL` (set at module import in the Globus integration test) and `ATLAS_ENV_FILE` (set on `os.environ` by a helper) were both still set at session end -- verified with a session-finish probe. `ATLAS_ENV_FILE` pointed at a `tmp_path` pytest later deletes, and every subprocess-spawning CLI test inherited it. Both now go through `monkeypatch`.
- `test_env_demo_server` inserted a CI-only path (`/home/runner/...`) into `sys.path` permanently and set/deleted env vars by hand, so the cleanup was skipped whenever an assertion above it failed.
- `TestGroupRestrictedModels._deps` called `patcher.start()` three times -- one of them on the shared `atlas.core.model_access.is_user_in_group` -- before its caller's `try` block existed. It is a fixture now, so teardown always runs.

## What else is in here

- An autouse `_isolate_module_singletons` fixture snapshotting and restoring 17 app singleton module globals (chat-history engine, process manager, portal/preset stores, hook manager, token storage, ...). 39 tests were leaving at least one of these populated; a leaked `HookManager` in particular makes later turns fire hooks they never asked for.
- Two guard tests in `test_env_isolation.py`: every redirected store resolves outside the checkout, and every entry in `_SINGLETON_GLOBALS` names an attribute that actually exists (the failure mode behind #2).
- `docs/developer/test-isolation.md` -- what conftest guarantees, the rules for test authors, and how to check for order dependence by hand.

## Verification

Measured at head (`fbdc988`). Counts moved 2334 -> 2371 as `main` merged in (#816) and this PR added its own tests. Reproduce the reordered runs with the plugin in the diff:

```bash
cd atlas
ROOT=$(git rev-parse --show-toplevel); export PYTHONPATH="$ROOT:$ROOT/scripts"
ATLAS_TEST_ORDER=reverse python -m pytest tests -q -p pytest_test_order
ATLAS_TEST_ORDER=1 python -m pytest tests -q -p pytest_test_order
ATLAS_TEST_ORDER=6 ATLAS_TEST_ORDER_SCOPE=test python -m pytest tests -q -p pytest_test_order
```

| Run | Before | After (head `fbdc988`) |
| --- | --- | --- |
| default order | 2334 passed, 17 skipped | 2371 passed, 17 skipped |
| reverse order | **1 failed**, 2333 passed | 2371 passed, 17 skipped |
| file shuffle (`ATLAS_TEST_ORDER=1`) | pass | 2371 passed, 17 skipped |
| test shuffle (`ATLAS_TEST_ORDER=6 SCOPE=test`) | **1 failed**, 2333 passed | 2371 passed, 17 skipped |
| `capability_tokens` + `llm_retry` + `rag_citations` + `tool_error_attribution` | **1 collection error, 54 failed** | 111 passed, 1 skipped |
| files inside the repo touched by a run | `data/*.db`, `data/agent_portal_audit.jsonl`, `runtime/feedback/*`, `logs/security_high_risk.jsonl`, `config/secure/` | none |
| `/tmp/atlas-test-*` left after four consecutive runs | 8 | 0 |

CI: all 10 checks green on `cd1a3df`, `67da2f7` and `a668295`. The new ordering leg does real work in-container rather than no-oping — from the `a668295` run:

```
Run tests in reverse collection order
  Re-running the suite in reverse order (test-isolation guard)
  [pytest_test_order] reversed 2386 tests
  2358 passed, 28 skipped, 620 warnings in 75.50s
```

(Container counts differ from local: it skips a few environment-dependent tests and runs the frontend suite separately. The step is `continue-on-error: true`, so read its own conclusion in the run's step list rather than the job's.)

No behavior change outside tests other than the `APP_LOG_DIR` fallback described above, a warn-once (previously debug-only) log when the security-audit write fails, and a one-time notice when a stale copy of the log remains at the pre-move location.


garland3:
## Summary
Isolates the pytest suite from the checkout: a session temp directory now backs the persistent stores, an autouse fixture snapshots and restores the module-level singletons of 17 modules, and `prompt_risk.high_risk_log_path()` is factored out so `security_high_risk.jsonl` follows `APP_LOG_DIR`.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. Each of the four leaks is reproduced with a failing command before its fix and the whole-suite effect is measured across six run orderings, which is well above the bar for a change of this size; the one gap is that the reordering plugin behind those runs is not in the diff, so the before/after table cannot be reproduced from the PR.

## Notes
- `_STATE_TMPDIR` (and the pre-existing `_TELEMETRY_TMPDIR`) are never removed — add a `pytest_sessionfinish` `rmtree` or use `tmp_path_factory`, else every run orphans a directory in `/tmp` (`atlas/tests/conftest.py:52`).
- Restoring `_engine` / `_idle_sweeper_task` by bare `setattr` drops live resources without `dispose()`/`cancel()`; give lifecycle-owning entries a per-entry teardown (`atlas/tests/conftest.py:186`).
- `high_risk_log_path()` reads `APP_LOG_DIR` from the environment only, unlike `telemetry_routes.py:163` and `admin_routes.py:150`, which also fall back to `app_settings.app_log_dir` — a deployment configuring it via `.env` splits the security log from the others, and the relocation is not noted in `docs/telemetry/README.md`.

---

garland3:
Thanks -- all three nits addressed in c0e8205, plus the reproducibility gap you flagged.

**Orphaned temp directories.** `pytest_sessionfinish` now `rmtree`s both `_STATE_TMPDIR` and the pre-existing `_TELEMETRY_TMPDIR`. `tmp_path_factory` is not an option here: the store redirects have to be in place before any test module imports app code (`AppFactory()` opens the chat-history engine at import), and pytest's factory does not exist that early. Verified by running the suite four times back to back -- `/tmp/atlas-test-*` count went 9 -> 9, where it previously grew by two per run.

**Lifecycle-owning entries.** `_isolate_module_singletons` now calls `_release()` on the value it is about to drop, but only when the test actually replaced one: `dispose()` for a SQLAlchemy engine, `cancel()` for a live (not already-done) asyncio task, wrapped so a half-torn-down object from a failing test cannot raise a second, confusing error. Restoring is still snapshot-based; only the discarded object is released.

**`app_settings.app_log_dir` fallback.** Good catch -- I mirrored `telemetry._tool_output_dir` (env only) rather than `telemetry_routes._log_base_dir` / `admin_routes._log_base_dir` (env, then setting). `high_risk_log_path()` now does env -> `app_settings.app_log_dir` -> repository `logs/`, with a lazy import so a config failure cannot break risk logging. The test now asserts all three levels of that order. I did not touch `docs/telemetry/README.md`: it documents the OpenTelemetry span/log files, and `security_high_risk.jsonl` is not among them -- the resolution rule is documented on the function and in `docs/developer/test-isolation.md`. Happy to add a line there if you would rather have it in both places.

**Reproducibility.** Fair -- the before/after table came from a scratch plugin that was not in the diff. It ships now as `scripts/pytest_test_order.py`, inert unless `ATLAS_TEST_ORDER` is set, with the exact invocations in `docs/developer/test-isolation.md`:

```bash
cd atlas
ROOT=$(git rev-parse --show-toplevel); export PYTHONPATH="$ROOT:$ROOT/scripts"
ATLAS_TEST_ORDER=reverse python -m pytest tests -q -p pytest_test_order
ATLAS_TEST_ORDER=1 python -m pytest tests -q -p pytest_test_order
ATLAS_TEST_ORDER=6 ATLAS_TEST_ORDER_SCOPE=test python -m pytest tests -q -p pytest_test_order
```

All four orderings: 2335 passed, 17 skipped. Note the plugin reorders execution, not imports -- the import-time leak in this PR needs the pairwise form (`pytest tests/test_capability_tokens_and_injection.py tests/test_llm_retry.py`), which is also documented.

garland3:
## Summary
Isolates the test suite from developer state: redirects persistent stores and the log directory to temp locations, adds an autouse fixture that snapshots and restores 17 module-level singletons, adds a guard test plus a reordering plugin, and makes the high-risk prompt-risk log path resolve from `APP_LOG_DIR` instead of a fixed repository path.

## Verdict
Request changes.

## Required changes
1. Release singletons by type rather than by method name in `_release()` at `atlas/tests/conftest.py:186` — `ProcessManager.cancel` is `async def cancel(self, process_id, *, sigkill_after=3.0)`, so the call raises `TypeError` that the bare `except Exception: pass` swallows; the one singleton owning real OS resources (live subprocesses and pty fds) is never released, which is the motivating case in the PR description.
2. Force `APP_LOG_DIR` like the six store redirects at `atlas/tests/conftest.py:32` instead of `setdefault` — `docs/getting-started/installation.md:129` tells developers to set it to a path inside the checkout, so anyone following the documented setup now fails the new guard at `atlas/tests/test_env_isolation.py:85` and the suite writes telemetry and `security_high_risk.jsonl` into the repository anyway.
3. Resolve paths before the containment comparison at `atlas/tests/test_env_isolation.py:76` — `Path(value.replace("duckdb:///", ""))` is compared unresolved, so a `..` component or a symlinked `TMPDIR` pointing back into the checkout passes a guard whose whole job is to catch that. Also skip the path-shaped assertions for non-file URLs, since `_resolve_db_url` already accepts `postgresql://`.
4. Make `test_server_imports` actually import something at `atlas/tests/test_env_demo_server.py:11` — it still targets a hard-coded `/home/runner/work/...` path, and `spec_from_file_location` does not stat the file, so the test passes on every machine even when the module is absent or broken. Resolve relative to the repository root, `exec_module` it, and skip when the demo server is not present.
5. Reconcile the log-directory resolution at `atlas/core/prompt_risk.py:173` with the three existing implementations (`core/telemetry.py:277`, `routes/admin_routes.py:151`, `routes/telemetry_routes.py:161`) — the new docstring claims equivalence with `admin_routes._log_base_dir`, which reads only the setting and not the env var, so the documented invariant is false under runtime env changes. Extract one helper or correct the docstring.
6. Raise the failure log level for the security-audit write at `atlas/core/prompt_risk.py:242` from `debug` to `warning`, and resolve the base directory with `.expanduser().resolve()` — an unwritable or CWD-relative audit path currently produces no operational signal, making a misconfigured log indistinguishable from "no attacks observed".

## Evidence
Evidence: 9/10 — each defect is reproduced before its fix and the suite-level effect is measured across six orderings with a re-runnable plugin, which is more than adequate for a non-visual change of this size. The one gap worth closing is the verification table, which reflects the earlier commit rather than the head; the temp-directory cleanup claim is also prose-only.

## Notes
- Await the cancelled `_idle_sweeper_task` after `cancel()` (`atlas/tests/conftest.py:229`) — a request alone never unwinds the coroutine once the owning loop is closed.
- Singletons for modules first imported *inside* a test escape the fixture entirely (`atlas/tests/conftest.py:213`); nine of the seventeen listed modules are absent at collection time under subset runs.
- `security_high_risk.jsonl` now follows `APP_LOG_DIR` but is not listed in the operator log layout (`docs/telemetry/README.md:32`).

---

garland3:
Round 2 addressed in af2cbcb + d7daea9, and the branch is merged with main (that is also why no CI ran on c0e8205 — the PR had gone `DIRTY` against #816, so GitHub never scheduled it; it is `MERGEABLE` and running again now).

**1. `_release()` dispatching by method name.** Correct and my bug. `ProcessManager.cancel` is `async def cancel(self, process_id, *, sigkill_after=3.0)`, so `value.cancel()` raised `TypeError` straight into the bare `except`. Dispatch is by type now — `Engine.dispose()`, `asyncio.Task.cancel()` — and anything else is dropped, which is what plain restoration already did. I did **not** add ProcessManager teardown: its shutdown path is async and there is no loop to run it on at that point, and the agent-portal fixtures already cancel the processes they launch. If you want the manager's subprocesses reaped by the fixture too, that wants an async teardown and is worth its own change.

**2. `APP_LOG_DIR` `setdefault`.** You are right, and this is the sharper form of the bug: `docs/getting-started/installation.md:129` tells contributors to set it to a path inside the checkout, so the documented local setup both failed the new guard and kept writing telemetry and `security_high_risk.jsonl` into the repository. Forced now, like the six store redirects.

**3. Unresolved containment comparison.** Fixed — the guard `.resolve()`s before comparing (a `..` component or a symlinked `TMPDIR` would have sailed through a check whose only job is to catch that), compares against a resolved project root, also rejects the root itself, and skips path assertions for server-backed URLs since `_resolve_db_url` only rewrites the `duckdb:///` form.

**4. `test_server_imports` importing nothing.** Correct — `spec_from_file_location` does not stat, and nothing called `exec_module`, so it passed on every machine. It now resolves `atlas/mcp/env-demo/main.py` from the repository root, `exec_module`s it, asserts the module exposes `mcp`, and skips when the demo server is absent.

**5. Log-directory resolution.** Half-accepted. The docstring was wrong and is corrected: the order is env → setting → repo `logs/`, which matches `telemetry_routes._log_base_dir`; `admin_routes._log_base_dir` reads only the setting and `core/telemetry._tool_output_dir` only the env var, and the docstring now says so. I did not extract a shared helper: three of the four call sites would change behavior (`admin_routes` would start honoring the env var, `_tool_output_dir` the setting), which is a runtime change for operators that does not belong in a test-isolation PR. Worth doing as a follow-up.

**6. Failure log level and path resolution.** `expanduser().resolve()` applied. On the level: `log_high_risk_event` runs per event, so an unconditional `warning` turns a persistently unwritable path into a flood in the log it is supposed to make legible. It warns **once** now — with the resolved path in the message — and drops to debug afterwards, the same warn-once shape as `_HMAC_WARNED` in `core/telemetry.py`. That gives the operational signal you are after without the spam.

**Notes.** Awaiting the cancelled `_idle_sweeper_task`: not possible from a sync teardown once the owning loop is closed, which is the normal case here; the comment now says the cancel is best-effort rather than implying the coroutine unwinds. Late-imported modules escaping the fixture: real, and fixed — a module absent at setup is snapshotted as `None`, the declared default of every entry in `_SINGLETON_GLOBALS`, so a singleton created by an import inside a test is cleaned up under subset runs too. `security_high_risk.jsonl` is now in the operator log layout at `docs/telemetry/README.md:35`.

**Verification table** refreshed to head in the PR description (counts moved 2335 -> 2353 with #816 merged in): default, reverse, file-shuffle and test-shuffle all 2353 passed / 17 skipped; the previously-broken four-file selection is 111 passed / 1 skipped; four consecutive runs left 0 `/tmp/atlas-test-*` directories behind.

garland3:
Inline review comments addressed in e5ad03b.

**CodeQL — unused global `_WRITE_FAILURE_WARNED`** (`prompt_risk.py:242`): the flag was read, but only inside the function that declares `global`, which is the shape that query flags. Replaced with a module-level `set` of paths already warned about — no global rebinding, and it is better behavior: a log location that is repaired and later breaks again warns again instead of staying silent for the life of the process. The warning also names the resolved path it failed to write.

**Copilot — CHANGELOG entry too long** (`CHANGELOG.md:10`): agreed, AGENTS.md:75 asks for 1-2 lines. Cut to a single summary bullet pointing at `docs/developer/test-isolation.md`, where the detail belongs.

**Copilot — doc implies guaranteed alphabetical ordering** (`test-isolation.md:9`): fair. Reworded to say collection order is *not* a contract pytest makes and today's alphabetical order is incidental — which is the point, since that is what a leak hides behind.

**Copilot — `APP_LOG_DIR` still uses `setdefault`** (`conftest.py:31`): already fixed in af2cbcb, before this comment landed. `conftest.py:39` is now `os.environ["APP_LOG_DIR"] = _TELEMETRY_TMPDIR`, forced like the six store redirects, with the reason (installation.md tells contributors to point it inside the checkout) in the comment above it.

Full suite after these changes: 2353 passed, 17 skipped.

garland3:
## Summary
Makes the pytest suite order-independent by adding import-time environment
redirects and a singleton-isolation fixture to `atlas/tests/conftest.py`, ships a
`scripts/pytest_test_order.py` reordering plugin plus docs, and moves
`security_high_risk.jsonl` behind a new `high_risk_log_path()` resolver that
honors `APP_LOG_DIR`.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — each of the four leaks is shown as a failing command with its
output before its fix, and the suite-level effect is tabulated across six
orderings using the plugin that now ships in the diff. The thin spot is the
round-2 additions: `_release()`, the not-yet-imported snapshot branch, and the
warn-once write-failure path have no test; a test asserting one WARNING then
DEBUG across two failed writes is the single most useful thing to add.

## Notes
- `log_high_risk_event`'s `except` block re-calls `high_risk_log_path()`, which
  can itself raise (`APP_LOG_DIR=~nosuchuser/logs`), so the never-raising audit
  sink propagates and aborts the per-chunk risk scan — `atlas/core/prompt_risk.py:242`.
- `_WRITE_FAILURE_WARNED` latches for the process life and is not in
  `_SINGLETON_GLOBALS`; a transient failure burns the only warning and leaks
  across tests — `atlas/core/prompt_risk.py:173`.
- Resolution order documented at `docs/telemetry/README.md:39` is the reverse of
  what the new resolver does (env var wins, not the setting), and the relocation
  of `security_high_risk.jsonl` has no operator-facing upgrade note.

---

garland3:
Round 3 addressed in c63a33e. Two of the three notes were already fixed in e5ad03b, before the review landed — details below.

**Evidence gap: the round-2 additions had no tests.** Agreed, and that is the same gap that let the original defects sit. `tests/test_isolation_fixtures.py` now covers all three, and I checked each test against the *pre-fix* behavior rather than assuming it discriminates:

| test | fails on old code with |
| --- | --- |
| `_release` never calls a domain method | `calls == ['dispose']` — the name-based dispatch called it |
| snapshot clears a singleton created by an import inside the test | `'created-during-test'` survives restore |
| one WARNING then DEBUG across two failed writes | `[DEBUG, DEBUG]` |
| a different path warns again | only one WARNING total |

Also covered: Engine disposed, pending Task cancelled, finished Task left alone, a raising `dispose()` swallowed, and a successful write logging nothing. `snapshot_singletons`/`restore_singletons` are module-level helpers now so they can be exercised directly instead of only through the autouse fixture.

**Note 1 — the `except` block re-calls `high_risk_log_path()`, which can itself raise.** Already gone in e5ad03b: the handler formats a local `failed_path` string captured before the write (`log_path` is initialized to `None` above the `try`, so a resolver that raises yields `"<unresolved>"`). Nothing in the handler can raise now, and the audit sink still never propagates.

**Note 2 — `_WRITE_FAILURE_WARNED` latches for the process and is not isolated.** The flag is gone in e5ad03b, replaced by `_WRITE_FAILURE_WARNED_PATHS`, a set keyed by resolved path — so a transient failure no longer burns the only warning for the life of the process, and a location that is repaired and later breaks again warns again. It is deliberately not in `_SINGLETON_GLOBALS`: that fixture snapshots by reference, which would restore the same mutable set object and undo nothing. The new tests reset it explicitly via `monkeypatch.setattr`, which is honest about the coupling.

**Note 3 — documented resolution order is reversed, and no upgrade note.** Fixed in `docs/telemetry/README.md`. The existing paragraph now scopes itself to `spans.jsonl` / `app.jsonl` / `tool_outputs/`, and `security_high_risk.jsonl` gets its own: env var → setting → `<project_root>/logs/`, matching `routes/telemetry_routes._log_base_dir`, so an exported env var takes effect without a config reload. It also records that the four implementations in the tree disagree (`otel_config` and `admin_routes` read the setting first, `core/telemetry` only the env var) and that unifying them is a behavior change for those call sites, not this PR. Plus the operator upgrade note you asked for: the file used to ignore `APP_LOG_DIR` entirely, so anyone with it set will find the log in the new location.

Verified at c63a33e: default, reverse and test-shuffle orders all 2364 passed / 17 skipped; the new file passes alone; four consecutive runs leave 0 `/tmp/atlas-test-*` directories.

garland3:
CodeQL's two "statement has no effect" annotations on `test_isolation_fixtures.py` (2197/2198) are its usual reading of a bare `await task` statement — the awaits are load-bearing, but the alerts are cheap to make go away honestly rather than dismiss. Fixed in eea6a27: the finished-task case now asserts `await task is None` (a real assertion — that is `asyncio.sleep(0)`'s result — rather than a suppression), and the cancelled case binds the result inside `pytest.raises`.

Suite unchanged at 2364 passed / 17 skipped.

garland3:
## Summary
Removes shared state and order dependence from the Python test suite: drops an import-time `sys.modules` fake for the LiteLLM caller and its four workarounds, adds a module-singleton snapshot/restore fixture and out-of-tree store redirects in `conftest.py`, and ships an ordering plugin plus isolation guard tests. One product change rides along — `security_high_risk.jsonl` now resolves through `APP_LOG_DIR` → `app_settings.app_log_dir` → `<project_root>/logs`.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10 — each leak is reproduced with its actual failing output before its fix, the suite-level effect is measured across six run configurations, and the reordering plugin behind those numbers ships in the diff. The main gap is that the headline verification table is measured at `d7daea9` rather than at head `c63a33e`, and the file-shuffle and four-file pairwise runs (the ordering that caught leak #1) are not re-reported at head.

## Notes
- The write-failure warn-once key collapses every resolver failure onto the constant `"<unresolved>"`, so a second bad `APP_LOG_DIR` is logged only at debug — `atlas/core/prompt_risk.py:244`.
- `test_restores_a_default_for_a_module_that_lacks_the_attribute` asserts the opposite of its docstring (the unconditional `delattr` only passes if restore *did* create the attribute), and its cleanup is not in a `finally` — `atlas/tests/test_isolation_fixtures.py:140`.
- `docs/getting-started/installation.md:129` still tells contributors to set `APP_LOG_DIR` inside the checkout, which is the recommendation `conftest.py:36` cites as the reason it overrides the variable.

---

garland3:
Round 4 addressed in cd1a3df.

**Warn-once key collapsing onto `"<unresolved>"`.** Real: when the resolver itself raises there is no path to key on, so every misconfiguration shared one entry and the second bad `APP_LOG_DIR` only reached debug. It now keys on the configured value in that case (`<unresolved:{APP_LOG_DIR}>`), with a test that two distinct unresolvable configs each produce a WARNING.

**`test_snapshot_skips_attributes_a_module_does_not_have` asserted the opposite of its docstring.** Correct, and the docstring had the right behavior: restore should not invent a global on a module that never declared one. It did, because `restore_singletons` called `setattr` unconditionally — so the test's unconditional `delattr` passed only because of the bug it was meant to catch. Fixed on both sides: `restore_singletons` skips a module lacking the attribute (every real entry declares its default at import, which `test_env_isolation` asserts), and the test now asserts `not hasattr(...)` with cleanup in a `finally`.

**`installation.md` still recommends a checkout-local `APP_LOG_DIR`.** Left the recommendation — it is correct for running the app, and `logs/` is gitignored — but the contradiction you spotted was that nothing said why the suite overrides it. That line now notes the suite substitutes a temp directory, so the recommended value cannot pollute test runs, and links to the isolation doc.

**Verification table** re-measured at head (`cd1a3df`) and updated in the description, including the file-shuffle and four-file pairwise runs you flagged as missing:

| Run | Before | After |
| --- | --- | --- |
| default | 2334 passed | 2365 passed |
| reverse | **1 failed**, 2333 passed | 2365 passed |
| file shuffle (seed 1) | pass | 2365 passed |
| test shuffle (seed 6) | **1 failed**, 2333 passed | 2365 passed |
| four-file pairwise | **1 collection error, 54 failed** | 111 passed, 1 skipped |
| `/tmp/atlas-test-*` after four runs | 8 | 0 |

Counts moved 2334 -> 2365 because `main` merged in (#816) and this PR added its own tests.

garland3:
## Summary
Makes the test suite order-independent and stops it writing into the developer's checkout: persistent stores and `APP_LOG_DIR` are redirected to a session temp directory, an autouse fixture snapshots and restores 17 module-level singletons, a session-finish hook cleans up temp dirs, and a reordering plugin plus guard tests and docs are added. `security_high_risk.jsonl` is also moved onto the `APP_LOG_DIR` -> `app_settings.app_log_dir` -> `<project_root>/logs/` resolution chain.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size: every leak is shown with real failing output before its fix, and before/after counts are measured at the head commit across six run configurations with the reordering harness shipped in the diff. The one missing artifact is a stated CI result on `cd1a3df`; the numbers are local-only.

## Notes
- `CHANGELOG.md:10` describes this as test-suite-only, omitting that `logs/security_high_risk.jsonl` now moves with `APP_LOG_DIR` — an operator-visible relocation of a security audit log that collectors may still be tailing at the old path.
- `docs/telemetry/README.md:41` contradicts itself: `tool_outputs/` is grouped under the setting-then-env order, but `atlas/core/telemetry.py:277` reads the env var only, as the next sentence says.
- The redirected stores are one file per *session*, not per test, so count/emptiness assertions can still be affected by earlier tests (`atlas/tests/conftest.py:60`); worth stating that limit in `docs/developer/test-isolation.md`.

---

garland3:
`build-and-test` went red once on cd1a3df and passed on rerun (all 10 checks green now). The failure was **not** from this branch's changes:

```
FAILED tests/test_calculator_tool.py::test_resource_bounds_reject_quickly
E   AssertionError: pow(9, 99999999) took 1.55s; guard ran too late
E   assert 1.549195667000049 < 1.0
```

That is the wall-clock assertion at `test_calculator_tool.py:219`, which nothing in this PR touches. Every earlier push on this branch passed the same job (acdc339, d387915, d7daea9, e5ad03b, c63a33e, eea6a27), and the test runs in 0.27s locally.

Worth flagging as a real flake rather than just re-running it, though. The assertion's intent is "the guard must reject before evaluating", and the margin it actually needs is enormous: I timed `pow(9, 99999999)` **unguarded** locally and killed it after 2 minutes without a result. So the failure mode it is protecting against is ~120s+, while the budget is 1.0s — on a shared runner under load, a rejection that took 1.55s is still three orders of magnitude away from "guard ran too late". Two options, whichever you prefer:

- raise the budget to something like 10s, which still fails loudly if the guard stops firing; or
- drop the wall clock and assert on the guard directly (e.g. that the rejection names the bound it hit), which is what the docstring actually cares about.

I have left it alone: it is outside this PR's scope, and re-tuning a security guard's test threshold deserves its own review. Happy to do either as a follow-up.

garland3:
Round 5 addressed in 67da2f7 — all three were documentation, all correct.

**CHANGELOG omitted the operator-visible relocation.** Added as a second bullet: `security_high_risk.jsonl` used to resolve to `<project_root>/logs/` unconditionally and now follows `APP_LOG_DIR`, with the note that a collector still tailing the repository path needs moving. That is the one product change in this PR and it belongs in the changelog rather than only in the telemetry doc.

**`tool_outputs/` grouped under the wrong resolution order.** My error, and self-contradicting two paragraphs apart. `core/telemetry._tool_output_dir` reads only `APP_LOG_DIR`, so the first paragraph now covers `spans.jsonl` / `app.jsonl` only, and `tool_outputs/` is called out explicitly: a directory configured solely through `app_settings.app_log_dir` leaves tool outputs behind in `<project_root>/logs/`. The list of which implementation reads what is now complete rather than a partial "the four disagree".

**Per-session, not per-test.** Correct and worth stating plainly, so `docs/developer/test-isolation.md` now says it: one temp directory backs the whole run, so the redirects stop real-state and cross-*run* pollution but do not hand each test an empty store — a test asserting emptiness or counting rows still sees what earlier tests in the session wrote, and should build its own store against `tmp_path` and pin it (as the agent-portal fixtures do) or reset the relevant singleton.

**CI result on `cd1a3df`**, which the evidence section noted was missing: all 10 checks green, `build-and-test` in 11m03s. It went red once on the first attempt with the pre-existing `test_calculator_tool` wall-clock flake — details and a suggested fix in the comment above — and passed on rerun.

Suite unchanged at 2365 passed / 17 skipped.

garland3:
## Summary
Makes the test suite order-independent by isolating module-level singletons, env vars and store paths in `conftest.py`, adds a `ATLAS_TEST_ORDER` reordering plugin, and reworks `high_risk_log_path()` so the security audit log follows the configured log directory.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is thoroughly demonstrated — each leak is reproduced with real failing output and the suite effect is measured across six run configurations with the reordering harness included in the diff. The only gap is that the verification table and the green-CI claim are labeled at `cd1a3df`, two commits behind head `67da2f7`; a one-line note or a re-run at head would close it.

## Notes
- Relocating `security_high_risk.jsonl` to follow `APP_LOG_DIR` silently moves the sink for deployments that already export it; a one-time log line naming old and new paths would stop a stale collector from reading as "no attacks" (`atlas/core/prompt_risk.py:194`).
- `_WRITE_FAILURE_WARNED_PATHS` is never cleared on a successful write, so the documented "fixed then broken again warns again" behavior does not hold — discard the key after the write, or correct the comment (`atlas/core/prompt_risk.py:174`).
- The order-independence claim is enforced nowhere: `ci.yml` runs default collection order only, so one non-blocking `ATLAS_TEST_ORDER=reverse` leg would keep the guarantee from decaying (`.github/workflows/ci.yml`).

---

garland3:
Round 6 addressed in 50e356c. Two of the notes were defects in my own round-5 work.

**`_WRITE_FAILURE_WARNED_PATHS` never cleared on success.** Correct, and it made my own comment false: I wrote that a repaired path warns again, but nothing ever discarded the key, so the first failure silenced that path for the life of the process. The key is now dropped after each successful append, and a test drives one path through broken -> repaired -> broken and asserts two WARNINGs. (Failure is injected by shadowing the module-global `open`, not by `chmod`, so it behaves the same in containers that run as root.)

**Silent relocation of the sink.** Fair — the docs upgrade note only helps operators who read it, and a stale collector reads as "no attacks observed", which is the exact failure mode this log exists to prevent. There is now a one-time WARNING naming both paths, emitted only when a file actually exists at the legacy location, so a deployment that never had one stays quiet. It is guarded so a notice can never break logging.

**Nothing enforces the order independence.** Agreed, and this was the real durability gap: without a CI leg, the guarantee decays the first time someone reintroduces a leak, and the reordering harness becomes documentation nobody runs. `ci.yml` now re-runs the suite with `ATLAS_TEST_ORDER=reverse` against the already-built debug image (the plugin ships in `scripts/`, which `Dockerfile-test` already copies). Deliberately `continue-on-error: true` to start, with the reason in the step comment: it inherits the suite's existing timing flakes (the `test_calculator_tool` wall clock that went red on cd1a3df), and this leg exists to surface *isolation* regressions rather than gate on timing noise. Flip it to blocking once it has a clean track record — I would rather it be advisory and stay than blocking and get deleted after the first unrelated flake.

**Evidence — table two commits behind head.** Re-measured at `50e356c` and relabeled: default / reverse / file-shuffle / test-shuffle all 2366 passed, 17 skipped; four-file pairwise 111 passed, 1 skipped; 0 `/tmp/atlas-test-*` left. CI was green on both `cd1a3df` and `67da2f7`; `50e356c` is in flight and I will confirm the new leg's output when it lands.

garland3:
CodeQL flagged `_RELOCATION_NOTICE_EMITTED` (2199) with the same "unused global" rule that caught the earlier warn-once flag — I had reintroduced the pattern I'd already fixed once. Now a set of already-noticed paths, matching `_WRITE_FAILURE_WARNED_PATHS`, which also means a runtime `APP_LOG_DIR` change gets its own notice instead of being swallowed by a process-wide latch. Fixed in a668295, with the three tests the notice was missing (fires once when a stale log remains, silent when there is none, silent when nothing moved).

**CI at head a668295: all 10 checks green**, `build-and-test` 12m30s — which now includes the new leg, and it did real work rather than no-oping:

```
Run tests in reverse collection order
  Re-running the suite in reverse order (test-isolation guard)
  ...
  2358 passed, 28 skipped, 620 warnings in 75.50s
```

(Counts differ from the local 2369/17 because the container skips a few environment-dependent tests and runs the frontend suite separately.) Worth knowing when reading it: because the step is `continue-on-error: true`, its failure would not turn the job red — read the step's own conclusion in the run's step list, not just the check. That is the trade for keeping it out of the merge gate while it shares the suite's existing timing flakes.

This closes the last evidence gap: the verification table is now labeled at head and the CI result is stated, both for the normal legs and for the ordering guard itself.

garland3:
## Summary
Makes the pytest suite order-independent: redirects persistent stores (chat history, agent portal, audit log, feedback/capture, MCP tokens) off the repository checkout, snapshots and restores memoized module singletons between tests, cleans up session temp directories, adds a reverse/shuffle ordering plugin plus a non-blocking CI leg, and moves the prompt-injection audit log to the configured `APP_LOG_DIR`.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — unusually well documented for a change this size, with per-leak reproductions and a before/after table across six ordering configurations. The one real gap is that the headline numbers and the green-CI claim are labeled at `50e356c`, two commits behind head, and the newly added reverse-order CI leg has no reported run output; pasting its `[pytest_test_order] reversed N tests` line would close it.

## Notes
- `_release()` at `atlas/tests/conftest.py:237` — the comment claims a closed-loop task is "marked cancelled", but `Task.cancel()` raises `RuntimeError: Event loop is closed` into the bare `except`, leaving the task `PENDING`; fix the comment (or cancel the sweeper before the loop closes) and cover the closed-loop branch.
- `_warn_once_if_relocated` at `atlas/core/prompt_risk.py:229` records the path before checking `legacy.exists()`, so a stale log that appears after the first resolution is never announced — move the `add` inside the `if`.
- `high_risk_log_path()` (`atlas/core/prompt_risk.py:184`) reads as a pure resolver but stats the filesystem, mutates module state and can warn; also the fifth divergent copy of log-directory resolution. Moving the notice into `log_high_risk_event` and extracting a shared `log_base_dir()` are cheap follow-ups. A mistyped `ATLAS_TEST_ORDER` exits 1 via bare `SystemExit` (`scripts/pytest_test_order.py:49`) — `pytest.UsageError` would distinguish it from a test failure in the new CI leg.

---

garland3:
Round 7 addressed in fbdc988. All three notes were correct; I verified the asyncio one before touching it.

**`_release()` comment was wrong about closed loops.** Confirmed empirically:

```python
loop = asyncio.new_event_loop(); task = loop.run_until_complete(make_task()); loop.close()
task.cancel()   # RuntimeError: Event loop is closed
task._state     # 'PENDING'
```

So the comment's claim that this "marks it cancelled" was false. It now says what actually happens -- effective only while the loop is open, the swallowed RuntimeError is the honest outcome, and cancelling properly means doing it before the loop closes, which is the owning fixture's job rather than teardown's -- with a test covering the closed-loop branch (teardown must not raise).

**`_warn_once_if_relocated` recorded the path before checking existence.** Right, and it defeated the notice for the case that matters most: a stale file that appears *after* the first write (restored backup, rolled-back deploy) would never be announced. The `add` moved inside the `if`, with a test that emits once with no legacy file, creates one, emits again, and expects exactly one notice.

**`high_risk_log_path()` was not a pure resolver.** Fixed by moving the notice into `log_high_risk_event`, which also answers the cost objection: the extra stat is paid only on a turn that actually writes a record, not on every resolution. The docstring now states the function is pure. On extracting a shared `log_base_dir()` -- still declining here for the same reason as rounds 5 and 6: `admin_routes` would start honoring the env var and `_tool_output_dir` the setting, which is a runtime change for operators that has no business riding in a test-isolation PR. Happy to open a follow-up issue for it if you want it tracked.

**`SystemExit` -> `pytest.UsageError`** in the ordering plugin: a mistyped `ATLAS_TEST_ORDER` now exits **4** (usage) instead of **1** (tests failed), which is exactly the distinction the new CI leg needs. Verified both paths.

**Evidence.** Table re-measured and relabeled at `fbdc988`: default / reverse / file-shuffle / test-shuffle all 2371 passed, 17 skipped; pairwise 111 passed; 0 temp dirs left. The CI leg's own output is now quoted in the description:

```
[pytest_test_order] reversed 2386 tests
2358 passed, 28 skipped, 620 warnings in 75.50s
```

garland3:
Heads-up: the last round of review fixes did not make it into this merge. `fbdc988` was pushed at 12:52 UTC and the squash-merge was taken at `a668295` at 12:57 UTC, so `main` is missing all four:

- `high_risk_log_path()` still stats the filesystem, mutates module state and can log, rather than being the pure resolver the round-7 note asked for;
- the relocation notice still records the path before checking existence, so a stale log that appears later (restored backup, rolled-back deploy) is never announced;
- `_release()` still carries the comment I verified false — a closed-loop `cancel()` raises `RuntimeError: Event loop is closed` and the task stays `PENDING`;
- the ordering plugin still raises bare `SystemExit`, so a mistyped `ATLAS_TEST_ORDER` exits 1 (looks like a failing suite) instead of 4 (usage) — which matters now that this PR added a CI leg reading that exit code.

Opened as #820, cherry-picked onto `d97a0ab` with no new work: https://github.com/sandialabs/atlas-ui-3/pull/820

Verified there — default / reverse / test-shuffle all 2371 passed, 17 skipped; 0 temp directories left; ruff clean.

PR #819, opened 2026-08-18
feat(workspaces): make the workspace flag switch prompt, RAG, and tool context

garland3:
## Why

`FEATURE_WORKSPACES_ENABLED` has existed since the initial feature-flag scaffold and was wired end to end **as data** — env var, `AppSettings` field, and a `features.workspaces` key in both `/api/config` and `/api/config/shell` — but nothing ever read it. The only frontend occurrence was the `DEFAULT_FEATURES` literal in `useChatConfig.js`, so turning the flag on changed nothing in the UI. `.env.example` said as much: *"Workspace selector (none configured yet)"*.

## What a workspace is

A named bundle of the selections that define a chat context — **active prompt + RAG data sources + MCP tools** — so switching between `Work` / `Home` / `Project A` is one click instead of re-picking every selection.

Two deliberate choices:

- **Applying replaces, never merges.** Carrying leftovers from the previous context would silently widen which tools and data sources a chat can reach. It also makes "clear this list" expressible, which a merge cannot do.
- **The stored config is a closed shape.** The API rejects unknown keys (422) and the repository dedupes/trims/caps the lists (≤500 entries, ≤300 chars), so a workspace row cannot become arbitrary client storage. A corrupt row degrades to empty selections rather than breaking the list endpoint for every other workspace the user owns.

A workspace stores *keys*, not permissions — switching selects things the user is already authorized for, and a stale key is inert.

## Shape of the change

Persistence mirrors the custom prompt library end to end:

| Layer | File |
| --- | --- |
| Table + migration | `models.py` (`UserWorkspaceRecord`), `alembic/versions/003_add_user_workspaces.py` |
| Repository | `atlas/modules/chat_history/workspace_repository.py` — owner-scoped queries, email normalized at every entry point |
| REST | `atlas/routes/workspace_routes.py` — `GET/POST/PUT/DELETE /api/workspaces`, 404 when disabled |
| Gate | `AppSettings.workspaces_effective` |
| Snapshot / apply | `frontend/src/hooks/chat/useSelections.js` |
| CRUD hook | `frontend/src/hooks/useWorkspaces.js` |
| UI | `frontend/src/components/WorkspaceSelector.jsx`, rendered from `Header.jsx` |

`features.workspaces` in the config payload now reports **`workspaces_effective`**, not the raw flag: workspaces live in the chat-history database, so `FEATURE_WORKSPACES_ENABLED=true` with `FEATURE_CHAT_HISTORY_ENABLED=false` has nowhere to persist and the feature stays off — same pattern as `custom_prompts_effective`.

## Verification

- Backend suite: **2374 passed, 17 skipped** (`bash test/atlas_tests.sh`), including 24 new tests in `atlas/tests/test_workspaces.py` — repository CRUD, config normalization, owner isolation with email case folding, corrupt-JSON degradation, route feature gate, and the `workspaces_effective` matrix.
- Frontend: **567 passed** across 52 files, including the new `workspace-selector.test.jsx` (12) and 4 new `useSelections` cases covering snapshot/apply semantics. `eslint` clean on all touched files; `npm run build` succeeds.
- Live server on the flag: created a workspace with duplicate/blank tool keys and confirmed the stored config came back deduped and trimmed; unknown config key → 422; `PUT` with empty lists cleared them (no merge); double `DELETE` → 200 then 404. With `FEATURE_CHAT_HISTORY_ENABLED=false`, `features.workspaces` reported `false` and `GET /api/workspaces` returned 404.

## Notes for review

- Default stays **off** in `.env.example` and `docker-compose.yml`; this is opt-in.
- Adding a selection dimension later is a two-file change: `WorkspaceConfig` in the routes and `normalize_config` in the repository. That is the cost of the closed shape, and it is intentional.
- Docs: `docs/user-guide/workspaces/README.md` (indexed; `scripts/check-docs.sh` passes).


garland3:
## Summary
Adds user workspaces: a new `user_workspaces` table plus Alembic migration, a repository and REST router under `/api/workspaces`, a `workspaces` feature flag, and a header switcher that saves and restores the active prompt, tools, data sources, and RAG toggle as a named bundle.

## Verdict
Request changes.

## Required changes
1. Fix the rename path at `frontend/src/components/WorkspaceSelector.jsx:106` — `dialog.prompt` resolves to a plain string unless `secondaryLabel` is passed (`ToastProvider.jsx:167-173`), so `answer.value` is `undefined`, `name` becomes `''`, and the function returns before calling `renameWorkspace`. Rename is currently a silent no-op with no error toast. Read `answer` as a string here (or normalize both shapes in one helper) and add a test case covering it.
2. Stop clearing the persisted active-workspace pointer on mount at `frontend/src/contexts/ChatContext.jsx:130` — the effect runs before `config.configReady` and before the workspace list has loaded (`useWorkspaces` starts at `loading: false`), so `chatui-active-workspace` is wiped on every page load and the header loses the active workspace after a refresh. Guard on `config.configReady` and on a "has fetched once" flag rather than on `loading`.
3. Render a distinct error state when `workspacesError` is set in `WorkspaceSelector.jsx:174` — a failed `GET /api/workspaces` currently shows "No workspaces yet…", telling a user with saved workspaces they have none and inviting duplicates. Show the error with a retry that calls the exported `fetchWorkspaces`.
4. Bound the selection lists in the request model at `atlas/routes/workspace_routes.py:41` — `selected_tools`, `selected_prompts`, and `selected_data_sources` are bare `list[str]`, so the 500-entry / 300-character caps in `normalize_config` only apply after the whole body is parsed into memory. Add `max_length` on the list and on each item, importing the constants from the repository so the two layers cannot drift.
5. Restore the original CRLF line endings on `atlas/routes/config_routes.py` — the file was rewritten wholesale (689 deletions / 689 insertions) to carry a two-line change, which resets `git blame` for the entire config payload and will conflict with any other branch touching it.

## Evidence
Evidence: 7/10. The backend is well demonstrated — specific suite counts, 24 new backend tests, 12 new component tests, and named live-server probes that match what the tests assert — but the change adds a visible header control and ships with no screenshot or recording of any of its states, and the `ChatContext` wiring between the switcher and the selection panels is verified at neither end. A screenshot set (active header, open dropdown, empty state) is the most valuable missing artifact.

## Notes
- Dropdown is not keyboard/screen-reader operable: no `aria-expanded`/`aria-haspopup`, no Escape-to-close, no focus return — `WorkspaceSelector.jsx:122-141`.
- No per-user workspace cap and no `LIMIT` on the list query — `workspace_repository.py:96-118`.
- `user_email` is logged unsanitized on create and delete, unlike the `sanitize_for_logging` convention used elsewhere — `workspace_repository.py:161,193`.

---

garland3:
## Browser walkthrough — and a bug it caught

Driven with Playwright against a live server (`atlas/main.py` on the built frontend, real `config/` with MCP servers and the mock RAG API). Same screenshots are now in [`docs/user-guide/workspaces/`](docs/user-guide/workspaces/README.md).

### The switcher

Two saved workspaces, `Project Atlas` active. Each row carries its description and a summary of what it holds, plus rename/delete.

Workspace switcher

### Switching actually swaps the context

One click, same session. Under **HR & Policy** — one data source (`Company Policies`), tools `canvas` + `evaluate`:

HR and Policy applied

Under **Project Atlas** — two data sources (`Technical Documentation`, `Product Knowledge Base`), tools `canvas`, `read_file`, `write_file`, `list_directory`:

Project Atlas applied

The filesystem tools do not survive into `HR & Policy` and `Company Policies` does not survive into `Project Atlas` — that is the replace-not-merge rule holding, which is what keeps a switch from silently widening tool and data-source reach.

### Save and delete

Save dialog

Delete confirmation

### Flag off

Restarted with `FEATURE_WORKSPACES_ENABLED=false`: no workspace button, `features.workspaces: false`, `GET /api/workspaces` → 404. Nothing else in the header moves.

Header with the feature disabled

## The bug this found — fixed in bb50659

The refresh screenshot came back wrong: after a reload the selections a workspace had applied were still in place, but the header had fallen back to the neutral **Workspace** label and `chatui-active-workspace` in `localStorage` read `null`.

The stale-pointer cleanup compared the persisted id against two facts that both arrive asynchronously:

1. **Feature flags default to off** until the config payload lands — so the effect read the feature as disabled and cleared the pointer on the very first render.
2. **The workspace list is empty** until its fetch resolves — so "not in the list" was not yet a meaningful test.

Either one alone wiped the pointer on every page load. The decision now lives in `isStaleWorkspacePointer`, gated on `configReady` and on a `loaded` flag that only a *successful* fetch sets, so a failed fetch cannot let the empty list count as authoritative either. Extracting it from the effect made the precondition testable without standing up the whole provider — 7 new cases in `workspace-pointer.test.js`.

Worth noting for reviewers: I chased this for a while against a **cached `index.html`**, re-testing a stale bundle and concluding my first fix had not worked. The fix was fine; the browser was not reloading it. Cache-bust the URL when verifying frontend changes against a local server.

### Suites after the fix

- Backend: 2374 passed, 17 skipped
- Frontend: 574 passed / 53 files (up from 567/52), `eslint` clean on touched files, build succeeds

garland3:
## Summary
Adds a saved-workspaces feature: a new table plus migration, an owner-scoped repository, a REST router behind a `workspaces` feature flag, and a header switcher that applies a saved set of prompt, tool, and data-source selections.

## Verdict
Request changes.

## Required changes
1. Fix the rename handler at `frontend/src/components/WorkspaceSelector.jsx:92` — `dialog.prompt` resolves to a bare string unless `secondaryLabel` is passed, so `answer.value` is `undefined`, the empty-name guard returns early, and rename is a silent no-op with no request and no toast. Normalize both answer shapes in one shared helper used by `handleRename` and `handleSaveNew` so the call sites cannot drift again.
2. Add a rename case to `workspace-selector.test.jsx` asserting `renameWorkspace` is called with `(ws.id, 'New name')` — rename is currently the only workspace action with no test, which is why the above shipped.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — six end-to-end screenshots across the switcher, applied contexts, save, delete, and flag-off states; named suite counts with deltas on both sides; and specific live-server probes for dedup, 422, empty-list clear, and double-delete. The one missing artifact is a post-reload screenshot confirming the active-workspace pointer fix, which is also what would settle the cache-staleness question raised below.

## Notes
- A failed `GET /api/workspaces` renders the "No workspaces yet" empty state; `workspacesError` is plumbed through `ChatContext.jsx:915` but never read — `WorkspaceSelector.jsx:174`.
- Selection lists are unbounded at the request model, so the 500-item / 300-char caps only apply after the full body is materialized; and `list_workspaces` has no `LIMIT` or per-user cap — `atlas/routes/workspace_routes.py:41`, `workspace_repository.py:96`.
- The dropdown has no `aria-expanded`/`aria-haspopup`, no Escape-to-close, and no focus return — `WorkspaceSelector.jsx:122`.

---

garland3:
## Summary
Adds saved "workspaces" — named bundles of active prompt, RAG sources, tool selections and RAG toggle — with a new `user_workspaces` table and migration, a CRUD REST surface at `/api/workspaces`, a `useWorkspaces` hook plus `ChatContext` orchestration, and a header dropdown to switch, save, update, rename and delete them. The feature is gated behind `workspaces_effective` (requires chat history).

## Verdict
Request changes.

## Required changes
1. Fix the rename action at `frontend/src/components/WorkspaceSelector.jsx:100` — `dialog.prompt` resolves to a bare trimmed string when no `secondaryLabel` is passed (`ui/ToastProvider.jsx:169-173`), so `answer.value` is `undefined`, the empty-name guard returns early, and `renameWorkspace` is never called: the dialog closes and nothing happens, with no error surfaced. Normalize both answer shapes in one helper shared with `handleSaveNew`, and add a `workspace-selector.test.jsx` case asserting `renameWorkspace` is called with `('ws-work', 'New name')` for the string-shaped answer.
2. Render a distinct error state in the workspace dropdown at `frontend/src/components/WorkspaceSelector.jsx:174` — `workspacesError` is exported from `ChatContext.jsx:915` but consumed nowhere, so a failed `GET /api/workspaces` shows the same "No workspaces yet" copy as a genuinely empty account and invites the user to recreate workspaces they already have. Show the error plus a retry wired to `fetchWorkspaces`, and reserve the empty state for `loaded && !error && length === 0`.
3. Make a missing repository a hard failure on read at `atlas/routes/workspace_routes.py:106` — `list_workspaces` returns `200 {"workspaces": [], "error": ...}` while the write verbs raise 503, and `useWorkspaces.fetchWorkspaces` only checks `res.ok`, so it sets `loaded = true` on an empty list and the stale-pointer effect then erases `chatui-active-workspace` permanently. Raise 503 via `_require_repo()` as the writes do (or have `fetchWorkspaces` treat a body carrying `error` as a failure and leave `loaded` false).
4. Bound the selection fields on the request model at `atlas/routes/workspace_routes.py:41` — `selected_tools`, `selected_prompts`, `selected_data_sources` are bare `list[str]` and `active_prompt_key` is an unbounded `Optional[str]`, so the 500-item / 300-char caps in `normalize_config` apply only after the whole body has been decoded and validated. Import `MAX_SELECTION_ITEMS` / `MAX_SELECTION_KEY_LEN` from `workspace_repository` and apply them as `Field(max_length=...)` / `StringConstraints` so oversized bodies are rejected at parse time.
5. Cap workspaces per user in `create_workspace` and add a matching `.limit(...)` to `list_workspaces` (`atlas/modules/chat_history/workspace_repository.py:96-118`) — nothing bounds the row count today, and the list endpoint runs on every page load, JSON-decoding and re-normalizing every row.
6. Add tests for `frontend/src/hooks/useWorkspaces.js` — the hook owns every network call and the `loading`/`loaded`/`error` state machine with no coverage. At minimum pin that a non-2xx `GET` leaves `loaded === false` (the invariant the pointer fix depends on) and that `enabled === false` does not fetch.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — six end-to-end screenshots across distinct states, suite counts with deltas on both sides, and live-server probes matching the assertions. The one missing artifact is a screenshot of the header still showing the active workspace after a hard reload, which is the exact state the most recent fix on this branch repaired.

## Notes
- `active_prompt_key` is stored untrimmed while every other selection key is trimmed — `atlas/modules/chat_history/workspace_repository.py:74`.
- `config_routes.py` was rewritten wholesale (689 lines, CRLF→LF) for a two-line change, resetting blame and guaranteeing conflicts; restore the line endings and land normalization separately.
- The dropdown is not fully operable without a mouse: no `aria-haspopup`/`aria-expanded`, no Escape-to-close, no focus return — `frontend/src/components/WorkspaceSelector.jsx:122`.

---

garland3:
## Prompts are part of a workspace — demonstrated end to end

Asked to confirm a workspace carries **prompts**, not just MCP tool selections. It always has: the config schema stores `active_prompt_key` (the prompt that *replaces* the system prompt) and `selected_prompts` (MCP prompts, which get *prepended*), and `applyWorkspace` restores both. Here is that working, plus one improvement the demo made obvious.

### What the backend actually stored

Two workspaces built through the UI, read straight back from `GET /api/workspaces`:

```json
{ "name": "HR & Policy",
  "config": { "active_prompt_key": "userprompt:b10cba59-…",
              "selected_tools": ["canvas_canvas", "calculator_evaluate"],
              "selected_data_sources": ["atlas_rag:company-policies"] } }

{ "name": "Project Atlas",
  "config": { "active_prompt_key": "userprompt:55ae2e40-…",
              "selected_tools": ["canvas_canvas", "filesystem_read_file",
                                 "filesystem_write_file", "filesystem_list_directory"],
              "selected_data_sources": ["atlas_rag:technical-docs", "atlas_rag:product-knowledge"] } }
```

Different prompt, different tools, different sources — per workspace.

### Improvement: rows now name the prompt

The switcher used to summarize any saved prompt as a generic `custom prompt`, so the one selection that changes how the model answers was the one you could not see before switching. Rows now name it:

Workspace switcher naming prompts

`2 tools · 1 source · Policy Explainer` and `4 tools · 2 sources · Terse Code Reviewer`. User-library prompts resolve by title, MCP prompts by their server key, and the generic label survives only as the fallback for a prompt that has since been deleted.

### The prompt switches with the workspace

Prompt picker under **Project Atlas** — `Terse Code Reviewer` marked active:

Project Atlas prompt

Same picker after switching to **HR & Policy** — `Policy Explainer` active, and the tools and source count moved with it:

HR and Policy prompt

### It reaches the model, not just the label

The same question — *"Is it acceptable to reuse the same password across systems?"* — asked in each workspace against a live LLM.

**HR & Policy** (prompt: *"plain language for a non-specialist, always cite which policy document, never speculate beyond the sources"*) — answers in plain language and cites `pol-001.txt` from `company-policies`:

Answer under HR and Policy

**Project Atlas** (prompt: *"terse senior code reviewer, at most three sentences, lead with the risk then the fix"*) — answers `Risk: … Fix: …` in two sentences and cites `tech-002.txt` from `technical-docs`:

Answer under Project Atlas

One click moved the system prompt *and* the retrieval corpus. Nothing was re-picked by hand between the two.

### Coverage note

This environment's MCP servers expose no prompts, so the demo uses user-library prompts (the `userprompt:<id>` path). The MCP path — `selected_prompts` plus an `<server>_<name>` active key — is covered by the snapshot/apply unit tests (`useSelections.test.js`: *"applyWorkspace loads an MCP active prompt into selected prompts"*) rather than by a screenshot. Worth a live check by anyone running a prompt-exposing server.

### Suites

- Frontend: **576 passed / 53 files** (up from 574), including 2 new cases for prompt naming and the deleted-prompt fallback
- `eslint` clean on touched files; build succeeds

garland3:
## Summary
Adds saved workspaces: a `user_workspaces` table and migration, a workspace repository with config normalization, a feature-gated CRUD REST surface, and a header switcher that snapshots and re-applies the active prompt, data sources, and tools.

## Verdict
Request changes.

## Required changes
1. Fix the rename no-op in `frontend/src/components/WorkspaceSelector.jsx:105` — `dialog.prompt` resolves to a bare trimmed string when `secondaryLabel` is omitted (`ui/ToastProvider.jsx:170`), so `answer.value` is `undefined`, the guard returns early, and `renameWorkspace` is never called. Normalize both answer shapes in one helper and use it in `handleRename` and `handleSaveNew`.
2. Add a rename test to `frontend/src/test/workspace-selector.test.jsx` — resolve `dialogPrompt` with the bare string `'New name'` and assert `renameWorkspace` was called with `('ws-work', 'New name')`; rename is the only workspace action with no coverage, which is why the bug shipped green.
3. Make the read path agree with the writes at `atlas/routes/workspace_routes.py:106` — call `_require_repo()` so a missing repository is a 503 instead of `200 {"workspaces": [], "error": ...}`. As written, `fetchWorkspaces` only checks `res.ok`, sets `loaded = true` on the phantom empty list, and the stale-pointer cleanup then erases `chatui-active-workspace`. Update `test_routes_503_without_repository` to match.
4. Render a distinct error state in `frontend/src/components/WorkspaceSelector.jsx:174` — `workspacesError` is plumbed from `ChatContext.jsx:915` but never read, so a failed load shows "No workspaces yet", inviting a user with saved workspaces to recreate them. Show the error plus a retry control and keep the empty-state copy for the loaded-and-truly-empty case.
5. Bound the selection lists at the request model in `atlas/routes/workspace_routes.py:41` — import `MAX_SELECTION_ITEMS` and `MAX_SELECTION_KEY_LEN` and apply them as `Field(max_length=...)` / `StringConstraints`, including on `active_prompt_key`, so oversized bodies are rejected at parse time rather than fully materialized and then normalized away.
6. Cap the per-user workspace count in `atlas/modules/chat_history/workspace_repository.py:96` and add a matching `.limit(...)` to `list_workspaces` — creation is unbounded and the list query has no LIMIT, so `GET /api/workspaces` on every page load costs the whole table for that user.
7. Strip `active_prompt_key` before storing it in `atlas/modules/chat_history/workspace_repository.py:82` — the emptiness check trims but the original value is persisted, so a padded `" userprompt:abc "` fails `isUserPromptKey` on the client and leaks into `selected_prompts` instead of the system prompt.

## Evidence
Evidence: 9/10 — unusually strong for a change this size: ten screenshots across distinct states, live-server probes matching the assertions, named suite counts with deltas, and a self-disclosed bug found during verification. The one gap worth closing is the post-reload header screenshot for the active-workspace pointer fix, since that state was previously mis-verified against a cached page.

## Notes
- `useWorkspaces` has no tests at all, including the failed-`GET`-leaves-`loaded === false` invariant the pointer fix depends on — `frontend/src/hooks/useWorkspaces.js`.
- Route tests monkeypatch both `_workspaces_enabled` and `_get_repo`, so `app_factory.workspace_repository` existing under the name the route reads is never confirmed — `atlas/tests/test_workspaces.py:210`.
- The dropdown has no `aria-haspopup` / `aria-expanded`, no Escape-to-close, and no focus return — `frontend/src/components/WorkspaceSelector.jsx:122`.

---

PR #820, opened 2026-08-18
fix(tests): land the review fixes that missed the #818 merge

garland3:
Follow-up to #818. These four fixes answer AGENT-REVIEW-BOT-3's last review round on that PR; they were pushed to the branch (`fbdc988`) about four minutes before the squash-merge was taken at `a668295`, so they did not land. This is that commit, cherry-picked onto the merged `main` — no new work.

## What was missed

**1. `high_risk_log_path()` was not a pure resolver** (`atlas/core/prompt_risk.py`)

It stat'd the filesystem, mutated module state and could log, all from a function that reads as a plain path resolver. The one-time relocation notice moved into `log_high_risk_event`, where a filesystem hit is already being paid and only on a turn that actually writes a record.

**2. The relocation notice could never fire for a stale log that appears later**

`_warn_once_if_relocated` recorded the path in its "already announced" set *before* checking whether a stale file existed, so the case that matters most — a legacy log restored from backup or left by a rolled-back deploy, after the process had already resolved the path once — was silently swallowed forever. The `add` now happens only when the notice is actually emitted.

**3. `_release()`'s comment was wrong about closed loops** (`atlas/tests/conftest.py`)

The comment claimed a task whose loop has closed is "marked cancelled". It is not:

```python
loop = asyncio.new_event_loop(); task = loop.run_until_complete(make_task()); loop.close()
task.cancel()   # RuntimeError: Event loop is closed
task._state     # 'PENDING'
```

The comment now says what actually happens — effective only while the loop is open, the swallowed `RuntimeError` is the honest outcome, and cancelling properly means doing it before the loop closes, which is the owning fixture's job rather than teardown's — with a test covering that branch.

**4. A mistyped `ATLAS_TEST_ORDER` looked like a test failure** (`scripts/pytest_test_order.py`)

The plugin raised bare `SystemExit`, so a typo exited **1** (tests failed) instead of **4** (usage error). That distinction matters now that #818 added a CI leg reading that exit code. Uses `pytest.UsageError`; both paths verified.

## Tests

Adds coverage for each behavior change: the notice fires once when a stale log is present, stays silent when there is none, stays silent when nothing moved, and now also fires for a stale log that appears after the first write; plus `_release` on a closed-loop task must not raise.

## Verification

Cherry-picked onto `main` at `d97a0ab`, then run with the ordering harness #818 shipped:

| Run | Result |
| --- | --- |
| default order | 2371 passed, 17 skipped |
| `ATLAS_TEST_ORDER=reverse` | 2371 passed, 17 skipped |
| `ATLAS_TEST_ORDER=6 ATLAS_TEST_ORDER_SCOPE=test` | 2371 passed, 17 skipped |
| `/tmp/atlas-test-*` left behind | 0 |

`ruff` clean on all four files.


garland3:
Both Copilot points addressed in e7b66c7.

**Per-event `legacy.exists()` stat when nothing moved.** Correct, and it matters more now that the notice runs per medium/high event rather than once per path resolution: `path != legacy and legacy.exists()` short-circuits the *warning* but Python still evaluated the stat for the un-relocated deployment, which is the common one. It early-returns on `path == legacy` and marks the path so later events skip even the comparison. Measured: **0** `exists()` calls on the legacy path across three events in the un-relocated case (was one per event).

**`ensure_future(..., loop=loop)`.** Simplified, though the deprecation premise did not hold up when I checked it — on 3.11 that call raises nothing under `-W error::DeprecationWarning`, and `loop` is still a documented parameter of `asyncio.ensure_future` (unlike the `loop=` arguments removed from `asyncio.sleep`, `gather`, etc. in 3.10). The helper is cleaner without it, so it now calls `ensure_future` from inside the running loop.

One caveat for anyone tempted to simplify further: `loop.create_task()` on a non-running loop does **not** reproduce this branch — the task has to have started before `cancel()` needs `call_soon`, so a never-scheduled task cancels without raising and the test would silently stop covering the closed-loop path. The test now asserts `task.get_loop() is loop` and `not task.done()` before releasing, so that regression would be caught.

Re-verified at e7b66c7: default / reverse / test-shuffle all 2371 passed, 17 skipped; 0 temp directories left; ruff clean.

garland3:
## Summary
Re-lands four follow-up fixes that missed the #818 merge: `high_risk_log_path()` becomes a pure resolver with the relocation notice moved into `log_high_risk_event`, the notice's memo is now recorded only when it actually fires (so a stale log appearing later is still announced), `_release()`'s closed-loop comment is corrected and covered by a test, and the ordering plugin raises `pytest.UsageError` instead of bare `SystemExit`.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — well demonstrated for a non-visual change of this size, with a named test for three of the four behavior changes and a suite run under three orderings. The one gap is the `pytest.UsageError` change: "both paths verified" is asserted but never shown, and neither invalid `ATLAS_TEST_ORDER` / `ATLAS_TEST_ORDER_SCOPE` value appears in the verification table.

## Notes
- `assert not task.cancelled()` at `atlas/tests/test_isolation_fixtures.py:85` passes whether or not the loop is closed, so it does not pin the documented behavior — wrap `task.cancel()` in `pytest.raises(RuntimeError, match="Event loop is closed")` before calling `_release`; the same test also abandons a PENDING task, so set `task._log_destroy_pending = False` (or a no-op loop exception handler) to keep `Task was destroyed but it is pending!` out of suite output.
- `scripts/pytest_test_order.py:53,69` has no automated coverage in either direction; the hook is a plain function, so `pytest.raises(pytest.UsageError)` against `pytest_collection_modifyitems` pins the exit-4 contract without a subprocess.
- `CHANGELOG.md:9` misstates the pre-fix behavior and the consumer: a mistyped value previously produced an `INTERNALERROR` traceback and exit 3, not exit 1, and the reverse-order CI leg at `.github/workflows/ci.yml:56` is `continue-on-error: true`, so nothing reads that code today.

---

garland3:
Round 8 addressed in 450076e. All three were right, including one place where my own PR text asserted something I had not measured.

**`assert not task.cancelled()` pinned nothing.** True — it holds whether or not the loop is closed, so the test passed for the wrong reason. It now asserts the documented behavior directly:

```python
with pytest.raises(RuntimeError, match="Event loop is closed"):
    task.cancel()
_release(task)          # same call, swallowed
assert not task.cancelled() and not task.done()
```

Also set `task._log_destroy_pending = False` as you suggested — `Task was destroyed but it is pending!` no longer appears in suite output (measured: 0 occurrences, was 1).

**Plugin UsageError paths had no coverage.** Added four direct calls to the hook — bad order value, bad scope, unset (order untouched), reverse — no subprocess needed, as you noted. One wrinkle worth recording: `scripts/` is only on `PYTHONPATH` when the ordering leg runs, so a plain `import pytest_test_order` fails in an ordinary suite run. The tests load the plugin by file path and skip if it is absent; verified passing both with and without `scripts/` on the path.

**Changelog misstated the old behavior.** Correct, and I should have measured rather than asserted "exit 1". I did now:

```
$ pytest ... -p sysexit_probe   # a hook raising bare SystemExit
INTERNALERROR> ...              # ~50 lines
exit code = 3
```

So it was exit **3** with an `INTERNALERROR` traceback, not exit 1. The entry is corrected to "one line and exit 4 (usage) instead of a ~50-line INTERNALERROR traceback and exit 3", and no longer implies the CI leg reads that code today — it is `continue-on-error`, so the value right now is the legible failure, and the exit code starts mattering when that leg becomes blocking.

Re-verified at 450076e: default / reverse / test-shuffle all 2375 passed, 17 skipped; 0 temp directories; 0 destroyed-pending warnings; ruff clean.

garland3:
## Summary
Re-lands review fixes dropped by the #818 squash-merge: makes `high_risk_log_path()` a pure resolver with the stale-log relocation notice moved to event emission, turns mistyped `ATLAS_TEST_ORDER`/`ATLAS_TEST_ORDER_SCOPE` values into a usage error instead of an `INTERNALERROR`, and adds tests for both.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10 — each behavior change has a named test, the suite is run under three orderings at each commit, and the performance and warning-suppression claims are backed by measurements rather than assertion. The main gap is that the measured "0 `exists()` calls in the un-relocated case" is not pinned by any test, so deleting the memo `add` in the `path == legacy` branch would go unnoticed.

## Notes
- `_emit`-based relocation tests pass vacuously if the emit fails, since `log_high_risk_event` swallows exceptions and logs a different message than `_notices()` filters on — assert the record landed in `_emit` (`atlas/tests/test_isolation_fixtures.py:297`, `:331`).
- `test_survives_a_task_whose_loop_is_closed` does not exercise `_release`'s swallow path: the second `task.cancel()` no longer raises, so the test stays green if the `except Exception` body is removed — use a second, untouched task (`atlas/tests/test_isolation_fixtures.py:85`).
- `ATLAS_TEST_ORDER_SCOPE` is validated only on the shuffle path, so `ATLAS_TEST_ORDER=reverse ATLAS_TEST_ORDER_SCOPE=sideways` exits 0 while the same typo with a seed exits 4 — read and validate scope before branching on order (`scripts/pytest_test_order.py:44-47`).

---

garland3:
Round 9 addressed in 1e39a62. All four were right, and chasing the first one turned up a correction to something **I** claimed earlier in this PR.

**The stat-count claim — mine was wrong, and the real defect is worse.** I wrote that the old code stat'd the legacy path on every event in the un-relocated case. It did not: `path != legacy and legacy.exists()` short-circuits, so when the paths matched there was never a stat. My new test proved that by *failing to fail* when I reverted the early-return.

What actually repeats the stat is a path mismatch: `path` comes back `resolve()`d and `legacy` did not, so a checkout reached through a **symlink** compares unequal, reports "relocated" forever, and stats the legacy file on every medium/high event. Both sides are resolved now, and the test exercises exactly that: log directory reached via symlink, 3 events — **2 stats before the fix, 0 after**. So the note was correct that the branch needed pinning; the mechanism was just not the one either of us named.

**`test_survives_a_task_whose_loop_is_closed` stopped exercising the swallow.** Verified: the first `cancel()` sets `_must_cancel`, so cancelling the same task again returns quietly — deleting `_release`'s `except` body kept the test green. It now releases a second, untouched task, and I confirmed removing the `except` body fails it.

**`_emit` could pass vacuously.** Right — `log_high_risk_event` swallows exceptions and logs a message `_notices()` does not match, so a broken emit made "stays silent" pass for the wrong reason. `TestRelocationNotice._emit` now asserts the record landed (line count grew by exactly one).

**`ATLAS_TEST_ORDER_SCOPE` validated only on the shuffle path.** Fixed by reading and validating scope before branching on `order`. Measured: `ATLAS_TEST_ORDER=reverse ATLAS_TEST_ORDER_SCOPE=sideways` now exits **4**; it exited 0 before.

Every new assertion was checked against the pre-fix code rather than assumed to discriminate — that is how the stat-count mistake surfaced.

Verified at 1e39a62: default / reverse / test-shuffle all 2376 passed, 17 skipped; 0 temp directories; ruff clean. CI was green on 450076e and is re-running now.

garland3:
## Summary
Re-lands four follow-up fixes that missed the #818 squash-merge: `high_risk_log_path()` becomes a pure resolver with the one-time relocation notice moved into `log_high_risk_event`, the notice now records a path only when it actually fires, `_release()`'s closed-loop comment is corrected, and a mistyped `ATLAS_TEST_ORDER` raises `pytest.UsageError` so it exits 4 rather than 1.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. Each behavior change has a named test, the suite is re-run under default/reverse/seeded-shuffle orderings with counts reported, and the numeric claims are measured rather than asserted; nothing material is missing.

## Notes
- `counting_exists` filters on the unresolved legacy path, so deleting the `path == legacy` early return leaves the test green — the memo it advertises is unpinned (`atlas/tests/test_isolation_fixtures.py:378`).
- `path` is used unresolved while `legacy` is resolved, so a symlinked log directory reports a false "no longer updated" (`atlas/core/prompt_risk.py:233`).
- `ATLAS_TEST_ORDER_SCOPE` is validated on the reverse path where it has no effect, and unvalidated when `ATLAS_TEST_ORDER` is unset (`scripts/pytest_test_order.py:39-56`).

---

garland3:
Round 10 addressed in 51585ba. The first two were right — and the second one caught my round-9 fix trading one false positive for another.

**Legacy path construction.** Round 9 resolved the whole legacy path. That fixes the symlinked-*checkout* case but breaks the symlinked-*log-file* case, which is what a rotation setup looks like:

```
path         .../logs/security_high_risk.jsonl      # dir resolved, filename appended
legacy(full) .../rotated/security_high_risk.jsonl   # resolves through the file symlink -> False
legacy(dir)  .../logs/security_high_risk.jsonl      # matched construction        -> True
```

So a deployment that symlinks its log file would get a "no longer updated" warning about a log that is very much still live, plus a stat on every event. The legacy path is now built exactly the way `high_risk_log_path` builds its return value — resolve the directory, append the filename — which is equal in both symlink shapes. New test covers the symlinked log file; verified it fails against the round-9 form.

**The stat counter left the memo unpinned.** Correct: it filtered on the symlinked spelling only, so deleting the `path == legacy` early return made the product stat the *resolved* spelling, which the filter ignored, and the test stayed green. It counts both spellings now. Verified: deleting the memo fails it (was silently green).

**`ATLAS_TEST_ORDER_SCOPE` validation — keeping it as is.** This one I am pushing back on, because it partly reverses round 9. That round's note was that `ATLAS_TEST_ORDER=reverse ATLAS_TEST_ORDER_SCOPE=sideways` exited 0 while the same typo with a seed exited 4, and asked for validation before the branch; that is what is there now. Validating a value the reverse path happens to ignore costs nothing, and the same typo behaving differently depending on an unrelated variable is the worse failure. It stays unvalidated when `ATLAS_TEST_ORDER` is unset because the plugin is completely inert then — erroring in that case would break ordinary runs for anyone who has the variable exported from an earlier session. Happy to flip either half if you disagree.

Verified at 51585ba: default / reverse / test-shuffle all 2377 passed, 17 skipped; 0 temp directories; ruff clean.

garland3:
## Summary
Follow-up to the high-risk security log relocation work: `high_risk_log_path()` becomes a pure resolver and the one-time relocation notice moves into `log_high_risk_event`, with the memo recorded only after the notice is actually emitted. Also hoists `ATLAS_TEST_ORDER_SCOPE` validation above the `reverse` branch in `scripts/pytest_test_order.py` and adds tests for both areas.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. Every behavior change has a named test, the suite is re-run under default/reverse/seeded orderings with counts reported, and several assertions were checked against the pre-fix code to confirm they discriminate. The one gap worth closing is a CI result at the head commit (`51585ba`) — the last reported green run is two commits back.

## Notes
- The relocation notice still fires falsely when the legacy path is a symlink to the live log and `APP_LOG_DIR` points at the link's target directory — `legacy.exists()` follows the link, so the message calls a file "no longer updated" while it receives every record. This is the exact shape #818's changelog recommends to operators. Consider `if legacy.exists() and not legacy.samefile(path)` at `atlas/core/prompt_risk.py:238`.
- The reverse + bad-scope combination this change exists to fix is untested: `test_invalid_scope_raises_usage_error` uses `ATLAS_TEST_ORDER=7`, which would pass with the validation in either position (`atlas/tests/test_isolation_fixtures.py:168`). The seeded/shuffle branch of the plugin also has no test or CI coverage.
- The new docstring at `atlas/core/prompt_risk.py:200` claims "no filesystem access", but `expanduser().resolve()` stats and follows links — the same claim-vs-behavior mismatch this PR fixes elsewhere.

---

garland3:
Round 11 addressed in 84fc3df. All three were right, and the first is the worst bug in this PR so far.

**The notice fired against the migration #818's own upgrade note recommends.** "move or symlink the old one" -- so an operator symlinks the repository path at the live log and sets `APP_LOG_DIR` to its directory. Reproduced:

```
path            .../var-log/security_high_risk.jsonl
legacy          .../repo-logs/security_high_risk.jsonl   # a symlink at the file above
legacy.exists() True        (follows the link)
would warn      True        -- about a file receiving every record
legacy.samefile(path)  True
```

Guarded with `samefile()`, and only when the new location actually has a file to compare against — if it does not, the old log genuinely is stale and the notice stands. New test covers it and fails without the guard.

**The `reverse` + bad-scope case had no test.** Correct — the existing one used `ATLAS_TEST_ORDER=7`, which passes with the check in either position, so the round-9 fix was unpinned. Parametrized over seed and `reverse`; the `reverse` case fails if the hoist is reverted.

**The shuffle branch had no coverage.** Added: same seed reproduces the same order, module scope keeps a file's tests contiguous, and both scopes preserve the item set.

**"No filesystem access" was wrong.** `resolve()` reads the filesystem to follow links. The docstring now says what callers actually need — free of logging and module state, callable as often as they like — rather than a purity claim the code does not honor. Fair catch given this PR exists partly to fix claim-vs-behavior mismatches.

**CI at head**, which the evidence note asked for: 51585ba was green on all 10 checks; 84fc3df is running now and I will confirm it here when it lands.

Verified locally at 84fc3df: default / reverse / test-shuffle all 2380 passed, 17 skipped; 0 temp directories; ruff clean.

garland3:
## Summary
Follow-up fixes to the high-risk log relocation notice in `atlas/core/prompt_risk.py` (the resolver no longer logs; a `samefile` guard suppresses the warning when the old path is a symlink at the live log) and to the `scripts/pytest_test_order.py` collection plugin (scope validation hoisted, `SystemExit` replaced with `pytest.UsageError`), plus new tests and a conftest teardown comment.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10 — every behavior change carries a named test, assertions were checked against the pre-fix code, and the suite counts are measured under three orderings. The one outstanding artifact is a green CI result at `84fc3df` itself rather than at `51585ba`.

## Notes
- `test_seeded_shuffle_is_deterministic_and_scope_aware` (`atlas/tests/test_isolation_fixtures.py:192`) stays green with `rng.shuffle(items)` deleted — seed 11 over two modules is the identity permutation, and the `scope="test"` branch is only checked by set equality. Use more modules and a seed known to reorder.
- The resolver-purity fix itself is unpinned: re-inserting `_warn_once_if_relocated` into `high_risk_log_path` leaves the suite green (`atlas/core/prompt_risk.py:214`). Assert that calling the resolver neither logs nor populates `_RELOCATION_NOTICE_PATHS`.
- In the relocated deployments (no stale file, or old path symlinked at the live log) `_warn_once_if_relocated` never memoizes, so it re-resolves and stats the legacy path on every medium/high event (`atlas/core/prompt_risk.py:238`); hoisting the resolved `legacy` to module scope removes most of it. Related: the `#820` changelog entry omits the now-fatal `ATLAS_TEST_ORDER_SCOPE` validation on non-shuffle orderings and the symlink guard (`CHANGELOG.md:9`).

---

garland3:
Round 12 addressed in 240b55b. All four were right, and two were tests of mine that passed for the wrong reason.

**The shuffle test passed with `rng.shuffle` deleted.** Confirmed — seed 11 is the identity permutation, for two modules *and* four:

```
seed 11, 4 modules: UNCHANGED
seed  6, 4 modules: ['test_d', 'test_c', 'test_b', 'test_a']
```

Exactly the failure a shuffle test exists to catch. It uses seed 6 over four modules now and asserts the order actually changed, that module scope keeps each file contiguous, and that the two scopes produce different orders. Verified it fails with the shuffle removed.

**Resolver purity was unpinned.** Right: re-inserting `_warn_once_if_relocated` into `high_risk_log_path` left the suite green, so the change this PR exists to make had nothing defending it. `TestResolverPurity` now asserts that resolving neither logs nor populates `_RELOCATION_NOTICE_PATHS`, and it fails if the call is reinstated.

**Relocated deployments re-resolved on every event.** Correct, and it follows from a deliberate choice: when nothing is stale the notice keeps re-checking, because a stale file can appear later (round 7). That means no memo, so the resolved legacy path was rebuilt — walking symlinks — per medium/high event. The resolved path is cached per default directory now, keyed so a test's monkeypatched directory cannot collide. Measured **3 resolves across 3 events before, 1 after**; the re-check itself still runs, so late-appearing stale logs are still announced.

I kept the re-check rather than memoizing the decision, since silently missing a stale log is the worse failure of the two.

**Changelog gaps.** Added: the `samefile` guard, and that `ATLAS_TEST_ORDER_SCOPE` validation is now fatal for every ordering rather than only the seeded shuffle — that second one is a new failure mode for anyone with a typo'd value exported, so it belongs in the entry.

Verified at 240b55b: default / reverse / test-shuffle all 2382 passed, 17 skipped; 0 temp directories; ruff clean. CI green at 51585ba; 84fc3df and 240b55b are in flight.

garland3:
`build-and-test` went red on 240b55b and I have re-run it. The failure is not from this PR:

```
FAILED tests/test_hooks_engine.py::TestFailureModes::test_output_over_cap_is_killed_and_treated_as_error
E   assert '1 MB' in "Hook 'f' errored: timeout"
```

`test_hooks_engine.py` is not in this PR's diff (`CHANGELOG.md`, `atlas/core/prompt_risk.py`, `atlas/tests/conftest.py`, `atlas/tests/test_isolation_fixtures.py`, `scripts/pytest_test_order.py`), the same job was green on 51585ba and 84fc3df, and the test passes locally in 0.05s across repeated runs, on this branch and on merged `main`.

What happened is that the hook flooding ~13 MB hit its **20 s timeout** before the 1 MB output cap could fire — the runner was slow enough that the cap never got the chance.

That is the **second** load-sensitive test to redden CI during this work, and the two share a shape worth naming: *a fast guard is asserted to fire before a slow fallback, with a budget tuned for an unloaded machine*.

| test | budget | what actually happened under load |
| --- | --- | --- |
| `test_calculator_tool::test_resource_bounds_reject_quickly` | `elapsed < 1.0` | rejection took 1.55 s (the unguarded expression runs >120 s) |
| `test_hooks_engine::test_output_over_cap_is_killed_and_treated_as_error` | `timeout_ms=20000` | timeout fired before the 1 MB cap |

Both would be steadier asserting the *mechanism* rather than racing the clock — for this one, cap the hook's runtime well below the timeout, or assert on the kill reason with a timeout large enough that only a genuinely broken cap loses the race. I have not touched either: they are outside this PR's scope and re-tuning them belongs with whoever owns those guards. Happy to take both as a follow-up if you want them fixed rather than re-run.

garland3:
## Summary
Hardens the test-ordering plugin (`ATLAS_TEST_ORDER_SCOPE` is now validated for every ordering and raises a usage error instead of an internal traceback), moves the high-risk-log relocation notice out of the path resolver into the write path, and adds a cache for the resolved legacy log directory, with accompanying tests and changelog/doc updates.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is adequately demonstrated for its size: every behavior change has a named test, the suite is re-run under default, reverse and seeded-shuffle orderings with counts reported, and claims that failed measurement were corrected rather than restated. The one gap is that the exit-4 half of the usage-error claim is asserted at the exception level only — a single pasted `ATLAS_TEST_ORDER=nonsense pytest ...; echo $?` run showing `4` would close it, along with a CI-green confirmation at this head commit.

## Notes
- Tests in `TestOrderingPluginUsageErrors` (`atlas/tests/test_isolation_fixtures.py:173`) don't clear `ATLAS_TEST_ORDER_SCOPE`; with it exported, two tests fail — the exact env leak this series exists to remove. An autouse `delenv` fixture on the class closes it.
- `_default_log_dir()` does a full `resolve()` before the new `_LEGACY_PATH_CACHE` lookup (`atlas/core/prompt_risk.py:246`), so the cache saves one of two resolves per medium/high event on a path reached inside an async RAG loop; `functools.lru_cache` on `_default_log_dir` makes it free.
- `_LEGACY_PATH_CACHE` is reset by only one of the eight relocation tests (`atlas/tests/test_isolation_fixtures.py:499`) and the `path.exists()` guard at `prompt_risk.py:270` has no test defending it — a shared reset seam and a first-event assertion would pin both.

---

garland3:
Round 13 addressed in a632680. The first note is the one that stung: **my own tests were reading the environment they were written to police.**

**Ordering tests inherited `ATLAS_TEST_ORDER_SCOPE`.** Reproduced:

```
$ ATLAS_TEST_ORDER_SCOPE=sideways pytest tests/test_isolation_fixtures.py
FAILED ...::test_invalid_order_value_raises_usage_error
FAILED ...::test_reverse_reverses
2 failed, 27 passed
```

Tests belonging to a series about ambient-environment leakage should not depend on ambient environment. An autouse fixture clears both variables; the file now passes with either exported (verified both).

**One reset seam for the module caches.** `_LEGACY_PATH_CACHE` and `_RELOCATION_NOTICE_PATHS` are cleared by a single autouse fixture for the whole relocation class instead of by whichever test remembered — they are process-wide memos, so one test's leftover was deciding the next test's answer.

**`_default_log_dir` is `lru_cache`d.** Correct that this was the other half of the per-event work: it resolves `__file__` and is consulted on every medium/high event, and the answer cannot change within a process.

**The `path.exists()` half of the guard had no test.** You were right, and my reasoning about why it was covered was wrong: removing it left the suite green because the notice merely slipped from the first event to the second. That matters on a quiet deployment where the second event may never come. Pinned: a stale log is announced on the first event; removing the guard now fails.

**Exit-code evidence**, measured rather than asserted:

```
$ ATLAS_TEST_ORDER=nonsense pytest tests/test_health_route.py -p pytest_test_order
ERROR: ATLAS_TEST_ORDER='nonsense' is not 'reverse' or an integer seed
$ echo $?
4
```

Verified at a632680: default / reverse / test-shuffle all 2383 passed, 17 skipped; 0 temp directories; ruff clean. CI: green at 51585ba and 84fc3df; the 240b55b run went red on `test_hooks_engine`'s output-cap-vs-timeout race (not in this diff, passes locally in 0.05s) and its rerun is in flight.

garland3:
**CI green at head `a632680`** — all 10 checks, `build-and-test` 12m23s. That closes the last evidence item (a green run at the head commit rather than two behind).

The reverse-order leg ran the full suite in-container at this head, not a no-op:

```
Run tests in reverse collection order
  [pytest_test_order] reversed 2400 tests
  2372 passed, 28 skipped, 620 warnings in 71.37s
```

Production-mode leg: 2365 passed, 35 skipped. Local at the same commit: default / reverse / test-shuffle all 2383 passed, 17 skipped, 0 temp directories.

For anyone picking this up: the branch is `MERGEABLE/CLEAN`. The only red run in this PR's history was `240b55b`, on `test_hooks_engine`'s output-cap-vs-timeout race — not in this diff, passes locally in 0.05s, and green on every other head here.

garland3:
## Summary
Makes the high-risk log path resolver pure by moving the relocation notice into `log_high_risk_event`, adds an `lru_cache` and a legacy-path memo to cut per-event filesystem work, hoists `ATLAS_TEST_ORDER_SCOPE` validation above the ordering branch in the pytest ordering plugin, and extends the isolation test suite to cover all of it.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10 — every behavior change is pinned by a named test, the performance claims are measured rather than asserted, and CI is green at the head commit. The one gap is the `@lru_cache(maxsize=1)` on `_default_log_dir`: tests monkeypatch that function wholesale, so the decorator is never on the measured path and deleting it leaves the suite green.

## Notes
- Set-but-empty `ATLAS_TEST_ORDER_SCOPE` (e.g. `docker run -e ATLAS_TEST_ORDER_SCOPE` pass-through) now raises `UsageError` instead of falling back to `module`; read it as `os.environ.get("ATLAS_TEST_ORDER_SCOPE") or "module"` to match how `ATLAS_TEST_ORDER` treats empty — `scripts/pytest_test_order.py:46`.
- `_reset_prompt_risk_caches` is autouse on `TestRelocationNotice` only, but `_warn_once_if_relocated` now runs on every event, so `TestHighRiskWriteFailureWarning` both inherits and leaves populated memos; promote the fixture to module scope — `atlas/tests/test_isolation_fixtures.py:362`.
- The symlinked-log shape never memoizes, so it repeats three stats per event; add `_RELOCATION_NOTICE_PATHS.add(str(path))` in the `samefile`-matched branch — `atlas/core/prompt_risk.py:273`.

---

garland3:

garland3:
## Summary
Hardens test isolation: the high-risk security log relocation notice no longer fires when the old path is a symlink at the live log, `_default_log_dir()` is memoized, and `ATLAS_TEST_ORDER_SCOPE` is validated for every ordering rather than only the seeded shuffle. Adds coverage in `atlas/tests/test_isolation_fixtures.py`.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. The four behavior changes are unusually well demonstrated — named tests, pasted measurements, and suite runs under default/reverse/shuffle orderings — but all of that attaches to `a632680`, three commits back; the current head is a merge that pulled in two unrelated feature PRs and has no test run or CI result at that SHA. Re-running the three orderings at `9f93040` is the missing artifact.

## Notes
- Set-but-empty `ATLAS_TEST_ORDER_SCOPE` now aborts collection with exit 4 instead of defaulting; `ATLAS_TEST_ORDER` treats empty as unset. Use `os.environ.get("ATLAS_TEST_ORDER_SCOPE") or "module"` — `scripts/pytest_test_order.py:46`.
- The `samefile`-matched branch never memoizes, so a symlinked legacy log re-stats three paths on every medium/high event forever — `atlas/core/prompt_risk.py:273`.
- The relocation notice attributes the move to `APP_LOG_DIR` even when the directory came from `app_settings.app_log_dir`, sending operators to an unset variable — `atlas/core/prompt_risk.py:277`.

---

PR #825, opened 2026-08-20
feat(rag): show backend reference label in the References section

garland3:
## Problem

The newest ATLAS-RAG response shape carries a human-readable `reference` string on each `metadata.references[]` entry:

```json
{
  "document_ref": 0,
  "filename": "string",
  "sections": [{"section_ref": 0, "text": "string", "relevance": 0}],
  "reference": "string"
}
```

`AtlasRAGClient._parse_metadata_newest` never read that key, so every entry in the UI's References section rendered as a bare filename (`tech-001.txt`) even when the backend supplied a proper source line. The loss was silent — `.get` returns `None` and the field is `Optional`, so the reference parsed "successfully" with a hole in it.

## Change

- `atlas/modules/rag/atlas_rag_client.py` — prefer `reference` as the displayed label, falling back to `filename` when it is absent, blank, or not a string. The value flows through the existing `DocumentMetadata.title` slot.
- `mocks/atlas-rag-api-mock/main.py` — emit `reference` alongside `citation` so the field is exercised locally.
- `mocks/atlas-rag-api-mock/README.md` — document the field and its fallback.
- Tests: 5 backend parser cases, 3 frontend label-extraction cases.

**No frontend code change.** The label reaches the UI through the markdown References block the backend already renders, so `ragCitations.js` picks it up unmodified. The frontend tests are there to pin the behavior, not because anything moved.

## Compatibility

Backends that send only `filename` render exactly as before — verified by test. The existing `citation` field is untouched and still renders as the italic line beneath each entry.

`DocumentMetadata.sanitize_title` (control-char strip, 200-char cap) and `_sanitize_label` (markdown/injection-char strip) both apply to the new value, so the attacker-influenced surface is unchanged.

## Verification

Full backend suite (`bash test/atlas_tests.sh`): **2448 passed, 17 skipped**.
Full frontend suite (`npx vitest run`): **579 passed, 53 files**.

End-to-end against the live mock (parser → markdown → `marked` → `ragCitations.js`):

```
--- extracted labels ---
 [1] API Authentication Guide, tech-001.txt
 [2] Deployment Pipeline, tech-003.txt
 [3] Microservices Architecture, tech-004.txt
--- collapsed summary (what the user sees closed) ---
[1] API Authentication Guide, tech-001.txt, [2] Deployment Pipeline, tech-003.txt, ...
--- anchors present ---
rag-ref-msg1-1, rag-ref-msg1-2, rag-ref-msg1-3
```

Previously those labels read `tech-001.txt`, `tech-003.txt`, `tech-004.txt`.

## Notes for reviewers

1. **Scoped to v1 `metadata.references`.** The v2 `hits`/`citations` shape already has a `title` field serving as the label, so it has no equivalent gap. If v2 also grows a `reference` key, `_parse_v2_documents` needs the same treatment.
2. **`_sanitize_label` strips parentheses and brackets** from labels before rendering (pre-existing). A backend `reference` of `PTO Policy (Rev. C)` renders as `PTO Policy Rev. C`. Not a regression, but more visible now that labels are richer than filenames. The mock deliberately uses a comma format to avoid it.
3. **Markdown round-trip fragility** (pre-existing): the frontend re-parses labels out of rendered HTML with a `N.\s` anchor. A `reference` containing a digit-dot-space sequence (`Doc 12. Overview`) could produce a spurious entry. Richer labels make this marginally more reachable. Fixing it properly means passing references as structured data rather than markdown — out of scope here.
4. `docs/admin/external-rag-api.md` still documents the older `rag_metadata` shape for v1 rather than `metadata.references`; that gap predates this change and I left it alone.

### DIFF
--- a/atlas/modules/rag/atlas_rag_client.py
+++ b/atlas/modules/rag/atlas_rag_client.py
@@ -26,6 +26,7 @@ Response body (RagResponse):
         "references": [
           {
             "citation": "IEEE format" | null,
+            "reference": "human-readable source line" | null,
             "document_ref": 1,
             "filename": "doc.pdf",
             "sections": [
@@ -769,6 +770,10 @@ class AtlasRAGClient:
         ``sections`` carry the actual snippet text matched to the query.
         The reference's top section relevance becomes ``confidence_score``
         so existing UI/sorting code still works without changes.
+
+        The displayed label prefers the backend's ``reference`` string and
+        falls back to ``filename``, so backends that only send a filename
+        render exactly as before.
         """
         try:
             references_raw = metadata.get("references") or []
@@ -789,6 +794,13 @@ class AtlasRAGClient:
 
                 top_relevance = max((s.relevance for s in sections), default=0.0)
                 filename = ref.get("filename") or ""
+                # Newer backends send a human-readable ``reference`` string
+                # (e.g. an IEEE-style source line) in place of the older
+                # ``citation`` field. Prefer it as the displayed label and
+                # fall back to the filename when it is absent.
+                reference = ref.get("reference")
+                if not isinstance(reference, str) or not reference.strip():
+                    reference = None
 
                 try:
                     doc_metadata = DocumentMetadata(
@@ -797,7 +809,7 @@ class AtlasRAGClient:
                         confidence_score=top_relevance,
                         chunk_id=None,
                         last_modified=None,
-                        title=filename or None,
+                        title=reference or filename or None,
                         url=None,
                         citation=ref.get("citation"),
                         document_ref=ref.get("document_ref"),
--- a/atlas/tests/test_atlas_rag_client.py
+++ b/atlas/tests/test_atlas_rag_client.py
@@ -464,6 +464,83 @@ class TestParseResponseMetadata:
         assert doc.confidence_score == 0.9
         assert [s.text for s in doc.sections] == ["first snippet", "second snippet"]
 
+    def test_reference_field_becomes_display_label(self, client):
+        """Newer backends send ``reference``; it becomes the shown label."""
+        data = {
+            "metadata": {
+                "response_time": 1,
+                "references": [
+                    {
+                        "reference": "API Authentication Guide, tech-001.txt",
+                        "document_ref": 1,
+                        "filename": "tech-001.txt",
+                        "sections": [
+                            {"section_ref": 1, "text": "snippet", "relevance": 0.5},
+                        ],
+                    }
+                ],
+            },
+        }
+        result = client._parse_response_metadata(data, "technical-docs")
+        doc = result.documents_found[0]
+        assert doc.title == "API Authentication Guide, tech-001.txt"
+
+    def test_filename_used_when_reference_absent(self, client):
+        """Backends that only send ``filename`` render exactly as before."""
+        data = {
+            "metadata": {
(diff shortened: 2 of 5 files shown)


garland3:
## Summary
Promotes the optional backend-supplied `metadata.references[].reference` string to the display label for RAG citations, falling back to `filename` when it is absent, blank, or non-string. The mock RAG server now emits the field, and parser plus frontend-extraction tests were added.

## Verdict
Request changes.

## Required changes
1. Neutralize `N.` sequences in labels in `_sanitize_label` (`atlas/modules/llm/litellm_caller.py:1377`) with the same `re.sub(r"(\d{1,2})\.\s", r"\1.", text)` that `_sanitize_snippet` already applies — a `reference` such as `Rev 7. Approved by security` on entry 1 registers itself as the label for citation `[7]` in `extractSourceLabels` (`frontend/src/utils/ragCitations.js:19`) and wins over the real source 7, so the user is shown a source label that belongs to a different document.
2. Replace `—` (and `–`) with `-` in `_sanitize_label` (`atlas/modules/llm/litellm_caller.py:1377`) — the frontend extractor treats the em dash as the label/details terminator, so `reference: "Report — Vol 12. Appendix, tech-001.txt"` renders truncated to `Report` and emits a phantom entry `[12]`. Em dashes were impossible in filenames and are common in curated source lines.
3. Keep the filename addressable rather than overwriting it: add a dedicated label field to `DocumentMetadata` (`atlas/modules/rag/client.py`) and have the References renderer prefer `label or title or source`, or at minimum append `filename` to the detail line in `_format_rag_references`. Today `source` is `data_source or filename` and `data_source` is effectively always set, so a backend sending `API Authentication Guide, Rev. 4` leaves the user with no way to tell which file the snippet came from.
4. Stop routing the prose label through the key named `filename` in `_tool_references` (`atlas/modules/mcp_tools/mcp_execution.py:63`) — the value is forwarded to the model as tool output, and `_sanitize_result_for_ui` (`atlas/application/chat/utilities/event_notifier.py:71`) basenames any `filename` key, so a label containing `/` is silently cut at the last slash in the tool-result panel. Fixed by item 3, or by renaming the payload key.
5. Add a test in `TestBuildReferencesSection` that asserts a `reference`-derived `title` renders exactly the `1. <label> — <details>` line the frontend tests feed to `extractSourceLabels`, plus `extractSourceLabels` cases for an em-dash label and a `Doc 12.` label — currently both layers are pinned independently and the middle hop can regress with every test still green.

## Evidence
Evidence: 8/10. The change is well demonstrated for its size — parser tests for the present/absent/blank/non-string/oversize cases, frontend extraction tests, full suites run with counts, and an end-to-end run against the mock with a before/after label comparison. The one missing artifact is a rendered view of the References section itself (collapsed and expanded), since the entire observable effect is a string a user reads.

## Notes
- Collapsed references summary has no length bound and labels grew from ~12 to up to 200 characters — clamp per label or line-clamp `.rag-references-summary` (`frontend/src/utils/ragCitations.js:92`).
- `sanitize_title` hard-slices at 200 characters with no ellipsis, unlike the adjacent `_sanitize_snippet` (`atlas/modules/rag/client.py:91`).
- The new field is documented only in the mock README; operators writing real backends read `docs/admin/external-rag-api.md`, which does not mention it or the parenthesis-stripping caveat.

---

PR #826, opened 2026-08-20
fix(chat-history): make conversation saving independent of the client

garland3:
## The problem

A WebSocket gets a fresh `session_id` on every connection (`main.py`), but the browser's `activeConversationId` outlives the socket — it is React state in `ChatProvider`, which does not remount when `WSProvider` reconnects. So after a temporary disconnect the next turn arrives naming a conversation the server holds no history for.

`handle_chat_message` bound that `conversation_id` to the empty session after checking **ownership only**; it never loaded the history. `save_conversation` then replaced the conversation's whole message set at commit time. A two-message history was written over the entire stored conversation — title included, which is why the damage leaves no obvious trace afterwards.

The user-visible symptom is a conversation that is on screen the whole time, that the model has no memory of, and that is shorter than it was the next time it is opened.

## The fix

**Rehydrate.** When a turn names a conversation the session is not already carrying, the server loads it from the store before running the turn. The model gets its context back and the save that follows is complete — no reload in the browser, nothing for the user to do. Skipped for incognito turns (never persisted, so there is no record to continue) and for a `conversation_id` the store has not seen (the ordinary first turn). A store that cannot be read logs and degrades to no context rather than failing the turn. The session is marked restored so the title is not regenerated from this turn's prompt.

**Refuse to shrink.** `save_conversation` now rejects a write carrying fewer messages than are stored, logs at ERROR, and reports a failed save rather than applying it — better to lose the newest turn than the conversation. This is the durable backstop, in the layer that owns the data, and it covers the un-hydrated turn that follows an unreadable store. Rewind / edit-and-resubmit passes `allow_shrink=True`, since dropping a prompt and everything after it is the point of that turn.

**One loader.** Restore and rehydrate now share `utilities/conversation_loader.py`, so both produce identical history. It also carries the persisted timestamp through, which the restore path dropped: every message was restamped at load time, and because a save rewrites the whole row set, that restamped value is what got persisted — each load/save cycle destroyed the real message times.

## Notes for review

- The guard compares against the actual row count rather than `message_count`, since the rows are what the replace deletes. A stored count of zero fails open — an empty conversation has nothing to lose.
- `allow_shrink` is read from `rewind_to_user_index` without popping it; the orchestrator still needs it. An out-of-range index is ignored downstream and simply leaves the guard relaxed for a turn that does not shrink anything.
- Rehydration only fills an *empty* history. A session already holding messages for the conversation is the live, authoritative thread; reloading underneath it would duplicate every message.
- No frontend change. Making the server self-sufficient is the point — a client-side re-restore on reconnect would have left persistence dependent on the client behaving.

## Testing

`atlas/tests/test_conversation_rehydration.py`, 20 tests: the guard (refuse / allow on rewind / unaffected when growing / first save / empty write), the loader (timestamps, `message_type` folding, unreadable roles), and the service path (fresh session rehydrates, second turn does not reload, conversation switch, unknown id not treated as restored, incognito skipped, failing store does not fail the turn, ownership enforced before any load).

Full backend suite green: 2463 passed, 17 skipped. `ruff check atlas/` clean apart from two pre-existing `F821`s in `modes/streaming_helpers.py`. `scripts/check-docs.sh` OK.


garland3:
## Summary
Adds server-side rehydration so a turn arriving on a fresh session with a client-held `conversation_id` reloads the stored thread, plus a no-shrink guard in `ConversationRepository.save_conversation` and a shared `conversation_loader` used by both the restore and rehydrate paths.

## Verdict
Request changes.

## Required changes
1. Clear `session.history` before binding a new `conversation_id` at `atlas/application/chat/service.py:349` — `_hydrate_session_from_store` returns early on a non-empty history, so a session carrying conversation A that receives a turn naming conversation B stays bound to B with A's messages, and the following save writes all of A over B. Rehydration makes this worse: the carried thread is now the full stored history, so it is longer than B and passes the count guard.
2. Derive `allow_shrink` from a rewind that actually happened rather than from field presence at `atlas/application/chat/service.py:342` — the orchestrator ignores out-of-range or non-integer `rewind_to_user_index` (`orchestrator.py:262-288`), so a stale client ordinal (`999`, `-1`, `"not-an-int"`) disables the guard for a turn that truncates nothing. Have the orchestrator report whether it truncated and use that.
3. Handle the incognito-then-resume slice at `atlas/application/chat/service.py:1129` — `_save_conversation` submits `messages[start_index:]` against the full stored row count, so once a session takes an incognito turn on an existing conversation every later save is refused and the user gets a `conversation_save_rejected` frame on every turn with no recovery. Persist the post-floor segment under a fresh `conversation_id`, or fail once with a distinct message.
4. Re-attempt hydration when a session is bound to a conversation but holds no history at `atlas/application/chat/service.py:1055` — after a transient store failure the session is locked out of hydration permanently, and the count guard only delays the overwrite: once the accumulated turns reach the stored count the write is accepted and replaces the original rows.
5. Add an end-to-end test wiring the real `ConversationRepository` into the service (`atlas/tests/test_conversation_rehydration.py`) — store a 50-message conversation, run a turn on a fresh `session_id` naming it, and assert the stored conversation is 52 messages with its original title and first-message timestamp. Today the guard and the rehydration are tested in disjoint halves and the seam between them is unpinned.
6. Drop the `history.messages.clear()` in `test_switching_conversations_loads_the_new_one` (`atlas/tests/test_conversation_rehydration.py:379`) and assert the session ends up holding the new conversation's thread — as written the test manufactures the state that hides item 1.
7. Offload the store read at `atlas/application/chat/service.py:1061` with `await asyncio.to_thread(...)` — `get_conversation` issues three synchronous queries and JSON-decodes every message's metadata on the event loop, before the model call, and it fires for every reconnecting client at once after a restart.

## Evidence
Evidence: 8/10. The change is well-demonstrated for its size — a precise problem statement, 20 new tests, full-suite and lint results, and operator docs with the log line to grep. The missing artifact is any demonstration of the reconnect sequence itself: the tests assert the premise (a new session arrives with an old conversation id) rather than establishing it through the WebSocket layer.

## Notes
- Sanitize `conversation_id` before the new ERROR log at `atlas/modules/chat_history/conversation_repository.py:91` — every neighbouring caller uses `sanitize_for_logging`; a client-registered id with CR/LF can forge log lines.
- Give the shrink refusal its own client copy naming the recovery ("reopen this conversation from the sidebar") instead of the generic `Conversation could not be saved` at `atlas/application/chat/service.py:667`, and suppress repeats within a session.
- `_make_service` passes a bare `MagicMock()` config manager (`atlas/tests/test_conversation_rehydration.py:243`), so each run mkdirs ten `MagicMock/mock.app_settings.runtime_capture_dir/...` directories in the repo root.

---

garland3:
Pushed one more commit after re-reading the switch case.

Hydration originally filled only an *empty* history, which left the other direction wrong: a session that changed conversation without going through restore kept the previous conversation's messages, ran the turn against them, and saved the spliced thread under the new id. The no-shrink guard does not catch that — the write is longer, not shorter.

The store lookup now decides. A conversation the store has **replaces** the live history; one it does not have leaves the live thread untouched. That second half matters for `local`/`none` save modes, which mint a client-side `local_*` id part-way through a conversation — those turns are incognito and never reach the loader today, but relying on that alone would have been a trap for anyone who later changes when incognito is set.

Adds `test_switching_to_an_unknown_conversation_leaves_the_history_alone`; suite is now 2464 passed, 17 skipped.

garland3:
Addressed the review feedback in `a0ef796`.

**Taken**

- **The shrink exemption was derived from a client-controlled field.** Correct, and the failure mode is worse than "an index that removes nothing": if rehydration had failed, the session holds a partial history, so a truncation measured against it says nothing about the stored conversation and a short write could replace a long one. The permission now follows the work. `ChatOrchestrator` sets `rewind_removed` where the truncation happens, once `truncate_at_user_index` has actually returned rows; a failed hydration sets `hydration_failed` and revokes the exemption for that turn outright. Both markers are cleared at the start of every turn so neither leaks into the next. Six new tests cover it, including the orchestrator contract for out-of-range, malformed, and absent indexes.
- **Missing PR validation script.** Added `test/pr-validation/test_pr826_conversation_persistence_independent_of_client.sh`. It drives the reconnect end-to-end against a real DuckDB file — 25 turns on one session, then a fresh session carrying the same `conversation_id` — and asserts the stored conversation grows to 52 messages with its original title, exercises the guard directly (a 2-message write over 50 is refused; a rewind is allowed), and proves the exemption tracks the truncation. Passes 6/6 locally.
- **`conversation_id` logged unsanitized in the guard's ERROR line.** Now goes through `sanitize_for_logging`, matching the rest of the codebase.

**Not taken, with reasons**

- **`datetime.fromisoformat` and a trailing `Z`.** That was true before Python 3.11, which learned to parse it; `pyproject.toml` sets `requires-python = ">=3.11"`. Added a test pinning `2026-08-20T12:00:00Z` so a future change to the parser is caught rather than silently restamping messages.
- **Unconditional `COUNT(*)` on every update.** The count is the safety property, not an optimization target. Gating it on `existing.message_count` reintroduces exactly the bug this guards against: a stale-low `message_count` (which is written by the same replace this guard protects, so a partial write that got through once leaves it wrong) would skip the count and let a short write destroy the rows. The same statement already issues a `DELETE` of every message row followed by a bulk re-insert, so one `COUNT` is noise against that.
- **Synchronous DB read on the event loop.** Real, but not new here and not this PR's scope: the whole `ConversationRepository` is synchronous and the save on this same path already blocks the loop the same way. Making it async is a repository-wide change worth doing on its own, where the migration can be tested as a unit.
- **CHANGELOG not updated.** It was, in `7671f78` — the review ran against `58bd932`.

The six CodeQL log-injection alerts are the repo's known false positive for this rule: `sanitize_for_logging` is not modeled as a sanitizer, so 53 alerts are open against existing code including `atlas/main.py:531`, which is sanitized. The lines flagged here were already sanitized except the repository one, now fixed.

Local: 2471 passed, 17 skipped. `check-docs.sh` OK. Ruff clean.

garland3:
Worked through the rest of the review list in `f29357a`. All seven required items are now addressed.

**3 — incognito-then-resume was wedged.** Confirmed and fixed. The reachable path is: open a conversation from the sidebar while incognito, take a turn off the record, switch saving back on. The save floor is the message count at that point, so every later save submits a slice against the full stored row count — before the guard that slice replaced the whole conversation, and with it every remaining turn of the session was refused with no recovery. The segment is now stored as its own conversation. That is the honest semantics: the messages between it and the original can never be persisted, so it is not a continuation of anything. The client already adopts the id from the `conversation_saved` frame (`websocketHandlers.js:459`), so the next turn names the branch and the loop closes.

**4 — a failed hydration locked the session out.** Fixed: a session bound to a conversation but holding no history re-attempts the load. Your read of the consequence is right and worse than a permanent no-op — the guard only delayed the overwrite until the accumulated turns reached the stored count. The extra lookup only fires while the history is empty, which after the first turn means a genuinely new conversation (one miss) or a session that needs the retry.

**5 — end-to-end test with the real repository.** Added three, all wiring `ConversationRepository` into the real `ChatService` over a throwaway DuckDB file: the 50-message reconnect (52 messages after, original title, first message's timestamp unchanged), the retry after a transient failure, and the branch. That closes the seam you flagged — the guard and the hydration are no longer tested in disjoint halves.

**6 — `history.messages.clear()` in the switching test.** Already gone in `7101a8b`; the test now asserts the session ends up holding only conv-2's thread.

**7 — offload the store read.** Taken, `await asyncio.to_thread(...)`. This supersedes what I said about the same point in my previous comment — I read it as a repository-wide async migration, but scoped to the one new call site it is a one-line change and the thundering-herd argument after a restart is the right one.

**Note — `_make_service`'s bare `MagicMock()` config manager.** Reproduced: `MagicMock/mock.app_settings.runtime_capture_dir/<id>/` in the repo root, one per service built, because `CaptureService._build_store` reads a truthy mock as a path. The fixture now points it at scratch space and turns the feature off.

**Note — distinct client copy for the shrink refusal.** Not taken. With 3 and 4 fixed a refusal is no longer a state the user is stuck in, so the copy would be describing a transient store failure rather than a recovery the user can perform. Distinguishing it from the ownership rejection also needs a new return contract from `save_conversation` (both are `None` today), which I would rather not bolt on inside this change.

Local: 2474 passed, 17 skipped. `test_pr826_conversation_persistence_independent_of_client.sh` 6/6. Ruff clean, `check-docs.sh` OK. `CHANGELOG.md` and `docs/admin/chat-history.md` cover the branch and the retry.

PR #832, opened 2026-08-21
fix(transfer-mcp): limit read_file_from_disk to head/tail preview

### LINKED ISSUE #831: transfer_read_file_from_disk should limit file head and tail sent to llm
in the mcp server for file transfer, the transfer_read_file_from_disk should not inject the whole file into the llm context. The start and end can be viewed, but the point is to put it in the LLM session context so the agent can decide to send the file to another tool if it needs to.

garland3:
## Summary

Closes #831.

The transfer MCP server's `read_file_from_disk` injected the entire file into the model context as decoded text (or base64 for binary files). For a large source file or log, a single read could consume a large fraction of the model's context window even with the 10 MiB size cap.

The tool-result text is now a **head+tail preview**: the first and last `MCP_TRANSFER_PREVIEW_LINES` lines (default 50 each) of a UTF-8 file, joined by an omission marker when the file exceeds twice the line budget. Short files (at most 2x the budget) are returned in full. The complete file still travels as a base64 artifact so the agent can forward it to a downstream tool (e.g. `write_file_to_disk`) without re-reading. Binary files no longer inject a redundant `content_base64` into the tool result -- the bytes travel only as the artifact.

### Design decisions (responding to the review-bot questions on #831)

- **Preview unit**: lines, not bytes -- more readable for source code, configs, and logs (the primary use case).
- **Default budget**: 50 lines head + 50 lines tail (100-line threshold). Configurable via `MCP_TRANSFER_PREVIEW_LINES`.
- **Full file as artifact**: always returned on every read, so a downstream tool can pick it up without an explicit request.

## Changes

- `atlas/mcp/transfer/main.py`: added `_preview_lines()` and `_text_preview()` helpers; `read_file_from_disk` now returns a head+tail preview instead of the full decoded text; removed redundant `content_base64` for binary files (bytes travel only as the artifact); removed unused `Optional` import.
- `.env.example`: documented `MCP_TRANSFER_PREVIEW_LINES`.
- `atlas/mcp/transfer/README.md`: updated tool description and added a Read preview section.
- `atlas/tests/test_mcp_transfer.py`: 5 new tests (long file truncation, short file full return, artifact completeness, binary file no text, env var budget control).
- `test/pr-validation/test_issue831_transfer_read_preview.sh`: end-to-end validation script.

## Test plan

- [x] Long UTF-8 file: tool-result text is trimmed to head+tail with omission marker; full file in base64 artifact
- [x] Short UTF-8 file (under 2x budget): full content returned, no truncation flag
- [x] Binary file: no text or base64 in tool result; bytes only in artifact
- [x] `MCP_TRANSFER_PREVIEW_LINES` env var controls the head/tail budget
- [x] Existing tests still pass (small file reads unchanged)
- [x] `ruff check` passes
- [x] `./test/run_tests.sh backend` passes
- [x] PR validation script passes (`bash test/pr-validation/test_issue831_transfer_read_preview.sh`)


garland3:
## Summary
`read_file_from_disk` in the transfer MCP server now returns a head+tail line preview in the tool-result text instead of the whole file, with the complete bytes still carried as a base64 artifact, gated by a new `MCP_TRANSFER_PREVIEW_LINES` env var. Binary files no longer emit `content_base64` in the result.

## Verdict
Request changes.

## Required changes
1. Add a byte ceiling alongside the line budget in `_text_preview` at `atlas/mcp/transfer/main.py:141` — the budget is measured only in `splitlines()` units, so a minified bundle, single-line JSON dump, or newline-free CSV is one line and is returned in full up to the 10 MiB `MCP_TRANSFER_MAX_BYTES` cap, which is the exact context flood this change is meant to prevent. Apply the cap to both the early-return path and the assembled head/tail, and set `truncated` when it bites.
2. Stop returning lossy text under the `content` key at `atlas/mcp/transfer/main.py:322` — `content` previously held the exact file bytes, so a caller chaining `read_file_from_disk` into `write_file_to_disk(content=...)` now silently writes a destination file containing the `... [N lines omitted ...] ...` marker instead of the real body. Emit the trimmed text under a distinct key (e.g. `content_preview`) and omit `content` whenever the text is incomplete, so a blind forward fails loudly.
3. Set `results["truncated"] = truncated` unconditionally at `atlas/mcp/transfer/main.py:325` and emit `total_lines`, `omitted_lines`, and `preview_lines` as integers — the counts currently exist only as prose inside the marker, so callers (and the new tests, which assert on `"16 lines omitted"`) must parse a human-readable string, and consumers must distinguish an absent key from `False`. `atlas/mcp/file_viewer/main.py:477` already emits the flag unconditionally.
4. Add a test covering `_preview_lines()` directly for unset, `""`, `"abc"`, `"0"`, and `"-5"` at `atlas/tests/test_mcp_transfer.py:191` — all three fallback branches are unexercised, and the non-positive fallback is documented user-facing behavior in `atlas/mcp/transfer/README.md`. A regression letting `0` through makes `2 * preview_lines == 0`, emptying `content` for every read.
5. Add a test exercising the shipped default budget of 50 against a file large enough to truncate, and one at the `2N + 1` boundary — every new test forces the budget to 1–5, so nothing pins `DEFAULT_PREVIEW_LINES` or the `<=` comparison at `atlas/mcp/transfer/main.py:145`.
6. Give the caller a way to reach the omitted middle at `atlas/mcp/transfer/main.py:258` — add optional line `offset`/`limit` parameters and name the recovery action in the marker. Today a request for the region around line 300 of a 900-line file is unsatisfiable by any sequence of calls, and the marker's "full content in artifact" points at bytes the model cannot read.
7. Harden `test/pr-validation/test_issue831_transfer_read_preview.sh:1` — add `set -uo pipefail`, use `python3` consistently (Test 5 uses `python`, which may not exist when the conditional venv `source` silently no-ops), and clean up the four leaked `tempfile.mkdtemp()` directories.

## Evidence
Evidence: 8/10. Well demonstrated for a backend change of this size — five new unit tests cover the behavioral branches, docs are updated in the same PR, and a committed validation script makes verification reproducible. The one missing artifact is a measured before/after size for a realistically large file, since the whole change is justified by a context-cost argument that is currently asserted rather than quantified.

## Notes
- `text.splitlines(keepends=True)` also breaks on `\f`, `\x85`, and U+2028, so line counts and the truncation trigger disagree with `wc -l` — `atlas/mcp/transfer/main.py:145`.
- Just above the threshold the preview can be larger than the file it replaces, and the marker reads "1 lines omitted" — `atlas/mcp/transfer/main.py:150`.
- `_preview_lines` duplicates `_max_read_bytes` verbatim; a shared `_positive_int_env` helper would make the next knob a one-liner — `atlas/mcp/transfer/main.py:120`.

---

garland3:
Addressed all review feedback from AGENT-REVIEW-BOT-3 in commit 39b7ebf:

**1. Byte ceiling** — Added `MCP_TRANSFER_PREVIEW_BYTES` (default 16 KiB). The preview text is now capped by both lines and bytes. A single-line minified file or JSON dump that bypassed the line budget now hits the byte ceiling. The cap applies to the early-return path and the assembled head/tail.

**2. `content_preview` key** — Trimmed text is now emitted under `content_preview`, not `content`. The `content` key is omitted when truncated, so `write_file_to_disk(content=result["results"]["content"])` raises `KeyError` instead of silently writing the omission marker.

**3. Structured integer metadata** — `truncated`, `total_lines`, `omitted_lines`, and `preview_lines` are now emitted as integers unconditionally for text files. No more parsing prose.

**4. `_preview_lines()` fallback tests** — Added direct tests for unset, `""`, `"abc"`, `"0"`, `"-5"`, and a valid positive value.

**5. Default budget + boundary tests** — Added tests exercising the shipped default of 50, the 2N+1 boundary (101 lines, first truncation), and the 2N boundary (100 lines, not truncated).

**6. offset/limit pagination** — Not addressed. The issue's design is head/tail preview + full artifact for forwarding to another tool, not middle-region pagination. The artifact carries the complete file bytes so the agent can forward it to `write_file_to_disk` or any other tool that accepts the artifact.

**7. Script hardening** — Added `set -uo pipefail`, use `python3` consistently, clean up all `tempfile.mkdtemp()` dirs with `shutil.rmtree` in `finally` blocks.

**Also addressed from the Notes section:**
- Refactored `_preview_lines`/`_max_read_bytes`/`_preview_bytes` to share a `_positive_int_env` helper.
- Fixed singular/plural in the omission marker ("1 line omitted" not "1 lines omitted").

25 unit tests pass, 7/7 PR validation checks pass, ruff clean.

garland3:
## Summary
Adds a head/tail preview to the transfer MCP `read_file_from_disk` tool so large files no longer return their full text in the tool result: the text channel is bounded by `MCP_TRANSFER_PREVIEW_LINES` and `MCP_TRANSFER_PREVIEW_BYTES`, `content` is renamed to `content_preview` when trimmed, and the complete bytes continue to travel as the artifact. Ships with unit tests, a validation script, and docs.

## Verdict
Request changes.

## Required changes
1. Enforce `MCP_TRANSFER_PREVIEW_BYTES` on encoded bytes at `atlas/mcp/transfer/main.py:158,171-174` — `len(text)`, `len(head) + len(tail)`, and the `head[:half]` / `tail[-half:]` slices all count code points, so a CJK or emoji file emits up to ~4x the configured ceiling (measured: 48 KiB against the 16 KiB default), which is the exact overrun the knob exists to prevent. Trim on the encoded string and decode with `errors="ignore"` so a multi-byte sequence is never split.
2. Recompute `omitted_lines` after the byte trim at `atlas/mcp/transfer/main.py:160,171-174` and build the marker from the recomputed value — the count is derived from the line budget before head/tail are sliced, so a 200-line file of long lines reports `omitted_lines: 100` while the preview actually contains 11 lines, making `total_lines - omitted_lines` wrong for any programmatic consumer.
3. Clamp the assembled preview to `preview_bytes` at `atlas/mcp/transfer/main.py:166-186` — `marker_reserve` is a fixed 160 while the marker is appended unconditionally, so any ceiling below ~160 returns a result larger than the configured limit and containing none of the file.
4. Add a multi-byte fixture and a combined line+byte truncation case to `atlas/tests/test_mcp_transfer.py:368` — the existing byte-cap test uses `"X" * 4000`, where characters and bytes coincide, so it passes against the defects in items 1 and 2.

## Evidence
Evidence: 8/10. The change is well demonstrated for a backend change of this size — 10 unit tests across the behavioral branches, a re-runnable validation script, and docs updated in the same commit; the notable gap is that the context-cost benefit motivating the change is asserted rather than measured, so a short before/after byte count on two or three representative files would close it.

## Notes
- The success result now has three shapes (`content`, `content_preview`, or neither for binary) distinguished only by key presence; an explicit `content_kind` discriminator, with `truncated` always present, would spare every caller a probe chain — `atlas/mcp/transfer/main.py:362-383`.
- `text.splitlines(keepends=True)` materializes every line to keep only 100 of them; a 10 MiB log costs tens of MB of transient allocation per read — `atlas/mcp/transfer/main.py:155`.
- `.env.example` does not mention that non-positive values fall back to the default, so `MCP_TRANSFER_PREVIEW_LINES=0` silently becomes 50 instead of disabling the preview.

---

garland3:
Addressed the second round of review feedback (commit 5114619):

**1. Byte ceiling on encoded bytes** — `len(text)` counts code points, so CJK/emoji content (3-4 bytes/char) could emit up to 4x the configured ceiling. The byte trim now works on `head.encode("utf-8")` / `tail.encode("utf-8")` and decodes with `errors="ignore"` so a split multi-byte sequence is not replaced with U+FFFD.

**2. Recompute `omitted_lines` after byte trim** — `omitted_lines` was derived from the line budget before head/tail were sliced, so a 200-line file of long lines reported `omitted_lines: 100` while the preview contained ~11 lines. Now counts complete lines shown (`head.count("\n") + tail.count("\n")`) and subtracts from `total_lines`, so `total_lines - omitted_lines` equals the lines in the preview.

**3. Final clamp** — The marker (up to ~160 bytes) was appended unconditionally, so any ceiling below ~160 returned a result larger than the limit. The assembled result is now encode+truncate+decoded to ensure it never exceeds `preview_bytes`.

**4. New tests** — Multi-byte CJK fixture, combined line+byte truncation with recomputed `omitted_lines`, and very-small-byte-ceiling edge case (50 bytes).

**5. .env.example** — Noted that non-positive values fall back to the default.

28 unit tests pass, 7/7 PR validation checks pass, ruff clean.

garland3:
## Summary
Adds a head+tail text preview to the `transfer` MCP server's `read_file_from_disk`, bounded by a line budget (`MCP_TRANSFER_PREVIEW_LINES`) and a byte ceiling (`MCP_TRANSFER_PREVIEW_BYTES`), with the full file still returned as a base64 artifact and structured truncation metadata (`truncated`, `total_lines`, `omitted_lines`, `preview_lines`) in the result.

## Verdict
Request changes.

## Required changes
1. Measure the byte ceiling in encoded bytes at `atlas/mcp/transfer/main.py:159`, `:171` and `:173-174` — `len()` on a `str` counts code points, so a CJK or emoji file returns up to ~4x `preview_bytes` (measured: 48,001 bytes against the shipped 16,384 default), and on the early-return path it escapes truncation entirely while reporting `truncated=False`. Encode each side, slice the byte strings, and decode back with `errors="ignore"` so a trim never splits a multi-byte sequence.
2. Recompute `omitted_lines` after the byte trim at `atlas/mcp/transfer/main.py:166` and build the marker from the new value — today it is derived from the line budget only, so a 200-line file reports `omitted_lines=100` while the preview contains 35 lines, and `total_lines - omitted_lines` is wrong for any consumer relying on the documented integer contract.
3. Add a multi-byte byte-ceiling test to `atlas/tests/test_mcp_transfer.py` (e.g. `"漢" * 20000` and `"🙂" * 20000` with `MCP_TRANSFER_PREVIEW_BYTES=2000`, asserting `len(preview.encode("utf-8")) <= 2000`) — the existing `test_read_byte_cap_trims_single_long_line` uses `"X" * 4000`, where code points and bytes coincide, so it passes against both defects above.
4. Add a case where the line budget and the byte ceiling both fire (200 lines of 500 characters, `PREVIEW_LINES=50`, `PREVIEW_BYTES=4000`) asserting `total_lines - omitted_lines` matches the lines actually present in the preview — the interaction of the two budgets is the path where the metadata goes wrong and it has no coverage.
5. Enforce the size ceiling as a post-condition at `atlas/mcp/transfer/main.py:169-187`: build the marker first, reserve its real length instead of the fixed `marker_reserve = 160`, and clamp the assembled `head + marker + tail` to `preview_bytes` — at `preview_bytes=50` the function currently returns 72 bytes of marker and none of the file.

## Evidence
Evidence: 8/10. The change is well demonstrated for its size — ten unit tests plus a committed, re-runnable validation script — but the byte ceiling is exercised only on ASCII payloads where the bug hides, and the context-cost saving that motivates the PR is asserted rather than measured. The single most useful addition is a before/after size figure for one realistic file (a large log or a minified bundle) in the description; the test plan checklist also still describes the state before the last two commits.

## Notes
- Emit a positive `content_kind` discriminator (`"full" | "preview" | "binary"`) and always emit `truncated`; the three result shapes are currently told apart only by key absence — `atlas/mcp/transfer/main.py:362-383`.
- Emit `preview_bytes` and a truncation reason so a `truncated=true, omitted_lines=0` result is interpretable without parsing the marker prose — `atlas/mcp/transfer/main.py:377`.
- `MCP_TRANSFER_PREVIEW_LINES=0` silently becomes 50 and there is no documented way to disable the preview; note the fallback in `.env.example:319-329` and state that both budgets must be raised together.

---

garland3:
## Summary
Adds a head/tail preview to `read_file_from_disk` in the transfer MCP server, gated by `MCP_TRANSFER_PREVIEW_LINES` and `MCP_TRANSFER_PREVIEW_BYTES`, so a large read no longer dumps a whole file into the tool result; the full bytes still travel as a base64 artifact, and the result gains `truncated`/`total_lines`/`omitted_lines` metadata plus a `content` → `content_preview` key split when trimmed.

## Verdict
Request changes.

## Required changes
1. Count shown lines in the same unit as `total_lines` at `atlas/mcp/transfer/main.py:186` — `shown_lines` counts `"\n"` while `total_lines` comes from `splitlines(keepends=True)`, so `omitted_lines` is wrong for three ordinary inputs: a file whose last line has no trailing newline (verified: 20-line file reports `omitted_lines=17` when 4 lines are shown, correct value 16), a byte-trimmed tail whose leading fragment is counted as a whole line, and files using `\r`/`\f`/`\x85`/U+2028 terminators (verified: 30 `\r`-separated rows report `omitted_lines == total_lines` while four rows are visible). Pick one line model — `splitlines()` on the emitted segments, or `split("\n")` everywhere — and use it for `total_lines`, the truncation trigger, and `shown_lines`.
2. Build the omission marker before allocating the byte budget at `atlas/mcp/transfer/main.py:172` and reserve `len(marker.encode("utf-8"))` instead of the hardcoded `marker_reserve = 160` — at any `preview_bytes` at or below 160 the per-side budget floors to zero and the final clamp slices the marker itself, returning zero bytes of file content and a sentence cut mid-word (verified at `preview_bytes=50`: `'\n... [1 line omitted; 1 total lines; full content '`). When the marker still does not fit, emit a short complete string rather than a fragment.
3. Assert the actual property in `test_read_combined_line_and_byte_truncation_recomputes_omitted` at `atlas/tests/test_mcp_transfer.py:415` — it documents `total_lines - omitted_lines == lines shown` but only asserts `shown < 100`, which passes even when zero complete lines survive; split `content_preview` on the marker, count complete lines in the head and tail segments, and assert the sum equals `total_lines - omitted_lines`. Add the same assertion for a fixture with no trailing newline, and extend `test_read_very_small_byte_ceiling_does_not_exceed_budget` (`:436`) to require the marker be complete and at least one byte of file content present at a mid-range ceiling.
4. Mark the removals as breaking in `CHANGELOG.md:9` with a migration line — `content_base64` is gone for binary reads (use `artifacts[0]["b64"]`) and `content` becomes `content_preview` when trimmed; the `content_base64` removal fails silently for callers using `.get()`. Mirror both lines in `atlas/mcp/transfer/README.md`.

## Evidence
Evidence: 8/10. Well demonstrated for a backend change of this size — 13 unit tests plus a standalone validation script cover the line budget, the byte ceiling, multi-byte input, and boundaries. The one gap that matters is the number the change exists to improve: the context saving is asserted, never measured; add before/after tool-result byte counts for one realistic file (a large log or a minified bundle) to the description, which is also still describing the state two commits ago.

## Notes
- The result reports `preview_lines` (the configured budget, not what was delivered) but never `preview_bytes`, so a caller cannot tell which budget fired without substring-matching the marker prose — `atlas/mcp/transfer/main.py:399`.
- Three success shapes are distinguished only by key absence (`content` / `content_preview` / neither, with `truncated` missing entirely for binary); an unconditional `content_kind` discriminator would make consumers positive rather than probing — `atlas/mcp/transfer/main.py:394-401`.
- Head and tail are joined and encoded in full before being sliced to the ceiling, costing ~45 MB peak on a 10 MiB single-line file vs ~28 MB if bounded before encoding — `atlas/mcp/transfer/main.py:171-174`.

---

Issue #153, opened 2025-12-05
allow users to set a custom system prompt at run time.


garland3:

garland3:
Could you clarify the intended scope here — should the custom system prompt be set per-conversation, saved per-user as a default, or configured at the workspace/admin level? Also, should it replace the existing default prompt or be combined with it?

---
_Posted automatically by **AGENT-REVIEW-BOT-3**_

garland3:
Related to #595 ("add a user info panel"). Both deal with user-supplied context being injected into the model. #153 is specifically a runtime **system prompt** override; #595 is a persistent **user info** panel (admin defaults + AI-injected knowledge, saved to Postgres, also injected into RAG/MCP calls). Keeping them separate, but they likely share a settings UI surface — worth coordinating the UX.

garland3:
One open design point: should the custom prompt fully replace the default system prompt, or be appended on top of it? And is this meant to be per-conversation only, or saved as a per-user default that carries across sessions?

---
_Posted automatically by **AGENT-REVIEW-BOT-3**_

garland3:
The db should maintain a collection of prompts that are for the user. Users can add, edit, del the prompt. The prompt is injected in place of the system prompt.

PR #833, opened 2026-08-22
feat(chat): inject current date/time into system prompt (closes #823)

### LINKED ISSUE #823: inject time and date into system prompt automatically.
inject time and date into system prompt automatically.
If more than n minutes pass in a conversation then re inject the time to allow the AI to know about how much time has elapsed...

garland3:
## What

Closes #823. The system prompt now carries the current date/time so the model knows what "now" is, and when a meaningful gap has opened between turns an explicit elapsed-time note tells it how much time has elapsed — a long pause can change what a correct answer looks like (a status may have resolved, a deadline may have passed).

## How

- A new helper, `atlas/application/chat/preprocessors/system_prompt_time.py`, appends a `### Current Date & Time` line to the rendered system prompt on every turn, and appends a `### Time Since Previous Message` note when the gap between this turn and the previous user message meets/exceeds `SYSTEM_PROMPT_TIME_REFRESH_MINUTES`.
- `MessageBuilder` calls the helper for both the default and a user-supplied custom system prompt; the orchestrator now passes `config_manager` to the `MessageBuilder` so the settings are available.
- The current time is **always** injected; the refresh setting only gates the elapsed-time note (`0` keeps the time, drops the note).
- The gap is derived from the conversation history's existing message **timestamps** (which the conversation loader already preserves across save/reload), so **no new per-session state** is required and no `session.context` mutation happens.
- Two new `AppSettings` fields, both env-configurable:
  - `SYSTEM_PROMPT_TIMEZONE` — IANA name (e.g. `UTC`, `America/Denver`), default `UTC`; unknown names fall back to UTC.
  - `SYSTEM_PROMPT_TIME_REFRESH_MINUTES` — default `5`; `0` disables the note.

## Product decisions (issue #823 was ambiguous; repo review bot flagged both)

- **Clock/timezone:** server time, default UTC, operator-configurable via `SYSTEM_PROMPT_TIMEZONE`. No per-user timezone plumbing was added (none exists in the codebase today). The frontend/timezone option would be a larger change; left for a follow-up if wanted.
- **"Re-inject after N minutes":** the system prompt is rebuilt every turn, so the time is always fresh. The "N minutes" gates an **explicit elapsed-time note** (the chosen semantics) so the model can reason about the gap, rather than gating the refresh of the timestamp itself.

## Contracts preserved

- Issue #153's "custom system prompt fully replaces the default" contract holds: the custom text still **leads** the message and the default template text does **not** leak in. The time/elapsed lines are appended **after** the custom text as runtime enrichment. The two existing custom-prompt tests were relaxed from exact-equality (`==`) to `startswith` + a positive "Current Date & Time" assertion, and still assert the default text does not leak.
- The packaged prompt templates (`atlas/config/prompts/system_prompt.md`) are **unchanged** — the time is appended by the builder, not baked into the template. The `PromptProvider` stays pure (its exact-equality test is untouched).
- MCP prompt overrides (prepended as separate system messages by `PromptOverrideService`) are unaffected.

## Files

- `atlas/modules/config/settings.py` — two new `AppSettings` fields.
- `atlas/application/chat/preprocessors/system_prompt_time.py` — new helper (<110 lines).
- `atlas/application/chat/preprocessors/message_builder.py` — wires the helper; new optional `config_manager` arg.
- `atlas/application/chat/orchestrator.py` — passes `config_manager` to `MessageBuilder`.
- `.env.example`, `docs/admin/configuration.md` — document the two knobs.
- `atlas/tests/test_system_prompt_time.py` — new focused tests.
- `atlas/tests/test_system_prompt_loading.py` — relaxed two custom-prompt exact-equality assertions.
- `test/pr-validation/test_issue823_system_prompt_time.sh` — end-to-end validation script.

## Test plan

- [x] `atlas/tests/test_system_prompt_time.py` (new) — helper + builder: time injected, timezone respected, unknown TZ → UTC, elapsed note fires only past threshold, `0` disables note, naive timestamps treated as UTC, custom prompt enriched, no-config → no injection, provider stays pure.
- [x] `atlas/tests/test_system_prompt_loading.py` — existing prompt tests still green; custom-prompt tests updated.
- [x] `ruff check` clean on all touched files.
- [x] `./test/run_tests.sh backend` — green.
- [x] Full unit suite: `python -m pytest atlas/tests/` → 2498 passed, 17 skipped.
- [x] `bash test/pr-validation/test_issue823_system_prompt_time.sh` → 12/12 passed.
- [ ] Manual smoke: rebuild frontend (`bash agent_start.sh`), send two turns >5 min apart in the running app, confirm the system prompt to the LLM carries the time and (after the gap) the elapsed note.


garland3:
## Summary
Appends a `### Current Date & Time` block to the system prompt on every turn, plus an "elapsed since previous message" note when the gap exceeds a threshold, controlled by two new settings (`SYSTEM_PROMPT_TIMEZONE`, `SYSTEM_PROMPT_TIME_REFRESH_MINUTES`). Backend-only: prompt assembly, config, docs, and a new test module.

## Verdict
Request changes.

## Required changes
1. Make `test/pr-validation/test_issue823_system_prompt_time.sh` capable of failing — every `grep -q ... || true` at lines 49–65 resets `$?` to 0, so the following `print_result $?` reports PASSED regardless of file contents. Capture the status directly (`if grep -q ...; then print_result 0; else print_result 1; fi`).
2. Delete the unconditional `print_result 0 "grep SYSTEM_PROMPT_TIMEZONE setting in settings.py"` at `test/pr-validation/test_issue823_system_prompt_time.sh:47` — it emits a hardcoded PASS that inflates the reported count.
3. Capture the exit status of the fallible `python` heredoc (line ~141) and the `pytest` subshells (lines ~143–148) as `cmd && rc=0 || rc=$?` — under `set -euo pipefail` a real failure kills the script before `print_result` runs, so `FAILED` is never incremented and `exit $FAILED` always returns 0. After the fix, break one target and confirm the script goes red.
4. Correct the elapsed-note wording at `atlas/application/chat/preprocessors/system_prompt_time.py:37` — `_previous_user_timestamp` measures from the previous *user* message, but the text asserts "since the previous message in this conversation". The literal previous message is the assistant reply, so in agent mode the reported figure absorbs generation time (a 40-minute tool loop makes a 5-minute-old exchange read as 45 minutes). Either reword to "since your previous prompt" or anchor to the last message of any role. Apply the same correction to the prose intro at `docs/admin/configuration.md:108`.

## Evidence
Evidence: 7/10. The unit coverage is substantial (229 lines over timezone resolution, fallback, threshold boundaries, naive timestamps, first-turn suppression) and full-suite results are reported with concrete numbers, so the change is adequately demonstrated for its size. The most valuable missing artifact is a sample of the actual rendered prompt tail — before/after, with and without a gap — since the entire output of this change is text; the unticked manual smoke in the test plan should also be run or explicitly withdrawn.

## Notes
- Placing a minute-granularity timestamp in `messages[0]` invalidates provider prefix caching on every turn, re-billing the whole history at uncached rates — moving the block to a trailing system message would fix it (`atlas/application/chat/preprocessors/message_builder.py:174`).
- No operator off-switch: `SYSTEM_PROMPT_TIME_REFRESH_MINUTES=0` only suppresses the elapsed note, so backing the feature out requires a revert despite 20+ `feature_*_enabled` toggles existing alongside it (`atlas/modules/config/settings.py:548`).
- A misconfigured timezone re-warns and reconstructs the zone once per turn forever, and is never surfaced at config-load time where an operator would see it (`atlas/application/chat/preprocessors/system_prompt_time.py:44`).

---

garland3:
Addressing the review feedback in commit 2643b73 (plus the main merge to clear the CHANGELOG conflict — PR is now mergeable).

**Required changes (from the request-changes review):**

1. **Validation script can now fail.** Every `grep -q ... || true` + `print_result $?` pair (which always reported PASSED) is replaced with a `check_grep` helper that puts the grep in an `if` and reports its real status. The unconditional `print_result 0 "grep SYSTEM_PROMPT_TIMEZONE setting in settings.py"` is gone.
2. **Fallible commands don't abort the script.** The `python` heredoc and the `pytest`/`run_tests.sh` subshells now run as `rc=0; cmd || rc=$?` so a real failure is recorded as FAILED and the summary still runs (previously `set -euo pipefail` killed the script before `print_result`). Verified by breaking one grep and one Python assertion — the script reports `9 passed, 2 failed` and exits `2`.
3. **Elapsed-note wording corrected.** `_previous_user_timestamp` measures from the previous *user* message, so the body now says "since your previous prompt" (not "since the previous message in this conversation", which would absorb generation time in agent mode). Applied to `system_prompt_time.py`, the `AppSettings` field description, and `docs/admin/configuration.md`. The `### Time Since Previous Message` header (which the unit tests assert on) is unchanged.

**Naming (from Copilot/Codex):**

4. **Renamed `test_issue823_system_prompt_time.sh` → `test_pr833_system_prompt_time.sh`** so `test/run_pr_validation.sh` actually discovers it (`test_pr*.sh` glob). The old name was never executed by the runner.

**Notes from the review I did not act on (deliberate):**
- *Prefix-cache invalidation from putting the time in `messages[0]`* — real concern, but moving the block to a trailing system message would change where every provider sees the system instructions and is a larger design change than this PR scope; left as a follow-up if the maintainers want it.
- *Operator off-switch* — `SYSTEM_PROMPT_TIME_REFRESH_MINUTES=0` only suppresses the note; a full feature toggle is possible but not added here to keep the surface minimal.
- *Validate timezone at config load* — the per-turn warning + UTC fallback is the existing pattern; surfacing it at load time is a separate improvement.

`./test/run_tests.sh backend` green, `ruff check` clean on touched files, focused unit suite 23 passed, the renamed validation script 11/11 passed.

garland3:
## Summary
Injects the current date/time into the system prompt on every turn, plus an "approximately N minutes have elapsed" note when the gap since the previous user message meets `SYSTEM_PROMPT_TIME_REFRESH_MINUTES`. Adds two settings, docs, focused unit tests, and a PR-validation script.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — strong for a backend prompt-assembly change: 229 lines of focused tests covering the meaningful branches, concrete suite numbers, and a validation script that was negatively verified (deliberately broken, confirmed red). The one missing artifact is a verbatim paste of the rendered system-prompt tail (first turn, post-gap, and with a custom prompt); the unticked manual-smoke item in the test plan should also be run or struck.

## Notes
- Minute-granularity timestamp lands in `messages[0]`, so provider prefix caching misses on essentially every turn and re-processes the whole history — `atlas/application/chat/preprocessors/message_builder.py:171`. Emitting it as a trailing system message, or coarsening to date+hour, removes the cost; worth sizing before this ships widely.
- `check_grep "config_manager=config_manager"` in `test/pr-validation/test_pr833_system_prompt_time.sh:104` also matches the unrelated `ToolAuthorizationService` call at `orchestrator.py:95`, so it passes even if the `MessageBuilder` wiring is reverted — anchor it to the construction under test.
- Elapsed note always renders truncated raw minutes ("4320 minutes" after three days), and `SYSTEM_PROMPT_TIME_REFRESH_MINUTES` gates the note rather than any refresh — scaling the unit and renaming the setting (with an alias) would both help operators — `atlas/application/chat/preprocessors/system_prompt_time.py:37`, `atlas/modules/config/settings.py:562`.

---

garland3:
Thanks for the re-review (approve with nits). Acting on the actionable nit and acknowledging the two design notes:

**Fixed in 5ae2d57 — grep anchor (nit #2):** `check_grep "config_manager=config_manager"` matched the unrelated `ToolAuthorizationService` call at `orchestrator.py:95` too, so it passed even if the `MessageBuilder` wiring was reverted. The grep now anchors to `prompt_provider=prompt_provider, config_manager=config_manager`, which is unique to the `MessageBuilder` construction. Negatively verified: with the `MessageBuilder` line reverted (its `config_manager=` dropped), the anchored grep fails; the old broad grep still matches the `ToolAuthorizationService` line and would falsely pass.

**Acknowledged, not actioned in this PR (design notes):**
- *Prefix-cache invalidation (nit #1)* — the minute-granularity timestamp in `messages[0]` does defeat provider prefix caching every turn. Moving it to a trailing system message (or coarsening to date+hour) is a real improvement, but it changes where every provider sees system instructions and is a larger design change than this PR's scope. Worth a follow-up issue to size before this ships widely.
- *Raw-minutes elapsed + setting name (nit #3)* — the note renders "4320 minutes" after 3 days, and `SYSTEM_PROMPT_TIME_REFRESH_MINUTES` gates the note rather than any refresh. Scaling the displayed unit (minutes→hours/days) and renaming the setting (with a back-compat alias) would both help operators; both expand the public surface and are better as a follow-up than a post-approval amendment here.

garland3:
## Summary
Appends a `### Current Date & Time` block to the system prompt on every turn, plus an optional "time since previous message" note once the gap meets `SYSTEM_PROMPT_TIME_REFRESH_MINUTES`, wired through `MessageBuilder` with two new settings, docs, and a focused test module.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — strong for a backend prompt-assembly change: a 229-line test module over the meaningful branches, concrete suite numbers (2498 passed; 11/11 validation checks), and a negatively-verified validation script that was proven able to fail. The one missing artifact is a verbatim paste of the rendered prompt tail (first turn, after a gap, with a custom prompt) — the entire product of this change is text nobody has shown as text. The test plan also still cites the pre-rename `test_issue823_system_prompt_time.sh` and a stale 12/12 count; fix that line so the numbers that are correct stay trustworthy.

## Notes
- Minute-granularity timestamp in `messages[0]` breaks automatic prefix caching for providers that key on an exact prefix, so long histories re-prefill every turn — coarsening to `%Y-%m-%d %H:00 %Z` or moving the block after the history would keep the prefix stable (`atlas/application/chat/preprocessors/message_builder.py:174`).
- `SYSTEM_PROMPT_TIME_REFRESH_MINUTES` gates only the elapsed note, not any refresh; both `.env.example` and the configuration doc spend a sentence correcting the name (`atlas/modules/config/settings.py:563`).
- Threshold tests use 30s and 10min against a 5-minute setting, so the documented "meets or exceeds" (`>=`) boundary is untested (`atlas/tests/test_system_prompt_time.py:73`).

---

PR #835, opened 2026-08-22
feat(workspaces): re-enable a conversation's workspace on reload

### LINKED ISSUE #829: Reloading an old conversation should re-enable its workspace
When an old conversation is loaded and it previously used a workspace, the loaded conversation should re-enable that workspace automatically.\n\nExpected behavior:\n- Loading the conversation restores the associated workspace state.\n- The user does not have to manually re-open or re-enable the workspace.\n\nThis should apply to conversations that were previously tied to a workspace and are later reopened from history.

garland3:
## Summary

Closes #829.

Loading a conversation from history now switches back to the workspace it was tied to, instead of leaving the user to re-pick the prompt, RAG sources, and tools. The active workspace id travels with each chat turn and is persisted in the conversation's metadata, so reopening the conversation restores its workspace automatically.

### How it works
- **Frontend** sends `workspace_id` (the current `activeWorkspaceId`) on every WebSocket chat frame.
- **Backend** (`ChatService.handle_chat_message`) stashes it in `session.context["workspace_id"]` and `_save_conversation` persists it in the conversation's `metadata_json` alongside the existing `agent_mode` field. No schema migration — the existing flexible `metadata_json` column carries it.
- **On reload**, `loadSavedConversation` reads `metadata.workspace_id` and calls a new `restoreWorkspace` helper that switches to that workspace.

### Edge cases (per the issue discussion)
- **A different workspace is already active** → switch automatically (the workspace's selections replace the current ones, consistent with the existing switcher).
- **The workspace has since been deleted** → best effort: silently skip the switch, leave the current selections untouched. The conversation's stored id is updated to the current (or null) value on the next save.
- **The workspace list hasn't finished loading** when the conversation opens → the switch is deferred until it does, so a slow fetch no longer drops the restore.
- **A conversation with no recorded workspace** (saved before this change, or started without a workspace) → leaves the active workspace untouched. Backward compatible.

The id is client-supplied and advisory only — it is never used for authorization, so a null or stale value is harmless.

## Test plan
- [x] Backend: repository persists/returns `workspace_id` in metadata; upsert updates it; null round-trips.
- [x] Backend: `handle_chat_message` captures `workspace_id` into `session.context`; `_save_conversation` persists it; omitting it defaults to null.
- [x] Frontend: loading a conversation tied to a workspace switches to it; a deleted workspace is skipped; no-workspace conversations leave the active one untouched; a deferred restore fires when the list loads; `sendChatMessage` forwards `workspace_id`.
- [x] PR validation script: `test/pr-validation/test_issue829_restore_workspace_on_load.sh` exercises end-to-end save/get and runs the suites.
- [x] `ruff check` and `npm run lint` clean (no new warnings).
- [x] `./test/run_tests.sh backend` passes.
- [x] Frontend builds (`npm run build`).

## Files changed
- `atlas/main.py` — forward `workspace_id` from the chat frame.
- `atlas/application/chat/service.py` — capture into `session.context`, persist in save metadata.
- `frontend/src/contexts/ChatContext.jsx` — send `workspace_id`; `restoreWorkspace` on load (with deferred/best-effort handling).
- `atlas/tests/test_workspace_conversation_binding.py` — new service-level tests.
- `atlas/tests/test_chat_history.py` — repository-level metadata tests.
- `frontend/src/test/workspace-restore-on-conversation-load.test.jsx` — new frontend tests.
- `test/pr-validation/test_issue829_restore_workspace_on_load.sh` — end-to-end validation.


garland3:
## Summary
Binds each conversation to the workspace active at the time of its turns and restores that workspace when the conversation is reopened, with the id persisted in conversation metadata and a deferred client-side restore for when the workspace list has not loaded yet.

## Verdict
Request changes.

## Required changes
1. Clear `pendingWorkspaceRestoreRef` on every entry into `restoreWorkspace` at `frontend/src/contexts/ChatContext.jsx:196`, not only on the resolved-switch branch — the falsy-`workspaceId` early return leaves a queued restore alive, so opening a workspace-bound conversation and then a workspace-less one (or a new chat) while the list is still loading applies the first conversation's workspace and overwrites the user's tools, prompt and RAG selections. Also clear it in `startNewChat` and `clearActiveWorkspace` so explicit user actions cancel a queued restore.
2. Add `set -o pipefail` (or capture the real exit status before `tail`) in `test/pr-validation/test_issue829_restore_workspace_on_load.sh:141` — steps 5–7 pipe pytest, vitest and the build into `tail`, so `$?` is `tail`'s status and those three checks report PASSED even when they fail.
3. Fix `frontend/src/test/workspace-restore-on-conversation-load.test.jsx:246` — the test named for the disabled-feature path hardcodes `features: { workspaces: true }` and asserts the workspace *was* restored, so the `!workspacesEnabled` guard has no coverage at all. Make the flag settable per test and assert `applyWorkspace` is not called.
4. Use a per-run temp database path instead of `duckdb:///:memory:` in the validation script and drop the `.gitignore:119` entry — that URL form creates a real file named `:memory:` in the repo root, so leftover rows from a prior run can carry into the next and mask a persistence bug.
5. Add a regression test for the sequence: workspace list unloaded → load a conversation with a workspace → load one without → list resolves → assert `applyWorkspace` was not called.

## Evidence
Evidence: 7/10 — the production surface is small and each named edge case has a matching unit test, but there is no end-to-end demonstration of the user-visible behavior. The most useful missing artifact is a short recording or before/after screenshot pair of opening a conversation saved under one workspace while a different one is active.

## Notes
- Absent vs. explicit-null `workspace_id` are indistinguishable at `atlas/application/chat/service.py:393`, so a client that omits the field clears an existing conversation's binding; consider an `_unset` sentinel.
- The WebSocket-to-service wiring (`atlas/main.py:702`) has no test — deleting `workspace_id=data.get("workspace_id")` breaks the feature end to end with every test still green.
- A recorded workspace that has since been deleted leaves the active-workspace indicator pointing at an unrelated workspace, and the next turn re-stamps the conversation with that id (`ChatContext.jsx:193`).

---

garland3:
Thanks for the reviews. The points raised on the first commit are addressed in the second (`debe50f`); the bots re-anchored the original comments to the new commit, so they read as still-open but the fixes are in. Summary:

- **Local-save metadata (Codex P1)**: the IndexedDB autosave now stores `metadata.workspace_id` alongside the server save path, so locally saved conversations also restore their workspace.
- **Cancel obsolete deferred restores (Codex P2)**: `restoreWorkspace` now clears any pending restore when a conversation has no workspace, and `clearChat` clears it on new chat — a queued restore from an earlier load can no longer fire into a later conversation.
- **Validation script naming (Codex P1)**: renamed `test_issue829_*.sh` → `test_pr835_restore_workspace_on_load.sh` so `run_pr_validation.sh` discovers it (it only scans `test_pr*.sh`).
- **Piped exit codes (Copilot + Codex P2)**: added `set -o pipefail` and `${PIPESTATUS[0]}` so a failing `pytest`/`vitest`/`build` no longer reports PASSED via `tail`'s exit code.
- **REST boundary (Codex P1)**: added `TestConversationRoutes::test_get_by_id_returns_workspace_id_in_metadata` exercising the real FastAPI route → repository path via `TestClient`; the validation script runs that suite.
- **Docs (Codex P1)**: added a "Reopening a conversation restores its workspace" section (with the deleted/unbound/deferred edge cases) to `docs/user-guide/workspaces/README.md`, plus a "Restore on reload" row in the code-location table.
- **Changelog (Codex P1 / Copilot)**: condensed to a single scannable bullet under `## [Unreleased]`.
- **Misleading test (Copilot)**: removed the test that claimed to cover the feature-disabled path while the mock hard-coded the feature on; replaced it with a deferred-restore cancellation test.

All 11 CI checks are green on `debe50f`.

garland3:
## Summary
Persists a conversation's `workspace_id` in its metadata (repository, chat service, and WebSocket frame) and restores the corresponding workspace on the frontend when a saved conversation is reopened, deferring the switch until the workspace list has loaded.

## Verdict
Request changes.

## Required changes
1. Stop the browser autosave effect from re-binding a conversation on open at `frontend/src/contexts/ChatContext.jsx:903` — in local save mode the effect fires ~1s after any conversation is loaded and rewrites `metadata.workspace_id` to whatever workspace is currently active, silently destroying the stored binding with no user action. Merge into the existing record, or only overwrite once the user actually sends a turn; removing `activeWorkspaceId` from the dependency array is not sufficient.
2. Cancel a queued deferred restore in `clearActiveWorkspace` (`frontend/src/contexts/ChatContext.jsx:175`) and `switchWorkspace` (line 144), and on the delete path that nulls the active workspace — today an explicit user workspace action taken while the list is still loading is overwritten when the pending restore fires.
3. Treat an absent `workspace_id` as a no-op in `atlas/application/chat/service.py:393` — `kwargs.pop("workspace_id", None)` collapses "omitted" and "explicit null", so any client that does not send the field (CLI, cached bundle, script) strips the binding on its next turn. Use an `_UNSET` sentinel and only assign into `session.context` when a value was supplied.
4. Defer rather than discard the restore while the config is still resolving at `frontend/src/contexts/ChatContext.jsx:187` — `workspacesEnabled` is `false` until the config fetch lands, so an early open throws the id away permanently. Gate on `config.configReady` and add it to the deferred-apply effect's dependencies.
5. Notify the user when a restore changes or fails to restore the workspace (`frontend/src/contexts/ChatContext.jsx:190`) — a deleted or inaccessible workspace currently leaves the header asserting an unrelated workspace, and a successful restore silently replaces hand-picked tools, prompt and RAG selections.
6. Fix the validation script's database URL at `test/pr-validation/test_pr835_restore_workspace_on_load.sh:29` — `duckdb:///:memory:` creates a real file literally named `:memory:`, leaking state between runs — and make step 5 (line 141) assert something rather than only printing.

## Evidence
Evidence: 7/10. Unit coverage is thorough for a change this size — repository, service, route, and five frontend cases — but nothing demonstrates the feature in the running application; the most valuable missing artifact is a screenshot or short recording of opening a workspace-bound conversation while a different workspace is active and seeing the header and selections flip.

## Notes
- The workspaces-disabled branch of `restoreWorkspace` has no test (`frontend/src/test/workspace-restore-on-conversation-load.test.jsx:47`), and `isStaleWorkspacePointer` is stubbed always-false, hiding that path (line 174).
- `workspace_id` is persisted with no length or format bound (`atlas/application/chat/service.py:393`).
- Conversation `metadata` is spelled out as an ad-hoc literal in three places with no shared definition; the local-save path drops `agent_mode` (`frontend/src/contexts/ChatContext.jsx:904`).

---

garland3:
## Second review round addressed + the requested end-to-end evidence

Both reviews asked for the same missing artifact — *"a screenshot or short recording of opening a workspace-bound conversation while a different workspace is active and seeing the header and selections flip."* Here it is, captured against the running app (backend on a scratch DuckDB, `FEATURE_WORKSPACES_ENABLED=true`, save mode *Server*):

Reopening a conversation restores its workspace

**Left:** the *Code Review* workspace is active, no conversation open.
**Right:** clicking *"Draft the Q3 revenue summary"* (saved under *Quarterly Report*) flips the workspace pill to **Quarterly Report**, restores that workspace's `canvas` tool into Active Tools, and shows the new notification: *"Switched to the "Quarterly Report" workspace this conversation was saved with."*

The same capture is now committed at `docs/user-guide/workspaces/screenshots/ws-11-restore-on-reload.png` and embedded in the user guide.

---

### Required changes

**1. Autosave re-binding a conversation on open** — fixed, and this was the real bug of the batch. The local-save effect fired ~1s after *any* load and rewrote `metadata.workspace_id` to whatever was active, so merely browsing history retagged old conversations. The conversation's binding is now tracked separately from the active workspace (`conversationWorkspaceIdRef`) and is only written by a load (which reads it from the saved metadata) or by the user actually **sending a turn**. `activeWorkspaceId` is out of the effect's dependencies because it is no longer read there at all. Covered by a regression test that loads a `ws-work` conversation while `ws-home` is active, advances fake timers past the debounce, and asserts the saved record still says `ws-work`.

**2. Cancel a queued restore on explicit user actions** — done for `switchWorkspace`, `clearActiveWorkspace`, the delete path, and `clearChat`. Two tests cover it (switch and clear); note the realistic trigger is a restore queued on **config**, not on the workspace list, since a switch is impossible while the list is still empty.

**3. Absent vs. explicit-null `workspace_id`** — fixed with an `_UNSET` sentinel: omission no longer touches `session.context`, while an explicit `null` still clears. That alone was not sufficient, so `handle_restore_conversation` now also seeds the session from the conversation's stored metadata — otherwise a fresh session's `_save_conversation` would still write `None` for a client that never sends the field. Three tests: omitted-is-noop, explicit-null-still-clears, and restore-seeds-then-omits.

**4. Defer instead of discard while the config resolves** — fixed. `restoreWorkspace` now gates on `config.configReady` *before* consulting `workspacesEnabled` (which is `false` for everyone until the fetch lands), and `configReady` is in the deferred-apply effect's dependencies. Tested both ways: deferred restore fires when the config arrives, and is dropped if the config says the feature is off.

**5. Notify the user** — added, via the existing toast system rather than a timeline event (a system message would be persisted into the transcript on the next save). A successful switch names the workspace; a recorded workspace that is gone says so instead of silently leaving the header pointing at an unrelated one. Both visible in the test suite and, for the switch case, in the screenshot above.

**6. Validation script database URL** — fixed. `duckdb:///:memory:` is not an in-memory URL for this driver; it created a real file named `:memory:` in the repo root (which is why a `.gitignore` entry had appeared to hide it). Each step now gets its own file under a `mktemp -d` cleaned up on exit, and the `.gitignore` line is removed. **Step 5 now asserts** — it runs the `TestConversationRoutes` route test through pytest with `${PIPESTATUS[0]}` instead of only printing.

### From the notes

- `workspace_id` is now length-bounded (128 chars) before it reaches metadata.
- The local-save path no longer drops `agent_mode` — it writes both metadata fields.

### Test-quality point

The frontend harness hardcoded `configReady: false` and `features.workspaces: true`, so the disabled-feature branch had no coverage and every restore test was passing for the wrong reason. Both are now settable per test, and a `does not restore when the workspaces feature is disabled` case asserts `applyWorkspace` is never called.

### Verification

- `bash test/atlas_tests.sh` → **2509 passed**, 17 skipped, 0 failed
- `npx vitest run` → **595 passed** across 54 files (16 in the workspace-restore suite, up from 10)
- `bash test/pr-validation/test_pr835_restore_workspace_on_load.sh` → **8/8 PASSED**, and no stray `:memory:` file afterwards
- `ruff check` and `eslint` clean on every touched file; `npm run build` clean
- Manual: the run pictured above, driven through the real WebSocket restore path

garland3:
## Summary
Persists a conversation's `workspace_id` in conversation metadata and restores the saved workspace (prompt, sources, tools) when that conversation is reopened, with a deferred-restore path for when the workspace list has not loaded yet.

## Verdict
Request changes.

## Required changes
1. Forward the omitted-vs-null distinction from the transport layer at `atlas/main.py:702` — the endpoint calls `data.get("workspace_id")`, so an omitted field arrives as an explicit `None` and the new `_UNSET` branch in `handle_chat_message` is never reached; the binding is still cleared on the next turn for any client that omits the field. Build the kwarg conditionally (or pass `data.get("workspace_id", _UNSET)`) and add a test at the endpoint boundary.
2. Send an explicit null from the client at `frontend/src/contexts/ChatContext.jsx:559` — `workspace_id: activeWorkspaceId || undefined` is dropped by `JSON.stringify`, so once change 1 lands a user who clears their workspace can never unbind the conversation. Use `?? null` and update the test that currently pins the omit behavior.
3. Clear `pendingWorkspaceRestoreRef` on the falsy-`workspaceId` early return in `restoreWorkspace` (`frontend/src/contexts/ChatContext.jsx:196`) — opening a workspace-bound conversation and then a workspace-less one while the list is still loading applies the first conversation's selections to the second.
4. Seed `session.context["workspace_id"]` from stored metadata in `_hydrate_session_from_store` (`atlas/application/chat/service.py:385`), or merge rather than replace `metadata_json` on save — a mid-conversation reconnect currently overwrites the stored binding with null.
5. Leave the existing binding untouched when an inbound `workspace_id` fails the type or 128-char length check (`atlas/application/chat/service.py:404`) — a single malformed frame currently persists a cleared binding that later well-formed turns cannot recover.
6. Add `set -o pipefail` (or check `${PIPESTATUS[0]}`) in `test/pr-validation/test_issue829_restore_workspace_on_load.sh:141` — steps 5–7 pipe through `tail`, so failing suites and builds are reported as PASSED, which invalidates the `8/8 PASSED` result.
7. Make the workspaces-disabled test case actually disable the feature at `frontend/src/test/workspace-restore-on-conversation-load.test.jsx:246` — it hard-codes `features: { workspaces: true }` and asserts a restore, so the `!workspacesEnabled` guard has no coverage.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — a committed before/after screenshot of the restore driven through the real WebSocket path, backend tests at the repository, service and route layers, and sixteen frontend cases; the only missing artifact is a capture of the "workspace no longer available" notification, whose outcome is covered by unit test only. Note that the validation script's pass/fail reporting is unreliable until item 6 is fixed.

## Notes
- The deleted-workspace toast also fires when `GET /api/workspaces` fails, telling users a workspace is gone when the fetch merely errored — `frontend/src/contexts/ChatContext.jsx:200`.
- `applyWorkspaceRestore` short-circuits when the restored id equals the active one, so hand-edited selections survive instead of being reset to the workspace's — `frontend/src/contexts/ChatContext.jsx:196`.
- The deferred-restore state machine is now read or written at seven sites in the context body; extracting it into a hook with `requestRestore`/`cancelRestore` would stop each new workspace-mutating action from being a fresh chance to leak a pending restore — `frontend/src/contexts/ChatContext.jsx:145-258`.

---

garland3:
## Third review round — items 1, 2, 4, 5 fixed; 3, 6, 7 were already in

**Item 1 was a genuine root gap and it invalidated last round's fix.** The `_UNSET` sentinel I added in `handle_chat_message` was unreachable: `atlas/main.py` collapsed an omitted field into an explicit null (`data.get("workspace_id")`) one layer above it, so the binding was still cleared for every client that omits the field. Good catch — fixed, and pinned by tests that drive the real WebSocket endpoint rather than the service in isolation. I verified the omitted-field test **fails against the previous code and passes now**, so it is not vacuous.

The sentinel is renamed `_UNSET` → `UNSET`, since the transport layer is a legitimate consumer.

**2. Client sends an explicit null.** Correct, and it only became a bug once item 1 landed: `workspace_id: activeWorkspaceId || undefined` is dropped by `JSON.stringify`, so a user who cleared their workspace could never unbind the conversation. Now `?? null`, and the test that pinned the omit behavior is updated to pin the null (plus an `'workspace_id' in payload` assertion so it cannot silently regress to omission).

**4. Reconnect path.** Real, and distinct from the restore path I fixed last round — `_hydrate_session_from_store` is a *second* entry point. It now seeds `session.context["workspace_id"]` from stored metadata. It runs before the turn's own `workspace_id` is applied, so a turn that carries one still wins. Covered by a test that continues a conversation from a brand-new session without sending the field.

**5. Malformed input no longer clears the binding.** A non-string, blank, or over-length id is now ignored with a warning, leaving any existing binding intact; an explicit `null` still clears. Test parametrises over `12345`, `{"id": "x"}`, `"  "`, and a 5000-char string and asserts the binding survives all four.

### Items already addressed in `2a7a9b4`

These three were fixed in the commit before this review was posted — I think the bot re-anchored to an older diff, since the line numbers cited (`ChatContext.jsx:196`, `test_issue829_...sh:141`, `test.jsx:246`) are from the first-round code:

- **3.** `restoreWorkspace`'s falsy-`workspaceId` early return already clears `pendingWorkspaceRestoreRef`, with a test for the exact A-then-B sequence described.
- **6.** The script already has `set -o pipefail` **and** `${PIPESTATUS[0]}` on every piped step, and was renamed `test_issue829_*` → `test_pr835_*` so `run_pr_validation.sh` discovers it. The `8/8 PASSED` figure was produced with those in place, so it is not the unreliable reporting described — I re-confirmed by checking a failing suite propagates.
- **7.** The workspaces-disabled case is real now: the harness no longer hardcodes `features.workspaces`, and the test asserts `applyWorkspace` is **not** called.

### From the notes

- **Deleted-workspace toast on a failed fetch** — good point, fixed. A failed `GET /api/workspaces` is indistinguishable from a deletion by the list alone, so the notification is now suppressed when `workspaces.error` is set. Test added.
- **`applyWorkspaceRestore` short-circuits when the ids match** — this one is deliberate. Re-applying the workspace the user is already on would discard edits they made on top of it, which reads as the app undoing their work; the conversation is already bound to that workspace, so nothing is lost. Happy to flip it if you disagree.
- **Extracting the restore state machine into a hook** — agreed in principle, declined for this PR. It is a refactor of code this PR is already changing heavily, and it would make the diff harder to review against the six behavioral fixes above. Worth a follow-up issue.

### Evidence — the deleted-workspace capture you asked for

The screenshot is now three panels, including the "workspace no longer available" case that was previously unit-test-only:

Workspace restore on reload

One correction to my previous comment, since it affects how much that capture proves: the first run's backend was resolving `atlas` to the main checkout via the editable install, not to this branch. I re-ran everything with `PYTHONPATH` pinned to the worktree and verified the serving process (`/proc/<pid>/environ`, `cwd`) before re-capturing, so all three panels above are this branch's code end to end.

### Verification

- `bash test/atlas_tests.sh` → **2514 passed**, 17 skipped, 0 failed
- `npx vitest run` → **596 passed** across 54 files
- `bash test/pr-validation/test_pr835_restore_workspace_on_load.sh` → **8/8 PASSED**
- Endpoint test verified to fail on the pre-fix `main.py` and pass after
- `ruff check` clean on every file I touched; `eslint` clean; `npm run build` clean

Note on `atlas/main.py`: ruff reports a pre-existing `I001` on its import block that also fires on `main` unmodified, and re-sorting would rewrite unrelated imports, so I added my import in place and left the block alone.

garland3:
## Summary
Persists a conversation's workspace binding in conversation metadata and restores it when the conversation is reopened, adding an omitted-vs-null (`UNSET`) distinction on the chat frame so non-browser callers don't clobber an existing binding, plus a deferred-restore path in `ChatContext` for cold loads.

## Verdict
Request changes.

## Required changes
1. Seed the workspace binding unconditionally in `_hydrate_session_from_store` at `atlas/application/chat/service.py:1193` — the current truthy-only assignment leaves a prior conversation's `workspace_id` in `session.context`, so a long-lived session that switches to an unbound conversation silently re-binds it via `_save_conversation` (line 1297). Newly reachable because omitted fields no longer clear the value. `handle_restore_conversation` (line 860) has the same shape; fix both and add a two-conversation/one-session test.
2. Cancel a pending restore on direct selection edits at `frontend/src/contexts/ChatContext.jsx:255` — a queued restore is cancelled only by workspace actions, so tools/prompt/RAG the user picks while config or the workspace list is still loading are silently overwritten seconds later. Clear `pendingWorkspaceRestoreRef` from `toggleTool`/`setSinglePrompt`/`toggleDataSource`, or bound the deferral with a timeout.
3. Move `conversationWorkspaceIdRef.current = activeWorkspaceId || null` at `frontend/src/contexts/ChatContext.jsx:571` below the `if (!sent)` guard — a send that never reached the wire currently re-binds the local (IndexedDB) record, and the wrong value is then durable.
4. Do the `workspaceList.find` lookup before the `workspaceId === activeWorkspaceId` short-circuit at `frontend/src/contexts/ChatContext.jsx:196` — a deleted workspace that still matches the active pointer returns success, so the "no longer available" notification never fires. If the no-re-apply behavior is intended, document it and pin it with a test.
5. Fix the two vacuous tests at `atlas/tests/test_workspace_conversation_binding.py:113` and `:~330` — both assert `session.context.get("workspace_id") is None`, which holds because the key is absent, not because anything was stripped; assert `"workspace_id" not in session.context` and rename to describe ignoring, or fold them into the parametrized malformed-id test.
6. Replace the `for _ in range(50): time.sleep(0.02)` busy-wait in `TestWebSocketChatFrameWorkspaceId._send` (`atlas/tests/test_workspace_conversation_binding.py:452`) with an `Event` wait plus explicit timeout — this is the only coverage of the `atlas/main.py:702` wiring, and a CI-load flake there surfaces as a misleading "not awaited".
7. Pin step 5 and step 7 of `test/pr-validation/test_pr835_restore_workspace_on_load.sh:~150` to explicit node IDs or add a collected-count assertion — a `-k` selector that matches nothing exits 0 and still prints PASSED.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size: a committed end-to-end screenshot of the restore in the running app, layered tests at every boundary including the WebSocket frame, and stated suite counts. The main gap is that those numbers were run against the previous head commit, not `22de1dc`; a second capture of the "workspace no longer available" toast would also close the only unshown user-visible path.

## Notes
- Client-supplied `workspace_id` is persisted without checking it names a workspace the user owns (`atlas/application/chat/service.py:404`); contained today, but worth an ownership check or a UUID-shaped allowlist.
- The deferred-restore state machine is an open-coded ref mutated from seven sites and the gating ladder is written twice (`frontend/src/contexts/ChatContext.jsx:150-264`); extracting a `useWorkspaceRestore` hook would make the cancel invariant enforceable.
- Restore toasts are full sentences with a literal `--` (`frontend/src/contexts/ChatContext.jsx:216`), unlike every other toast in the app.

---

garland3:
## Fourth round — all seven addressed; two were defects in my last fix

Items 1 and 3 were bugs I introduced last round, and item 4 was a hole in a behavior I had defended as intentional. Fixed at `ea989a1`.

**1. Seeded binding leaking between conversations.** Correct, and the sharpest find in this round. The seeding was truthy-only, so a long-lived session that switched to an unbound conversation kept the previous conversation's `workspace_id` in `session.context`, and `_save_conversation` stamped it onto a conversation the user never bound — the exact failure my round-3 change was supposed to prevent, moved one step sideways. Both `_hydrate_session_from_store` and `handle_restore_conversation` now assign unconditionally. The regression test switches conversations inside one session and asserts the unbound one stays unbound; **verified it fails against the truthy-only assignment**.

(For the record: on the `handle_restore_conversation` path the session is recreated first, so a stale value could not survive there — but the shapes should not differ, so both are unconditional now.)

**2. Direct selection edits cancel a queued restore.** Fixed. `toggleTool`, `addTools`, `removeTools`, `togglePrompt`, `setSinglePrompt`, `makePromptActive`, `toggleDataSource`, `addDataSources` and `clearDataSources` all cancel it now, rather than only the workspace actions. I went with cancellation over a bounded timeout: a timeout would still overwrite the user inside the window and would make the behavior depend on how fast the fetch happened to be.

**3. Re-bind before the `if (!sent)` guard.** Correct — a send that never reached the wire durably re-bound the local record. Moved below the guard, with a test.

**4. Lookup before the already-active short-circuit.** You are right, and this is a real hole in the behavior I defended in my last comment. My reasoning ("the conversation is already bound to it, so nothing is lost") only holds when the workspace still *exists*; a workspace deleted while still being the active pointer returned success and never notified. The resolve now happens first, so the deleted case notifies regardless of the pointer. The no-re-apply behavior itself is still deliberate — re-applying would discard selection edits made on top of the workspace — and is now documented in the code and pinned by its own test, as you asked.

**5. Vacuous tests.** Both fixed: they asserted `.get("workspace_id") is None`, which held because the key was absent, so they would have passed with the ignore logic deleted. They now assert `"workspace_id" not in session.context` and are renamed to describe ignoring rather than stripping.

**6. Busy-wait.** Replaced with a `threading.Event` signalled from the mock plus an explicit 10s timeout and a real failure message, so a CI stall reports as a timeout rather than a misleading "not awaited".

**7. `-k` selector.** Fixed — the route step is pinned to the explicit node id `atlas/tests/test_chat_history.py::TestConversationRoutes::test_get_by_id_returns_workspace_id_in_metadata`. I also added collected-count assertions to the backend and frontend steps, since both can exit 0 having run nothing.

### On the evidence note

You are right that the previous comment's counts predated the head commit — I ran the suites against the working tree that became `22de1dc`, then pushed a docs-only commit on top, so the numbers described the code but not the head SHA. Fresh figures for `ea989a1` below, and the deleted-workspace toast is already captured as panel 3 of the committed screenshot (added in `a5223b6`).

### Notes

- **Ownership check on `workspace_id`** — agreed this is worth doing, declined for this PR. It is advisory-only and never used for authorization (a forged id at worst restores a workspace the attacker already controls, in their own session), and a real check means giving `ChatService` a workspace-repository dependency plus a query per turn. Better as its own change; happy to open the issue.
- **`useWorkspaceRestore` hook** — still declining, same reason as last round: it is a refactor of the code under active review. Now that the cancel invariant has genuinely leaked twice, I think you are right that it should be extracted, and I'd rather do it as a focused follow-up than fold it into this diff.
- **Toast wording** — fixed, the literal `--` is gone.

### Verification on `ea989a1`

- `bash test/atlas_tests.sh` → **2515 passed**, 17 skipped, 0 failed
- `npx vitest run` → **601 passed** across 54 files (22 in the workspace-restore suite)
- `bash test/pr-validation/test_pr835_restore_workspace_on_load.sh` → **8/8 PASSED**, now with collected-count assertions so the figure means something
- Negative controls: the conversation-leak test fails against the truthy-only seeding, and the endpoint test fails against the pre-fix `main.py`
- `ruff` clean on files I touched, `eslint` clean, `npm run build` clean

garland3:
## Summary
Binds conversations to a workspace by persisting `workspace_id` in conversation metadata and restoring it on load, with deferred restore while config/workspace list resolve and cancellation when the user edits selections.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well-demonstrated for its size — a committed end-to-end screenshot covering the switch and the "workspace no longer available" case, tests at the repository, service, REST, WebSocket and frontend-context layers, and named negative controls showing the key tests fail against the pre-fix code. The only gap is the deferred-restore timing path, which is covered by unit tests but never shown in the running app.

## Notes
- Stale binding can still leak: the `session.context["workspace_id"]` seeding in `_hydrate_session_from_store` sits below the `not conv` and exception early returns, so a session that names an unknown conversation id keeps the prior workspace and stamps it on save — `atlas/application/chat/service.py:1193`.
- `selectAllServerTools` / `deselectAllServerTools` / `selectAllServerPrompts` bypass `withRestoreCancelled`, so a bulk selection made during a queued restore is silently overwritten — `frontend/src/contexts/ChatContext.jsx:494`.
- `test_oversized_workspace_id_is_rejected` asserts `context.get("workspace_id") is None`, which passes even with the validation removed; use `"workspace_id" not in session.context` like its siblings — `atlas/tests/test_workspace_conversation_binding.py:288`.

---

garland3:
## Nits fixed — thanks for the approve

All three were real; `60147a2`.

**1. Binding still leaking past the early returns.** Correct, and my previous fix was in the wrong place: I cleared the carried binding on the success branch only, so an unknown conversation id (a new conversation, or a client-side `local_*` id) and an unreadable store both returned early with the previous conversation's workspace still in `session.context`. It is now cleared up front, before any early return, with the stored value assigned afterwards when there is one. The regression test **fails against the previous placement**.

**2. Bulk selection helpers bypassed cancellation.** Fixed. While writing the test I found the obvious fix was still wrong: routing them through the wrapped toggles only cancels when a key actually changes, so "Deselect All" with nothing selected stayed a no-op. The cancel is now on the action itself — that is the user stating what they want, whether or not it changes a key. Test covers both select-all and deselect-all.

**3. Vacuous oversized-id assertion.** Fixed the same way as its siblings: `"workspace_id" not in session.context`.

### On the remaining evidence gap

You are right that the deferred-restore timing path is unit-tested but never shown in the app. I did not add a capture for it because the honest version is unilluminating — the deferral window is the config/workspace fetch, so a screenshot is either "the restore hasn't happened yet" (indistinguishable from a plain cold load) or "it has", with the interesting part being that nothing overwrites a selection in between. A recording would need artificial fetch latency to be legible. If you want it, the useful artifact is probably a throttled-network capture of picking a tool during the window and seeing it survive — say the word and I'll produce it rather than guess.

### One observation, not a change

While testing item 1 I hit a pre-existing edge unrelated to this PR: when a session switches to an unknown conversation id, the rehydrate path deliberately leaves the live history alone, so the previous conversation's messages get re-saved under the new id and trip a primary-key constraint on `messages.id`. It predates this change and the workspace binding has nothing to do with it, so I left it alone — noting it in case it is news. The test asserts on `session.context` (what the nit is about) rather than the save, with a comment explaining why.

### Verification on `60147a2`

- `bash test/atlas_tests.sh` → **2516 passed**, 17 skipped, 0 failed
- `npx vitest run` → **602 passed** across 54 files (23 in the workspace-restore suite)
- `bash test/pr-validation/test_pr835_restore_workspace_on_load.sh` → **8/8 PASSED**
- Negative control: the new unknown-conversation test fails against the previous clear placement
- `ruff` and `eslint` clean; `npm run build` clean

For the record, CI came back **green on all 11 checks for `ea989a1`**; these nit fixes are the only delta since.

garland3:
## Summary
Persists a conversation's `workspace_id` in conversation metadata and restores it when the conversation is reopened, deferring the restore until config and the workspace list have loaded and cancelling it on any deliberate selection change. Adds an omitted-vs-null distinction on the chat frame's `workspace_id` so a turn can leave, clear, or set the binding.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size, with an end-to-end capture, layered frontend and backend tests, and named negative controls for each cancellation path. The one gap worth closing: the quoted suite counts are stated for `60147a2`, two commits behind the head, and `05c9bc6d` edits the config mock those frontend numbers came from — re-run and paste the counts against the head commit.

## Notes
- Sending a turn while a restore is still queued sends `workspace_id: null`, unbinding the conversation server-side, then the queued restore fires and swaps selections underneath the turn — `frontend/src/contexts/ChatContext.jsx:607` is the one deliberate action that does not cancel the pending restore.
- `does not re-bind the local record when the send never reached the wire` wraps its only assertion in `if (h.saveLocalConv.mock.calls.length)`, which is never true in that scenario, so the test passes against the regression it names — `frontend/src/test/workspace-restore-on-conversation-load.test.jsx:519`.
- A stored conversation with an empty or malformed message list returns from hydrate with its binding cleared and the null is written back over the stored id; read the binding right after `get_conversation` rather than after the message guard — `atlas/application/chat/service.py:1149`.

---

garland3:
## Nits fixed — `3881e9d`

**1. Sending a turn now cancels a queued restore.** Correct, and it was the one deliberate action I had missed. Worse than a stale selection: the frame carries the binding (including an explicit null), so the queued restore both swapped the selections out from under the turn *and* contradicted the binding that turn had just written. Cancelled on send now, with a test that sends under `ws-home` while a `ws-work` restore is queued and asserts the restore never fires.

**2. The guarded assertion.** You are right, and it is the fourth vacuous test of mine this review has caught. My reasoning when I wrote the guard was wrong in an interesting way, so worth recording: I assumed a failed send meant no user message and therefore no autosave. It does not — the *loaded* conversation's messages are already there, so the autosave runs and its binding is observable. The guard was not just redundant, it would have hidden a wrong value as readily as a missing call. Now asserted unconditionally, **verified to fail when the re-bind is moved back above the `!sent` guard**, with a positive case alongside it so both halves are pinned.

**3. Binding read after the message guards.** Fixed — read as soon as the record is in hand, before the empty/malformed-message returns, so a stored conversation with a bad message list no longer has its binding nulled out.

### Suite counts on the head commit (`3881e9d`), as asked

You were right to push on this — the previous counts predated `05c9bc6`, which edited the very config mock the frontend numbers came from. Re-run against the head:

- `bash test/atlas_tests.sh` → **2516 passed**, 17 skipped, 0 failed
- `npx vitest run` → **604 passed** across 54 files (25 in the workspace-restore suite)
- `bash test/pr-validation/test_pr835_restore_workspace_on_load.sh` → **8/8 PASSED**
- `ruff` clean on the files I touched; `eslint` clean; `npm run build` clean
- Negative controls, each confirmed to fail against the pre-fix code: the WebSocket omitted-field test, the conversation-leak test, the unknown-conversation test, and now the failed-send re-bind test

CodeQL also flagged a duplicate `prompts` key I introduced in that config mock — fixed in `05c9bc6` before this.

### A note on the pattern, since it seems worth saying

Across these rounds the review has found four vacuous tests of mine and three fixes placed a branch or a layer away from where they belonged. The common thread is that this feature writes one piece of state from many call sites, and each fix I made was locally correct while leaving a neighbouring path wrong. That is the case for the `useWorkspaceRestore` extraction you suggested twice — not as cleanup, but because the cancel invariant has now leaked four times and would be enforceable in one place instead of remembered at each new call site. I still think it belongs in its own PR rather than folded into this diff, but I'd support it as the immediate follow-up and I've said so in my hand-off notes.

garland3:
## Summary
Persists a conversation's workspace binding in conversation metadata and restores it when the conversation is opened, with a deferred-restore path for when config or the workspace list has not loaded yet, and a cancel-on-deliberate-edit rule so a queued restore never overwrites a user's own selections.

## Verdict
Request changes.

## Required changes
1. Send the conversation's binding, not the active pointer, while a restore is still queued — `frontend/src/contexts/ChatContext.jsx:607`. Sending a turn during the deferral window puts `workspace_id: activeWorkspaceId ?? null` on the frame, so the backend overwrites the conversation's stored workspace (commonly with `null`) and the binding is unrecoverable. Use `pendingWorkspaceRestoreRef.current ?? activeWorkspaceId ?? null`, or omit the key entirely so the `UNSET` branch leaves the stored value alone. The existing test "sending a turn cancels a queued restore" pins the current behavior and needs updating with it.
2. Route the remaining selection mutators through `withRestoreCancelled` — `frontend/src/contexts/ChatContext.jsx:1060-1092`. `addPrompts`, `removePrompts`, `clearActivePrompt`, `clearToolsAndPrompts` and `toggleRagEnabled` are exported unwrapped; at least "Default Prompt" (`PromptSelector.jsx:110`), "remove selected prompts" (`PromptSelector.jsx:137`) and the RAG toggle (`ChatArea.jsx:1106`) are live controls whose effect a queued restore silently reverts. Better: apply the guard once at the boundary (map over `selections` against a `MUTATING_ACTIONS` set) so a future action is guarded by default.
3. Preserve the binding when a store read fails — `atlas/application/chat/service.py:1149`. `_hydrate_session_from_store` clears `session.context["workspace_id"]` before `get_conversation` and the `except` at line 1163 returns without restoring it, so a transient read error can persist `null` over a good binding whenever the no-shrink guard does not block the write. Snapshot and restore on the exception path, and add a test that patches `get_conversation` to raise and asserts the stored metadata is unchanged.
4. Assert the active workspace pointer in the restore suite — `frontend/src/test/workspace-restore-on-conversation-load.test.jsx:143`. The `usePersistentState` mock returns a throwaway setter, so no case asserts `setActiveWorkspaceId(ws.id)`; a regression that applies the selections but leaves the header pill on the old workspace passes all 25 cases. Hoist the setter onto the shared handle and assert it.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size, with a committed three-panel end-to-end capture, layered tests, and named negative controls. The one gap worth closing is panel 3 of `docs/user-guide/workspaces/screenshots/ws-11-restore-on-reload.png`, which still shows the pre-`ea989a1` toast wording and so was not captured against this head.

## Notes
- Give the session's workspace binding one owner — `handle_restore_conversation` and `_hydrate_session_from_store` implement the same seed-from-record step in two shapes (`atlas/application/chat/service.py:866, 1186`).
- Conversation `metadata` is rebuilt as a fresh literal by each writer and replaces the stored record wholesale, so one writer's keys drop another's (`atlas/application/chat/service.py:1307`).
- The deleted-workspace toast is suppressed on the shared, sticky `workspaces.error`, so an unrelated earlier CRUD failure makes the restore fail silently (`frontend/src/contexts/ChatContext.jsx:217`).

---

garland3:
## All four fixed — `a3cc80a`

**1. The frame carried the wrong binding during the deferral window.** Correct, and this one is a bug my *previous* round's fix created: cancelling the queued restore on send was right, but I left the frame carrying `activeWorkspaceId`, which during the deferral window is still the previous conversation's pointer — usually null. So the turn overwrote the stored binding and the cancel made it unrecoverable. Strictly worse than before I "fixed" it.

The frame and the local record now both carry `pendingWorkspaceRestoreRef.current ?? activeWorkspaceId ?? null`, so a turn sent mid-deferral re-states the conversation's own workspace rather than clobbering it. I used the pending-ref form over omitting the key because it keeps the frame explicit about what the conversation is bound to, rather than relying on the backend's leave-alone branch to paper over a client that knows the answer. The test that pinned the old behavior is updated to assert `ws-work`.

**2. Guarded at the boundary, as you suggested.** All mutating selection actions are now wrapped by name from one `MUTATING_SELECTION_ACTIONS` list, so a new action is guarded by default instead of being the next leak. That covers the five you named — `addPrompts`, `removePrompts`, `clearActivePrompt`, `clearToolsAndPrompts`, `toggleRagEnabled` — plus the bulk helpers, which were still calling `selections.*` directly. `applyWorkspace` is deliberately excluded, with a comment: it is how a restore applies, and guarding it would cancel the restore mid-flight.

**3. Read failure no longer unbinds.** Fixed by snapshotting the carried binding before the read and restoring it on the exception branch (distinguishing "was absent" from "was null" via `UNSET`). Test patches `get_conversation` to raise and asserts both the session context and the stored metadata survive; **verified to fail without the restore**.

**4. The pointer was never asserted.** Correct, and it is the fifth blind spot of this kind you have found in my tests. Hoisted `setActiveWorkspaceId` onto the shared handle and asserted it in both positive restore cases plus the deleted case. **Verified**: dropping `setActiveWorkspaceId(ws.id)` from `switchWorkspace` — exactly the regression you described — now fails 2 tests, where before it passed all 25.

### On the sticky-error note

Fixed, and investigating it turned up dead code of mine. Gating on `workspaces.error` was wrong for the reason you gave. But gating on `workspacesLoaded` instead turned out to be *trivially true*: both callers already gate on it, so `applyWorkspaceRestore` is only ever reached with an authoritative list. The guard I added last round could never have been false. Removed it and documented why the branch is safe to speak from, rather than leaving a condition that reads like a real check.

### Screenshot

Good catch — panel 3 predated the `ea989a1` toast rewording. All three panels re-captured against this head with a freshly built bundle, and I re-verified the serving process was running branch code first.

### Verification on `a3cc80a`

- `bash test/atlas_tests.sh` → **2517 passed**, 17 skipped, 0 failed
- `npx vitest run` → **605 passed** across 54 files (26 in the workspace-restore suite)
- `bash test/pr-validation/test_pr835_restore_workspace_on_load.sh` → **8/8 PASSED**
- Negative controls, each confirmed to fail against the pre-fix code: WebSocket omitted-field, conversation-leak, unknown-conversation, failed-send re-bind, store-read-failure, and the new active-pointer assertions
- `ruff` and `eslint` clean; `npm run build` clean
- CI was **green on all 11 checks for `3881e9d`**, the previous head

### The remaining notes

Both are the same structural point and I agree with it: `handle_restore_conversation` and `_hydrate_session_from_store` implement the same seed step in two shapes, and each `metadata` writer rebuilds a literal that replaces the stored record wholesale. Between them they are the reason this PR has needed seven rounds — the binding has one meaning and five writers. I would rather land the behavior with the coverage it now has and do the consolidation (single owner for the seed, a shared metadata builder, and the `useWorkspaceRestore` hook on the frontend) as one focused follow-up than keep widening this diff. If you would rather see them here, say so and I will.

garland3:

garland3:
## Verified the conflict resolution in `7266d16`

Since the branch was rewritten by a bot after my last push, I checked that nothing of this PR's was lost in the merge rather than assuming it:

- `git diff a3cc80a 7266d16` over every file this PR touches (`service.py`, `main.py`, `ChatContext.jsx`, both test files, the validation script, the docs and screenshot) is **empty** — the merge only brought in #833's files (`system_prompt_time.py`, `message_builder.py`, `orchestrator.py`, `settings.py`, `.env.example`, its tests and docs) plus the CHANGELOG.
- Spot-checked the four load-bearing markers on the merged tree: the deferral-window binding on the frame, `MUTATING_SELECTION_ACTIONS`, the `previous_workspace_id` snapshot/restore, and `data.get("workspace_id", UNSET)` in `main.py`. All present.

Re-ran everything against the merged head:

- `bash test/atlas_tests.sh` → **2531 passed**, 17 skipped
- `npx vitest run` → **605 passed** across 54 files
- `bash test/pr-validation/test_pr835_restore_workspace_on_load.sh` → **8/8 PASSED**

One caveat worth recording rather than glossing: my first full backend run on the merged head had `test_hooks_engine.py::TestFailureModes::test_output_over_cap_is_killed_and_treated_as_error` fail. It passes 3/3 in isolation and passed on the immediate re-run of the full suite, and this PR touches neither the hooks engine nor that test — so it looks like a load-sensitive flake in that output-cap test rather than anything from the merge. Flagging it because it will presumably bite someone else too; happy to open a separate issue if it is not already known.

PR #839, opened 2026-08-24
feat(ui): combine tools, settings, theme, and admin into one panel (closes #836)

### LINKED ISSUE #836: feature: combine menus
Combine the 'settings' gear with the tools and integrations. The tools and integrations can be another tab. 
The light or dark mode toggle can be moved into the settings panel and off the top bar. 
if admin, in a new tab show the most used admin tools which are:  MCP Configuration & Controls section , Banner Messages, , and the Feedback section.  Give a link to the full admin page. 

in the selection or prompts below the chat input, for custom user prompts, show a 'edit' button next to each prompt to quickly take users to the edit page in the config. 
Also, add a small 'create new system prompt' button in that area. 
Both of these are to  help user find their custom prompts quikcly.

garland3:
Closes #836.

The top bar carried three separate entry points -- a wrench for tools, a gear for settings, and a sun/moon for the theme. They are now one **Tools and Settings** wrench button opening a single tabbed panel.

## What changed

- **Tools & Integrations is a tab.** `ToolsPanel` gained an `embedded` mode that drops its own modal chrome. It stays mounted while the panel is open, so staged tool selections survive a tab switch, and closing the panel with unsaved selections still raises the existing save/discard dialog.
- **Theme toggle moved** off the top bar into **General -> Appearance**.
- **Admin tab** (admins only) with the three most-used dashboard cards -- MCP Configuration & Controls, Banner Messages, User Feedback -- fully featured rather than read-only, plus a **Full Admin Page** link. The notification / system-status / edit-modal plumbing those cards need moved into `hooks/useAdminConfigActions.js`, which `AdminDashboard` now shares instead of keeping its own copy.
- **Prompt shortcuts.** The picker under the chat input shows a pencil next to each custom prompt and a **New system prompt** button; both open the Prompts tab with that prompt's editor already open (via an `atlas:open-settings` window event, since the picker sits far below the panel in the tree).

## Screenshots

One wrench button in the header:

Header

Tools & Integrations tab:

Tools tab

General tab with the Appearance toggle:

General tab

Admin quick controls:

Admin tab

Prompt picker with edit / create buttons, and the editor it opens:

Prompt picker

Prompt editor

## Verification

- `npm run lint` clean (only pre-existing warnings).
- `npx vitest run`: 613 passed. New suite `combined-tools-settings-panel.test.jsx` (8 tests) covers the panel title, tools-tab gating on `features.tools`, admin-tab gating on `isInAdminGroup`, the theme toggle, prompt-intent pass-through, and both picker events. The SettingsPanel test now wraps in `ThemeProvider`; the PromptSelector test's lucide mock gained `Pencil`/`Plus`.
- `./test/run_tests.sh frontend` and `backend` pass; `scripts/check-docs.sh` OK.
- Ran the built frontend against a live `atlas.main` and clicked through every flow -- the screenshots above are from that run (admin tab shown against 10 connected MCP servers, light/dark verified to actually flip).

## Notes / caveats

- `test_e2e/tests/marketplace.spec.js` was updated for the renamed button but not run; that suite also still expects older UI strings ("Go to Marketplace", "of 7 servers selected") that no longer match `ToolsPanel`, so it looks stale independent of this change.
- Behaviour change: canvas content arriving no longer auto-closes the tools surface, since it may now hold unsaved selections inside a modal.
- The standalone (non-embedded) `ToolsPanel` modal path is kept -- nothing renders it any more, but its 22 unit tests exercise it unchanged. Say the word if you'd rather delete it.
- The header still has its separate admin Shield button; the issue didn't ask to remove it.
- The wrench is icon-only (tooltip + aria-label "Tools and Settings"); a visible label wrapped to three lines and crowded the top bar.


garland3:
## Summary
Consolidates the tools modal, settings modal, prompt library, appearance toggle and an admin quick panel into a single five-tab "Tools and Settings" panel, wired from the header, the marketplace return path and a new `atlas:open-settings` window event, with admin config logic extracted into a shared `useAdminConfigActions` hook.

## Verdict
Request changes.

## Required changes
1. Fix the tab selection that is discarded when the panel opens before `/api/config` resolves — `frontend/src/components/SettingsPanel.jsx:93` rewrites `activeTab` to `'general'` because `DEFAULT_FEATURES` hides the Tools tab, and the `[isOpen, initialTab]` effect never re-applies the request once the flags arrive, so the panel is stuck on General for the rest of the session. This is on the normal path: `App.jsx:65` calls `openSettings('tools')` from a mount effect on the marketplace return. Derive the active tab from `visibleTabs` until the user explicitly picks one, or retain the pending `initialTab` in a ref and re-apply it when it becomes visible.
2. Preserve the prompt draft across tab switches at `frontend/src/components/PromptManager.jsx` — it renders only under `activeTab === 'prompts'`, so switching to General mid-compose unmounts it and destroys `editingId`/`draft` with no warning. Keep it mounted behind `hidden` the way `ToolsPanel` is, or lift the draft into `SettingsPanel`.
3. Clear `promptIntent` once it has been applied (`frontend/src/components/PromptManager.jsx:38`) — it is only cleared on panel close while `appliedIntentRef` resets on unmount, so leaving and re-entering the Prompts tab yanks the editor back to the originally-clicked prompt. Add an `onIntentConsumed` callback that clears the intent in `App`.
4. Stop the tools tab's **Save Changes** / **Cancel** from dismissing the whole panel (`frontend/src/components/ToolsPanel.jsx:163,172`) — embedded, both call the parent's `onClose`, so a per-tab save exits Prompts, General, User Info and Admin too. Pass a commit-and-stay handler in `embedded` mode and reserve dismissal for the header X and backdrop.
5. Pass `requestClose` rather than the raw `onClose` as `onNavigate` to `AdminQuickPanel` (`frontend/src/components/SettingsPanel.jsx:239`) — "Full Admin Page" currently drops staged tool selections without the prompt the X button gives.
6. Re-sync the tools tab when saved selections change underneath it (`frontend/src/components/ToolsPanel.jsx:63`) — the Admin tab's **Reload MCP** prunes non-existent keys from `selectedTools`, but the still-mounted pending set keeps them and **Save Changes** writes them back. Re-seed pending state when there are no local edits, and intersect pending keys with the current tool list in `handleSave`.
7. Add tab and dialog semantics to the panel (`frontend/src/components/SettingsPanel.jsx`) — the strip is plain buttons with no `role="tablist"`/`role="tab"`/`aria-selected`, and the container has no `role="dialog"`, `aria-modal`, focus move, focus trap or Escape handling. This panel is now the only route to tools, theme and admin controls.

## Evidence
Evidence: 9/10 — six screenshots against a live server with 10 connected MCP servers, a new 8-test suite, quoted full-suite and lint results, and a new user guide, all proportionate to a 25-file change; known gaps are disclosed rather than hidden. The one artifact worth adding is a test or screenshot of the unsaved-selection dialog raised from the *combined* panel: the new suite mocks `ToolsPanel` and the existing 22 tests cover the standalone modal path, so the `closeGuardRef`/`onDirtyChange`/`requestClose` plumbing has no direct coverage.

## Notes
- `useAdminConfigActions` has no direct tests despite owning the per-endpoint method/payload branching for both the dashboard and the quick panel — `frontend/src/hooks/useAdminConfigActions.js:1`.
- Notification `setTimeout`s are never cleared and the admin tab unmounts on every tab switch, so admins never see save results and timers fire against a dead hook — `frontend/src/hooks/useAdminConfigActions.js:24`.
- The hidden-but-mounted tools tab re-renders its unmemoized server/tool tree on every slider tick in General; `React.memo` plus a `useMemo` on `filteredServers` would be close to free — `frontend/src/components/SettingsPanel.jsx:470`.

---

garland3:
@cmlanca

garland3:
Refreshed against `main` (merge commit \`fdd4eac\`) and worked the review list.

**Merge.** Three conflicts: `CHANGELOG.md` (both sides added entries — kept both, #839 in PR-number order), `Header.jsx` (main added the container-width breakpoint + `useTheme` for the header theme toggle; this PR removed that toggle and the separate tools button, so the breakpoint logic is kept and the now-unused `useTheme` import and `onToggleTools` prop are dropped), and `test_e2e/tests/marketplace.spec.js` (kept this PR's `Tools and Settings` button name, kept main's `atlas` consolidated-server assertions from #856).

**Review items 1-7, all addressed:**

1. A requested tab is held in a ref and applied the moment its feature flag makes it visible; a tab the user picks supersedes it, and it is dropped on close.
2. The Prompts tab is kept mounted behind `hidden`, like the tools tab, so a draft survives a tab switch.
3. `onIntentConsumed` fires once an intent has opened its editor; `App` clears `promptIntent` there.
4. `commitChanges` / `revertChanges` are split out from dismissal. Embedded, **Save Changes** and **Cancel** commit or revert and stay; the standalone modal still closes on save, and the unsaved-changes dialog still dismisses.
5. `AdminQuickPanel` gets `onNavigate={requestClose}`.
6. Pending state is re-seeded when the saved selections change under an un-edited panel, and `commitChanges` intersects pending keys with the live tool/prompt lists (skipped when those lists are empty, so a not-yet-loaded list cannot wipe a selection).
7. `role="dialog"` + `aria-modal`, a `tablist` with `role="tab"` / `aria-selected` / `aria-controls`, roving tabindex with arrow-key + Home/End navigation, `role="tabpanel"` on each body, focus move to the close button on open, a focus trap, and Escape routed through the same unsaved guard as the X.

Also from the notes: admin notification timers are tracked and cleared on unmount, and the admin tab is mounted on first visit and kept mounted after that (mounting it eagerly would cost every admin a config fetch), so its save-result notifications are actually seen.

**Verification.** `npm run lint` clean (4 pre-existing warnings). `npx vitest run`: **672 passed / 59 files**, up from 662 — 6 new SettingsPanel tests (pending-tab application after the flag arrives, prompts tab staying mounted, dialog/tab semantics, Escape, focus move, arrow-key nav) and 4 new ToolsPanel tests (embedded save and cancel do not dismiss, standalone save still does, a key for a vanished server is not written back). `npm run build` and `scripts/check-docs.sh` OK.

Not done: the unmemoized tools tree re-render note, and direct unit tests for `useAdminConfigActions` — say the word if you want either in this PR.

garland3:
## Summary
Merges the tools modal, settings gear, theme toggle and admin quick controls into a single five-tab "Tools and Settings" panel behind one header wrench, keeping the tools tab mounted so staged selections survive tab switches.

## Verdict
Request changes.

## Required changes
1. Stop the tools footer from dismissing the whole panel — `handleSave`/`handleCancel` at `frontend/src/components/ToolsPanel.jsx:163,172` call `onClose`, which in embedded mode is the panel's own close, so a per-tab "Save Changes" tears down Prompts, General, User Info and Admin. Add an `onCommit`/`onRevert` prop that defaults to `onClose` only in the standalone path.
2. Re-apply the requested tab once config resolves at `frontend/src/components/SettingsPanel.jsx:93` — `DEFAULT_FEATURES.tools` is false, so `openSettings('tools')` fired from the marketplace-return mount effect (`App.jsx:65`) is overwritten to `general` by the tab-validity effect and never restored. Hold the requested tab as pending state and apply it when it appears in `visibleTabs`.
3. Preserve the in-progress prompt draft at `frontend/src/components/SettingsPanel.jsx:245` — `PromptManager` renders only under `activeTab === 'prompts'`, so any tab switch unmounts it and discards `editingId`/`draft` silently. Render it behind a `hidden` wrapper like the tools tab, or lift the draft into `SettingsPanel`.
4. Make `promptIntent` one-shot — add an `onIntentConsumed` callback from `PromptManager` (`frontend/src/components/PromptManager.jsx:38`) that clears the intent in `App`, so re-entering the Prompts tab does not yank the editor back to the originally-clicked prompt after `appliedIntentRef` resets on unmount.
5. Route "Full Admin Page" through the unsaved-changes guard — pass `onNavigate={requestClose}` instead of `onClose` at `frontend/src/components/SettingsPanel.jsx:239`, and defer `navigate('/admin')` until the guard resolves, so staged tool selections are not dropped without the prompt the X and backdrop raise.
6. Prevent pruned tool keys from being resurrected at `frontend/src/components/ToolsPanel.jsx:63` — the pending sets seed only on the closed→open transition, so after an Admin-tab MCP reload prunes `selectedTools`, Save re-adds the dead keys. Re-seed pending state when `hasChanges` is false and intersect pending keys against the current tool list inside `handleSave`.
7. Add dialog and tab semantics to the panel (`frontend/src/components/SettingsPanel.jsx:224,236`) — `role="dialog"`, `aria-modal`, an accessible name, focus move/trap on open, Escape-to-close, and `role="tablist"`/`role="tab"`/`aria-selected`/`role="tabpanel"` with arrow-key navigation; this panel is now the only route to tools, prompts, theme and admin controls.

## Evidence
Evidence: 8/10. Six live screenshots, a new 8-test suite, updated suites and a full 613-test run make this well demonstrated for a change of this size; the one notable gap is the unsaved-selection guard from the combined panel, which the new suite skips because it mocks `ToolsPanel` outright.

## Notes
- Clear the auto-dismiss `setTimeout` handles in `frontend/src/hooks/useAdminConfigActions.js:24`; the admin tab unmounts on every tab switch, so save confirmations are lost and timers fire against dead state.
- The always-mounted tools tree re-renders unmemoized on every General-tab slider tick — `React.memo` the embedded panel and `useMemo` `filteredServers` (`frontend/src/components/ToolsPanel.jsx:330`).
- Validate the `atlas:open-settings` event payload at `frontend/src/App.jsx:63` against a known tab list and intent shape rather than spreading `event.detail` through.

---

garland3:
## Summary
Consolidates the tools modal, settings, prompts and admin quick-controls into a single tabbed "Tools and Settings" panel, adds dialog focus/Escape handling and an unsaved-selection guard, and extracts shared admin config actions into a hook.

## Verdict
Request changes.

## Required changes
1. Defer navigation until the unsaved-selection guard resolves in `goToDashboard` at `frontend/src/components/admin/AdminQuickPanel.jsx:44` — `onNavigate?.()` (now `requestClose`) only raises the save/discard dialog when the tools tab is dirty, and the following `navigate('/admin')` unmounts the panel and silently discards the staged selections. Give `requestClose` a completion callback (or predicate) and navigate only from it.
2. Stop the document-capture Escape handler at `frontend/src/components/SettingsPanel.jsx:301` from closing the whole panel over nested UI — attach it to the dialog node in the bubble phase and ignore Escape while `UnsavedChangesDialog`, `TokenInputModal` or `AdminModal` is open, so half-entered token/config input and the unsaved-changes prompt are not destroyed.
3. Extend the close guard to cover an in-progress prompt draft (`frontend/src/components/PromptManager.jsx`), reporting dirty state the same way `onDirtyChange` does for tools — today Escape while typing in the Prompts editor closes the panel and loses the draft with no warning.
4. Make the re-seed effect at `frontend/src/components/ToolsPanel.jsx:63` a no-op when nothing changed — it depends on `savedSelectedTools`/`savedSelectedPrompts` Set identity and now runs in the steady state (`hasChanges === false`), so a context that republishes fresh Sets writes new state on every render. Compare contents, or key off a stable scalar/revision.
5. Fix the focus-trap visibility filter at `frontend/src/components/SettingsPanel.jsx:352` — `el.offsetParent !== null` is null for elements inside `position: fixed` ancestors, which is how the nested dialogs render, so Tab escapes behind a modal the user must answer. Use `el.getClientRects().length > 0` or scope the trap to the innermost dialog.
6. Split the dialog effect at `frontend/src/components/SettingsPanel.jsx:295` so the initial `closeButtonRef.focus()` is keyed on `[isOpen]` only — it currently depends on `requestClose`, which changes with `toolsDirty`, yanking focus to the Close button the moment a keyboard user toggles their first tool.
7. Remove `node_modules/.vite/vitest/da39a3ee5e6b4b0d3255bfef95601890afd80709/results.json` from the branch and add `node_modules/` to `.gitignore` — it is machine-local build cache, it records the new suite as `"failed":true`, and `npm ci` will leave a dirty tree.

## Evidence
Evidence: 8/10 — well demonstrated for a 27-file user-facing change, with six live screenshots, an 18-test suite, and quoted full-suite/lint/build results. The main gap is that the screenshots predate the follow-up rework (save-and-stay, Escape/focus handling) and the `closeGuardRef`/`requestClose` seam has no end-to-end coverage in either suite; one integration test rendering the real `ToolsPanel` inside `SettingsPanel` would close it.

## Notes
- Embedded "Save Changes" gives no confirmation the save happened, only a greyed-out button — `frontend/src/components/ToolsPanel.jsx` (`handleSave`).
- `aria-controls` on the General/User Info/Admin tabs points at panels that are not in the DOM until selected — `frontend/src/components/SettingsPanel.jsx` tab strip.
- `useAdminConfigActions` has no direct tests despite owning three-way per-endpoint request shaping, and `loadSystemStatus` ignores `response.ok` — `frontend/src/hooks/useAdminConfigActions.js:59`.

---

garland3:
Second review round — all 7 required items addressed in `1f7b056`.

1. **Deferred navigation.** `requestClose(afterClose)` now holds the action and runs it only from `finishClose`, once the panel is really dismissed. `AdminQuickPanel.goToDashboard` passes `() => navigate('/admin')`; the tools guard reports back on cancel so a user who backs out of the dialog does not get navigated, and the pending action is cleared rather than left to fire on some later close. With no host (the full dashboard renders these cards too) it navigates immediately as before.
2. **Escape scoping.** The handler moved off `document`-capture onto the dialog node in the bubble phase. `UnsavedChangesDialog`, `TokenInputModal`, `AdminModal` and the disconnect confirm now each declare `role="dialog"`/`aria-modal`, and the panel resolves the innermost one — while a nested modal is up it owns both Escape and the trap.
3. **Prompt draft guard.** `PromptManager` reports dirty via `onDirtyChange` (creating with any text, or editing text that differs from what is stored) and stops reporting on unmount. `requestClose` asks about the draft first, then falls through to the tools guard, so a close with both kinds of unsaved work asks about each. `UnsavedChangesDialog` took optional `title`/`message` and an optional `onSave` so it could be reused with discard/keep-editing.
4. **Re-seed.** Compares selection *contents* via a `sameKeys` helper and returns the previous `Set` when nothing changed, so React bails out — identity churn from the context can no longer write state every render.
5. **Focus trap.** `el.getClientRects().length > 0` instead of `offsetParent !== null`, and the trap is scoped to the innermost dialog.
6. **Focus effect split.** The initial `closeButtonRef.focus()` is its own effect keyed on `[isOpen]`; only the key handler depends on `requestClose`.
7. **`node_modules/.vite/.../results.json` removed** — my mistake, an over-broad `git add -A` after I ran vitest once from the repo root. Added `/node_modules/` and `frontend/node_modules/` to `.gitignore`. `test_e2e/node_modules/` predates this branch and is left tracked.

From the notes: the embedded save now shows a "Tool selections saved" status (it no longer dismisses, so the greyed-out button was the only feedback), `aria-controls` is omitted for tabs whose panel is not mounted, and `loadSystemStatus` checks `response.ok`.

**Evidence gap closed.** New `frontend/src/test/combined-panel-close-guard.test.jsx` renders the **real** `ToolsPanel` inside `SettingsPanel` — 8 tests over the `closeGuardRef`/`requestClose` seam: the guard raised on close with staged selections, closing once answered, Escape deferring to the nested modal, navigation withheld until the guard is answered, the deferred navigation dropped on cancel (and not resurrected by a later close), and the three prompt-draft paths.

**Verification.** `npm run lint` 0 errors (same 4 pre-existing warnings). `npx vitest run`: **682 passed / 60 files** (was 672). `npm run build` and `scripts/check-docs.sh` OK. Working tree clean, no build artifacts.

Still open, deliberately: direct unit tests for `useAdminConfigActions`, and memoizing the tools tree. Happy to add either.

garland3:
## Summary
Merges the tools surface, prompts, general settings, admin quick controls and user info into a single tabbed panel, adds a two-stage close protocol with unsaved-work guards, and extracts admin config actions into a hook.

## Verdict
Request changes.

## Required changes
1. Stop deleting saved selections for servers missing from the live tool list in `frontend/src/components/ToolsPanel.jsx:196` — compute `toolsToRemove` (and `promptsToRemove`) as saved minus the *unpruned* pending set and use the live-key intersection only to filter additions; today an MCP reload that has republished some servers but not others wipes the user's selection for the absent server on the next unrelated Save.
2. Restore the document-level Escape/Tab handling in `frontend/src/components/SettingsPanel.jsx:340`, scoping by event target to the innermost dialog instead of relying on the listener sitting on the dialog node — clicking any non-focusable text in the panel moves focus to `document.body`, after which Escape does nothing and Tab walks the page behind the modal.
3. Render the admin notification stack outside the `hidden` tabpanel at `frontend/src/components/SettingsPanel.jsx:440` (lift `notifications`/`removeNotification` into `SettingsPanel`, or portal `Notifications` to `document.body`) — `hidden` applies `display: none` to the ancestor, so admin save successes and errors are invisible from any other tab.
4. Do not answer the prompt guard by mutating the child's dirty flag in `frontend/src/components/SettingsPanel.jsx:116` — track the discard answer separately (reset when the close is cancelled) or expose an imperative `discardDraft()` on `PromptManager`; after Discard → Cancel the guard stays off for the panel's lifetime and the next close silently destroys the draft.
5. Defer the General-tab revert in `frontend/src/components/SettingsPanel.jsx:263` until the close actually completes (pass it as the `afterClose` action) — Cancel currently discards in-progress General settings even when a later guard aborts the close and the panel stays open.
6. Add a unit test over `useAdminConfigActions` covering the `banners` / `help-config` / default request shapes and the unmount timer cleanup; the hook shapes three different endpoints and has no direct coverage.

## Evidence
Evidence: 8/10 — proportionate to a 31-file user-facing change, with six live screenshots, a new user-guide page, two new test suites and quoted lint/test/build results. The main gap is that the screenshots predate two rounds of behavioural rework; a shot of the embedded Save Changes confirmation and of the Unsaved Prompt dialog at the current head would close it.

## Notes
- `ToolsPanel` is always mounted and re-renders in full on every panel state change, redoing the server merge per slider tick (`frontend/src/components/ToolsPanel.jsx:425`).
- "Cancel" reverts-and-stays on the Tools tab but reverts-and-dismisses on General (`frontend/src/components/ToolsPanel.jsx:236`).
- The panel is now a fixed `h-[85vh]` with a five-tab horizontally scrolling strip; Admin and User Info are off-screen with no affordance on small viewports (`frontend/src/components/SettingsPanel.jsx:357`).

---

garland3:
Third review round — all 6 required items addressed in `924dce2`.

1. **Saved selections for absent servers.** Correct, and the worse of my two failure modes. The live-key intersection now filters **additions** only; `toolsToRemove`/`promptsToRemove` are computed against the **unpruned** pending set, so a server merely missing from the current list keeps its saved selection. Test: `keeps a saved selection for a server missing from the live list` asserts `removeTools` is not called at all when an unrelated tool is toggled and saved.
2. **Escape/Tab back on the document, scoped by target.** You are right that a click on plain text drops focus to `document.body` and strands the dialog-node listener. The handler is on `document` again, returns early when a nested `role="dialog"` is open or when the target is outside the panel (body/documentElement excepted), and Tab from outside now pulls focus back to the first/last focusable rather than letting it walk the page. Two tests cover Escape and Tab from `document.body`.
3. **Admin notifications portalled to `document.body`.** `hidden` is `display: none` on the ancestor, so keeping the admin tab mounted — which I added precisely so its save results survive a tab switch — was hiding them anyway. `createPortal` fixes the thing the mount was for.
4. **Prompt guard no longer mutates the child's dirty flag.** `discardPromptAndClose` leaves `promptDirty` alone; the draft is only really gone when the panel unmounts, and a later guard can still abort. Test: `asks again after a discard whose close was then aborted` — draft + staged tools, discard the draft, cancel the tools guard, and the next close asks about the draft again.
5. **General revert deferred.** `handleCancel` passes `revertGeneralSettings` as the `afterClose` action, so an aborted close leaves in-progress General settings intact.
6. **`useAdminConfigActions` tested** — new `frontend/src/test/use-admin-config-actions.test.js`, 8 tests: the `banners` split-message POST, the `help-config` PUT, the default `content` + inferred `file_type` shape (json/yaml/text), the non-ok error notification, auto-dismiss timing (5s success / 8s error), timer cleanup on unmount, cancellation on manual dismiss, and `loadSystemStatus` rejecting a non-ok response.

**Verification.** `npm run lint` 0 errors (same 4 pre-existing warnings). `npx vitest run`: **694 passed / 61 files** (was 682). `npm run build` and `scripts/check-docs.sh` OK.

On the evidence note: the screenshots do predate the last two rounds. I have not re-shot them — that needs a live `atlas.main` and I did not want to attach stale-looking evidence; happy to if you want the embedded save confirmation and the Unsaved Prompt dialog captured at head.

The remaining notes are unaddressed and I think reasonably out of scope for this PR, but say the word: memoizing the tools tree, the Tools-vs-General "Cancel" asymmetry (Tools now reverts-and-stays by design from round 1, General still reverts-and-dismisses), and the five-tab strip scrolling off-screen on small viewports.

garland3:
## Summary
Consolidates the tools, prompts, general, user and admin surfaces into a single tabbed "Tools and Settings" panel, embedding `ToolsPanel` inside `SettingsPanel`, adding unsaved-work close guards, a window `CustomEvent` channel for opening the panel, and a user-guide page.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — the change is well demonstrated for its size (six live-server screenshots, three new test suites, quoted lint/unit/build/docs results), with gaps disclosed rather than hidden. The most useful missing artifact is a capture at head of the two newest surfaces — the embedded "Tool selections saved" confirmation and the "Unsaved Prompt" dialog — plus one narrow-viewport shot of the five-tab strip.

## Notes
- Escape and Tab-trapping both switch off while a nested dialog is open (`frontend/src/components/SettingsPanel.jsx:322`), and none of `UnsavedChangesDialog`, `AdminModal`, `TokenInputModal` or the disconnect-confirm implements either — so focus walks out of both overlays and Escape is a no-op on the unsaved-changes prompt. Route the key to the innermost dialog instead of returning early.
- Embedded "Cancel" discards staged tool selections with no prompt (`frontend/src/components/ToolsPanel.jsx:236`), while every other exit path asks; rename it "Discard Changes" and route it through the same dialog.
- `ToolsPanel` stays mounted and unmemoized (`frontend/src/components/ToolsPanel.jsx:353`), so every General-tab slider step re-merges, re-sorts and re-filters the full server tree; `React.memo` plus `useMemo` on `allServers`/`serverList` closes it.

---

garland3:
Thanks — took the three nits too, in `a4f911b`.

1. **Escape/Tab over nested overlays.** You were right that standing down entirely was worse than the bug it fixed. New `frontend/src/hooks/useEscapeKey.js` gives each overlay its own Escape (capture phase, stops propagation, so the innermost wins): `UnsavedChangesDialog`, `TokenInputModal`, `AdminModal`, and the disconnect confirm. The panel's Tab trap now runs **scoped to the innermost open dialog** instead of switching off, and pulls focus back into that scope when it has fallen to the body. The panel only claims Escape when no nested overlay handled it. Test: Escape on the unsaved-changes prompt dismisses the prompt and leaves the panel open.
2. **Embedded "Cancel" → "Discard Changes"**, routed through the same dialog (titled "Discard Changes", discard/cancel only — there is nothing to save on that path), and disabled when there is nothing staged. The standalone modal keeps "Cancel" and its close-on-cancel behaviour. Test updated to the new flow.
3. **`ToolsPanel` memoized.** Wrapped in `React.memo`, and the server merge/sort/filter moved into a `useMemo` keyed on the tool and prompt lists, so the always-mounted tools tab no longer redoes the tree per General-tab slider step.

`npm run lint` 0 errors (same 4 pre-existing warnings); `npx vitest run` **695 passed / 61 files**; build and `check-docs.sh` OK.

On the evidence gap: the screenshots still predate the last three rounds, and I have not re-shot them — that needs a live `atlas.main`. If you want the "Tool selections saved" confirmation, the "Unsaved Prompt" dialog, the new "Discard Changes" flow, and a narrow-viewport shot of the tab strip captured at head, I can do that; otherwise this is ready as far as I can take it.

The narrow-viewport tab strip (Admin and User Info scrolling off with no affordance) is the one note I have not touched — it is a pre-existing property of the five-tab layout rather than something the last three rounds introduced, and fixing it properly means a layout decision. Happy to take it here or leave it for a follow-up.

garland3:
## Summary
Consolidates the tools modal, prompt manager, general settings, user info and admin quick controls into a single tabbed Settings panel, with a shared close guard for unsaved work, a focus trap scoped to the innermost dialog, and a new `atlas:open-settings` window-event entry point.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — the change is well demonstrated for its size (six end-to-end screenshots, three new test suites, quoted lint/test/build/docs runs, and a named test for each behavioural seam added during rework). The most important gap is that the screenshots predate three rounds of rework and ship as user-guide documentation, so the embedded save confirmation, the "Unsaved Prompt" dialog, and the renamed "Discard Changes" flow have no current visual evidence.

## Notes
- Marketplace navigation from the embedded tools tab bypasses the unsaved-selection guard and calls `clearToolsAndPrompts()`, dropping saved selections without a prompt — `frontend/src/components/ToolsPanel.jsx:395`.
- An `edit` prompt intent that arrives before `userPrompts` resolves is neither consumed nor deferred, so it can overwrite a draft the user opened in the meantime — `frontend/src/components/PromptManager.jsx:57`.
- Escape in the admin config editor discards an in-progress textarea edit with no confirmation, unlike every other unsaved-work surface in this PR — `frontend/src/components/AdminModal.jsx:8`.

---

cmlanca:
Sources feels separated from the search tool now and adds extra clicks; would be better to combine source options within the same tool view. Additionally, when the search tool is enabled, have an ability to change and toggle the search. Ideal would be a short scroll menu in the chat bar that, if needed to expand, would pull up a popup window of the the settings menu.You should also see what datasets are enabled from the chat bar, with a small number being an expanded pill or similar. When too many datasets, it can be a short summary or something that can expand. It could be that the tools/prompts/data sources for intensive tasks get stored under the gear. Below are a few options for how that modal might look in the chat bar for sources using Claude that can serve for inspiration for what I think that might look like (ignore the numbering and color schemes; that's just from iteration):

Screenshot 2026-08-31 at 1 15 52 PM
Screenshot 2026-08-31 at 1 16 51 PM
Screenshot 2026-08-31 at 1 17 17 PM


Also if you can trigger a new prompt from the chat bar and bring up the prompt menu, it would be good to be able to pull up the tool window from the chat bar as well. I would prefer some degree of toggling for the tools rather than a long, indistinguiable list of tool names. this way you could add icons and a short description. It should still show the icon as enabled or the name in the chat bar, but would also function like prompt selection where you can see more for small tasks without changing views. 

The tool bar at the top has weird sizing on the menu selection when browser views change. They should have dynamic resizing rather than set sizes that get stuck. New chat has this problem in particular. 

The server selection for incognito, saved, saved local should be placed next to the chat history pane on the lefthand side. 

If adding the admin mode to tools and settings, you should remove the shield icon to reduce clutter in the toolbar. . 

You could combine the buttons for workspace with the saved files from chats so it's associated with a given workspace. This could further reduce the number of buttons. Consider if this reduces clutter or creates too much nesting. 

Put chat model selection in the chat bar rather than at the top; this is where people do most of their work and should be the focus.


Use a clearer help icon so you can reduce the words and use more icons in the tool bar. 

Make the custom prompt selection in the chat bar more obvious that you cannot select "custom prompt" and that is the menu header. Still looks selectable. You can change the wording so it's "Add Custom Prompts" or some of the images suggested here. 

image
image


If the datasets stay as a list on the side, then there needs to be better text wrapping. The names of datasets get cutoff and cannot be read fully, especially if anything gets appended to them. Make the datasets reformat to the amount of content included.

garland3:
@cmlanca thanks -- this was a really useful review. Pushed in 09f0ab7. Here is what I took, and what I deliberately did not. **Screenshots are attached in a follow-up comment** rather than committed to `docs/` (same rule as #860/#853 -- review evidence stays on the thread).

## Addressed

**Remove the shield icon.** Gone from both the desktop cluster and the compact menu. Admin controls are the Admin tab, and the full dashboard is one click on from there.

**Clearer help icon, fewer words.** Help and Agent Portal are icons only now (`CircleHelp`, tinted, instead of the outline circle sitting in a row of grey glyphs).

**Dynamic resizing of the top bar.** You were right that this was a fixed-size problem. The labels were on Tailwind viewport breakpoints (`hidden md:inline`) while the header itself sits beside the sidebar and the canvas panel, so the query kept labels at widths where they no longer fit and dropped them at widths where they did. They are now driven by the header's own measured width:

| Header width | Result |
| --- | --- |
| >= 1080px | Full desktop button cluster |
| >= 760px | Buttons keep their text labels |
| < 760px | Icon-only, cluster collapses into the menu |

Removing the model picker and the two word-labels also let the cluster threshold come down from 1320px to 1080px, so the compact menu kicks in much later.

**Combine sources into the search tool view.** Sources are only ever read by `atlas_search`, so the picker is now rendered inline at the top of **Tools & Integrations**, collapsible, with a live "N enabled" count. No extra clicks. The left-hand drawer still exists and is unchanged; both render the same new `DataSourcesSelector` component, so there is one implementation rather than two.

**Chat-bar tool control, with toggling instead of a flat list.** New **Tools** control under the message box: searchable, grouped by server, each row an icon + tool name + one-line description + an on/off switch. Small changes never need a trip into the panel; the footer link opens the full panel for everything else.

**Dataset visibility in the chat bar.** Enabled datasets show as pills above the message box. Each pill removes its dataset; past three the rest collapse into an expandable `+N more`, and the **Data Sources:** label opens the picker.

**Model selection in the chat bar.** Moved out of the top bar entirely -- same behaviour (capability icons, expandable model card, per-model API key button), menu now opens upward.

**"Custom Prompt" header looks selectable.** Now a muted, uppercase, non-interactive **ADD CUSTOM PROMPTS** label, with "Pick one below to customize AI behavior" under it. "My Prompts" got the same treatment for consistency.

**Dataset text wrapping.** `truncate` replaced with `break-words` on the name, description and server line, and the row is `items-start` so it grows to fit rather than clipping. Worth flagging honestly: the demo corpora in this environment all have short names, so the screenshot does not itself demonstrate a long name wrapping -- the change is in `DataSourcesSelector.jsx` and the row will now grow instead of cutting off.

## Deferred, with reasons

- **Server selection (incognito / saved / saved local) next to chat history.** Feasible, but it is a move across a component boundary the rest of this PR does not touch -- the save-mode control is header state and the sidebar is a sibling, and it also has to keep working in the compact menu where there is no sidebar. This PR is already large; I would rather do it as its own change than bolt it on.
- **Combine the workspace button with saved files from chats.** You flagged the nesting risk yourself, and I think it is real: workspace is a selection context and saved files is a browser, and merging them buries one behind the other. Wants a design pass, not an implementation.
- **The fuller "short scroll menu that expands into a popup" spec from your screenshots.** What I built is the compact-menu-plus-escape-hatch shape of that, but not a faithful reproduction of the reference designs, which I could not see. Happy to iterate if the chat-bar menus above are close enough in spirit to be worth refining rather than redoing.

## Verification

`npm run lint` clean (4 pre-existing warnings), `npx vitest run` 708 passed / 62 files, `npm run build` clean. 13 new tests cover the chat-bar tool toggles, the dataset pills, and the sources section inside the tools tab. All screenshots are new, taken against a live backend running this branch.

garland3:
## Screenshots

All captured against a live backend running this branch (`feat/combine-menus-836` @ 09f0ab7). Hosted on the orphan branch `pr-evidence/839-combine-menus`, same pattern as #862 -- not merged, delete once this lands.

### Top bar: no shield, icon-only Help and Portal

top bar

### Top bar: width-driven labels

Labels now key off the header's own measured width, not a viewport query. Medium (header >= 1080px, full cluster + labels) and narrow (header < 760px, icons only, cluster collapsed):

medium

narrow

### Data sources inside the Tools & Integrations tab

Collapsible, with a live "N enabled" count, above the installed tools list:

sources in the tools view

After **Enable All**, without leaving the tools view:

sources enabled

### Chat-bar tool toggles

Searchable, grouped by server, icon + name + one-line description + on/off switch, with an escape hatch to the full panel at the bottom:

chat-bar tool toggles

### Chat bar: model picker, tool control, prompt picker, dataset pills

Five datasets enabled -- three pills plus an expandable `+2 more`. The model picker (bottom left) is the one that used to live in the top bar:

dataset pills

### "Add Custom Prompts" heading

Muted, uppercase, non-interactive -- no longer reads as a selectable row:

add custom prompts

### Data sources drawer

Still available and unchanged in behaviour; renders the same `DataSourcesSelector` as the tools tab. As noted above, the demo corpora here all have short names, so this shot does not itself show a long name wrapping -- the `truncate` -> `break-words` change is in the diff.

sources drawer

garland3:
## Summary
Reworks the Tools and Settings surface: `ToolsPanel` becomes an embedded tab of `SettingsPanel` with a staged save/discard protocol, the model picker and tool picker move into the chat bar, and a new enabled-data-sources pill strip plus an `atlas:open-settings` event channel wire the surfaces together.

## Verdict
Request changes.

## Required changes
1. Route "Browse Marketplace" through the panel's close guard at `frontend/src/components/ToolsPanel.jsx:400` — `navigateToMarketplace` calls `clearToolsAndPrompts()` and navigates directly, so staged tools, saved selections and an in-progress prompt draft are discarded with no prompt. Move the `clearToolsAndPrompts()` call into the deferred after-close action so an aborted navigation does not still wipe saved selections.
2. Resolve the split commit model in the Tools tab at `frontend/src/components/ToolsPanel.jsx:684` — the inline `DataSourcesSelector` applies instantly while every other control in the same view is staged, so "Discard Changes" reverts only half of what the user touched. Either stage sources through `commitChanges`/`revertChanges`/`hasChanges`, or label the section as applying immediately.
3. Filter the chat-bar tool menu through the same source as the panel at `frontend/src/components/ToolSelector.jsx:42` — it builds rows from the raw `tools` list, so tools excluded by the compliance filter or by `selectedServers` are listed with working toggles.
4. Make the prompt `edit` intent single-shot at `frontend/src/components/PromptManager.jsx:57` — the effect re-runs when `userPrompts` resolves and overwrites a draft the user started while waiting; also call `onIntentConsumed` when the target no longer exists so the stale intent clears.
5. Handle Escape and focus return in the chat-bar popovers at `frontend/src/components/ToolSelector.jsx:31` and `frontend/src/components/ModelSelector.jsx:88` — both close only on outside click, leaving keyboard users no dismissal path.
6. Add coverage for `ModelSelector` (currently none) and fix `frontend/src/test/header-container-width.test.jsx:22`, which re-declares `DESKTOP_ACTIONS_MIN_WIDTH` as 1320 while `Header.jsx` now uses 1080 — import the thresholds instead of mirroring them.

## Evidence
Evidence: 9/10 — unusually complete for a change this size: nine end-to-end screenshots at head, four new test suites, and quoted lint/unit/build/docs results. The one real gap is that the committed user-guide screenshots (`docs/user-guide/tools-and-settings/screenshots/836-0*.png`) predate four rounds of rework and no longer match the shipped UI; re-capture them.

## Notes
- `filteredServers` is recomputed on every context-driven re-render (`ToolsPanel.jsx:464`); `memo()` does not help since the component consumes `useChat()` directly.
- Validate `event.detail` in the `atlas:open-settings` listener (`App.jsx:63`) — `tab` and `promptIntent` are spread into panel state unchecked.
- The line-ending normalization of `ChatArea.jsx` rewrites the whole file in this diff; worth splitting into a mechanical commit with a `.gitattributes` entry.

---

garland3:
CI is green for the first time on this branch -- all 10 checks pass and the PR is `MERGEABLE` / `CLEAN` at 92585c3.

Worth explaining the three extra commits, since the PR had been sitting in a conflicted state with no checks running at all:

- **48abb02** -- `ChatArea.jsx`, `RagPanel.jsx` and `MarketplaceContext.jsx` are CRLF in this tree, and my earlier commit rewrote them through tooling that normalizes line endings on read. That turned a 15-line change into a 1,244-line whole-file diff, which collided with #868 (also touching `ChatArea.jsx`) and left the PR `CONFLICTING` -- and a conflicted PR never gets CI scheduled, which is why no checks had ever run here. Restoring the original endings collapsed `ChatArea.jsx` from 2,499 diff lines back to 15. Done as a forward commit rather than a force-push so existing review anchors survive.
- **b386abb** -- merge of `main`, which then applied with zero conflicts.
- **92585c3** -- #871 landed on main while this was open and made the Active Tools strip a single non-wrapping row. The Data Sources pill strip this PR adds sits directly beneath it and was still on the old wrapping layout, so on a narrow viewport it would have wrapped into extra rows right under a strip that had just been fixed not to. Same treatment now (`flex-nowrap` + horizontal scroll, icon-only label below `sm`, `+N more` outside the scroll region), with three regression tests mirroring `enabled-tools-indicator.test.jsx`.

Local gates after the merge: lint 0 errors (4 pre-existing warnings), 746 tests / 66 files passing, `npm run build` and `check-docs.sh` clean.

@cmlanca the UX changes themselves are unchanged from my earlier write-up -- screenshots are in the comment above.

garland3:
## Summary
Merges the tools panel into a combined Tools and Settings panel (tabs for tools, prompts, theme, admin, user info), adds chat-bar tool/model/prompt/data-source controls, a close guard with staged save/discard, and an accompanying user-guide page.

## Verdict
Request changes.

## Required changes
1. Defer `clearToolsAndPrompts()` until the navigation actually proceeds in `navigateToMarketplace` (`frontend/src/components/ToolsPanel.jsx:400`) and route the click through the close guard — today "Add from Marketplace" / "Browse Marketplace" silently wipe *saved* tool and prompt selections with no unsaved-changes prompt.
2. Build `ToolSelector`'s rows from `getComplianceFilteredTools(complianceLevelFilter)` / `getFilteredTools()` (`frontend/src/components/ToolSelector.jsx:42`) — the chat bar currently lists and toggles tools the panel hides for compliance level or unselected servers, and add the parity case to `chat-bar-controls.test.jsx`.
3. Union externally-added keys into `pendingSelectedTools` instead of skipping the re-seed while `hasChanges` is true (`frontend/src/components/ToolsPanel.jsx:211`) — a chat-bar toggle made with staged edits open is removed again by the next Save.
4. Consume the prompt intent once the list resolves and the target id is absent, and on the first user-driven `setEditingId` (`frontend/src/components/PromptManager.jsx:55`) — a late-arriving `edit` intent overwrites a draft the user has already started.
5. Import `DESKTOP_ACTIONS_MIN_WIDTH` / `ACTION_LABELS_MIN_WIDTH` from `Header.jsx` in `frontend/src/test/header-container-width.test.jsx:22` — the test's local copy says 1320 while the component now ships 1080, so it asserts nothing.
6. Re-capture `docs/user-guide/tools-and-settings/screenshots/836-01..06` at head — they predate four rounds of rework and ship as documentation showing a header, model picker and footer that no longer exist.
7. Either stage data-source selections through `commitChanges`/`revertChanges` or label the section as applying immediately (`frontend/src/components/ToolsPanel.jsx:659`) — "Discard Changes" currently reverts tools but keeps datasets.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size (nine end-to-end screenshots at head, four new test suites, green CI). The one material gap is the stale committed user-guide screenshots, covered above.

## Notes
- `memo(ToolsPanel)` cannot skip re-renders for a direct context consumer; the always-mounted tools tree still re-renders on all chat-context churn (`frontend/src/components/ToolsPanel.jsx:1179`).
- Chat-bar popovers close only on outside mousedown — no Escape and no focus return (`frontend/src/components/ToolSelector.jsx:31`, `ModelSelector.jsx:88`).
- Validate `event.detail` (`tab`, `promptIntent`) before spreading it into panel state (`frontend/src/App.jsx:63`).

---

garland3:
## Summary
Consolidates tools, prompts, data sources, general settings and admin quick actions into a single tabbed `SettingsPanel`, moves the model/tool/data-source pickers into the chat bar, and reworks the header's width-driven layout. Large change: 41 files, +4694/-2352.

## Verdict
Request changes.

## Required changes
1. Route "Add from Marketplace" through the close guard at `frontend/src/components/ToolsPanel.jsx:400` — `navigateToMarketplace` currently calls `clearToolsAndPrompts()` and navigates unconditionally, wiping *saved* tool/prompt selections and any in-progress prompt draft with no unsaved-changes dialog. Move both the clear and the navigate into the deferred `afterClose` action so a cancelled close leaves state intact.
2. Merge externally-added selections into the pending set at `frontend/src/components/ToolsPanel.jsx:211` — while the panel is dirty the re-seed effect returns early, so a tool enabled from the new chat-bar menu is treated as a removal by `commitChanges` and switched back off on Save. Diff against a `prevSavedRef` snapshot and union in newly-added keys instead of skipping the re-seed.
3. Build the chat-bar tool rows from the marketplace filters at `frontend/src/components/ToolSelector.jsx:42` — use `complianceEnabled ? getComplianceFilteredTools(complianceLevelFilter) : getFilteredTools()` as `ToolsPanel.jsx:388` does, so tools hidden by compliance level or by unselected servers are not toggleable from the chat bar, and add a parity assertion to `chat-bar-controls.test.jsx`.
4. Fix the header width test at `frontend/src/test/header-container-width.test.jsx:22` — it re-declares `DESKTOP_ACTIONS_MIN_WIDTH = 1320` while `Header.jsx:42` ships 1080, so the test cannot fail. Export both thresholds from `Header.jsx`, import them, and add a case covering the new `ACTION_LABELS_MIN_WIDTH` label collapse.
5. Consume the prompt edit intent once the list resolves at `frontend/src/components/PromptManager.jsx:55` — a late-arriving intent currently re-applies after `userPrompts` loads and overwrites a draft the user has already started, and an intent for a deleted id is never consumed. Mark it applied on the first user-driven `setEditingId` and when the target is absent from a resolved list.
6. Resolve the mixed commit models at `frontend/src/components/ToolsPanel.jsx:659` — `DataSourcesSelector` writes through immediately while everything else in the tab is staged, so **Discard Changes** half-reverts. Either fold source selections into `hasChanges`/`commitChanges`/`revertChanges`, or label the section as applying immediately.
7. Re-capture the six committed screenshots in `docs/user-guide/tools-and-settings/screenshots/` at head — `836-01-header.png` still shows the admin shield, the header model picker and the word-labelled Help/Portal buttons, all removed here, and the tools shots predate the inline data-sources section and the Discard footer.

## Evidence
Evidence: 8/10 — well-demonstrated for a change of this size, with nine end-to-end screenshots, four new test suites, and quoted lint/unit/build results. The one material gap is that the screenshots that ship as user documentation were captured before several rounds of rework and no longer depict the shipped UI (covered above).

## Notes
- `atlas:open-settings` detail is spread into panel state unvalidated — validate `tab` against known ids and `promptIntent` shape (`frontend/src/App.jsx:89`).
- Chat-bar popovers close only on outside `mousedown`: no Escape, no focus return, and `aria-label="Select chat model"` hides the current model name (`frontend/src/components/ToolSelector.jsx:31`, `ModelSelector.jsx:53`).
- `memo(ToolsPanel)` cannot short-circuit while the component reads `useChat()`/`useMarketplace()` directly, and `filteredServers` re-filters the whole tree on every context publish (`frontend/src/components/ToolsPanel.jsx:464`).

---

garland3:
## Triage of the two AGENT-REVIEW-BOT-3 reviews

Both reviews raise the same seven required changes. **All seven are real and all seven are fixed** — nothing rejected. Commits: `b577d39` (code), `5f4aa92` (screenshots), `2be85d5` (a11y note).

### Fixed

| # | Finding | Fix |
|---|---|---|
| 1 | `navigateToMarketplace` cleared **saved** selections and navigated with no guard | Both the clear and the navigation are deferred behind the close guard. Embedded, it defers to the whole panel's `requestClose`, so an unfinished prompt draft is asked about too. Cancelling leaves saved selections untouched. |
| 2 | Chat bar listed/toggled tools the panel hides | `ToolSelector` now builds rows from `getComplianceFilteredTools(complianceLevelFilter)` / `getFilteredTools()`, via `useOptionalMarketplace` — matching `ModelSelector`, its sibling from this PR. |
| 3 | Re-seed skipped while dirty, so a chat-bar toggle was removed by the next Save | The effect now replays the *delta* in the saved selections onto the staged edits against a `prevSaved` snapshot, unioning additions and dropping external removals. |
| 4 | Prompt intent re-applied over a started draft; intent for a deleted id never retired | Consumed on the first user-driven `setEditingId`, and when the target is absent from a **resolved** list (gated on `userPromptsLoading`, so a still-loading list still waits). |
| 5 | `header-container-width.test.jsx` re-declared `DESKTOP_ACTIONS_MIN_WIDTH = 1320` vs the shipped 1080 | Both thresholds are exported from `Header.jsx` and imported. Added cases pinning both values and covering each threshold's boundary plus the `ACTION_LABELS_MIN_WIDTH` label collapse. |
| 6 | Mixed commit models — "Discard Changes" half-reverted | Took the "label it" option: `DataSourcesSelector` is shared with the RAG panel, where nothing is staged, so folding it into `commitChanges` would have meant staging in one host and not the other. The section now says data source changes apply immediately, and the user guide says so too. |
| 7 | Screenshots predated several rounds of rework | All six re-captured against a live server at head. The tools shot now collapses the data-sources section so the installed tools *and* the Save/Discard footer are both visible; the picker shot has a saved prompt so the pencil is actually shown. |

Also took two of the three **Notes**:

- `atlas:open-settings` detail is now validated (`parseOpenSettingsDetail`) — unknown tab ids and malformed prompt intents are dropped rather than spread into panel state.
- Escape now closes both chat-bar popovers and returns focus to the trigger, and the model button's `aria-label` names the current model instead of the fixed "Select chat model".

### Deferred (not rejected)

The `memo(ToolsPanel)` note is correct — `memo` cannot short-circuit a component that reads `useChat()`/`useMarketplace()` directly, so it is dead weight, and `filteredServers` re-filters on every context publish. Fixing it properly means splitting the context reads out of the memoized boundary, which is a refactor with real regression surface for a panel this PR has already reworked several times. I would rather not land that in the same change; happy to do it as a follow-up if you would prefer it here.

### Verification

- `npm run lint` — 0 errors (4 pre-existing warnings, none in touched files).
- `npx vitest run` — **782 passed, 69 files** (was 613 before this round).
- Each new regression test confirmed to **fail without its fix** — reverted each change in turn and checked the specific test went red, so none of them are no-ops. That was the point of finding #5.
- `npm run build` clean; `scripts/check-docs.sh` OK.
- Screenshots captured from a real browser run against a live backend, with zero console/page errors; Escape + focus-return also verified there rather than only in jsdom.

garland3:
### Re-captured screenshots (finding #7)

These are the six files now committed under `docs/user-guide/tools-and-settings/screenshots/`, captured at head against a live backend.

Header — one wrench, no admin shield, no top-bar model picker, Help/Portal icon-only:

Header

Tools tab — collapsed data-sources row, installed tools, and the Save / Discard footer:

Tools tab

General, with the Appearance toggle:

General tab

Admin quick controls:

Admin tab

Prompt picker — now with a saved prompt, so the pencil and "New system prompt" are both visible — and the editor it opens:

Prompt picker

Prompt editor

One thing worth a maintainer's call: the Admin card shows the local config path, which in my capture is a worktree path (`.../atlas-ui-3-wt-839/config/mcp.json`), and four locally-failing MCP servers in red. That is honest current state from the machine that captured it, but if you would rather the docs show a clean `atlas-ui-3` path and no failed servers, say so and I will re-shoot from a plain checkout.

garland3:
## Summary
Consolidates settings, tools, prompts, data sources and admin quick-controls into a single tabbed settings panel, and adds chat-bar surfaces for tool selection and enabled data sources. Large change: 43 files, roughly +5200/-2400.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well-demonstrated for its size — re-captured user-guide screenshots at head, nine end-to-end captures, four new test suites, and quoted lint/test/build gates, with known gaps disclosed rather than hidden. The only missing artifact worth adding is a screenshot of the open panel at a narrow viewport, so the deferred tab-strip overflow is a decision made on evidence.

## Notes
- Chat-bar tool toggles apply immediately while the Tools tab stages the same checkboxes until Save; a panel Discard then reverts them — add the "applies immediately" note used by data sources (`frontend/src/components/ToolSelector.jsx:158`).
- `openSettingsPanel({ tab: 'tools' })` from the data-sources strip lands on a different tab when `features.tools` is false, leaving no route to the picker (`frontend/src/components/EnabledDataSourcesIndicator.jsx:42`).
- Chat-bar tool count uses the unfiltered `selectedTools`, so it can exceed the rows the menu renders (`frontend/src/components/ToolSelector.jsx:88`).

---

garland3:
## Summary
Consolidates the tools, prompts, general, user-info and admin surfaces into a single "Tools and Settings" panel, moves the model and tool pickers into the chat bar, and adds a window-event channel plus a shared Escape/focus-trap hook to coordinate the new overlays.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — screenshots re-captured at head, four new suites, 782 passing unit tests with each new regression test confirmed to fail without its fix, and green CI. The one artifact still missing is a capture of the open panel at a narrow viewport (~600–760px), where the five-tab strip's overflow behavior was deferred as a layout decision.

## Notes
- Escape from the message composer is swallowed while a chat-bar popover is open, so the `@`-tool autocomplete cannot be dismissed — `stopPropagation` on a document capture listener does not suppress sibling listeners; use `stopImmediatePropagation` (`frontend/src/hooks/useEscapeKey.js:20`).
- The "Data Sources:" shortcut always opens the `tools` tab, which is hidden when `features.tools` is false, leaving no route to the picker (`frontend/src/components/EnabledDataSourcesIndicator.jsx:42`).
- The chat-bar tool count reads the unfiltered `selectedTools` set while the rows come from the compliance/marketplace-filtered list, so the label can exceed the menu (`frontend/src/components/ToolSelector.jsx:90`).

---

PR #840, opened 2026-08-25
fix(sessions): pluggable session repository + disconnect persistence test (#760)

### LINKED ISSUE #755: Agent mode: stopping the agent discards the turn - prior tool calls are not persisted and not in the next turn's context

garland3:
## Summary

Addresses the **prerequisite** identified in issue #760 for decoupling session identity from the WebSocket connection: the session store was hardcoded to `InMemorySessionRepository` in `app_factory.py`, forcing sticky sessions on multi-replica deployments.

### What changed

1. **Pluggable session repository** (`SESSION_REPOSITORY_TYPE`, default `memory`): a new setting and factory function (`atlas.infrastructure.sessions.factory.create_session_repository`) select the session store implementation at startup. The default `memory` behavior is unchanged. An unknown type raises `ValueError` at startup instead of silently falling back to the in-memory store (which would re-introduce the sticky-session requirement the setting is meant to remove). A deployment that needs a distributed store (Redis, shared DB, etc.) registers it by extending the factory -- the `SessionRepository` protocol in `atlas.interfaces.sessions` is the contract.

2. **End-to-end disconnect persistence test**: `test_disconnect_persists_in_flight_turn` drives the real `cleanup_disconnected_session` entry point against a real `ChatService` and verifies the turn is saved. This closes the gap between the persistence unit tests (which call `handle_chat_message` directly, added by PR #776 for #755) and the disconnect tests (which used a plain `never_ending` coroutine). PR #776's `except asyncio.CancelledError` handler commits completed work before re-raising; this test proves it works through the actual disconnect path.

3. **Factory unit tests** (`test_session_repository_factory.py`): verify default, explicit `memory`, unknown-type error, and that `AppSettings` exposes the field.

### What's NOT in this PR

The issue describes two separable parts. **Part 1** (commit what already happened on every cancel path) was done by PR #776 (closed #755). This PR adds the test that proves Part 1 works for disconnects specifically.

**Part 2** has three sub-goals:
- **Pluggable session repository** -- done here (the prerequisite the issue explicitly identifies).
- **Decouple session identity from the connection** (key session on user+conversation rather than `uuid4()` per socket) -- NOT done. The issue says this "deserves deliberation."
- **Continuation setting** (opt-in server-side continuation with step/wall-clock caps) -- NOT done. The issue says this is a policy decision: an agent-mode turn running unattended executes tools with no one watching, and the blast radius depends on which MCP servers are enabled.

### Files changed

- `atlas/modules/config/settings.py` -- `SESSION_REPOSITORY_TYPE` setting
- `atlas/infrastructure/sessions/factory.py` -- `create_session_repository` factory (new)
- `atlas/infrastructure/app_factory.py` -- use factory instead of hardcoded `InMemorySessionRepository()`
- `atlas/tests/test_websocket_disconnect_resilience.py` -- `test_disconnect_persists_in_flight_turn`
- `atlas/tests/test_session_repository_factory.py` -- factory unit tests (new)
- `.env.example`, `docker-compose.yml` -- document the new env var
- `CHANGELOG.md` -- entry
- `test/pr-validation/test_pr840_pluggable_session_repository.sh` -- PR validation script (new)

## Test plan

- [x] Factory returns `InMemorySessionRepository` for `memory` (default and explicit)
- [x] Unknown type raises `ValueError` mentioning the extension point
- [x] `AppSettings.session_repository_type` defaults to `memory`
- [x] End-to-end disconnect persists the in-flight turn (the completed tool call survives)
- [x] `docker-compose.yml` env sync test passes
- [x] Backend tests pass (2537 passed, 17 skipped)
- [x] Frontend tests pass (605 passed)
- [x] `ruff check` clean on all changed/added files
- [x] PR validation script passes (8/8)

Refs #760 (Part 2 prerequisite; Part 1 was closed by #776).


garland3:
## Summary
Adds a `SESSION_REPOSITORY_TYPE` setting and an `atlas/infrastructure/sessions/factory.py` factory so the session repository is selected at startup instead of hardcoded, and adds a disconnect-resilience test that drives the real `cleanup_disconnected_session` path.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is backend-only and adequately demonstrated — the headline claim (a mid-turn disconnect persists completed work) is asserted against the real entry point, and the factory has default/explicit/unknown-type coverage. The only missing artifact is verifiable output for the quoted suite counts (a pasted run tail or CI link).

## Notes
- Normalize the setting before dispatch at `atlas/infrastructure/sessions/factory.py:38` — `Memory`, `MEMORY`, or an empty `SESSION_REPOSITORY_TYPE=` currently raises during package import rather than falling back to the documented default.
- `test_app_settings_has_session_repository_type` at `atlas/tests/test_session_repository_factory.py:38` reads the ambient environment; `monkeypatch.delenv("SESSION_REPOSITORY_TYPE", raising=False)` before constructing `AppSettings()`.
- The error message and env/compose comments say a new store "must be registered in" the factory, but there is no registry — either add one or reword to say it requires a code change (`atlas/infrastructure/sessions/factory.py:43`).

---

garland3:
## Review feedback addressed

**Codex P1 -- Persist the session mutated by the orchestrator**: Valid concern. The in-memory repo returns the same object reference on every `get()`, so `ChatService` and `ChatOrchestrator` share the same `Session` instance. A distributed adapter that deserializes a fresh copy on each `get()` would break this. I've documented the constraint in `docs/admin/session-store.md` (under "Current Limitations -- Object identity matters") so a future implementer knows to either cache the session per request or pass it through the call chain. This is a concern for the future distributed implementation, not the current in-memory default.

**Codex P1 -- Document the new session repository configuration**: Added `docs/admin/session-store.md` with the setting, the sticky-session constraint, how to register a distributed implementation, and the current limitations. Indexed in `docs/admin/README.md`.

**Copilot -- Changelog entry too long**: Shortened to a 1-2 line summary per AGENTS.md.

**Copilot -- Unused `session` parameter**: Removed from `_make_real_chat_service`.

**Copilot -- `asyncio.sleep(30)` on failure**: Replaced with `await asyncio.wait_for(block_forever.wait(), timeout=30)` where `block_forever` is an `Event` that never fires. The task is immediately cancellable and fails fast if the cancel never arrives.

**CodeQL -- Statement has no effect** (`await chat_task` inside `pytest.raises`): This is a false positive -- the `await` is the statement that should raise `CancelledError`. Left as-is.

garland3:
## Summary
Makes the session store pluggable via a new `SESSION_REPOSITORY_TYPE` setting and a `create_session_repository` factory (default `memory`, unchanged behavior), documents the multi-replica constraint, and strengthens the WebSocket disconnect persistence test.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. The change is well demonstrated for its size — unit tests cover the factory and a PR-validation script exercises the wiring. The one gap is that the quoted suite counts are asserted rather than shown, and the follow-up commit that rewrote `test_disconnect_persists_in_flight_turn` has no re-run result attached; pasting the pytest tail at the current head would close it.

## Notes
- Normalize the setting before dispatch — `SESSION_REPOSITORY_TYPE=`, `Memory`, or a trailing space currently raise at startup instead of resolving to the documented default (`atlas/infrastructure/sessions/factory.py:38`).
- `ChatService.__init__` still falls back to a hardcoded `InMemorySessionRepository()`, so the setting is not universal (`atlas/application/chat/service.py:127`); the docs' "registered in ...factory" wording also promises a registration API that does not exist.
- The docs sample does not run as written — no `import os`, and `SESSION_REPOSITORY_URL` is not a defined setting (`docs/admin/session-store.md:31`). `test_app_settings_has_session_repository_type` also reads the ambient env/`.env` rather than asserting the declared default (`atlas/tests/test_session_repository_factory.py:38`).

---

PR #841, opened 2026-08-25
fix(ui): sleep tool shows progress clock, not 'taking longer' (closes #838)

### LINKED ISSUE #838: sleep tool should not say it is taking a long time
When the sleep tool is used, then UI says, "▶
atlas_agent_sleep
5m 05s
- taking longer than expected"

but this is expected. 

Maybe for this special case, show a timer of the sleep amount, like 05:12 of 20:00.

garland3:
## Problem (issue #838)

`atlas_agent_sleep` is a built-in pseudo-tool whose purpose is to wait minutes to hours (polling long-running external work). The generic 30-second slow-tool threshold in `ToolElapsedTime` appends `- taking longer than expected` to every active tool call past 30s, so a normal 20-minute sleep shows `5m 05s - taking longer than expected` — which reads like a bug, not a wait.

## Fix

The active `atlas_agent_sleep` row now shows a **progress clock against the requested duration** instead of the generic warning:

- `MM:SS of MM:SS` while the wait is in progress (e.g. `05:12 of 20:00`)
- `HH:MM:SS of HH:MM:SS` for hour-plus waits (e.g. `00:05:12 of 02:00:00`), with both sides kept on the same clock style once either crosses the hour boundary
- Switches to a `- completing...` hint once the requested wait has elapsed and the `tool_complete` frame has not arrived yet (the genuinely-anomalous case the issue comment asked about)
- The generic `- taking longer than expected` warning is **suppressed** for the sleep tool; every other tool keeps the existing timer + 30s threshold

A missing, non-positive, or non-numeric `seconds` argument falls back to the generic timer so the row never renders `of NaN:NaN`.

## Why the small surface

The change is scoped to `ToolElapsedTime.jsx` (the live ticker extracted from `Message.jsx`). `Message.jsx` just passes two new props (`toolName`, `arguments`) it already had on hand. No backend, WebSocket, or persistence changes — the `tool_start` frame already carries the parsed `arguments` (including `seconds`), so the frontend already had everything it needed to show progress against the requested duration.

## Screenshots / verification

I could not drive a live agent-mode turn (no LLM API key on this host, and the Playwright browser was held by another concurrent run), so verification is via:

- 14 new unit tests in `ToolElapsedTime.test.jsx` covering the generic timer, the sleep progress clock, the `completing...` hint, hour-plus formatting, numeric-string `seconds`, and the three fallback paths (missing / non-positive / non-numeric)
- 3 new `Message`-level integration tests in `sleep-tool-timer.test.jsx` rendering the full tool-call row (compact + classic mode) and asserting the new copy appears and the old warning does not
- Full frontend suite: **622/622 pass** (was 619; +3 Message-level, +14 unit). `npm run lint` clean on the touched files (no new warnings)

## Notes for reviewers

- The displayed total is the **requested** `seconds`, not the backend-clamped duration. If a model asks for 99999s and the per-call cap clamps it to 7200s, the row reads `MM:SS of 1666:39` and flips to `completing...` after 2h. Showing the requested value is honest about what the model asked for; showing the clamped value would require the backend to emit the actual (clamped) duration in the `tool_start` frame, which is a larger change I left for a follow-up if a reviewer wants it.
- I did not retroactively edit the dated design note `docs/developer/design-notes/agent-sleep-tool-2026-08-18.md`, whose "Known limits" section already said the UI showed no elapsed timer (that was stale before this PR). Design notes are point-in-time records per `AGENTS.md`; the CHANGELOG entry documents the new behavior.

Closes #838.


garland3:
## Summary
Replaces the generic "taking longer than expected" warning on `atlas_agent_sleep` tool rows with a progress clock (`05:12 of 20:00`, plus a `completing...` hint once the wait runs over), and adds 17 unit and Message-level tests covering the formatting, thresholds, and fallback paths.

## Verdict
Approve with nits.

## Evidence
Evidence: 7/10. Test coverage is unusually thorough for a change this size and the description states the trade-offs up front, but the deliverable is on-screen copy and no screenshot of the rendered row is included — one mid-wait and one overdue capture would close the gap. Also worth reconciling the "622/622 (was 619)" count in the description with the 17 tests actually added.

## Notes
- The clock counts toward the *requested* seconds, but the backend clamps twice — per call and against the remaining per-turn budget (`sleep_tool.py:159-163`) — so a second sleep in a turn can vanish before reaching its total; either emit the effective duration on the `tool_start` frame or label the total as "requested" (`ToolElapsedTime.jsx:70`).
- `seconds: true` coerces to `1` and renders a bogus one-second countdown for a call the backend rejects outright; mirror the backend's boolean guard before coercing (`ToolElapsedTime.jsx:58`).
- The ticking clock sits inside the tool row's `<button>`, so its accessible name changes once a second for the full multi-hour wait — mark it `role="timer" aria-live="off"` or move it out of the button's labelled content (`Message.jsx:302`).

---

garland3:
Thanks for the review feedback — both points addressed in c11ccb4:

**@chatgpt-codex-connector (P2, stale `display_args` after approval edit):** Confirmed and fixed. The user-edit path in `execute_single_tool` recomputed `filtered_args` (the args that actually run) but left `display_args` (the UI-facing copy) stale, so `tool_start` still carried the pre-approval request — meaning the sleep progress clock would show `MM:SS of 20:00` while the backend actually slept for 5s. Added `display_args = _sanitize_args_for_ui(dict(filtered_args))` right after the user-edit re-filter, matching the existing hook-rewrite paths (tool_executor.py:499 and :537). New regression test `atlas/tests/test_tool_start_edited_args.py` fails with `assert 1200 == 5` without the fix and passes with it.

**@Copilot (`completing...` boundary):** Fixed — switched `elapsed > sleepSeconds` to `>=` so the hint appears the instant the requested wait is reached instead of lagging up to 1s. Added a boundary test (`tick(5)` against `seconds: 5`) to lock it in.

Both fixes were verified locally: backend pytest (`test_tool_start_edited_args.py` + the existing telemetry/approval/hooks suites, 94 passed), frontend vitest (623/623), and `ruff` / `npm run lint` clean.

garland3:
## Summary
Replaces the static "taking longer than expected" warning on an active `atlas_agent_sleep` tool row with a live progress clock (`05:12 of 20:00`, plus a `- completing...` hint past the requested duration), and fixes a backend path where approval-edited arguments were not re-sanitized before the `tool_start` frame was emitted.

## Verdict
Approve with nits.

## Evidence
Evidence: 7/10. Unusually thorough for the size — 17 new frontend tests plus a backend regression test with a stated failure signature — but the deliverable is on-screen copy and no rendered capture of the row is included; two fixture screenshots (mid-wait and overdue) would close that. Note also that the test-count arithmetic in the description does not reconcile (619 + 17 ≠ 622).

## Notes
- The clock's denominator is the *requested* `seconds`, but `sleep_tool.py:159-163` clamps the wait to `min(max_seconds, remaining turn budget)`; when a clamp fires the row advertises a total the backend never intends to reach. Either emit the post-clamp duration on `tool_start`, or label it `of 20:00 requested` — `ToolElapsedTime.jsx:70`.
- The ticker renders inside the disclosure `<button>`, so the row's accessible name changes once a second for the entire multi-hour wait. Mark it `aria-hidden` inside the button (or expose it as a labelled `role="timer"`) — `Message.jsx:302`.
- `Number(rawSeconds)` accepts values the backend rejects (`true` → `1`, `[1200]` → `1200`), rendering a countdown for a call that fails immediately; accept only numbers and numeric strings — `ToolElapsedTime.jsx:58`.

---

PR #845, opened 2026-08-25
fix(rag): tell the LLM and user when a RAG query fails instead of failing silently (closes #844)

### LINKED ISSUE #844: silent failure on RAG is bad user experience
If querying an atlas-rag. atlas.modules.llm.litellm_caller - ERROR - [RAG] Failed to query server atlas_rag-prod: 500: RAG service error
Then, then the LLM should be told the query failed and told to tell the user. 
right now, it fails silently, so the user never knows there was an error and the LLM just tries to answer which is bad for the user.

garland3:
## Problem (issue #844)

When a selected RAG data source errored -- e.g. `atlas.modules.llm.litellm_caller` logging `[RAG] Failed to query server atlas_rag-prod: 500: RAG service error` -- the turn failed **silently**: it logged the error and fell back to a plain LLM answer as if nothing had happened. The user was never told the retrieval failed, and the LLM had no idea the RAG context it was (not) given was missing because of an error. The issue calls this out as bad UX.

## Fix

Surface RAG query failures to the LLM so it tells the user, instead of dropping them silently.

- `_query_all_rag_sources` now returns a **3-tuple** `(successful, exclusions, failures)`. A non-permission error (e.g. a 500 from the RAG service) populates `failures` with one user-facing message per failed server batch naming the **corpus** the user selected. It still does **not** raise (an unreachable backend is not an authorization problem); the hard `DataSourcePermissionError` is reserved for the all-groups-denied case, unchanged.
- New `_build_rag_failure_notice(failures)` (mirrors the existing `_build_rag_exclusion_notice`) renders the notice for the RAG system message and instructs the model to **tell the user in its first sentence** that the sources could not be reached.
- **Partial failure** (some sources ok, some broken): the failure notice rides into the RAG context block alongside the retrieved context, on both the streaming and non-streaming paths.
- **All sources failed**: instead of a silent plain/tools-only fallback, a system message is injected telling the model the retrieval failed and to begin its reply by telling the user; the LLM is still called (no raise) so the user gets a meaningful response.
- Applies uniformly to `call_with_rag`, `call_with_rag_and_tools`, `stream_with_rag`, and `stream_with_rag_and_tools`.

### Safety properties preserved
- **Raw error text never reaches the model or user** -- only the corpus name is named; the `500: RAG service error` detail stays in the server log.
- **Permission denials** behave exactly as before (`exclusions` + `DataSourcePermissionError` when all are denied).
- **LLM domain errors** (`RateLimitError`, `LLMTimeoutError`, `LLMBadRequestError`, `ContextWindowExceededError`, `DataSourcePermissionError`, `HookBlockedError`) still re-raise and are **not** masked by the new failure-injection path.
- The `is_completion` direct-return shortcut on the RAG-only path is unchanged.

## Tests

- Updated `test_rag_partial_compliance_exclusion.py`: the old `test_non_permission_failures_still_degrade_quietly` (which encoded the silent behavior as intended) is replaced by `test_non_permission_failures_are_reported_not_silent` -- still no raise, but the failure is now returned. Added tests for `_build_rag_failure_notice`, partial-failure notice reaching the RAG system message, and the all-sources-failed path informing the LLM on all four call paths (non-streaming, +tools, streaming, streaming+tools).
- Updated the 3-tuple mock return values in `test_rag_context_insert_position.py`, `test_rag_tools_is_completion.py`, `test_tool_error_attribution.py`.
- Added `test/pr-validation/test_pr844_rag_failure_notifies_llm.sh` exercising the fix end-to-end (8 checks including the LLM-error passthrough guarantee).

## Verification

- `ruff check` clean on all changed files.
- Affected unit tests: 52 passed.
- `bash test/run_tests.sh backend`: passed.
- `bash test/pr-validation/test_pr844_rag_failure_notifies_llm.sh`: 8/8 passed.

## Risk

Low. The change is confined to the RAG fan-out and the four RAG-aware LLM call paths. The failure path is additive: when `failures` is empty (the common case) behavior is identical to before. The one behavior change -- no longer silently degrading on a total RAG failure -- is exactly what the issue asks for.

## Checklist
- [x] CHANGELOG entry added under `## [Unreleased]`
- [x] PR-validation script added (`test/pr-validation/test_pr844_rag_failure_notifies_llm.sh`)
- [x] Lint clean on changed files
- [x] Backend tests pass


garland3:
## Summary
When RAG retrieval fails, the failure is now surfaced instead of silently degrading: `_query_all_rag_sources` returns a third `failures` list, and all four LLM call paths (plain/tools, sync/streaming) inject a notice instructing the model to tell the user which data sources could not be searched.

## Verdict
Request changes.

## Required changes
1. Stop echoing the raw client-supplied source identifier into the injected `system` message at `atlas/modules/llm/litellm_caller.py:645` — `selected_data_sources` arrives unvalidated from the WebSocket payload and an unknown corpus name reliably takes the error path, so multi-line imperative prose can be smuggled into the highest-trust message role. Resolve the identifier against the user's discovered sources and emit the canonical display name, or strip control characters/newlines, collapse whitespace, and truncate before interpolation.
2. Render the exclusion notice in the no-successful-responses branch at `atlas/modules/llm/litellm_caller.py:1124` (and `:1359`, `litellm_streaming.py:326`, `:429`) — when one source is permission-denied and another errors, `DataSourcePermissionError` is not raised, so `rag_exclusions` is dropped and the model is told "every query failed" and to advise a retry, which is the wrong remedy for an authorization problem.
3. Cap the number and length of identifiers rendered by `_build_rag_failure_notice` (`litellm_caller.py:645`) — an unbounded list of long nonexistent source names is reflected verbatim into a real upstream prompt, turning cheap client input into billed tokens and, at size, a context-window error on an otherwise valid turn.
4. Pluralize and separately quote the names in the failure message at `atlas/modules/llm/litellm_caller.py:640` — sources are batched per server, so the singular template currently produces `The data source 'policies, techdocs, handbook' could not be queried`, which the model paraphrases to the user verbatim.
5. Add a regression test asserting an LLM domain error still propagates through the new injection path (all sources fail, `call_plain` raises `RateLimitError`, expect it to re-raise) to `atlas/tests/test_rag_partial_compliance_exclusion.py` — this stated safety property is currently checked only by `test/pr-validation/test_pr844_rag_failure_notifies_llm.sh`, which CI does not run.
6. Extend the partial-failure notice test to the other three call paths (`call_with_rag_and_tools`, `stream_with_rag`, `stream_with_rag_and_tools`) — only `call_with_rag` is covered, so dropping the one-line insertion on any other path would silently restore the old behavior.

## Evidence
Evidence: 9/10. The change is backend-only and well demonstrated — unit tests cover the new return shape, the notice builder, and the partial and total failure paths, and the replaced test makes the intent change explicit. The one gap worth closing is a rendered example of the injected system message in the PR description, since the exact prompt wording is what the fix actually delivers.

## Notes
- Retitle the CHANGELOG heading to `### PR #845` — `CHANGELOG.md:9` uses the issue number, unlike neighbouring entries.
- The total-failure block is copy-pasted across four call paths (`litellm_caller.py:1119`, `:1354`, `litellm_streaming.py:326`, `:429`); a single helper alongside `_build_rag_failure_notice` would stop the prompt text drifting.
- User-facing copy says "the RAG service returned an error" (`litellm_caller.py:645`); product wording would serve users better and keep the acronym in the log line.

---

garland3:
## Summary
When retrieval fails, the RAG fan-out now returns the per-server failures alongside successes, and all four LLM call paths inject a system message telling the model to disclose the failure instead of silently answering without context.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — proportionate for a backend-only change, with unit tests across the new return shape and all four total-failure paths. The one gap worth closing is that the actual rendered system message is never shown; pasting the fully-rendered text for a partial failure and a two-batch total failure into the description would make the deliverable (the prose itself) reviewable.

## Notes
- Batched failures render a singular template over a comma-joined list — `The data source 'policies, techdocs' could not be queried` names a corpus that does not exist; quote and pluralize each name at `atlas/modules/llm/litellm_caller.py:640`.
- The four total-failure branches (`litellm_caller.py:1124`, `:1359`, `litellm_streaming.py:326`, `:429`) render only `_build_rag_failure_notice` and drop `rag_exclusions`, so a mixed denial + error turn tells the user to retry when the real remedy is a permission request.
- Client-supplied source identifiers reach a `system`-role message with no length, count, or control-character bounds (`litellm_caller.py:643`); the module's existing `_sanitize_label` plus a cap would close it.

---

PR #848, opened 2026-08-26
fix(rag): align mock v2 and client to the v0.8.0 OpenAPI schema

garland3:
## Summary

The v2 RAG wire contract used a bespoke shape that did not match the published v0.8.0 OpenAPI schema. Both the mock and the ATLAS client now speak the schema contract.

## Wire contract (v2)

**Before:** Request `{query, corpora, mode, top_k, filters, synthesis_params}` -> Response `{query, mode, results: {hits|answer+citations}, metadata: {response_time_ms, corpora_searched}}`

**After (v0.8.0):** Request `{query, corpora, search_kwargs}` -> Response `{response, metadata: {response_time, references: [{filename, sections: [{text, relevance}], reference}]}}`

### Mock (mocks/atlas-rag-api-mock/main.py)
- New SearchKwargs, QuerySection, QueryReference, RagQueryRequest, RagQueryResponse models matching the v0.8.0 schema.
- v2 discovery returns a plain DataSource list (no api_version field, matching the schema).
- v2 query returns {response, metadata: {response_time, references}}.
- Version bumped to 0.8.0. v1 endpoints unchanged.

### Client (atlas/modules/rag/atlas_rag_client.py)
- query_v2 now sends {query, corpora, search_kwargs} (no mode/top_k/filters/synthesis_params on the wire).
- mode is a **client-side interpretation knob**: synthesized uses the response verbatim (is_completion=true); raw builds an evidence block from metadata.references (is_completion=false). This preserves the existing product behavior while conforming to the schema.
- top_k maps to search_kwargs.top_k_final.
- Response parsing rewritten for {response, metadata: {response_time, references}} (seconds, not ms).

### Models (atlas/modules/rag/client.py)
- Section.section_ref is now Optional[int] so both the v1 mock shape (which numbers sections) and the v0.8.0 shape (which does not) parse.

### Docs
- Rewrote docs/admin/external-rag-api.md v1 + v2 sections to match the actual contract (the old v1 examples showed an OpenAI ChatCompletion shape the mock never produced; the v2 section showed mode/top_k/results which the schema does not have).
- Updated planning doc, mock README, AGENTS.md RAG discovery line, and planning README index.

## Tests
- atlas/tests/test_atlas_rag_v2.py -- updated fixtures and assertions to the v0.8.0 shape (35 tests).
- test/pr-validation/test_issue791_rag_api_v2.sh -- updated all 10 checks.

## Verification
- ruff check passes on all changed Python files.
- scripts/check-docs.sh passes (no orphaned docs or broken links).
- Full backend unit suite: 2545 passed, 17 skipped.
- PR validation script: 10 passed, 0 failed.
- Mock smoke-tested via curl: v2 discovery and query return the correct v0.8.0 shape; 400/403/401 error paths verified.

## Backward compatibility
- v1 endpoints and client path are unchanged.
- The service layer (UnifiedRAGService) still passes mode/top_k to query_v2; the client handles them (mode = client-side, top_k -> search_kwargs), so no service-layer changes were needed.
- filters/synthesis_params params are still accepted by query_v2 for signature compatibility but are no longer sent (the v0.8.0 schema has no such fields).


garland3:
## Summary
Realigns the v2 RAG wire contract with the v0.8.0 schema across the mock, the HTTP client, its models, tests and docs: requests move to `search_kwargs`, responses to `{response, metadata.references}`, and `Section.section_ref` becomes optional so both shapes parse.

## Verdict
Request changes.

## Required changes
1. Guard the section marker in `_format_rag_references` at `atlas/modules/llm/litellm_caller.py:1626` — `Section.section_ref` is now `None` for every v0.8.0 reference, so each snippet renders `data-section-ref="None"` and a visible `§None (91%): …` prefix; fall back to a positional index from `enumerate(doc.sections, start=1)` or omit the marker.
2. Parse `response_time` as a float and round instead of truncating at `atlas/modules/rag/atlas_rag_client.py:697` — `int(0.4) * 1000` reports `0 ms` for a real sub-second query in the UI footer.
3. Drop the `+ 1` bias in `mocks/atlas-rag-api-mock/main.py:735` — `max(1, int(elapsed + 1))` makes every sub-second mock query advertise 1000 ms.
4. Use one numbering scheme for the whole evidence block at `atlas/modules/rag/atlas_rag_client.py:667` — mixing backend `document_ref` with the `enumerate` fallback lets two documents emit the same `[N]`, which the citation matcher then mis-links. Assigning the ordinal once in `_parse_v2_documents` when the wire supplies none also fixes the parse/format coupling.
5. Add a migration subsection to `docs/admin/external-rag-api.md:186` mapping the old v2 fields to the new ones (`top_k` → `search_kwargs.top_k_final`, `results.hits[]` → `metadata.references[]`, `results.answer` → `response`, `response_time_ms` → `response_time` in seconds) — operators on the previously documented v2 shape get a silent "No response from RAG system." placeholder with no guidance.
6. Raise the dropped-`filters`/`synthesis_params` message at `atlas/modules/rag/atlas_rag_client.py:450` from `debug` to a throttled `warning`, and mark both parameters deprecated in the signature/docstring — an admin who configured filters currently gets unfiltered results with no signal at default log levels.
7. Fix the CHANGELOG heading at `CHANGELOG.md:9` — it reads `### PR #849` but this is PR #848.

## Evidence
Evidence: 8/10. The change is well demonstrated for a wire-contract rewrite — 35 updated unit tests pinning both request body and response parsing, a 10-check end-to-end validation script against the mock, a full backend suite run, and explicit before/after wire shapes in the description. The one gap is the user-visible surface: no rendered citation output against a v2 source in either `raw` or `synthesized` mode, which is also where the `§None` and marker-collision defects above would have shown up.

## Notes
- `_to_query_reference` can produce a null required `reference` and 500 on response validation when both `reference` and `filename` are empty — `mocks/atlas-rag-api-mock/main.py:685`.
- `search_kwargs={}` is treated as an explicit override and silently discards the configured `top_k` — `atlas/modules/rag/atlas_rag_client.py:438`.
- `title` now carries the combined `"PTO Policy, pto-policy.pdf"` label, duplicating the filename in citation chips — `atlas/modules/rag/atlas_rag_client.py:633`.

---

garland3:
## Summary
Realigns the v2 RAG client and its mock to the v0.8.0 wire contract — `search_kwargs.top_k_final` replaces `top_k`, `metadata.references[]` replaces `results.hits[]`, `response` replaces `results.answer`, and `response_time` is now seconds — with matching updates to models, tests, and the admin docs.

## Verdict
Request changes.

## Required changes
1. Parse `response_time` as a float in `atlas/modules/rag/atlas_rag_client.py:697` — `int(raw_seconds) * 1000` truncates any sub-second value to `0 ms`, so every fast query reports zero elapsed time in the UI footer. Use `max(0, int(round(float(raw_seconds) * 1000)))`, keeping the existing `(TypeError, ValueError)` guard. The head commit message states this was fixed; the code still truncates.
2. Add a test pinning the sub-second `response_time` → `query_processing_time_ms` conversion — the existing coverage only uses whole-integer seconds, which is why the truncation survived.

## Evidence
Evidence: 7/10. The wire contract itself is well demonstrated (35 updated unit tests pinning both the outbound body and the inbound parse, a live-mock end-to-end script, and a full suite run), but the user-visible rendering path is not: there is no captured or screenshotted references panel for a v2 source in either `raw` or `synthesized` mode, which is the single most useful missing artifact given this PR changes the `[N]`/`§N` markers and the citation label.

## Notes
- Assign document ordinals once in `_parse_v2_documents` rather than in each formatter — `atlas/modules/rag/atlas_rag_client.py:667` mixes backend-supplied `document_ref` with positional indices, which can emit `[N]` twice if a backend populates the field for only some references.
- `search_kwargs={}` is treated as an explicit override and silently drops the configured `top_k` (`atlas/modules/rag/atlas_rag_client.py:438`); `if search_kwargs:` plus a merge would be safer. No caller does this today.
- `filters` / `synthesis_params` are accepted and discarded at `debug` level (`atlas/modules/rag/atlas_rag_client.py:450`) — worth a throttled warning and a deprecation note in the docstring — and the CHANGELOG entry is filed under `### PR #849` (`CHANGELOG.md:9`).

---

PR #856, opened 2026-08-27
Consolidate canvas, sleep and search into one built-in `atlas` server (#855)

### LINKED ISSUE #855: consolidate atlas tools
Move the 'canvas', 'sleep' and 'search' into a single ATLAS server like tool. 
remove the magnified glass search button. If the 'search' tool is on, then use the selected data sources to do a search. 
search should have one arg -- query . when the AI Agent uses the search tool it

garland3:
Closes #855.

## What

ATLAS grew three built-in, non-MCP "pseudo-servers" one at a time:

| Server | Tool(s) |
| --- | --- |
| `canvas` | `canvas_canvas` |
| `atlas_agent` | `atlas_agent_sleep` |
| `atlas_rag` | `atlas_rag_discover_data_sources`, `atlas_rag_query` |

In the tools panel that read as three servers for four tools, and each carried its own copy of the "special-case this pseudo-server" branch through discovery, schema building, the tool index, execution dispatch, the authorization allow-list and `/api/config`.

They are now **one** built-in server, `atlas`:

| Tool | Arguments |
| --- | --- |
| `atlas_canvas` | `content` |
| `atlas_sleep` | `seconds`, optional `reason` |
| `atlas_search` | `query` |

`atlas/modules/mcp_tools/atlas_server.py` (mirrored by `frontend/src/constants/atlasTools.js`) is the single source of truth for the names, schemas, gating and legacy aliases.

## `atlas_search` takes only a query

`data_sources` and `mode` are gone from the model-facing schema:

- **Sources** come from the user's RAG panel selection, falling back to every source the user is authorized for when nothing is selected.
- **Mode** is always `raw` — a tool call should hand the model evidence to reason over, not a backend-written answer.

Model-supplied `data_sources`/`mode` on an `atlas_search` call are ignored. That is defence in depth (the authorization gate already intersects any requested source with the user's discovered set), but a model should not be able to reach past the user's selection to another source the user happens to be authorized for.

`atlas_rag_discover_data_sources` is no longer advertised — with search reading the UI selection there is nothing left for a discover step to feed — but it still executes so a saved conversation replaying it does not come back broken.

## The magnifying-glass button is gone

It did two unrelated things at once (enable RAG, open the sources panel) and duplicated a decision that belongs in the tools panel. Selecting `atlas_search` is now how a user turns search on; the sources panel is still reachable from the header. A turn counts as RAG-activated whenever `atlas_search` is selected, so "search on, nothing picked" means "everything I can reach" rather than "nothing".

## Backwards compatibility

Tool selections live in `localStorage` and in saved conversations, so the old names are still accepted and normalized at the edges on both sides:

```
canvas_canvas      -> atlas_canvas
atlas_agent_sleep  -> atlas_sleep
atlas_rag_query    -> atlas_search
```

A browser holding `['canvas_canvas', 'math_add']` gets `atlas_canvas` selected on first render — no reset, no lost selection.

## Gating

The built-in server is authorized for every user; its tools are gated as before, except a disabled tool now drops out of the tool list instead of the whole server disappearing (`atlas_canvas` does not depend on either flag):

- `atlas_sleep` — `AGENT_SLEEP_MAX_SECONDS > 0`
- `atlas_search` — `feature_rag_enabled` **and** `feature_atlas_rag_tools_enabled`

Disabled built-ins are omitted from the schema rather than advertised and then refused: agent mode reaches the loop without ACL filtering, so a tool in the schema costs a step before execution can reject the call.

## Ambiguity in the issue

The issue text trails off after "The `search` tool should take a single argument — `query`." Two judgement calls were made in that gap (both written up in the design note):

1. **Empty selection means "all authorized sources"**, matching what the old RAG toggle did when on with nothing picked. The alternative — empty means search nothing — would make a freshly-enabled search tool return nothing until the user opens the sources panel.
2. **`atlas_rag_discover_data_sources` was dropped from the advertised set** rather than kept as a fourth tool, since it has no job left once search reads the UI selection.

Happy to flip either if the intent was different.

## Verification

- `bash test/atlas_tests.sh` — 2617 passed, 17 skipped
- `npm test -- --run` (frontend) — 644 passed, 58 files
- `npm run lint` — 0 errors (4 pre-existing warnings)
- `npm run build` — clean

New tests: `atlas/tests/test_atlas_consolidated_tools.py` (alias resolution, single-argument search schema, disabled-tool omission, canvas executing under both names) and `frontend/src/test/atlas-tools-consolidation.test.js` (name migration, `useSelections` upgrading a pre-consolidation persisted selection), plus `atlas_search` source-selection and ignored-argument cases in `test_atlas_rag_agent_tools.py`.

Docs: `docs/developer/design-notes/atlas-tool-consolidation-2026-08-27.md`, CHANGELOG, AGENTS.md, `docs/admin/configuration.md`.


garland3:
Thanks — all three addressed.

**P1, hook matchers (`mcp_discovery.py:597`)** — correct and the important one: a `matcher` is an operator-written regex over the tool name, and a `PreToolUse`/`PermissionRequest` hook may be a deny or approval policy, so renaming the tool would have retired it silently. `HookConfig.matches` now also tests the pre-#855 spelling of a built-in, so `^canvas_canvas$` keeps firing on `atlas_canvas`. The aliasing runs **one way** — a matcher written against the new name does not start matching the old one — so no hook's reach widens. Tests: `test_a_hook_matcher_written_against_the_old_name_still_fires`, `test_a_hook_matcher_on_the_new_name_does_not_widen`.

**P2, a configured server named `atlas` (`mcp_discovery.py:136`)** — also correct: it would connect, get discovered, and then be unreachable, since the built-in names short-circuit the tool index, `/api/config` and execution dispatch. Took the "reject with a clear error" option: `_drop_reserved_servers` removes `atlas`/`canvas`/`atlas_agent`/`atlas_rag` at config load (initial load *and* hot reload) and logs an error naming the collision, rather than half-working. Tests: `test_reserved_server_names_are_dropped_from_mcp_config`, `test_drop_reserved_servers_leaves_a_clean_config_untouched`.

**P2, `useSelections` batch removals** — real bug, and it was already fixed in b361ac2 (the review ran against the first commit e9b9da7). `addTools`/`removeTools`/`toggleTool` now migrate before writing, so a write also converges storage onto the new names instead of leaving a legacy name beside its replacement. Test: `converges storage onto the new names when a selection is written`.

garland3:
## Summary
Consolidates the `canvas`, `atlas_agent` and `atlas_rag` built-in servers into a single `atlas` pseudo-server with three tools (`atlas_canvas`, `atlas_sleep`, `atlas_search`), adds a legacy-name alias table on both sides, and removes the magnifying-glass RAG toggle from the composer in favour of selecting the search tool.

## Verdict
Request changes.

## Required changes
1. Normalize the tool name before hook matching at `atlas/hooks/models.py:165` (and at the `matcher_value` sites `atlas/application/chat/tool_executor.py:473` and `:511`) — matcher aliasing is one-directional, so an operator deny rule on `^atlas_search$` is evaded by a call that emits the still-executable legacy name `atlas_rag_query`.
2. Stop expanding legacy aliases for approval-shaped events at `atlas/modules/mcp_tools/atlas_server.py:214` — `PermissionRequest` matches auto-approve, so a grant written for the narrow `atlas_rag_query` now silently covers `atlas_search`, which fans out to every authorized source.
3. Remove `ragEnabled` from the activation condition at `frontend/src/contexts/ChatContext.jsx:578` and migrate the persisted `chatui-rag-enabled` key off on read — the button removed at `frontend/src/components/ChatArea.jsx:1095` was its only writer, so returning users with the flag set have RAG force-enabled over all sources with no way to turn it off.
4. Decouple "sources the search tool may read" from `selected_data_sources` at `frontend/src/contexts/ChatContext.jsx:578` — selecting `atlas_search` with no sources picked sends every authorized source, which triggers implicit RAG augmentation on every turn at `atlas/application/chat/modes/tools.py:252`, including turns where the model never calls the tool. Resolving the fallback server-side in `_execute_atlas_rag_tool` is the cleaner fix.
5. Ignore model-supplied `data_sources`/`mode` for the legacy spelling too at `atlas/modules/mcp_tools/mcp_execution.py:787` — `is_consolidated_search` keys off `tool_call.name`, so emitting `atlas_rag_query` reaches past the user's selection; extend `test_atlas_search_ignores_model_supplied_sources` to cover it.
6. Update `frontend/src/test/rag-activation-gating.test.js:18` to mirror the new three-term `ragActivated` expression — the hand-copied helper still uses the old formula, so the primary activation path and the all-sources fallback are green but untested.
7. Reconcile the frontend and backend alias tables for `atlas_rag_discover_data_sources` (`frontend/src/constants/atlasTools.js:24` vs `LEGACY_TOOL_ALIASES`) — the frontend maps it to search while the backend deliberately does not, so the same stored name resolves to two different tools; add a parity assertion or serve the table via `/api/config`.

## Evidence
Evidence: 6/10. The backend half is well demonstrated — named tests for alias resolution, gating and reserved-name handling, plus suite, lint and build results — but nothing shows the reworked UI actually rendering; the most important gap is a before/after screenshot of the consolidated tools panel, the change's stated motivation.

## Notes
- `get_available_tools` advertises gated built-ins unconditionally while every other accessor filters them — `atlas/modules/mcp_tools/mcp_discovery.py:528`.
- Turn-budget messages hand the model the retired `atlas_agent_sleep` spelling — `atlas/modules/mcp_tools/sleep_tool.py:150`.
- `_drop_reserved_servers` is untested on the hot-reload path, the realistic route to a name collision — `atlas/modules/mcp_tools/client.py:291`.

---

garland3:
Second round addressed.

**`atlas_search` had no execution-time gate (`mcp_execution.py:471`)** — correct, and the important one. Schema omission stops the model being *offered* a disabled tool, but a replayed conversation or a non-UI client can name it anyway. `execute_tool` now refuses `atlas_search` when `feature_rag_enabled`/`feature_atlas_rag_tools_enabled` are off, with an error naming the switch — matching how `atlas_sleep` has always been refused at `AGENT_SLEEP_MAX_SECONDS=0`. Test: `test_atlas_search_refuses_to_execute_when_rag_is_disabled`. The RAG tool tests now run with the flags on via an autouse fixture, since every one of them exercises that tool; the two disabled-path tests patch after it and win.

**Misindented `_drop_reserved_servers` call (`client.py:140`)** — fixed; the custom-config-path branch now matches the default branch.

**Two `atlas_server` import blocks (`mcp_execution.py:24`)** — merged into one (the `as` alias sorts fine inside the block).

**Changelog labelled "PR #856" (`CHANGELOG.md:10`)** — the number is right: this PR *is* #856, and it closes issue #855, which is what the entry's first bullet says. I did drop the "review feedback on #856" self-reference, which was noise inside #856's own entry.

Local verification on the head commit: `bash test/atlas_tests.sh` 2622 passed / 17 skipped; frontend `npm test` 645 passed; `npm run lint` 0 errors; `npm run build` clean.

garland3:
## Summary
Consolidates the built-in ATLAS tools from three pseudo-servers into a single `atlas` server, renames `atlas_rag_query` to a single-argument `atlas_search` behind a legacy-alias table, and removes the composer's magnifying-glass RAG toggle in favour of selecting the search tool.

## Verdict
Request changes.

## Required changes
1. Normalize the tool name before hook matching in `atlas/hooks/models.py:165` — a call spelled `atlas_rag_query` yields only itself as a candidate, so a `PreToolUse` deny rule written on `^atlas_search$` never fires even though the call resolves to and executes as `atlas_search`. Build the candidate set from `normalize_tool_name(matcher_value)` so aliasing is symmetric.
2. Restrict legacy-alias expansion to deny-shaped hook events, or gate it behind a flag on `matches` — expanding candidates for approval-shaped events (`PermissionRequest` / auto-approve) means a grant scoped to the narrow `atlas_rag_query` now auto-approves `atlas_search`, which fans out to the whole selected or authorized source set.
3. Derive `is_consolidated_search` from the resolved name at `atlas/modules/mcp_tools/mcp_execution.py:787` (`normalize_tool_name(tool_call.name) == SEARCH_TOOL_NAME`) — keying off the raw name lets a call spelled `atlas_rag_query` keep model-supplied `data_sources`/`mode`, reaching past the user's selection to any other source they are authorized for.
4. Drop `ragEnabled` from the `ragActivated` expression at `frontend/src/contexts/ChatContext.jsx:578` and clear the persisted `chatui-rag-enabled` key on read — the deleted button at `ChatArea.jsx:1095` was its only writer, so any browser holding `true` has retrieval over every authorized source permanently on with no control to clear it.
5. Reconcile `atlas_rag_discover_data_sources` between `frontend/src/constants/atlasTools.js:24` (maps it to `atlas_search`) and the backend `LEGACY_TOOL_ALIASES` (deliberately omits it), then add a parity check so the two alias tables cannot drift again.
6. Add tests covering the fixes above: a deny hook on `^atlas_search$` firing for a call named `atlas_rag_query`, `test_atlas_search_ignores_model_supplied_sources` under the legacy spelling, and the new `searchToolSelected` term in `frontend/src/test/rag-activation-gating.test.js`, whose hand-copied helper still encodes the old two-term formula.

## Evidence
Evidence: 6/10. Backend coverage is proportionate and specific for a 32-file change, and full-suite, lint and build results are reported at the head commit — but the UI change that motivated the work is undemonstrated; the single most useful missing artifact is a before/after screenshot of the tools panel collapsing from three built-in servers to one.

## Notes
- `get_available_tools` extends with all `ATLAS_TOOL_NAMES` unconditionally while every sibling accessor honours the feature gates — `mcp_discovery.py:528`.
- Discoverability: nothing signposts that search is now enabled via Tools → `atlas` → `search`, and the header pill still reads "Sources" when the implicit-all fallback is active — `ChatArea.jsx:1095`, `Header.jsx:198`.
- The disabled-search error does not name the settings to change, unlike the sibling sleep message — `mcp_execution.py:485`.

---

garland3:
## Summary
Consolidates the built-in ATLAS tools (`canvas_canvas`, `atlas_agent_sleep`, `atlas_rag_query`) into a single `atlas` server exposing `canvas`, `sleep` and `search`, with a legacy-alias table for backward compatibility, and removes the composer's magnifying-glass RAG toggle in favour of selecting the search tool.

## Verdict
Request changes.

## Required changes
1. Normalize before matching in `HookConfig.matches` at `atlas/hooks/models.py:165` — build candidates from `normalize_tool_name(matcher_value)` so a `PreToolUse` deny rule spelled `^atlas_search$` also fires on a call named `atlas_rag_query`, which currently executes as search while evading the policy. Add a test asserting that direction.
2. Do not expand legacy aliases for approval-shaped events (`PERMISSION_REQUEST`) in `atlas/hooks/models.py:165` — an operator grant scoped to the narrow `atlas_rag_query` currently auto-approves the consolidated `atlas_search`, which fans out to the whole authorized source set.
3. Derive the flag from the resolved name at `atlas/modules/mcp_tools/mcp_execution.py:794` — `is_consolidated_search = normalize_tool_name(tool_call.name) == SEARCH_TOOL_NAME`; keying off the raw name lets a call spelled `atlas_rag_query` keep model-supplied `data_sources`/`mode` and reach past the user's selection. Extend `test_atlas_search_ignores_model_supplied_sources` with the legacy spelling.
4. Drop `ragEnabled` from the `ragActivated` expression at `frontend/src/contexts/ChatContext.jsx:578` and clear/ignore the persisted `chatui-rag-enabled` key on read in `useSelections.js:27` — its only writer was the deleted button, so a browser holding `true` has retrieval forced on every turn over all authorized sources with no way to clear it. Note `useSelections.js:159` also re-sets it from a restored workspace.
5. Keep `searchToolSelected` out of `dataSourcesToSend` at `ChatContext.jsx:580` and resolve the empty-selection fallback server-side in `_execute_atlas_rag_tool` — otherwise selecting the search tool with no sources attaches every authorized source to every turn, including turns the tool is never called on.
6. Reconcile the two alias tables for `atlas_rag_discover_data_sources` — `frontend/src/constants/atlasTools.js:24` maps it to search while the backend `LEGACY_TOOL_ALIASES` keeps it as a separate executable tool — and add a parity assertion (or serve the table from `/api/config`) so they cannot drift.

## Evidence
Evidence: 6/10. The backend half is well demonstrated — new alias/gating/hook suites, extended RAG execution tests, and full-suite, lint and build numbers at the head commit — but the user-facing half of a large change is undemonstrated. The single most important missing artifact is a before/after screenshot of the tools panel and the composer with the search button removed, since UI simplification is the stated goal.

## Notes
- `get_available_tools` at `atlas/modules/mcp_tools/mcp_discovery.py:528` ignores the feature gates every sibling accessor honours.
- Turn-budget messages at `atlas/modules/mcp_tools/sleep_tool.py:150,187` still name the retired `atlas_agent_sleep`.
- The disabled-search refusal at `mcp_execution.py:485` names no setting, unlike the sleep refusal that cites `AGENT_SLEEP_MAX_SECONDS=0`.

---

garland3:
## Follow-up: discovery back, two search knobs, ATLAS pinned first (`83dd6e8`)

### 1. `atlas_discover_sources` is a tool again

Dropping it assumed its only job was feeding source names into a search that now reads the UI selection. It has another job: saying **which** corpus an answer came from, and whether the thing being asked about is indexed anywhere the user can reach — the difference between "I found nothing" and "you have no source that would contain this". It takes no arguments (the authenticated user decides the result, and the user is never a model input) and is gated by the same RAG flags as search. `atlas_rag_discover_data_sources` now normalizes onto it rather than onto search.

### 2. `atlas_search` gains `max_results` and `depth`

`query` is still the only **required** argument.

| Argument | Type | Effect |
| --- | --- | --- |
| `max_results` | integer | Passages per source, clamped to 50 → `search_kwargs.top_k_final` |
| `depth` | `quick` \| `standard` \| `deep` | Retrieval effort → `rerank`, `top_k_vector`, `top_k_full_text`, `expanded_window` |

The line worth stating explicitly: **these change how much comes back, never which sources are reachable.** `data_sources` was removed because it was a *reach* control; these two are not, which is why they are safe to take from the model. Both are coerced and clamped in `search_kwargs_for` before they leave the process, so a nonsense value degrades to the source default instead of reaching a backend.

### 3. ATLAS is pinned to the top of the tools panel and the marketplace

`sortAtlasFirst` lifts it out of the list and leaves everything else in the order the backend sent.

---

## Does this use the ATLAS RAG v1 and v2 specs?

Both, and the difference matters for the new knobs. `atlas_search` goes through `unified_rag.query_rag` → `_query_http_client`, which speaks whichever contract the source is configured for:

- **v1** (`POST /api/v1/rag/completions`) — request body is the conversation plus a model name. **No retrieval knobs exist on the wire**, so `max_results` and `depth` are silently ignored and the source's configured behaviour applies. `mode` is equally meaningless here: v1 always synthesizes.
- **v2** (`POST /api/v2/rag/query`) — request body is `{query, corpora, search_kwargs}`. This is where both knobs land.

That silent degradation is deliberate: asking a v1 source for 20 deep results should return results, not an error. One subtlety worth flagging — `query_v2` treats an explicit `search_kwargs` as the *whole* block, so `_query_http_client` folds the source's configured `top_k` in as the `top_k_final` default first. Without that, a `depth`-only call would quietly drop the configured `top_k`. Covered by `test_v2_search_kwargs_keep_the_configured_top_k_as_a_floor` and `test_v1_ignores_search_kwargs`.

---

## Verified in a browser against a live RAG backend

Not just unit tests — a real turn against `corporate_cars` with five sources.

**The model calls both tools, and both are labelled `Server: atlas`:**

atlas_search call

atlas_discover_sources call

Note the arguments the model actually chose: `{"query": "executive fleet", "depth": "deep", "max_results": 3}`, and `{}` for discovery.

**This turned up a real bug, now fixed in the same commit.** The "Server:" label was derived by splitting the fully-qualified name on its last underscore, so `atlas_discover_sources` displayed as the server `atlas_discover`. The bug predates this PR — `pptx_generator_markdown_to_pptx` has always displayed as `pptx_generator_markdown_to` — because *both* halves of a `server_toolName` can contain underscores, so the split can never be right in general. `notify_tool_start` now asks the tool manager which server owns the tool, keeping the split as a fallback only (it is a display label, so a lookup failure must not fail the call).

**ATLAS first in the tools panel, four tools:**

tools panel

**`search` shows `query` as the only required argument:**

search schema

**`discover_sources` takes nothing, and both select cleanly:**

discover selected

**ATLAS first in the marketplace too:**

marketplace

**The magnifying-glass button is still gone from the composer:**

composer

---

### Tests

`2641 passed, 17 skipped` backend (`bash test/atlas_tests.sh`), `648 passed` frontend (`vitest`), `check-docs: OK`. New coverage: discovery under both names, refusal when RAG is off, the knobs reaching `search_kwargs`, clamping/coercion of model-supplied values, v1-ignores/v2-honours routing, `sortAtlasFirst`, and the server-label lookup.

The screenshots live on the `assets/pr-856-screenshots` branch so they stay out of this PR's diff.

garland3:
## Summary
Consolidates the built-in ATLAS tools behind a single `atlas` server — `atlas_search`, `atlas_discover_sources`, `atlas_sleep`, `atlas_canvas` — with `atlas_server.py` owning names, schemas and legacy aliases, adds `depth`/`max_results` knobs on search, and replaces the composer's magnifying-glass RAG toggle with tool selection in the Tools panel.

## Verdict
Request changes.

## Required changes
1. Derive `is_consolidated_search` from the resolved name at `atlas/modules/mcp_tools/mcp_execution.py:806` — `tool_call.name != _ATLAS_RAG_QUERY_TOOL` keys off the raw spelling, so a call named `atlas_rag_query` (which normalizes to and executes as `atlas_search`) still honours model-supplied `data_sources` and `mode`, defeating the stated "sources are server-side" contract. Parameterize `test_atlas_search_ignores_model_supplied_sources` over both spellings.
2. Make hook alias matching symmetric at `atlas/hooks/models.py:165` — `matcher_candidates` only expands current names to legacy ones, so a `PreToolUse` deny matcher `^atlas_search$` never fires on an executable call named `atlas_rag_query`. Build the candidate set from `normalize_tool_name(matcher_value)` plus its legacy spellings, and restrict the widening half to deny-shaped events so an approval scoped to a narrow legacy name is not broadened. Add a test asserting the deny case fires.
3. Remove `ragEnabled` from the activation expression at `frontend/src/contexts/ChatContext.jsx:578` and ignore/clear the persisted `chatui-rag-enabled` key on read (`useSelections.js:27`, and the workspace-restore path at `useSelections.js:148`) — the only control that ever wrote it was deleted at `ChatArea.jsx:1095`, so any browser holding `true` has retrieval forced on for every turn with no way to clear it.
4. Keep `searchToolSelected` out of `dataSourcesToSend` at `frontend/src/contexts/ChatContext.jsx:580` and resolve the empty-selection fallback inside `_execute_atlas_rag_tool` — as written, merely selecting the tool attaches every authorized source to every turn, triggering implicit retrieval (`modes/tools.py:251`) on turns where the model never calls search.
5. Fix the `max_results` description at `atlas/modules/mcp_tools/atlas_server.py:181` (and the matching docs/CHANGELOG lines) — it says "per source" but maps to `top_k_final`, which bounds the combined result set across corpora.
6. Filter `get_available_tools` at `atlas/modules/mcp_tools/mcp_discovery.py:528` through `_atlas_tool_flags()` — it advertises `atlas_search`, `atlas_discover_sources` and `atlas_sleep` on deployments where those flags are off, disagreeing with every sibling accessor.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size, with an end-to-end run against a live RAG backend, screenshots of each distinct UI state, and named automated coverage at the head commit. The one gap worth closing is a demonstration of a pre-upgrade browser profile (legacy tool names plus `chatui-rag-enabled: true`) loading into the new UI, which is the same path items 3 and 4 above concern.

## Notes
- Feature gating for the built-ins is re-implemented in three places (`atlas_server.py`, `config_routes.py:52`, `mcp_execution.py:478`); a single predicate table would keep the flags in one spot.
- The disabled-RAG refusal at `mcp_execution.py:485` hardcodes `atlas_search` and names no setting, so a refused `atlas_discover_sources` call misreports the tool.
- Sleep turn-budget messages at `sleep_tool.py:149` still instruct the model to use the retired `atlas_agent_sleep` name.

---

PR #861, opened 2026-08-28
fix(sleep): emit heartbeat progress during long sleeps to prevent apparent hang

garland3:
## Problem

The `atlas_sleep` tool did a single `asyncio.sleep(seconds)` for up to 7200 seconds (2 hours by default) with **zero `tool_progress` events**. If the `tool_complete` frame was lost (WebSocket disconnect, backend error, task cancellation without `tool_interrupted`), the frontend row stayed in `calling` status forever with the clock ticking.

**Evidence from the screenshot:** a 900-second sleep showed 79 minutes elapsed with the generic "taking longer than expected" warning instead of the sleep progress clock -- the `tool_complete` never arrived, so the row spun indefinitely.

## Root Cause

`execute_sleep_tool` in `atlas/modules/mcp_tools/sleep_tool.py` called `await asyncio.sleep(seconds)` as a single blocking call. Between `tool_start` and `tool_complete`, nothing was emitted on the WebSocket. The frontend has no staleness watchdog (documented gap in `websocketHandlers.test.js:353-355`), so a row that never receives `tool_complete` stays in `calling` with the clock counting up forever.

The sleep tool also bypassed the progress-handler wiring in `mcp_execution.py` (it returns before the generic MCP path that constructs `_progress_handler`), so no heartbeat was emitted.

### Backend (`sleep_tool.py`)

- Break the wait into **heartbeat intervals** (capped at 30s, proportional for shorter waits) that emit `tool_progress` frames via the `update_callback` already present in the tool execution context dict.
- The heartbeat `total` field reflects the **clamped** duration (per-call cap and turn budget), not the raw request -- so the frontend clock is accurate even when the model asks for more than the cap allows.
- Short sleeps (<=10s) and calls without an `update_callback` still sleep in a single shot (no overhead).
- Cancellation is preserved: `asyncio.sleep(chunk)` raises `CancelledError` immediately on cancel, same as the single long sleep.

### Frontend (`ToolElapsedTime.jsx`, `Message.jsx`)

- `ToolElapsedTime` now accepts `progressRaw` (set on the message by the `tool_progress` handler) and uses `progressRaw.total` as the clock denominator when available, falling back to `args.seconds` before the first heartbeat arrives.
- Added a **"connection may be lost" (red)** hint when elapsed exceeds the wait by more than 60 seconds, distinguishing a finishing sleep (yellow "completing...") from a genuinely stuck one (red "connection may be lost").

## Tests

**Backend (8 new tests in `test_agent_sleep_tool.py`):**
- `_heartbeat_interval` returns correct values for short/medium/long waits
- Long sleep with callback emits multiple `tool_progress` frames with correct elapsed/total
- Short sleep (<=10s) does not emit heartbeats
- Sleep without callback does not emit heartbeats
- Heartbeat loop is cancellable (raises `CancelledError`)
- Heartbeat total reflects clamped duration (10000s request clamped to 300s -> total=300)

**Frontend (6 new tests in `ToolElapsedTime.test.jsx`):**
- Uses `progressRaw.total` as the clock denominator when available
- Falls back to `args.seconds` before the first heartbeat
- Works with the consolidated tool name (`atlas_sleep`)
- Shows "connection may be lost" past the overdue threshold (sleepSeconds + 60)
- Shows "completing..." before the stale threshold
- Does not show stale hint for a sleep still in progress

All 67 backend and 35 frontend related tests pass. Lint passes with 0 errors.

## Files Changed

| File | Change |
|------|--------|
| `atlas/modules/mcp_tools/sleep_tool.py` | Heartbeat loop, `_sleep_with_heartbeats`, `_heartbeat_interval` |
| `atlas/tests/test_agent_sleep_tool.py` | 8 new heartbeat tests |
| `frontend/src/components/ToolElapsedTime.jsx` | `progressRaw` support, stale hint |
| `frontend/src/components/Message.jsx` | Pass `progressRaw` to `ToolElapsedTime` |
| `frontend/src/test/ToolElapsedTime.test.jsx` | 6 new frontend tests |
| `CHANGELOG.md` | Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fix - atlas_sleep heartbeat - 2026-08-27
+- **The `atlas_sleep` tool no longer appears to hang on long waits**: a 900-second sleep could show 79 minutes of elapsed time because the tool did a single `asyncio.sleep(seconds)` with zero `tool_progress` events — if the `tool_complete` frame was lost (WebSocket drop, backend error), the frontend row stayed in `calling` status forever with the clock ticking and the generic "taking longer than expected" warning. The wait is now broken into heartbeat intervals (capped at 30s) that emit `tool_progress` frames, so the frontend knows the sleep is alive, transitions to `in_progress`, and shows the actual (clamped) duration. The frontend `ToolElapsedTime` clock uses the heartbeat's `total` field (which reflects per-call/turn clamping) instead of the raw requested `seconds`, and escalates from "completing..." to "connection may be lost" (red) when the clock runs more than 60s past the wait without a `tool_complete` — so a genuinely stuck sleep is visually distinct from one that is wrapping up.
+
 ### Repo cleanup - 2026-08-27
 - **Removed 11 PR-evidence screenshots that no doc or code references** (~1.1 MB): `docs/developer/images/pr500-citations-screenshot.png`, `pr637-tiff-vision-proof.png`, `pr650-chat-export-active-prompt-ui.png`, `pr650-chat-export-prompt-e2e.png`, `pr664-agent-multi-tool-chain-2.png`, `rag-snippets-collapsed.png`, `rag-snippets-expanded.png`, `rag-snippets-live-pipeline.png`, `wormhole-e2e-console-2026-06-09.png`, `wormhole-e2e-dashboard-2026-06-09.png`, and `docs/readme_img/screenshot-10-24-2025.png`. These were review evidence attached to past PRs that landed in the tree rather than staying on the PR thread; every image still referenced by a doc or test is untouched. Same cleanup as #853.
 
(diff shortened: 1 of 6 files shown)


garland3:
## Summary
Adds periodic `tool_progress` heartbeats to the sleep tool so a long wait reports liveness, and updates the frontend elapsed-time row to use the heartbeat total as its denominator with "completing..." and "connection may be lost" hints.

## Verdict
Approve with nits.

## Evidence
Evidence: 6/10. The fourteen enumerated tests cover interval selection, emission, cancellation, and the clamped-total path, which is proportionate for a change this size; the main gap is that nothing demonstrates the change in a running system — no screenshot or captured frames of a sleep row actually advancing through its three states.

## Notes
- Bound the heartbeat emit — an unresolved `await notify_tool_progress(...)` on a half-open socket blocks the turn indefinitely (`atlas/modules/mcp_tools/sleep_tool.py:180`); `asyncio.wait_for` with a small timeout keeps a dropped beat harmless.
- `_sleep_with_heartbeats` sums nominal chunks rather than tracking a monotonic deadline, so each beat's latency is added to the real wait and the reported `elapsed` drifts behind the frontend clock (`atlas/modules/mcp_tools/sleep_tool.py:170`).
- A sleeping row now renders both the purpose-built clock and the generic progress bar with a unitless `30/7200` label (`frontend/src/components/Message.jsx:338`); the red "connection may be lost" hint is also driven by overshoot alone, ignoring heartbeat recency (`frontend/src/components/ToolElapsedTime.jsx:74`).

---

PR #862, opened 2026-08-28
feat(rag): make atlas_search an explicit tool call instead of silent RAG injection

garland3:
## The problem

`atlas_search` was defined as a tool, but the model almost never had to call it: retrieval ran on its own, before the model was asked anything.

Any turn carrying `data_sources` was routed through `LiteLLMCaller.call_with_rag_and_tools` (or `stream_with_rag_and_tools`). Those methods queried every selected source in parallel, built a `Retrieved context from ...` system message, inserted it ahead of the user turn, and only then called the LLM with the tools schema.

The consequences were all the same shape -- the search was invisible and nobody chose it:

- **The user could not see it happen.** No tool call was published, so the chat showed an answer with citations and no account of where they came from.
- **The model could not choose the query.** It was derived from the conversation by `_query_all_rag_sources`, not decided.
- **The model could not choose *whether* to search.** Every turn paid for a retrieval round trip.
- **One search per turn, always first.** The model could never read a result and search again with a better query -- the thing that makes a search tool useful.

## The change

Retrieval is now an ordinary tool call, the way a Tavily or Perplexity MCP server works.

- `agentic_loop._call_llm` / `_call_llm_streaming` and both tools-mode call sites always take the plain `call_with_tools` / `stream_with_tools` path. The `data_sources` branch is gone.
- `call_with_rag_and_tools` and `stream_with_rag_and_tools` are **deleted**, along with their `LLMProtocol` declarations and the RAG branch in `error_handler.safe_call_llm_with_tools`. Leaving them in place would have left a second, silent retrieval path one call site away from returning.
- The execution path is unchanged and already worked: `tool_executor` -> `ToolManager.execute_tool` -> `_execute_atlas_rag_tool`, with its own feature gate and its own authorization intersection against the user's discovered sources.
- RAG mode (`call_with_rag` / `stream_with_rag`, no tools) is untouched. With no tools schema and no model turn to spend, pre-injection is the design there rather than a hidden step.

### The data-source selection still means two things

1. **A ceiling** -- `mcp_execution` intersects requested sources with the selection and then with the user's authorized set. Unchanged.
2. **Availability** -- a selection now also *offers* `atlas_search` (`application/chat/utilities/search_tool_selection.py`), exactly as if the user had ticked it.

(2) is a deliberate compromise, called out in the design note. A literal reading of "search only happens when the model calls the tool" would leave a user who picks a source but never ticks the tool with a selection that does nothing and says nothing about it -- trading one silent behaviour for another. It is gated on `FEATURE_RAG_ENABLED` and `FEATURE_ATLAS_RAG_TOOLS_ENABLED` and never duplicates a tool already selected under its current or pre-#855 name.

## Compatibility

Legacy names still normalize (`atlas_rag_query` -> `atlas_search`), saved conversations replay unchanged (they hold rendered tool-call rows, not re-executed calls), and `atlas_discover_sources` was always explicit and is unaffected.

## Tests

`2657 passed, 17 skipped` (`bash test/atlas_tests.sh`); `ruff check` and `npm run lint` clean against the pre-existing baseline; `scripts/check-docs.sh` OK.

Tests that pinned the old injection behaviour were removed with the methods they covered. What replaces them asserts the *absence* of injection -- `TestAgenticLoopSearchIsATool` and `test_data_sources_do_not_inject_context_and_offer_the_search_tool` both assert the model saw exactly the messages it was given. New unit tests cover `with_search_tool` in `test_search_tool_selection.py`.

## Docs

New design note `docs/developer/design-notes/search-as-an-explicit-tool-2026-08-28.md`; `external-rag-api.md`, `error-flow-diagram.md` and `error-handling-improvements.md` updated where they described the deleted path. CHANGELOG entry added.

Screenshots of the working flow are in a follow-up comment.


garland3:
## Verified in the app

Backend on `:8020` (`atlas/main.py`, `FEATURE_RAG_ENABLED=true`, `FEATURE_ATLAS_RAG_TOOLS_ENABLED=true`), frontend built by `agent_start.sh`, driven with Playwright. Model `openrouter-gpt-oss-120b`, data source `corporate_cars:west_region` (the bundled MCP RAG demo server -- the HTTP mock's port 8002 was occupied by an unrelated local service on this box).

**1. `atlas_search` selected as a tool, one data source picked**

tool selected

**2. The model decides to search: `atlas_search` appears as a visible tool call in the chat**

Prompt: *"Which corporate cars are in the West Region fleet, and who are they assigned to? Search the data sources."*

visible tool call

**3. Expanded -- the model's own query, and the retrieved passages as the tool result**

```json
// Input Arguments
{ "query": "West Region corporate cars", "max_results": 10, "depth": "standard" }
```
```json
// Output Result
{ "results": { "query": "West Region corporate cars",
  "answers": [ { "data_sources": ["corporate_cars:west_region"],
    "content": "Alice Johnson's 2022 Toyota Camry is in San Francisco, CA.\nCEO Vehicle's 2023 Mercedes-Benz E350 is in San Jose, CA.",
    "is_completion": false } ] } }
```

tool result

The model then answers *from* that tool result and cites it. Note the query is the model's -- `"West Region corporate cars"`, not the user's sentence -- and `max_results`/`depth` were chosen by the model, which is exactly what the old pre-injection path made impossible.

The screenshots live on the throwaway branch `pr-evidence/862-atlas-search` (not merged into `main`, delete after this lands) rather than in the tree -- #860 cleaned up exactly this kind of committed PR evidence.

garland3:
## Summary
Replaces the implicit RAG-context injection path with an explicit `atlas_search` tool: `call_with_rag_and_tools` / `stream_with_rag_and_tools` are deleted from the protocol and both callers, and a new `with_search_tool` helper offers the search tool in tools and agent mode whenever data sources are selected and the RAG feature flags are on.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — the change is large and user-visible, and the PR carries screenshots, a named test environment, full suite results and a design note, which is proportionate. The main gap is that the screenshots are hosted on a branch the author plans to delete, so re-upload them as comment attachments before deleting it.

## Notes
- `with_search_tool` runs after `ToolAuthorizationService.filter_authorized_tools`, so an ACL failure that deliberately drops all tools still yields a turn offering `atlas_search` (`atlas/application/chat/modes/tools.py:114`); applying the helper before the filter would close this.
- The GH #844 guarantee that a total retrieval failure is named in the answer no longer holds in tools/agent mode, and inline citations plus the references block are gone there — worth stating in the design note and CHANGELOG if intended (`atlas/modules/llm/litellm_caller.py:1327`).
- `test_all_sources_fail_streaming_informs_llm` was removed although it covered `stream_with_rag`, a path this PR does not touch; restore it verbatim (`atlas/tests/test_rag_partial_compliance_exclusion.py`). Also fix the placeholder header `### PR #TBD` at `CHANGELOG.md:9`.

---

garland3:
## Conflict resolved + review feedback addressed

Merged `origin/main` (conflict was the CHANGELOG `[Unreleased]` section only — #861's `atlas_sleep` heartbeat entry landed alongside this one; kept both). PR is `MERGEABLE` again.

### 1. `FEATURE_RAG_ENABLED=true` + `FEATURE_ATLAS_RAG_TOOLS_ENABLED=false` (P1) — fixed

Real, and narrower than stated: with sources and **no** tools the turn routes to RAG mode, which still reads them. It only strands the sources when tools are *also* selected, so the turn runs in tools mode with nothing that can read them.

Rather than keep a silent injection path alive for one flag combination, that turn now warns:

> **Your data sources were not searched.** The built-in `atlas_search` tool is turned off (`FEATURE_ATLAS_RAG_TOOLS_ENABLED`) …

The warning is about a *missing tool*, not a disabled flag — a turn that already names search, including as pre-#855 `atlas_rag_query`, is never warned about. That distinction matters: without it the warning fired on every replayed RAG conversation, which `test_orchestrator_mode_routing.py::test_agent_mode_with_selected_atlas_rag_tool_routes_to_agent` caught.

### 2. Agent-mode guards run before the implied tool existed (P2) — fixed

Correct and well spotted. Resolution moved into `ChatOrchestrator._resolve_search_tool`, which runs *ahead* of the `agent_mode and not selected_tools` guard, and the frontend guard now counts selected sources. The source-only agent workflow is actually reachable now. The loops still call the helper (idempotent) so direct callers and tests stay honest.

### 3. PR-validation scripts (P2) — fixed

`test_pr389`, `test_pr276`, `test_pr844`, `test_pr850` each keep every check that still describes live behaviour; the RAG+tools ones are replaced with a note naming what superseded them. `test_pr389` now guards that the pre-injection path *stays* deleted. The dead `stream_with_rag_and_tools` stubs in `test_issue824` / `test_pr798` are gone, and `test_pr850`'s `_call_llm_streaming` call was updated for the dropped `data_sources` parameter.

Results on this branch: **`test_pr389` PASSED, `test_pr844` PASSED, `test_pr850` PASSED, new `test_pr862` PASSED (6/6)**.

`test_pr276` still fails 3 checks — but it **already fails 5 on unmodified `main`** (verified in a detached worktree at `origin/main`). The two it no longer fails are exactly the `call_with_rag_and_tools` ones this PR retired. The remaining 3 (`AtlasRAGClient` completion detection, `call_with_rag` multi-source, and the mock-API E2E, which needs port 8002) are pre-existing and untouched by this change.

### 4. `PR #TBD` in the CHANGELOG — fixed

Now `### PR #862 - 2026-08-28`.

---

`2669 passed, 17 skipped`; `ruff check atlas/` back at the exact pre-change baseline; `scripts/check-docs.sh` OK. The design note gained a section on where the selection resolves and on the one configuration that loses retrieval.

garland3:
## Summary
Replaces pre-injected RAG context in tools and agent mode with an explicit `atlas_search` tool call, deleting the `*_rag_and_tools` LLM caller methods and adding a shared `with_search_tool` helper plus a user-facing warning when the search feature flag is off.

## Verdict
Request changes.

## Required changes
1. Resolve the implied search tool before the tool-capability strip at `atlas/application/chat/orchestrator.py:479` — a model without `supports_tools` has `selected_tools` nulled and the user warned, then `_resolve_search_tool` re-adds `atlas_search`, routing the turn to tools mode with a schema the model cannot use, where it previously ran RAG mode and answered.
2. Stop re-deriving the tool list after authorization: drop the `with_search_tool` calls in `atlas/application/chat/modes/tools.py:114` and `:248` (and the equivalent in `agentic_loop.py`), or pass an explicit "already resolved" flag — `filter_authorized_tools` fails closed by returning `[]` on a missing tool manager or an ACL exception, and this re-application puts `atlas_search` back into the schema for exactly those turns.
3. Emit the "your data sources were not searched" warning based on the routing the turn will actually take, at `atlas/application/chat/orchestrator.py:206` — with `FEATURE_ATLAS_RAG_TOOLS_ENABLED=false` an agent-mode, sources-only turn resolves to an empty list, publishes nothing, and silently answers without reading the sources; passing the agent-mode flag in and warning on `resolved or agent_mode` covers it.
4. Skip the warning when `only_rag` is set (`atlas/application/chat/orchestrator.py:206`) — that turn routes to RAG mode and does read the sources, so the current message is false.
5. Restore `test_all_sources_fail_streaming_informs_llm` in `atlas/tests/test_rag_partial_compliance_exclusion.py` — it covered `stream_with_rag`, an untouched path, and was removed alongside tests for deleted methods, leaving the "total retrieval failure is named to the user" guarantee unverified on the streaming RAG path.
6. Rephrase the published warning at `atlas/application/chat/orchestrator.py:245` in user terms and keep `FEATURE_ATLAS_RAG_TOOLS_ENABLED` in the log line only — an env-var name and "plain RAG" are not actionable for a chat user.

## Evidence
Evidence: 8/10 — proportionate for a change of this size, with screenshots of the end-to-end tool call, a named test environment, and full-suite results. The main gap is that the second-round behaviours (the new warning as rendered, and a source-only agent-mode turn) are described but not shown, and the screenshots are hosted on the `pr-evidence/862-atlas-search` branch the author plans to delete — re-upload them as comment attachments so they survive.

## Notes
- Frontend guard relaxation is untested in the file that exists to test it — `frontend/src/test/agent-mode-requires-tools.test.jsx:117` hardcodes an empty `selectedDataSources`, so the new clause is unreachable.
- Non-streaming `ToolsModeRunner.run` gained the same `with_search_tool` change as the streaming path with no coverage (`atlas/application/chat/modes/tools.py:110`).
- `safe_call_llm_with_tools` drops a middle positional parameter (`atlas/application/chat/utilities/error_handler.py:157`); making the trailing arguments keyword-only would turn a silent mis-binding into an error for out-of-tree callers.

---

PR #865, opened 2026-08-28
fix: remove USE_NEW_FRONTEND flag and old-frontend code path (closes #864)

### LINKED ISSUE #864: remove flag USE_NEW_FRONTEND
USE_NEW_FRONTEND is always used, 
no need for backward compatibility, this is the only use case.

garland3:
## Summary

Closes #864.

Per maintainer guidance in the issue ("Delete everything related to the old frontend"), this removes the `USE_NEW_FRONTEND` flag entirely and deletes the old-frontend code path it was protecting — not just the flag checks.

The flag was always `true` and the skip-build branch it guarded (skip the frontend build in `agent_start.sh` / `ps_agent_start.ps1`) referenced a frontend that no longer exists in the repo. The new React/Vite frontend in `frontend/` is the only frontend, so the build should always happen.

## Changes

- **`agent_start.sh`**: Removed the `use_new_frontend` variable and the `if [ ... != "true" ]` skip-build block from `build_frontend()`. The function now always builds the frontend.
- **`ps_agent_start.ps1`**: Same removal in `Build-Frontend` (PowerShell parity) — the `` variable and the `if ( -ne "true")` skip-build block are gone.
- **`.env.example`**: Dropped the `USE_NEW_FRONTEND=true` declaration.
- **`docker-compose.yml`**: Dropped the `USE_NEW_FRONTEND=true` environment entry.
- **`Dockerfile-test`**: Dropped the `echo "USE_NEW_FRONTEND=true" >> /app/.env` line (and its comment).
- **`CHANGELOG.md`**: Added an entry under `## [Unreleased]`. The historical entry for PR #863 (which documented adding the flag to the PowerShell script) is left intact as a record.

## What was NOT changed

- No Python code changes were needed — `atlas/main.py` already serves only from `atlas/static/` (package) or `frontend/dist/` (dev), with no old-frontend serving logic.
- The `CHANGELOG.md` historical entry for PR #863 is preserved (it is a record of a shipped PR, not live code).

## Verification

- `ruff check atlas/` — no new errors (pre-existing import-organization warnings in unrelated test files only)
- `python -m pytest atlas/tests/test_docker_env_sync.py -v` — 4/4 passed (confirms `.env.example` and `docker-compose.yml` stay in sync after removing the var from both)
- `bash -n agent_start.sh` — syntax OK
- `./test/run_tests.sh backend` — 5/5 passed
- `cd frontend && npm run lint` — 0 errors (4 pre-existing warnings, unrelated)


garland3:

## Summary
Removes the dead `USE_NEW_FRONTEND` feature flag and the branches it guarded across `.env.example`, `docker-compose.yml`, `Dockerfile-test`, `agent_start.sh`, and `ps_agent_start.ps1`. The frontend is now always built at startup; no user-visible behavior changes.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. The change is well-demonstrated for its size — the description names each removed line, and verification includes `bash -n agent_start.sh` plus the env-sync test. The one gap is the PowerShell edit, which has no parse check; recording a `Parser::ParseFile` result would close it.

## Notes
- Operators with `USE_NEW_FRONTEND=false` in an existing `.env` will now get an unexpected frontend build — one CHANGELOG sentence telling them to delete the var would cover it (`CHANGELOG.md:9`).
- No regression guard that the flag stays gone; a test asserting the string is absent from the config/script files would prevent quiet reintroduction (`atlas/tests/test_docker_env_sync.py`).
- Pre-existing, now always reached: `build_frontend` runs `rm -rf atlas/static` without checking that `npm run build` succeeded (`agent_start.sh:241`).

---

garland3:

## Summary
Removes the always-true `USE_NEW_FRONTEND` flag and the skip-build branch it guarded across the env files, `Dockerfile-test`, and both startup scripts, adding a CHANGELOG note and a regression test asserting the flag string is gone.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — per-file accounting of every removed line, a machine-checked regression test, and named verification commands. The one missing artifact is a parse check on `ps_agent_start.ps1`, which was edited symmetrically but never syntax-checked the way `agent_start.sh` was via `bash -n`.

## Notes
- The `best_effort` branch in `test_docker_env_sync.py:262` silently skips `agent_start.sh` and `ps_agent_start.ps1` — `Dockerfile-test` never copies them, so in CI the guard checks only the two env files and passes even if the flag returns to a startup script. Either `pytest.skip` when none are found, or `COPY` the scripts into the test image.
- `test_docker_env_sync.py:245` asserts only on the absence of a literal string; a skip-build branch under a different name would restore the removed behavior and still pass.
- `build_frontend` (`agent_start.sh:241`, and `ps_agent_start.ps1:517`) does not check the build exit status before `rm -rf atlas/static`, so a failed build now unavoidably wipes the served assets — pre-existing, but the removed flag was the only way to bypass it.

---

garland3:

## Summary
Removes the `USE_NEW_FRONTEND` flag and the skip-frontend-build branch it guarded from `.env.example`, `docker-compose.yml`, `Dockerfile-test`, `agent_start.sh`, and `ps_agent_start.ps1`, adds a regression test asserting the string stays gone, and copies the two startup scripts into the test image so that test can run there.

## Verdict
Request changes.

## Required changes
1. Fix the multi-source `COPY` at `Dockerfile-test:44` — change the destination to `./` or split into two single-source lines; Docker rejects `COPY a b .` ("destination must be a directory and end with /"), so the test image fails to build and the entire containerized suite stops running, not just the new guard. Rebuild with `docker build -f Dockerfile-test .` to confirm.
2. Extend the regression guard at `atlas/tests/test_docker_env_sync.py:245` to assert the build functions have no early-return skip path (not just that one literal string is absent) — otherwise the same behavior returns under any other variable name and the test still passes.
3. Update the docstring at `atlas/tests/test_docker_env_sync.py:250` — it still says files are checked best-effort because they are not copied into the container, which this PR contradicts for the two startup scripts, and make the best-effort branch `pytest.skip` instead of silently `continue`ing so a container run reports reduced coverage.

## Evidence
Evidence: 8/10. The change is well demonstrated for its size — per-file accounting of every removed line plus five named verification commands — but the verification list predates the last two commits, so neither the `Dockerfile-test` `COPY` nor the 5th test has a recorded run; a container run showing the suite green is the one missing artifact, and it would have caught the build break above.

## Notes
- `ps_agent_start.ps1:517` got no parse check while the bash edit got `bash -n`; a `Parser::ParseFile` run would restore symmetry.
- `CHANGELOG.md:10` adds and removes the same flag inside one `[Unreleased]` block, and the operator note omits compose/`-e` overrides where the variable was also declared.
- `agent_start.sh:258` now unconditionally `rm -rf`s `atlas/static` before checking that `npm run build` succeeded — a failed build leaves no assets, and the flag was the last escape hatch.

---

garland3:

## Summary
Removes the `USE_NEW_FRONTEND` flag so the frontend is always built, updating `agent_start.sh`, `ps_agent_start.ps1`, `Dockerfile-test` and the CHANGELOG, and adds a regression test asserting the flag no longer appears.

## Verdict
Approve with nits.

## Evidence
Evidence: 7/10. The change is well described and the backend suite, lint and `bash -n` runs are recorded, but the verification list predates the last commits — nothing exercises the modified `Dockerfile-test`, and the PowerShell edit has no parse check. Re-running the containerized suite against the new image (showing 5/5 in `test_docker_env_sync.py`) is the one artifact worth adding.

## Notes
- `build_frontend` deletes `atlas/static` before checking that `npm run build` succeeded, and the flag was the only way to avoid that path — `agent_start.sh:258` (mirrored at `ps_agent_start.ps1:536`). Gating the `rm -rf` on build success is worth doing while this code is being touched.
- `Dockerfile-test` is not copied into the test image, so the `best_effort` loop's `pytest.skip` marks the whole guard SKIPPED in container CI even after the required assertions pass — `atlas/tests/test_docker_env_sync.py:296`. Splitting that check into its own test makes the pass visible.
- The guard matches the literal string `Skipping frontend build` over a fixed five-file list, so a differently-worded skip path or a flag reintroduced elsewhere still passes — `atlas/tests/test_docker_env_sync.py:266`.

---

garland3:

## Summary
Removes the `USE_NEW_FRONTEND` flag and its skip-build branch from `agent_start.sh` and `ps_agent_start.ps1`, adds regression tests asserting the flag is gone, and makes the frontend build capture `npm run build`'s exit status so a failed build no longer deletes `atlas/static/`.

## Verdict
Request changes.

## Required changes
1. Check `build_frontend`'s status at `agent_start.sh:368` — the `--frontend-only` path prints "Frontend rebuilt successfully" and `exit 0` immediately after a failed build, so operators and CI see success.
2. Check it again at `agent_start.sh:389` — full startup proceeds into `start_mcp_mock`/`start_backend` after a failed build, serving stale or missing assets with a normal-looking startup banner.
3. Give `Build-Frontend` a failure signal at `ps_agent_start.ps1:536` — the bare `return` is indistinguishable from success, so no caller can detect the failure even if it checks; use `throw` or `return $false` and test it at `:593` and `:625`.
4. Verify `frontend/dist` exists before `rm -rf "$PROJECT_ROOT/atlas/static"` at `agent_start.sh:258` and check `cp`'s status — a build that exits 0 without producing `dist/` still wipes the served assets, which is the failure this PR set out to close.
5. Add a test for the new failure path (stub `npm` exiting non-zero, assert return 1 and that a sentinel under `atlas/static/` survives) — the only new test asserts string absence and passes whether or not the guard works.
6. `COPY Dockerfile-test .` into the test image at `Dockerfile-test:40` so `test_use_new_frontend_flag_is_gone_in_dockerfile_test` runs in CI instead of permanently skipping on the file most likely to reintroduce the flag.

## Evidence
Evidence: 6/10. The flag removal is documented per-file with named verification commands, but that list predates the last several commits and covers none of the added surface. The most important gap is a transcript of a deliberately failing build showing the error message and `atlas/static/` left intact.

## Notes
- `test_docker_env_sync.py:266` keys on the literal log string `Skipping frontend build`; any reworded early return passes. Soften the docstring or assert the shape of `build_frontend` instead.
- The shell/PowerShell startup-script assertions sit in a module otherwise about `.env.example` ↔ `docker-compose.yml` parity (`test_docker_env_sync.py:245`); a dedicated module would age better.
- `test_docker_env_sync.py:271` re-asserts `script_path.exists()` for paths the first loop already checked.

---

garland3:

## Summary
Removes the `USE_NEW_FRONTEND` flag so the frontend is always built, adds failure handling to `build_frontend` / `Build-Frontend` so a failed build no longer wipes `atlas/static/`, and adds string-absence regression tests plus supporting `COPY` lines in `Dockerfile-test`.

## Verdict
Approve with nits.

## Evidence
Evidence: 6/10. The change is clearly described, but the Verification section predates the commits that added the failure handling (it reports 4/4 tests for a file that now has 6), and the most consequential new behavior — a failing build preserving `atlas/static/` and aborting — has no transcript or test. The single most useful addition would be a run of `agent_start.sh --frontend-only` with `npm` stubbed to fail, showing the error line, non-zero exit, and `atlas/static/` intact.

## Notes
- `Build-Frontend` throws while `build_frontend` returns non-zero, and neither PowerShell call site catches (`ps_agent_start.ps1:597`, `:629`) — the same failure prints one clean line on bash and a stack trace on Windows; the CHANGELOG at `CHANGELOG.md:11` also describes a return-code contract PowerShell does not have.
- `npm install` and the `cp -r frontend/dist atlas/static` are still unchecked (`agent_start.sh:243`, `:272`), leaving the same class of hole the guard was meant to close.
- The `Dockerfile-test` docstring at `atlas/tests/test_docker_env_sync.py:295` says the file is not copied into the image, which this PR's `COPY Dockerfile-test .` contradicts; the three new `COPY` lines also sit above the dependency layers and will bust build cache on every startup-script edit.

---

garland3:

## Summary
Removes the `USE_NEW_FRONTEND` flag from `.env.example`, `docker-compose.yml` and `Dockerfile-test`, makes `build_frontend` / `Build-Frontend` abort startup when the frontend build fails, and adds two string-absence regression tests plus the `COPY` lines needed for them to run in the test image.

## Verdict
Request changes.

## Required changes
1. Add `-ErrorAction Stop` to the `Remove-Item` and `Copy-Item` calls in `Build-Frontend` (`ps_agent_start.ps1:553-556`) — `Copy-Item` errors non-terminatingly by default, so a failed copy leaves `atlas/static` deleted and startup continues, which is the exact failure this PR set out to close.
2. Emit the caught exception in both `catch` blocks (`ps_agent_start.ps1:601`, `:638`), e.g. `Write-Host "ERROR: $($_.Exception.Message)"` — the `npm install` failure path throws with no prior output, so on Windows the only record of the cause is discarded.
3. Copy to a temp path and swap in `build_frontend` (`agent_start.sh:271`) — `rm -rf atlas/static` still runs before `cp -r`, so a copy failure destroys the served assets despite the CHANGELOG claiming they are preserved.
4. Add a test that runs `build_frontend` with `npm` stubbed to exit non-zero and asserts return status 1 with `atlas/static/` intact — the two added tests assert only string absence and pass whether the guard works, is inverted, or is deleted.
5. Reword `CHANGELOG.md:11` to state the actual per-script contract (bash returns non-zero, PowerShell throws) — as written, a caller checking a return value from `Build-Frontend` would never see the failure.

## Evidence
Evidence: 6/10. The flag removal is well documented, but the newer build-failure handling — now the riskiest part of the diff — has no demonstration: the recorded `4/4 passed` predates the two added tests, and there is no `docker build -f Dockerfile-test .` output despite the Dockerfile changing. The most valuable missing artifact is a transcript of a failing build showing a non-zero exit with `atlas/static/` still populated.

## Notes
- Move the three new `COPY` lines below the dependency installs in `Dockerfile-test:41-46` — as placed, editing a startup script busts the `npm ci` and dependency layers.
- Guard the directory change: `cd "$PROJECT_ROOT/frontend" || return 1` (`agent_start.sh:243`), and the same at `ps_agent_start.ps1:520`, so `npm install` cannot run lifecycle scripts from an unintended directory.
- Send the new failure messages to stderr (`agent_start.sh:244`, `:257`, `:265`, `:271`) — they are the first output meaning "the run failed".

---

PR #867, opened 2026-08-29
fix: open chat-response links in a new tab (closes #859)

### LINKED ISSUE #859: any links in llm output, should automatically be converted to open a new tab
right now they just, open which navigates you away from ATLAS.

garland3:
## Summary

Links rendered in LLM chat responses opened in the same tab, navigating the browser away from ATLAS. This adds `target="_blank"` preservation to the assistant-message rendering pipeline.

## Root cause

The marked link renderer (`frontend/src/utils/markdownRenderer.js`) already adds `target="_blank" rel="noopener noreferrer"` to external links. However, the rendered HTML is then passed through `DOMPurify.sanitize(html, DOMPURIFY_CONFIG)` before being injected via `dangerouslySetInnerHTML` in `Message.jsx`. DOMPurify 3.x **strips the `target` attribute by default** (it is no longer in the built-in allowlist), so `target="_blank"` was silently removed before the HTML reached the DOM — only `rel="noopener noreferrer"` survived. The result: every link in a chat response opened in the same tab.

## Fix

Add `target` to `DOMPURIFY_CONFIG.ADD_ATTR` so DOMPurify preserves the `target="_blank"` the renderer emits. One-line config change; no change to the renderer, the sanitization call sites, or any rendering logic.

- External links (`https://...`) and bare/autolinked URLs keep `target="_blank" rel="noopener noreferrer"` and open in a new tab.
- In-app fragment links (`#section`) are unchanged: the renderer already omits `target` for those, so they stay same-tab (e.g. citation-badge / references scrolling).

## Scope

Per the issue clarification, this applies to links rendered in the **chat response**. The same DOMPurify-strips-`target` behavior technically also affects the operator-configured `HelpPage` and `SplashScreen` markdown (they call `DOMPurify.sanitize(html)` with no config), but those are not LLM output and are out of scope for this issue.

## Verification

- New vitest suite `frontend/src/test/llm-link-target.test.js` exercises the full pipeline (`marked.parse` -> `DOMPurify.sanitize` with `DOMPURIFY_CONFIG`):
  - markdown link keeps `target="_blank"`
  - bare/autolinked URL keeps `target="_blank"`
  - `rel="noopener noreferrer"` present on external links
  - in-app `#fragment` link does **not** get `target="_blank"`
  - multiple links each keep `target="_blank"`
- `npx vitest run` — 659 tests pass (59 files), including the 5 new tests.
- `npm run lint` — 0 errors (4 pre-existing warnings unchanged).
- Built the frontend with `agent_start.sh` and verified live in the browser on a running backend: an assistant reply containing `[link to Example](https://example.com)` and a bare `https://github.com` both rendered as `<a href="..." target="_blank" rel="noopener noreferrer">`.

## Changes

- `frontend/src/utils/markdownRenderer.js` — add `target` to `DOMPURIFY_CONFIG.ADD_ATTR` (+ explanatory comment).
- `frontend/src/test/llm-link-target.test.js` — new test file guarding the pipeline.

Closes #859.


garland3:
## Summary
Adds `target` to `DOMPURIFY_CONFIG.ADD_ATTR` in `frontend/src/utils/markdownRenderer.js` so the `target="_blank"` emitted by the custom link renderer survives sanitization, plus a new vitest suite and a CHANGELOG entry.

## Verdict
Request changes.

## Required changes
1. Keep same-origin links in the current tab at `frontend/src/utils/markdownRenderer.js:75` — the renderer only exempts `#`-prefixed hrefs, so an assistant reply containing `[settings](/settings)` now opens a second copy of the app in a new tab instead of navigating in place. Extend the same-tab branch to `/`, `./`, `../`, and hrefs resolving to `window.location.origin` (a small `isExternalHref` helper keeps it testable).
2. Constrain the widened `target` allowlist at `frontend/src/utils/markdownRenderer.js:59` — `ADD_ATTR` is element-agnostic, so raw HTML in model output such as `<a href="https://evil.example" target="_blank">` (no `rel`) or `target="_top"` now survives sanitization. Add a DOMPurify `afterSanitizeAttributes` hook that drops `target` on non-anchors, normalizes it to `_blank`, and forces `rel="noopener noreferrer"`.
3. Add a test for raw-HTML anchors in `frontend/src/test/llm-link-target.test.js` pinning the behavior from item 2, and one for `[x](/settings)` covering item 1 — the current suite exercises only renderer-generated external links.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size: the new suite drives the real `marked.parse` -> `DOMPurify.sanitize` pipeline and covers the markdown link, autolink, `rel`, fragment, and multi-link cases, with full-suite and lint results reported. The manual browser check is described in prose rather than captured, which is the only thin spot.

## Notes
- Test helper `renderAssistant` (`frontend/src/test/llm-link-target.test.js:16`) skips the citation/LaTeX stages of the real `Message.jsx` path, so a call site losing `DOMPURIFY_CONFIG` would not fail these tests.
- Assertions are whole-document `toContain` substring checks (`frontend/src/test/llm-link-target.test.js:22`); parsing the output and asserting per anchor would be more precise.
- New-tab links carry no "opens in a new tab" cue for assistive tech (`frontend/src/utils/markdownRenderer.js:75`).

---

garland3:
Addressed all three required changes from @garland3 (AGENT-REVIEW-BOT-3) in 9889f27.

**1. Same-origin links stay in the current tab.** Added an `isExternalHref(href)` helper (`frontend/src/utils/markdownRenderer.js`) and switched the renderer to use it instead of the old `href.startsWith(#)` check. In-app navigation now stays same-tab: fragments (`#x`), server-relative (`/settings`), relative (`./`, `../`), query strings, and absolute URLs that resolve to `window.location.origin`. Only protocol-relative (`//host`) and genuinely cross-origin URLs get `target="_blank"`. The helper reads `window.location.origin` lazily and degrades safely (treated as in-app) when there is no window.

**2. Constrain the widened `target` allowlist.** Added a DOMPurify `afterSanitizeAttributes` hook registered once at module load. It keeps `target` only on `<a>`, normalizes any anchor target to `_blank`, forces `rel="noopener noreferrer"`, and strips `target` from every other element. This closes the `target="_top"`/clickjacking path, the raw `<a target="_blank">`-without-`rel` tabnabbing path, and `target` on non-anchors. It applies to every `DOMPurify.sanitize` in the app; I checked `CanvasPanel` (`IFRAME_SANITIZE_CONFIG`) — the hook only touches `target`/`rel`, so iframe `src`/`sandbox` handling is unaffected.

**3. Tests.** Expanded `frontend/src/test/llm-link-target.test.js` (5 -> 11 tests) and switched from whole-document `toContain` substring checks to per-anchor DOM assertions. Added the raw-HTML anchor cases (`target="_top"` -> `_blank` + rel; `target="_blank"` missing rel -> rel forced; `target` stripped from `<div>`) and the same-tab in-app cases (`/settings`, `./`, `../`, same-origin absolute). Full suite: 665 pass.

**On the notes:**
- *Test helper skips citation/LaTeX stages* — intentional; these tests target the link-target pipeline specifically. Noted it in the test file header so a future reader knows the scope.
- *Substring -> per-anchor assertions* — done (see above).
- *"opens in a new tab" a11y cue* — deliberately deferred. Adding it cleanly means handling both renderer-emitted links and raw-HTML anchors the hook normalizes, and an `aria-label` would replace the link's accessible name. I think it's worth doing but as a focused a11y follow-up rather than expanding this PR's rendered output; happy to add a visually-hidden cue here if you'd prefer it in-scope.

Live browser check on a rebuilt backend: an assistant reply with `[Google](https://google.com)`, `[settings](/settings)`, `[#section](#section)`, and a bare `https://example.com` renders external/bare links with `target="_blank" rel="noopener noreferrer"` and the in-app `/settings` + `#section` links with `target=null` (same tab).

PR #868, opened 2026-08-29
fix(chat): stop the transcript shifting while typing in the composer (#866)

### LINKED ISSUE #866: Previous chat response jumps while typing into multi-line composer
## Summary

In an existing conversation, while typing a new message, the previous chat response can visibly jump downward by a few lines and then snap back up. This happens during typing only. It seems more likely once the composer grows to at least 2 lines tall, and is easy to notice around 3 lines. The effect can occur on each keystroke while the input remains multi-line.

## Reproduction notes

The issue is somewhat intermittent, but it still reproduces. It was just reproduced while entering a 3-line message. It has been seen across desktop and mobile, and on multiple operating systems.

## Current repro

1. Open an existing conversation with a prior response visible.
2. Type a new message until the composer reaches 2–3 lines.
3. Keep typing.
4. Observe the previous response jump down a few lines and then scroll/snap back up.

## Expected behavior

Typing in the composer should not cause previously rendered chat messages to shift vertically.

## Notes

This may be related to composer auto-resize, scroll anchoring, or prior viewport re-anchor work, but that relationship is not yet confirmed.

garland3:
Closes #866.

## Problem

While typing a new message in an existing conversation, the previous response visibly jumps down a few lines and then snaps back up — on each keystroke, and most noticeably once the composer has grown to 2–3 lines.

## Two per-keystroke causes

**1. The memoized `Message` never actually memoized.**

`Message` is `export default memo(Message)`, and all its props were referentially stable except one:

```jsx
onCorrect={correctionsEnabled ? () => handleOpenCorrection(index) : null}
```

A fresh arrow per render meant every `Message` re-rendered and re-reconciled on every keystroke — the whole transcript, including markdown/KaTeX/highlight subtrees. Any DOM churn there is also picked up by ChatArea's `MutationObserver`, which calls `scrollToBottom` with `behavior: 'smooth'` — an animated scroll, i.e. the "snap back up".

The handlers are now built in a `useMemo` keyed on the message list and the feature gate. Measured with the new test: typing a 3-line message went from **58 `Message` renders to 2**.

Note this path is only live when the fine-tune capture feature *and* user consent are on, which fits the issue's "somewhat intermittent".

**2. The composer resize displaced the transcript's `scrollTop`.**

```js
textarea.style.height = 'auto'
textarea.style.height = Math.min(textarea.scrollHeight, 128) + 'px'
```

The composer is a flex sibling of the scrollable message list, so collapsing it to `auto` momentarily *grows* that list's viewport, which shrinks its scrollable range. The browser clamps `scrollTop` into the smaller range at the forced layout (the `scrollHeight` read), and the clamp survives the re-expansion. I confirmed the clamp in headless Chromium with a minimal flex harness: `scrollTop` went 1275 → 1259 (one line) mid-measurement.

Extracted to `utils/composerAutoResize`, which snapshots the scroll offset — or the pinned-to-bottom state, so a transcript that is following along stays pinned — and restores it after measuring. It all runs in one task, so nothing is painted in the collapsed state.

## Tests

- `src/test/composer-typing-no-transcript-rerender.test.jsx` — renders `ChatArea` with corrections enabled, types a 3-line message, asserts zero additional `Message` renders. Fails on `main` (`expected 58 to be 2`).
- `src/test/composer-auto-resize.test.js` — 8 cases against a scroll container that models a browser's permanent `scrollTop` clamp, including a regression case running the pre-fix code verbatim to prove the harness reproduces the displacement.

## Verification

- `npx vitest run` — 60 files, **663 passed**
- `npx eslint src/` — 0 errors (4 pre-existing warnings, none in changed files)
- `npx vite build` — clean


garland3:
## Summary
Fixes the transcript shifting while typing by extracting composer auto-resize into `utils/composerAutoResize.js`, which snapshots and restores the message list's scroll offset around the height measurement, and by memoizing the per-message `onCorrect` handlers so the transcript no longer re-renders on every keystroke.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — well above the bar for a change this size: a component render-count test that fails on `main`, an 8-case unit suite including a pre-fix regression case, and headless-browser measurements quoted in the description. The one missing artifact is a before/after recording of the actual visual symptom in a real conversation.

## Notes
- `handleSubmit` still writes `textarea.style.height = 'auto'` inline in both the success and catch branches (`frontend/src/components/ChatArea.jsx:274`, `:284`), leaving two sizing paths outside the new util; route them through a shared reset so the invariant holds everywhere.
- The 128px cap is duplicated three times — util default, call site (`ChatArea.jsx:151`) and the inline `maxHeight` style (`ChatArea.jsx:1137`); export one `COMPOSER_MAX_HEIGHT` so they cannot drift.
- No test invokes the rewired `onCorrect` prop (`ChatArea.jsx:844`), so a wrong index mapping would open the correction modal against the wrong turn without failing the suite.

---

garland3:
Thanks — addressed all four points in 5a8c087.

**Changelog** (@chatgpt-codex-connector, P1): added the `### PR #868 - 2026-08-29` entry under `## [Unreleased]`.

**Two sizing paths outside the util** (`handleSubmit:274/:284`): both now go through a new `resetComposerHeight`. Worth flagging a subtlety this surfaced — the reset can't just delegate to `autoResizeComposer`. It runs immediately after `setInputValue('')`, *before* React writes the empty value to the DOM, so a measurement there would still see the sent text and leave the composer tall. It clears the inline height instead and lets the stylesheet's `rows=1` / `min-height` take over, which is both correct and independent of when the value lands. There's a test for exactly that (`does not re-measure the not-yet-cleared value`). Both entry points share a `withPreservedScroll` wrapper, so there is now one place that mutates the composer's height.

**Duplicated 128px cap**: exported `COMPOSER_MAX_HEIGHT`, now used by the util default, the call site, and the inline `max-height` style.

**`onCorrect` untested**: added a case over a four-message transcript that clicks each correction button and asserts the index it resolves to. Verified it catches the failure it's meant to — with `handleOpenCorrection(index + 1)` it fails `expected [ 2, 4 ] to deeply equal [ 1, 3 ]`.

One thing the new `resetComposerHeight` tests made explicit: collapsing the composer on send genuinely grows the transcript's viewport, so a near-bottom offset can legitimately fall outside the new scroll range. That clamp is the composer correctly shrinking, not the #866 displacement, and the test asserts `min(before, maxScrollTop)` rather than pretending the old offset is reachable.

On the missing artifact you noted: agreed, there's no before/after recording of the symptom in a real conversation. I did not stand up the server + auth for this; the evidence is the headless-Chromium measurement (`scrollTop` 1275 → 1259 mid-measurement) and the render-count test that fails on `main`.

Verification: `npx vitest run` → 60 files, **669 passed**; `npx eslint` clean on all changed files; `npx vite build` clean.

garland3:
## Summary
Fixes the transcript "jump" reported in #866 by extracting composer sizing into a `composerAutoResize` utility that preserves the transcript's scroll offset across height mutations, and by memoizing the per-message `onCorrect` handlers so typing no longer re-renders every message.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — the change is well demonstrated for its size: a render-count regression test that fails on `main` (58 → 2), an 8-case unit suite modelling the browser `scrollTop` clamp, and a headless measurement of the root cause. The one missing artifact is a before/after recording of the user-visible symptom itself; the proxy evidence is strong enough that this strengthens rather than gates the change.

## Notes
- `correctionHandlers` still rebuilds on every streamed token (`ChatArea.jsx:119` depends on `messagesWithOrdinals`/`handleOpenCorrection`, both keyed on `messages`), so the memo is defeated during streaming; keying on `messages.length` with the handler held in a ref would close that path.
- The `MutationObserver` at `ChatArea.jsx:248` still answers any transcript mutation with a smooth `scrollToBottom(false)`, while the streaming path deliberately uses instant — worth aligning.
- `Message` calls `onCorrect(message)` but the bound closure ignores the argument and uses an index (`ChatArea.jsx:122`); pin one contract before anyone wires the handler directly.

---

garland3:

garland3:
## Summary
Fixes the transcript shifting while typing in the composer (#866) by memoizing the per-message correction handlers and moving composer height measurement into a new `utils/composerAutoResize` that preserves the message list's scroll offset. Adds unit and render-count tests plus a changelog entry.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. The change is well demonstrated for its size — a measured root cause, a before/after render count (58 → 2) from a test that fails on `main`, and a regression case running the pre-fix code. The one missing artifact is a before/after recording of the reported visual symptom in a real conversation; the author notes explicitly that no server was stood up.

## Notes
- The memo at `frontend/src/components/ChatArea.jsx:119` is keyed on `messages`, whose identity changes per streamed token, so the fix holds only while idle — typing during a stream still re-renders the transcript. Holding `messages` in a ref and keying on `messages.length` would cover it.
- `onCorrect` has two contracts: `Message.jsx:744` calls `onCorrect(message)` but the closures at `ChatArea.jsx:122` ignore the argument and use a captured index. Pick one shape before a second call site is added.
- `wasPinnedToBottom` uses a bare `<= 1` at `frontend/src/utils/composerAutoResize.js:47`; fractional `scrollTop` under zoom or non-integer DPR can exceed it and stop the transcript auto-following. Consider a small named tolerance with a fractional test case.

---

garland3:

PR #869, opened 2026-08-29
fix: keep in-app chat links same-tab; harden target allowlist (#867 follow-up)

### LINKED ISSUE #859: any links in llm output, should automatically be converted to open a new tab

garland3:
## Summary

Follow-up to **#867** (merged) that lands the review feedback from `AGENT-REVIEW-BOT-3` which did not get pushed before #867 was merged. #867 closed #859 by making external links open in a new tab, but the merged version has two gaps the reviewer flagged:

1. **In-app links still open a new tab.** #867 only exempted `#`-prefixed hrefs from `target="_blank"`, so a chat link like `[settings](/settings)` opens a second copy of ATLAS instead of navigating in place.
2. **The widened `target` allowlist is unconstrained.** `DOMPurify.ADD_ATTR` is element-agnostic, so raw model HTML such as `<a target="_top">`, `<a target="_blank">` without a `rel`, or `target` on a non-anchor element survives sanitization (a `target="_top"`/clickjacking path and a tabnabbing path).

## Changes

`frontend/src/utils/markdownRenderer.js`:
- New `isExternalHref(href)` helper. In-app navigation stays same-tab: fragments (`#x`), server-relative (`/settings`), relative (`./`, `../`), query strings, and absolute URLs that resolve to `window.location.origin`. Only protocol-relative (`//host`) and cross-origin URLs get `target="_blank"`. Reads `window.location.origin` lazily and degrades safely with no window.
- New DOMPurify `afterSanitizeAttributes` hook (registered once at module load). Keeps `target` only on `<a>`, normalizes any anchor target to `_blank`, forces `rel="noopener noreferrer"`, and strips `target` from every other element. Applies to every `DOMPurify.sanitize` in the app; `CanvasPanel` (`IFRAME_SANITIZE_CONFIG`) is unaffected since the hook only touches `target`/`rel`.

`frontend/src/test/llm-link-target.test.js`:
- Grows 5 -> 11 tests and switches from whole-document `toContain` substring checks to per-anchor DOM assertions. Adds: protocol-relative new-tab; same-tab for `/settings`, `./`, `../`, and same-origin absolute; raw `<a target="_top">` -> `_blank`+rel; raw `<a target="_blank">` missing rel -> rel forced; `target` stripped from `<div>`.

## Verification

- `npx vitest run` — 665 tests pass (59 files), including the 11 link-target tests.
- `npm run lint` — 0 errors (4 pre-existing warnings).
- Live browser check on a rebuilt backend (during #867): an assistant reply with `[Google](https://google.com)`, `[settings](/settings)`, `[#section](#section)`, and a bare `https://example.com` rendered external/bare links with `target="_blank" rel="noopener noreferrer"` and the in-app `/settings` + `#section` links with `target=null` (same tab).

## Note

This PR is intentionally additive on top of the merged #867; it does not revert anything in main. The follow-up CHANGELOG entry sits above the shipped `### PR #867` entry under `## [Unreleased]`.

Addresses the three "Required changes" from the review on #867.


garland3:
## Summary
Adds an `isExternalHref` classifier to the markdown renderer so in-app links render without `target`, and registers a global `afterSanitizeAttributes` DOMPurify hook that restricts `target` to anchors and forces `rel="noopener noreferrer"`. Test coverage grows from 5 to 11 cases with per-anchor DOM assertions.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — the change is well demonstrated for its size, with a named test behind each claimed behavior plus stated lint/test runs. The one gap is that the live-browser check is explicitly scoped to the predecessor PR (#867), so the new helper and hook have not been shown working in a running app.

## Notes
- `isExternalHref` returns in-app for `/\evil.example` because the `href.startsWith('/')` check runs before URL resolution; browsers normalize the backslash and navigate cross-origin in the current tab. Normalize backslashes (or resolve through `new URL`) before the prefix tests — `frontend/src/utils/markdownRenderer.js:81`.
- `mailto:`/`tel:` hrefs resolve to origin `"null"` and get `target="_blank"`, leaving an orphan tab; add an early return for non-navigational schemes — `frontend/src/utils/markdownRenderer.js:88`.
- The hook mutates the shared DOMPurify singleton as an import side effect, so its reach and its registration order are implicit; consider an exported `sanitizeMarkdown()` wrapper plus an idempotence guard — `frontend/src/utils/markdownRenderer.js:231`.

---

garland3:
Thanks @chatgpt-codex-connector and Copilot -- addressed all three points in 19af7a9.

**CHANGELOG heading (Codex + Copilot).** Reformatted to the required `### PR #869 - 2026-08-28` format and shortened the entry to a single bullet.

**Raw-HTML external anchor bypass (Copilot).** Real gap -- good catch. The `afterSanitizeAttributes` hook previously only acted when a raw `<a>` already carried a `target`, so a raw `<a href="https://example.com">` (no target) slipped through and opened same-tab. The hook now classifies every surviving anchor with `isExternalHref(href)`: external anchors get `target="_blank" + rel="noopener noreferrer"` enforced (even when the model omitted `target`), and in-app anchors have any injected `target` stripped. Non-anchor elements still lose `target`. So raw-HTML external links can no longer bypass the new-tab behavior, and a model-injected `target="_blank"` on an in-app `/settings` link can no longer open a second copy of the app.

**Tests (Copilot).** Added the two cases: a raw external `<a href>` with no target now asserts `target="_blank"` + `rel`; a raw `<a href="/settings" target="_blank">` asserts the injected target is stripped. Suite is now 13 tests, all per-anchor DOM assertions. Full suite: 667 pass; `npm run lint`: 0 errors.

Re-running CI on the updated head.

garland3:
## Summary
Chat markdown links are now classified as in-app vs external, with external anchors forced to `target="_blank" rel="noopener noreferrer"`, enforced both in the marked renderer and in a global `afterSanitizeAttributes` sanitizer hook.

## Verdict
Request changes.

## Required changes
1. Normalize backslashes before the prefix tests in `isExternalHref` at `frontend/src/utils/markdownRenderer.js:87` — `/\evil.example` hits `href.startsWith('/')` and is classified in-app, so a cross-origin link navigates the current tab with no `rel`, which is the exact case this PR set out to prevent. Add a regression test. (A fix was reported in commit `43613b9`, which is not in this head.)
2. Return `false` early for non-navigational schemes (`mailto:`, `tel:`, `sms:`) at `frontend/src/utils/markdownRenderer.js:94` — `new URL('mailto:…').origin` is `"null"`, so these get `target="_blank"` and leave an orphan blank tab after the handler opens.
3. Scope the sanitizer policy or state its true reach at `frontend/src/utils/markdownRenderer.js:234` — the hook mutates the shared DOMPurify singleton at import, so `HelpPage.jsx:35`, `SplashScreen.jsx:89` and `CanvasPanel.jsx:321/330/363/385` gain added `target`/`rel` past `IFRAME_SANITIZE_CONFIG`'s allowlist, and only if some other module imported this one first. Export a `sanitizeMarkdown(html)` wrapper that owns the hook lifecycle, or keep it global with an idempotence guard plus an updated description/CHANGELOG.
4. Add a test that sanitizes through `IFRAME_SANITIZE_CONFIG` and a bare `DOMPurify.sanitize` at `frontend/src/test/llm-link-target.test.js` — the "CanvasPanel is unaffected" claim is asserted nowhere, and every existing case goes through `renderAssistant`.

## Evidence
Evidence: 8/10 — the behavior is non-visual, and each claimed case maps to a named test, with the suite strengthened from substring checks to per-anchor DOM assertions. The main gap is that the browser-level check is inherited from #867, which predates both `isExternalHref` and the hook; the test counts quoted across the description and comments (11/665, 13/667, 15/669) also do not agree on which commit passed.

## Notes
- Protocol-relative same-origin URLs are forced to a new tab by the `//` short-circuit at `frontend/src/utils/markdownRenderer.js:86`; dropping the special case lets the origin comparison decide.
- The `catch` at `frontend/src/utils/markdownRenderer.js:92` classifies unparseable hrefs as in-app; external is the conservative direction here.
- `renderer.link` at `frontend/src/utils/markdownRenderer.js:107` now duplicates the target/rel policy the hook already applies; pick one enforcement point.

---

PR #873, opened 2026-08-31
fix: use the AWS credential chain when no S3 keys are configured

garland3:
## Problem

The S3 credential settings defaulted to `minioadmin` and were passed straight into `boto3.client(...)`. boto3 does **not** fall back to its default credential chain when explicit credentials are supplied — even empty strings disable it. Only omitted or `None` values let boto3 resolve credentials via env vars → shared config → container credentials → instance metadata.

That made AWS-native credential mechanisms (IAM roles, EKS Pod Identity, IRSA, EC2 instance profiles) unusable: the client always authenticated as `minioadmin`.

## Fix

- `atlas/modules/config/settings.py`: `s3_access_key` / `s3_secret_key` are now `Optional[str] = None`.
- `atlas/modules/file_storage/s3_client.py`: falsy credentials are passed to boto3 as `None`, so the standard credential chain applies. `endpoint_url`, `region_name`, `use_ssl` and the `Config(...)` block are unchanged.

## Compatibility

Existing MinIO deployments that set `S3_ACCESS_KEY` / `S3_SECRET_KEY` are unaffected — those values are still passed through to boto3 verbatim.

## Tests

New `atlas/tests/test_s3_client_credentials.py` asserts that:
- no configured credentials (`None` or `""`) → `boto3.client` receives `aws_access_key_id`/`aws_secret_access_key` as `None`, with the rest of the client config intact;
- explicitly injected credentials are passed through;
- credentials coming from settings (the env-var path) are passed through;
- the settings defaults are `None`.

Full backend suite: 2678 passed, 17 skipped.


garland3:
Addressed both review points in ecaaff7:

- **Copilot (partial credentials):** the two credentials are no longer normalized independently. The pair is passed to boto3 only when *both* values are set; if exactly one is set it is ignored with a warning and the default credential chain is used, so a half-configured pair can't reach boto3 and fail signing in a confusing way. Added a parametrized test covering all four half-set combinations.
- **Codex (documentation):** `docs/admin/file-storage.md` now has a "Credentials" subsection explaining that `S3_ACCESS_KEY`/`S3_SECRET_KEY` are optional, what boto3 resolves when they are omitted (env vars → shared config → container credentials incl. ECS task roles, EKS Pod Identity and IRSA → EC2 instance metadata), and that both must be set together. The CHANGELOG `[Unreleased]` entry for this PR was added after the reviewed commit and now mentions the docs and the pairing rule.

Backend suite: 2682 passed, 17 skipped.

garland3:
## Summary
Makes `S3_ACCESS_KEY`/`S3_SECRET_KEY` optional so `S3StorageClient` omits the credential kwargs and lets boto3's default credential chain (IAM role, IRSA, instance metadata) resolve them, warns when only one of the pair is set, and documents the behavior in the admin file-storage page.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — well demonstrated for a small backend change, with unit tests covering explicit, absent, and all four half-set credential combinations plus a green backend suite. The one gap is that nothing shows the chain actually resolving end to end (e.g. a test that sets `AWS_ACCESS_KEY_ID` in the environment and asserts the client picks it up).

## Notes
- Local (non-Docker) MinIO loses its working default: `.env.example:443-444` still ships `S3_ACCESS_KEY`/`S3_SECRET_KEY` commented out, so the documented host-against-compose-MinIO setup now has no credentials. Uncomment them and add a migration sentence to `docs/admin/file-storage.md:31` noting the `minioadmin` default is gone.
- The init log at `atlas/modules/file_storage/s3_client.py:110` reports only endpoint and bucket; add `credentials=explicit|default-chain` so a dropped secret that silently falls through to the instance role is visible in deployment logs.
- Passing `s3_access_key=None` explicitly cannot request the chain — `s3_access_key or config.s3_access_key` at `s3_client.py:63` still substitutes configured keys. A distinct `_UNSET` default for the credential pair would separate "omitted" from "deliberately none", and resolving the pair from a single source would also stop an injected key from splicing with a half-set env pair.

---

PR #874, opened 2026-08-31
fix(rag): restore citations for atlas_search as structured message data

garland3:
## The problem

Selecting data sources and asking a question produced an answer with **no sources attached**. Nothing in the UI said what had been read.

#862 made `atlas_search` an explicit tool call and deleted `call_with_rag_and_tools`. That method did **three** things, and the #862 design note accounts for one:

1. injected the retrieved context as a system message — the silent retrieval #862 set out to remove;
2. appended `_build_citation_instructions()` — the text telling the model to write `[1]`, `[2]`;
3. appended `_format_rag_references()` — a literal `**References**` markdown block — to the assistant's own message.

(2) and (3) went with (1). Neither is mentioned in that PR.

What made it invisible rather than obviously broken: the citation UI was never a data path. `Message.jsx` triggered on `latexRestoredHtml.includes('References')` and `ragCitations.js` recovered the source list with **regexes over the rendered HTML**. With nothing writing that block, the extractor found nothing and there was no error to notice. A second, independent blocker sat underneath — `VITE_FEATURE_RAG_CITATIONS` defaulted to `false` in `.env.example`, both Dockerfiles, `quay-publish.yml`, and both start scripts.

This was a **design gap**, not a rewiring job. The data existed (`_tool_references()` since #791) and reached the browser, where `processToolResult` had no branch for it and it was pretty-printed as JSON inside a collapsed tool row.

## Why scraping could not be repaired

Scraping worked only because exactly one retrieval happened per turn, before the model was asked anything. A tool has none of those properties: the model may search zero times or six, the answer is written across those calls, and two calls can return the same document. There is no "the RAG response" to append to.

It was also why `_sanitize_snippet` existed — passage text sat in the same string the extractor parsed, so a snippet starting `1. ` could masquerade as a reference entry. Structured data removes that surface.

## The design

A **turn-scoped citation register**, carried on the tool execution context exactly like the sleep tool's `TURN_BUDGET_KEY`: `tool_executor` forwards it to every call, the turn's owner reads it back. Numbers are:

- **Identity-keyed** — a document returned by two searches in one turn keeps one number. Identity is `url`, then `filename`, then `document_ref`, **in that order**. `document_ref` sounds authoritative but indexes *one response's* result list, so the same document is 3 from one search and 5 from the next. Trusting it first made a repeated hit take two numbers — caught against the reference mock, where two searches each returned `Code of Conduct, pol-003.txt` and it was numbered twice.
- **Scoped by data source** — filename is in that chain, and two corpora may each hold a `policy.pdf` that are not the same document.
- **Stable across turns** — seeded from citations persisted on earlier messages, carrying both the count (a new document never reuses a number) and the identities (a document cited again keeps its number). `[3]` names one document for the whole transcript.

The numbers are echoed into the tool result — the model can only cite a number it has been shown — and `SEARCH_TOOL_DESCRIPTION` carries the instruction. That lives on the tool, not the system prompt, so it travels with the schema wherever the tool is offered, agent mode included.

At turn close the register is **persisted** (message metadata: reload renders sources without re-running retrieval; the next turn continues the numbering) and **published** (a `citations` event). One event per turn, not per search: a turn may search three times and cite two documents, and a per-call event would show sources for a call the model discarded.

`get_messages_for_llm` folds an identity-only line of the conversation's numbers into the next turn — `agent_digest`'s `_MAX_RESULT_CHARS = 400` is right for passage text and wrong for citation numbers.

## Also fixed

- `_query_mcp` attaches references, so MCP-backed RAG sources stop being silently uncitable.
- The tool row summarizes `references` instead of dumping raw JSON.
- `VITE_FEATURE_RAG_CITATIONS` defaults to on.

## RAG mode is untouched

RAG mode (sources selected, no tools) still pre-injects and still appends `**References**` — it has no tools schema and no model turn to spend, so pre-injection is the design there, exactly as #862 left it. The markdown extractor is **kept, not deleted**; the structured path wins when both are present.

## Verification

- `bash test/atlas_tests.sh` — **2711 passed**, 17 skipped
- `npx vitest run` — **699 passed** (63 files)
- `npm run lint` — 0 errors; `ruff check atlas/` clean
- `scripts/check-docs.sh` — OK
- New: `test_citation_register.py` (26 cases) and `structured-citations.test.js` (10 cases), plus `test_two_searches_in_one_turn_share_one_numbering` through the real execution path.

**End to end**, against the reference RAG mock with a live model — six `atlas_search` calls over two turns. Screenshots and the recorded `citations` events are in the comments below.

Design note: `docs/developer/design-notes/citations-for-tool-based-search-2026-08-31.md`


garland3:
## End-to-end verification (live model + reference RAG mock)

Driven with Playwright against a real backend on the branch build, with the bundled `mocks/atlas-rag-api-mock` serving three corpora and `openrouter-gpt-oss-120b` as the model. Two data sources selected, auto-approve on, no tools ticked — so `atlas_search` reaches the model purely because sources are selected (the #862 implied-tool path).

### 1. Two data sources selected

Sources selected

### 2. Turn 1 — four `atlas_search` calls, one references block

The model searched four times. Every document is numbered once, and the answer carries inline chips.

First turn with references expanded

The expanded block shows the label, the data source, the backend's citation string (with its own conflicting `[n]` prefix stripped — the register owns the numbering), and the matched passage text.

Recorded `citations` event:

```
turn 1: n = [1, 2, 3, 4, 5]
   [1] Remote Work Policy, pol-001.txt
   [2] Expense Reimbursement Policy, pol-002.txt
   [3] Code of Conduct, pol-003.txt
   [4] PTO and Leave Policy, pol-004.txt
   [5] Microservices Architecture, tech-004.txt
```

Four searches, **one** event, five distinct numbers — the multi-call case.

garland3:
### 3. Turn 2 — numbering survives the turn boundary

Second turn, two more `atlas_search` calls. One of the documents (`Microservices Architecture, tech-004.txt`) was already cited in turn 1.

Second turn

```
turn 2: n = [5, 6, 7]
   [5] Microservices Architecture, tech-004.txt      <- kept its number from turn 1
   [6] Deployment Pipeline, tech-003.txt
   [7] Database Schema Documentation, tech-002.txt
```

The re-cited document comes back as **`[5]`**, not a fresh `[8]`, and the genuinely new documents continue at `[6]`/`[7]`. The model wrote `⁶` and `⁷` inline for the new material and `⁵` for the architecture doc — it used the replayed numbering from `get_messages_for_llm` without being told again in the prompt.

Totals for the run: **6 `atlas_search` calls, 2 turns, 2 reference blocks, 8 reference entries, 0 duplicates.**

### The two bugs this run caught

Both were found by running against real data, not by the unit tests:

1. **`document_ref` is not a document id.** It indexes *one response's* result list, so the same document came back as `document_ref` 3 from one search and 5 from the next and was numbered twice — `[3]` and `[5]` were both `Code of Conduct, pol-003.txt`. Identity now prefers `url`, then `filename`, then `document_ref`.
2. **Re-citing across turns minted a new number.** Seeding the register with only the *count* meant turn 2 gave `Microservices Architecture` a fresh `[8]`, so `[5]` and `[8]` were the same document — the same misleading-scrollback problem the design set out to avoid, one turn later. The register now carries forward the prior *identities* too.

Both are pinned by tests (`test_a_repeated_hit_is_one_number_even_as_document_ref_shifts`, `TestNumberingSurvivesTheTurnBoundary`).

garland3:
## Review feedback addressed in d533b99

**`results.citations`, not `documents`/`sources`** — correct, and this was the important one. `RagMCPService.synthesize` collects each server's citations under `results["citations"]` (`atlas/domain/rag_mcp_service.py:489-494`, returned at :519-524) and never emits `documents` or `sources`, so the MCP branch I added was a silent no-op and MCP-backed sources stayed exactly as uncitable as before the PR.

**And the label field, also correct.** I was reading `doc.get('filename')` back off the server's own dict when the label had been resolved from `title`/`source`, so the echoed reference would usually have carried no label at all. Worse than that: the MCP citation shape is *server-defined*, not `DocumentMetadata` — the bundled `corporate_cars` server returns `resourceId`, `snippet`, `car`, `location`, `vin` and nothing resembling a filename. Assuming the HTTP field names would have left every such entry unlabelled and therefore unnumbered (the register drops entries with nothing identifying).

So the label is now taken from whichever of `title`, `filename`, `name`, `label`, `source`, `resourceId` the server actually filled; the echoed reference is read back off the *registered* entry rather than the raw dict; the `snippet` rides along as evidence for the references panel; and identity is scoped by the citation's own `server` stamp when the aggregator set one.

Two tests pin it, both of which fail on the previous commit:
- `test_mcp_sources_are_citable` — the ordinary `title`/`url` case
- `test_mcp_citations_are_labelled_from_whatever_field_the_server_filled` — the `corporate_cars` shape, asserting `{"n": 1, "filename": "vin-1HGCM"}`

This also closes an open question I had flagged for the operator: I could not tell from the repo whether MCP RAG backends returned per-document identity, and had guessed at the key names. The aggregator answers it.

**The three stale comments** — all correct, fixed. `.env.example`, `agent_start.sh:250`, `ps_agent_start.ps1:532` still said the flag defaults to off after I flipped the values. The `.env.example` comment now also says *why* being build-time matters: a build that does not export the variable compiles the citation rendering out entirely, which is the trap that made this invisible in the first place.

Re-verified: **2713 backend tests pass** (up from 2711), `check-docs` OK, no line-ending churn.

garland3:
## Review round 2 — three more real findings, fixed in 1347fae

All three from the Codex review were genuine. (Its fourth, on `results.citations`, was already fixed in `d533b99` — it reviewed the earlier commit.)

### P1 — response-local markers colliding with stable numbers

The most serious of the three, and it undercut the whole point of the register. `AtlasRAGClient._format_raw_evidence` (`atlas_rag_client.py:650-678`) writes `[document_ref] heading`, falling back to a 1-based index — both **response-local**, so the second search of a turn restarts at `[1]`. Prefixing my stable numbers without touching those left a single tool result asserting both `[4] document.pdf` and `[1] document.pdf`. A model picking the wrong one cites a source it never read.

Passage markers are now rewritten to the register's numbers. The rewrite is anchored on the whole `[old] heading` line the formatter emits, **not** a bare `[N]` — `test_renumbering_leaves_bracketed_numbers_in_passage_prose_alone` pins that a passage saying "`[1] is cited by the paper as prior work`" is untouched.

### P1 — citations leaking across turns in agent mode

Correct, and I had missed the asymmetry: `AgentModeRunner` ends on `agent_completion` and **never** emits `response_complete` (confirmed — no `publish_response_complete` anywhere in `modes/agent.py` or `agent/`). So the pending list survived the turn and the *next* turn's `response_complete` attached it to a different answer. `agent_completion` now settles it. Two tests, including one asserting a later no-search turn gets nothing.

### P2 — batched HTTP citations attributed to the first corpus

Correct. `query_rag_batch` returns one response spanning several corpora, and I scoped every document to `group[0]`, so later corpora were displayed as belonging to the first and same-named files across them merged into one number. Batched responses now scope each document by its own `source`. Single-source queries keep the group scope deliberately — `DocumentMetadata.source` is used as a document *heading* elsewhere in the client, so I only lean on it where the group name is provably insufficient.

### Verification

- **2715 backend** (up from 2713), **701 frontend**, lint/check-docs clean, no line-ending churn.
- Re-ran the full live end-to-end afterwards, since finding 1 changes the text the model reads. Turn 2 now re-cites **all three** of turn 1's documents under their original numbers and mints none:

```
turn 1: n = [1, 2, 3, 4, 5, 6, 7]
turn 2: n = [5, 6, 7]     <- all three carried over, no new numbers
```

Second turn

The model wrote `⁷` and `⁶` inline, matching turn 1's numbering exactly.

One structural change worth flagging: `_tool_references` now returns the renumbering alongside the references, so `register_documents` moved out of the register into `mcp_execution` — the RAG payload shape belongs there, and the register keeps only the numbering primitive.

garland3:
## Review round 2, remaining two findings — fixed in 8856b94

I missed these two on my first pass through the review (both on `citation_register.py`). Both are real, and they are two sides of one bug.

### P2 — the ceiling was per-turn, not conversation-wide

Correct, and this one was a genuine hole in my own stated design. The check read `len(self._entries) >= MAX_CITATIONS`, but `_entries` holds only what *this turn* registered — it is always empty at the start of a turn. A conversation that had already issued 100 citations would start the next turn with a fresh register and carry straight on at 101, 102, and so on. The bound existed in the docstring and nowhere in the behaviour.

It now tests `self._next`, which *is* seeded from the conversation's persisted numbers, so the ceiling holds across turns.

### P2 — numbers outside what the renderer can match

Also correct. `processCitationBadges` matches `[\d{1,2}]`, so a model-emitted `[100]` would stay plain text rather than becoming a chip that scrolls to its entry — a silent failure of exactly the kind this PR exists to remove.

I took the reviewer's second option rather than the first. Widening the regex to three digits would make it match three-digit array indices in prose (`arr[100]`), and the `<code>`/`<pre>` guard does not cover inline prose. So the ceiling gives way instead: `MAX_CITATIONS = 100` becomes **`MAX_CITATION_NUMBER = 99`**, renamed for what it actually bounds, with the renderer constraint written down as the reason so nobody raises it later without checking the regex. `MAX_RENDERED_REFERENCES` on the frontend follows to 99.

One behaviour worth naming: at the ceiling, a document that already has a number is still recognised and keeps it. A long conversation stops *adding* sources rather than renumbering the ones it has — renumbering would be worse than dropping.

Four tests added, including `test_the_ceiling_is_conversation_wide_not_per_turn` (fails on the previous commit) and `test_a_seeded_document_is_still_recognized_at_the_ceiling`.

**2718 backend passing**, 701 frontend, lint and check-docs clean. Design note updated with the reasoning.

---

That is all six findings from this review round addressed — four in `1347fae`/`d533b99`, these two here. Thanks; the two P1s in particular (`results.citations`, and the response-local marker collision) were bugs I would not have caught from the tests alone.

garland3:

PR #875, opened 2026-08-31
fix(windows): stale POSIX-style APP_CONFIG_DIR no longer blanks the MCP config lookup; Build-Frontend matches intended body

garland3:
## Problem

Two Windows startup issues, owned end to end here:

1. **0 MCP servers load on Windows even though `config/mcp.json` exists.** The logs show:

   ```
   JSON config not found in any of these locations: ['C:\\home\\agarlan\\git\\atlas-ui-3\\config\\mcp.json']
   ```

   followed by `Created empty MCP config (no configuration file found)`.

2. **`Build-Frontend` in `ps_agent_start.ps1` drifted from the intended function body.**

## Root cause (1): POSIX-style `APP_CONFIG_DIR` is drive-less on Windows

`AppSettings.app_config_dir` picks up `APP_CONFIG_DIR` from the process env / `.env`. A value like `/home/agarlan/git/atlas-ui-3/config` (a leftover from a WSL/Linux setup of the same checkout) is:

- **absolute on POSIX**, so nothing breaks on Linux;
- **not absolute on Windows** — pathlib sees no drive letter (`is_absolute() == False`, `root='\', drive=''`), so `ConfigManager._search_paths` takes the `project_root / config_dir` branch, and joining a rooted drive-less path onto `C:\...\<repo>\atlas` *replaces everything after the drive letter*:

  ```python
  PureWindowsPath("C:/Users/agarlan/git/atlas-ui-3/atlas") / PureWindowsPath("/home/agarlan/git/atlas-ui-3/config")
  # -> C:\home\agarlan\git\atlas-ui-3\config
  ```

  which reproduces the logged `C:\home\agarlan\git\atlas-ui-3\config\mcp.json` byte for byte. The real `config\mcp.json` next to the checkout is never searched, so Atlas logs "not found", builds an empty `MCPConfig`, and starts with **0 MCP servers**.

### Fix

- `ConfigManager._search_paths` now also searches the conventional repo-root `config/` directory **between** the configured dir and the packaged `atlas/config/` defaults. A stale `APP_CONFIG_DIR` (POSIX path on Windows, deleted dir, typo) can no longer blind the lookup. With the default relative `config`, the new candidate deduplicates to the existing list, so default behavior is unchanged; with a valid absolute dir, that dir still wins (first match).
- The "config not found in any of these locations" warning now appends a hint naming `APP_CONFIG_DIR` when that directory does not exist on the host, so this failure is self-diagnosing next time.
- `ps_agent_start.ps1`'s `Start-Backend` (the Windows startup path from the report) now falls back to `<project_root>\config` with a loud `Write-Warning` when the inherited `APP_CONFIG_DIR` does not exist on the machine — this covers the operator's flow even before the backend fix is deployed.

## Build-Frontend matches the intended body

- `$env:VITE_APP_NAME` fallback is now the documented example default **`"ATLAS"`** (`.env.example` ships `VITE_APP_NAME=ATLAS`; the comment already said "fall back to the example default"). `agent_start.sh` is aligned so the parity pair stays consistent — flagging that this also changes the Linux fallback from `"Chat UI 13"`; easy to drop if unintended.
- The `frontend/dist not found` message uses `;` instead of an em dash (matches the intended body; also avoids mojibake on non-UTF8 Windows consoles).

## Verification (Linux — explicit limits below)

- **PowerShell 7.4.7** (Core, runs on Linux): full `ps_agent_start.ps1` parses with 0 syntax errors; existing `test/pr-validation/test_pr863_ps_env_parser.ps1` regression suite passes.
- **`Build-Frontend` equivalence**: token-for-token equality with the intended body (98 meaningful tokens via the PowerShell tokenizer — only blank-line placement differs). Behavior exercised with a stubbed `npm`: happy path (`npm install` → `npm run build` → `dist` copied to `atlas/static`, `VITE_APP_NAME` defaults to `ATLAS`, `VITE_FEATURE_RAG_CITATIONS` to `false`), install failure throws the exact message, build failure **preserves existing `atlas/static`**, missing `dist` throws and preserves assets.
- **Config resolution**: new unit tests in `atlas/tests/test_config_manager_paths.py`, including a `PureWindowsPath` simulation (platform-independent, runs on Linux) that reproduces the exact logged `C:\home\agarlan\...\mcp.json` path and asserts the repo-root fallback precedes package defaults; native-POSIX fallback + loader tests; default-dedup regression test.
- End-to-end on Linux: `ConfigManager.mcp_config` with a stale POSIX `APP_CONFIG_DIR` and `config/mcp.json` in a fixture checkout loads 1 server (previously: 0).
- `./test/run_tests.sh backend` green (includes live E2E); `ruff check` clean on changed files (48 pre-existing warnings on `main` untouched); `npm run lint` matches the `main` baseline (4 pre-existing warnings, 0 errors — no frontend files changed).

## What could NOT be verified (no Windows host)

- Actual `windows-pathlib` behavior was validated via `PureWindowsPath` (the platform-independent Windows flavour) rather than a real `WindowsPath`; concrete `WindowsPath.exists()` resolution against the current drive was not executed on a Windows kernel.
- The `ps1` changes were parsed and the extracted functions executed under PowerShell 7.4.7 on Linux, but not under Windows PowerShell 5.1 (`Start-Backend`'s `Test-Path -LiteralPath` and `Start-Process` semantics are equivalent for these calls, but this was not observed on Windows).
- The operator-side `.env` still contains the stale value; after merge the app tolerates it (repo-root fallback), but removing/commenting `APP_CONFIG_DIR=/home/...` in the Windows `.env` is the clean fix and is now warned about at startup.

Docs updated: `docs/getting-started/python-package.md` (stale `APP_CONFIG_DIR` behavior + fallback). Follow-up candidate (not in this PR): `PromptProvider._resolve_base_paths` has the same theoretical weakness if `PROMPT_BASE_PATH` is set to a POSIX-absolute path on Windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### PR #875 - 2026-08-31
+- **A stale POSIX-style `APP_CONFIG_DIR` no longer blanks the MCP config lookup on Windows**: a value like `/home/<user>/git/atlas-ui-3/config` (left in a Windows `.env` from a WSL/Linux setup) has no drive letter under Windows pathlib, so it is treated as relative and joined onto the current drive -- the backend searched a nonexistent `C:\home\...` location, logged "JSON config not found in any of these locations", and started with 0 MCP servers even though `config/mcp.json` existed next to the checkout. `ConfigManager._search_paths` now also searches the conventional repo-root `config/` directory (between the configured dir and the packaged `atlas/config/` defaults), the not-found warning names `APP_CONFIG_DIR` when it does not exist on the host, and `ps_agent_start.ps1` `Start-Backend` falls back to `<project_root>\config` with a loud warning. **Operators:** remove or fix any `APP_CONFIG_DIR=/home/...` line in your Windows `.env`; the app now tolerates it, but the clean setup is an unset value or a Windows path.
+- **`Build-Frontend` in `ps_agent_start.ps1` matches the intended body**: the `VITE_APP_NAME` fallback is the documented example default `"ATLAS"` (was `"Chat UI 13"`; `agent_start.sh` aligned so the parity pair stays consistent), and the `frontend/dist not found` error message uses `;` instead of an em dash to avoid mojibake on non-UTF8 Windows consoles.
+
 ### PR #873 - 2026-08-31
 - **S3 storage can now authenticate with IAM roles, EKS Pod Identity, IRSA or EC2 instance profiles**: `s3_access_key`/`s3_secret_key` defaulted to `minioadmin` and were passed straight to `boto3.client(...)`, which disables boto3's credential chain whenever explicit credentials (even empty strings) are given; they now default to `None` and falsy values are passed as `None`, while deployments setting `S3_ACCESS_KEY`/`S3_SECRET_KEY` are unchanged. An incomplete key pair is ignored (with a warning) rather than passed on to fail signing, and `docs/admin/file-storage.md` documents the credential-chain path.
 
--- a/agent_start.sh
+++ b/agent_start.sh
@@ -245,7 +245,7 @@ build_frontend() {
     # Use VITE_* values from the environment / .env instead of hardcoding.
     # If VITE_APP_NAME is not already set, fall back to the example default.
     if [ -z "$VITE_APP_NAME" ]; then
-        export VITE_APP_NAME="Chat UI 13"
+        export VITE_APP_NAME="ATLAS"
     fi
     # RAG citations UI is opt-in at build time; default to off if unset.
     if [ -z "$VITE_FEATURE_RAG_CITATIONS" ]; then
--- a/atlas/modules/config/config_loader.py
+++ b/atlas/modules/config/config_loader.py
@@ -50,6 +50,17 @@ class ConfigManager:
         Two-layer lookup:
         1. User config dir (APP_CONFIG_DIR, default "config/") - user customizations
         2. Package defaults (atlas/config/) - always available as fallback
+
+        The conventional repo-root ``config/`` directory is also searched in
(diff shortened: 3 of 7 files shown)


garland3:
Thanks for the review — all four unique findings are addressed in 5850663:

1. **Copilot (hint gating / CWD resolution)**: the hint now fires only when `APP_CONFIG_DIR` was explicitly provided (checked via `model_fields_set` with a `getattr` guard), and relative values are resolved against the project root before the existence check. Unit tests updated (stubs carry `model_fields_set`) plus a new test asserting the hint is skipped for the pydantic default.
2. **Codex (relative `APP_CONFIG_DIR` in `Start-Backend`)**: the guard now resolves relative values against `$PROJECT_ROOT` (not the current location, which is `atlas/`) before the `Test-Path` staleness check, matching the backend's `_search_paths` semantics. PowerShell behavior tests cover the relative-existing, relative-missing, and stale-POSIX branches.
3. **Codex (unconditional repo-root fallback)**: agreed — the rescue now applies **only when the resolved configured directory does not exist**. An existing (even deliberately partial) custom dir keeps its old semantics: that dir, then package defaults; the repo-root layer is never consulted. New unit test + validation-script check cover the no-padding rule; the Windows stale-POSIX target case still recovers because its resolved directory is missing.
4. **Codex (CHANGELOG)**: added under `## [Unreleased]` as PR #875 in the second commit (the review ran against the first commit only).
5. **Codex (doc freshness)**: `Last updated` bumped to 2026-08-31.
6. **CodeQL (empty except)**: the diagnostic `except` now carries an inline comment, and the search-path rescue probe uses `os.path.isdir` (never raises) instead of a bare pass.

Re-verified after the changes: backend suite green, `bash test/run_pr_validation.sh 875` green (10 checks), `ruff` clean on changed files, full `ps_agent_start.ps1` parses under PowerShell 7.4.7, and `Build-Frontend` remains token-identical (98 tokens) to the intended body.

PR #883, opened 2026-09-02
feat: canvas width and orientation controls (#754)

### LINKED ISSUE #754: canvas width options
add button to enable canvas to take full width and then jump back to half or none. ... helpful when viewing docs on mobile where the split screen makes both chat and canvas to narrow

garland3:
Closes #754.

## Problem

The canvas opened beside the chat at a fixed ~half width with no way to change it. On a phone that split leaves both the chat and the document too narrow to read, which is exactly what the issue reports. The follow-up comments ask for a vertical (canvas-above-chat) option on desktop too, and for the choice to persist in browser storage per device.

## What changed

Two controls in the canvas header, next to the existing close button:

| Control | Behavior |
| --- | --- |
| Expand / Shrink | half <-> full. At full the canvas takes the whole content area and the chat is hidden. |
| Move | beside the chat <-> above the chat (stacked). |
| Close (existing) | the "none" state. |

Details worth calling out:

- **The chat is hidden, not unmounted**, at full size (`hidden` class rather than a conditional render), so an in-progress draft survives expanding and shrinking.
- **Below 768px the layout is forced to stacked** and the Move control is hidden, since a horizontal split at that width is the complaint in the issue. The *stored* preference is deliberately not overwritten, so widening the window (or rotating a tablet back to landscape) restores the user's side-by-side choice. `useCanvasLayout` returns both `orientation` (stored) and `effectiveOrientation` (what renders).
- **Persistence** is per browser via the existing `usePersistentState` helper (`chatui-canvas-size`, `chatui-canvas-orientation`), with unrecognized stored values falling back to the defaults. Phone and desktop therefore keep separate preferences, as requested in the issue thread.
- **Reopening from "none"** is unchanged: the canvas button in the app header, or new canvas content arriving from a tool.
- The drag-to-resize handle is suppressed when it cannot do anything (full size, stacked, or mobile).

## Verification

- `npx vitest run` - 64 files, 710 tests pass, including 7 new ones in `frontend/src/test/canvas-layout.test.jsx` covering defaults, full-size and stacked rendering, the header callbacks, localStorage persistence and round-trip, bad stored values, the narrow-viewport override plus its restore on resize, and the hidden Move control.
- `npm run lint` - 0 errors (4 pre-existing warnings, untouched files).
- `npm run build` - clean.
- `bash scripts/check-docs.sh` - OK, new doc indexed.

## Docs

New user-guide page `docs/user-guide/canvas-layout/README.md`, indexed in `docs/user-guide/README.md`, plus a CHANGELOG entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### PR #TBD - 2026-09-02
+- **The canvas can now take the full window or stack above the chat** (#754): the canvas header gains an expand/shrink control (half <-> full width, with the chat hidden at full) and a move control (beside the chat <-> above the chat), alongside the existing close button. Below 768px the layout is forced to stacked -- a side-by-side split is unreadable on a phone -- and the move control is hidden without overwriting the stored preference, so widening the window restores it. Both choices persist per browser in local storage.
+
 ### PR #876 - 2026-09-01
 - **The Active Tools strip's "+N more" / "Show less" toggle expands visibly again**: making the strip a single row (#870) turned the pill row into `flex-nowrap` + `overflow-x-auto`, so expanding it only lengthened a row that already scrolled horizontally -- the extra pills rendered off-screen and clicking "+25 more" looked like it did nothing. Expanded, the row now wraps into a height-capped (`max-h-24`), vertically scrolling block; collapsed, it stays the single horizontally scrolling row from #870. The toggle also exposes `aria-expanded`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
-### PR #TBD - 2026-09-02
+### PR #883 - 2026-09-02
 - **The canvas can now take the full window or stack above the chat** (#754): the canvas header gains an expand/shrink control (half <-> full width, with the chat hidden at full) and a move control (beside the chat <-> above the chat), alongside the existing close button. Below 768px the layout is forced to stacked -- a side-by-side split is unreadable on a phone -- and the move control is hidden without overwriting the stored preference, so widening the window restores it. Both choices persist per browser in local storage.
 
 ### PR #876 - 2026-09-01
--- a/docs/user-guide/README.md
+++ b/docs/user-guide/README.md
@@ -7,6 +7,7 @@ first-run setup, see [getting-started/](../getting-started/README.md).
 
 ## Features
 
+- [Canvas Layout Controls](canvas-layout/README.md) - Expand the canvas to full width, stack it above the chat, and keep the choice per device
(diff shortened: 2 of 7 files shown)


garland3:
### Verified against a running instance

Built the branch, served it from the backend (bundle hash confirmed against the local build so nothing stale was being served), and drove the layout controls in Chromium. Measurements of the canvas panel's bounding box and whether the chat composer is visible:

| Step | size / orientation | canvas w x h | chat input visible |
| --- | --- | --- | --- |
| open (defaults), 1400px viewport | half / right | 700 x 831 | yes |
| click Expand | full / right | 1144 x 831 | no |
| click Shrink | half / right | 700 x 831 | yes |
| click Move | half / top | 1144 x 416 | yes |
| reload + reopen | half / top | 1144 x 416 | yes |
| resize to 390px | half / **top** (forced) | 390 x 396 | yes |
| click Expand at 390px | full / top | 390 x 791 | no |

`localStorage` after the Move click: `{"chatui-canvas-size":"\"half\"","chatui-canvas-orientation":"\"top\""}`, and the stacked layout came back after a full page reload.

At 390px the Move control is absent from the DOM (count 0), as intended -- the layout is locked to stacked there, and the stored `right` preference is left intact so widening the window restores side-by-side.

No console errors during the run.

garland3:
Thanks — all three addressed in 22b8bec.

- **Unused `MAX_WIDTH`** — removed. Worth noting it had already been dead on `main` before this branch; the point about it being computed once at module load is right, which is exactly why the live drag clamp recomputes `window.innerWidth * 0.9` inline. Nothing now reads the stale constant.
- **Duplicate `@testing-library/react` import** in the new test — consolidated into one.
- **`### PR #TBD` in the changelog** — this was already fixed in 7b71f2a (it now reads `### PR #883 - 2026-09-02`); the Codex review ran against 3c97537, before that commit landed.

Re-verified after the changes: 710 tests pass across 64 files, `npm run lint` 0 errors, build clean.

garland3:
## Summary
Adds a `useCanvasLayout` hook that owns persisted canvas size (half/full) and orientation (right/top), forces stacked layout below 768px, and hides the chat with a `hidden` class at full size so an in-progress draft survives expand/shrink.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. The change is well demonstrated for its size — 7 new tests plus a measured browser run covering each layout state, persistence, and the narrow-viewport lock; the only missing artifact is a rendered screenshot of the new full-size and stacked layouts.

## Notes
- Chat scroll is not re-anchored when the canvas leaves full size; `scrollIntoView`/`scrollHeight` are inert inside the `hidden` subtree, so the list stays parked at its old offset (`frontend/src/components/ChatArea.jsx:183`, `frontend/src/App.jsx:160`).
- The 768px breakpoint exists twice — `NARROW_VIEWPORT_PX` in the hook and a hardcoded `768` in `CanvasPanel`'s `checkMobile`; export the constant so both move together (`frontend/src/hooks/useCanvasLayout.js:11`, `frontend/src/components/CanvasPanel.jsx:82`).
- Persisted `full` size means a later auto-opened canvas hides the whole chat with no visible cue; consider session-scoping `size` or labeling the Shrink control (`frontend/src/hooks/useCanvasLayout.js:26`).

---

PR #886, opened 2026-09-04
Fix MCP prompt result text extraction

### LINKED ISSUE #783: MCP prompt override injects raw GetPromptResult repr instead of extracted text
## Summary

MCP prompts registered via custom/third-party MCP servers are followed much less closely by the model than prompts registered via the bundled `prompts` MCP server (`atlas/mcp/prompts/main.py`), even when the prompt text itself is equivalent. This is caused by an extraction bug, not intentional behavior — the bundled server's prompts happen to survive the bug better due to how their text is authored.

## Root cause

`PromptOverrideService._extract_prompt_text()` in `atlas/application/chat/preprocessors/prompt_override_service.py:90-118` assumes the object returned by `ToolManager.get_prompt()` (`atlas/modules/mcp_tools/mcp_execution.py:337-405`, a passthrough of the MCP client's `get_prompt()`) exposes a top-level `.content` attribute:

```python
if hasattr(prompt_obj, "content"):
    ...
# Fallback: string dump
return str(prompt_obj)
```

The actual MCP protocol response type for `prompts/get` is `mcp.types.GetPromptResult`, which has a `.messages` list (each a `PromptMessage` with its own `.content`) — there is no `.content` on the result itself. Confirmed via SDK source (`mcp/types.py`) and by running the bundled `prompts` server through a real `fastmcp.Client`:

```
meta=None description='...' messages=[PromptMessage(role='user', content=TextContent(type='text', text='System: You are a financial technology wizard...\n\nUser: Please adopt this personality...', annotations=None, meta=None))]
```

Because `hasattr(prompt_obj, "content")` is always `False`, extraction always falls through to `str(prompt_obj)`, and this raw Pydantic repr — not the clean prompt text — gets wrapped as a `{"role": "system", ...}` message and prepended to the conversation (same file, lines 74-86). This happens identically for every MCP server; there is no server-name special-casing anywhere in the codebase (checked `mcp_execution.py`, `mcp_discovery.py`, `orchestrator.py`, `service.py`).

## Why the bundled `prompts` server "works" anyway

Its prompt bodies (`atlas/mcp/prompts/main.py`, e.g. `financial_tech_wizard` at lines 16-26, `expert_dog_trainer` at 30-41, `creative_writer` at 44-56) are hand-authored with literal `"System: ..."` / `"User: Please adopt this ... for our conversation."` role-cue text baked directly into the returned string. Even buried inside the garbled repr dump, these markers still read as a salient directive to the model and survive reasonably well.

A typical custom server's prompt, written the normal MCP way (plain instruction text, no fake role headers), gets the same repr-dump treatment but has no salient marker to stand out from the surrounding `PromptMessage(role='user', content=TextContent(...), annotations=None, meta=None)` noise, so the model weights it far less. The apparent "bundled server prompts are followed more closely" effect is incidental compensation for this bug via prompt-text formatting, not a designed difference in handling.

## Suggested fix

In `_extract_prompt_text`, extract from `prompt_obj.messages[i].content.text` (joining all message contents in order) instead of checking `.content` on the top-level result.

## Repro

```python
from fastmcp import Client
from atlas.mcp.prompts.main import mcp
from atlas.application.chat.preprocessors.prompt_override_service import PromptOverrideService

async with Client(mcp) as client:
    result = await client.get_prompt("financial_tech_wizard")
    svc = PromptOverrideService(tool_manager=None)
    print(svc._extract_prompt_text(result))  # prints the raw GetPromptResult repr, not clean text
```

mcc0nnell:
## Summary

- extract MCP prompt text from `GetPromptResult.messages[*].content`
- preserve message order when joining text content
- skip structured prompt results with no text instead of injecting their raw repr
- retain the existing legacy `.content` extraction path

## Tests

- adds a regression test using real MCP `GetPromptResult`, `PromptMessage`, and `TextContent` types
- adds coverage ensuring non-text structured prompt results return `None`

Fixes #783

### DIFF
--- a/atlas/application/chat/preprocessors/prompt_override_service.py
+++ b/atlas/application/chat/preprocessors/prompt_override_service.py
@@ -122,6 +122,21 @@ class PromptOverrideService:
         if isinstance(prompt_obj, str):
             return prompt_obj
 
+        # MCP prompts/get returns a GetPromptResult whose text lives under
+        # messages[i].content, not on the result itself.
+        if hasattr(prompt_obj, "messages"):
+            messages = getattr(prompt_obj, "messages")
+            if isinstance(messages, list):
+                texts = []
+                for message in messages:
+                    content = getattr(message, "content", None)
+                    content_items = content if isinstance(content, list) else [content]
+                    for item in content_items:
+                        text = getattr(item, "text", None)
+                        if isinstance(text, str):
+                            texts.append(text)
+                return "\n".join(texts) if texts else None
+
         # FastMCP PromptMessage-like: may have 'content' list with text entries
         if hasattr(prompt_obj, "content"):
             content_field = getattr(prompt_obj, "content")
--- a/atlas/tests/test_multi_prompt.py
+++ b/atlas/tests/test_multi_prompt.py
@@ -147,6 +147,35 @@ class TestPromptTextExtraction:
     def test_extract_string(self, service):
         assert service._extract_prompt_text("hello") == "hello"
 
+    def test_extract_get_prompt_result_messages(self, service):
+        """Extracts text from the actual MCP prompts/get response shape."""
+        from mcp.types import GetPromptResult, PromptMessage, TextContent
+
+        prompt_obj = GetPromptResult(
+            messages=[
+                PromptMessage(
+                    role="user",
+                    content=TextContent(type="text", text="Part 1."),
+                ),
+                PromptMessage(
+                    role="assistant",
+                    content=TextContent(type="text", text="Part 2."),
+                ),
+            ]
+        )
+
+        assert service._extract_prompt_text(prompt_obj) == "Part 1.\nPart 2."
+
+    def test_get_prompt_result_without_text_does_not_inject_repr(self, service):
+        """Structured MCP prompt results without text are skipped, not stringified."""
+        from types import SimpleNamespace
+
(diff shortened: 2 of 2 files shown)


garland3:
## Summary
Adds a branch to `_extract_prompt_text` that pulls text out of `GetPromptResult.messages[*].content`, joins it with newlines, and returns `None` (instead of a raw object repr) when a structured result carries no text. Two unit tests cover the new extraction path.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — the change is small and its four stated behaviors each map to an assertion, using real `mcp.types` classes rather than stubs. The one gap is any confirmation that the originally reported symptom (#783) is gone against a real `prompts/get` call.

## Notes
- Add an `else: logger.warning(...)` after `if prompt_text:` at `atlas/application/chat/preprocessors/prompt_override_service.py:76` — a prompt that resolves to no text is now dropped with no log line and no UI signal.
- `messages` present but not a `list` falls through to the legacy `.content` path and then to `str(prompt_obj)`, reintroducing the repr injection this PR removes; pin the intended behavior at `prompt_override_service.py:127`.
- The list-valued and mixed text/non-text `content` cases are only exercised indirectly; one assertion in `atlas/tests/test_multi_prompt.py:150` would cover the normalization branch.

---

PR #890, opened 2026-09-04
fix(ui): make the Data Sources tab keyboard-operable and scroll the opened tab into view

garland3:
Two follow-ups to #889, which moved the data source picker out of the Tools & Integrations tab into its own **Data Sources** tab. Both were found by driving the merged main build in a browser.

## 1. The Data Sources tab was not keyboard-operable

Source rows were click-only `<div>`s with no `role`, `tabindex`, or key handler. That was survivable in the left-hand Sources drawer, but the new tab promotes the picker into the Tools and Settings modal — a focus-trapped ARIA tabs pattern whose every other tab (Tools, Prompts, General, Admin) is fully keyboard-operable.

Measured on `main` with the panel open on the Data Sources tab:

```
rowCount: 5, rowTag: "DIV", rowRole: null, rowTabIndex: null
focusableInPanel: ["INPUT:Search data sources...", "BUTTON:Enable All", "BUTTON:Clear All"]
```

A keyboard user could reach the tab and then not toggle a single source — "Enable All" / "Clear All" were the only way in, and they are all-or-nothing.

The rows are now real `<button type="button">` toggles carrying `aria-pressed`. Out-of-boundary rows use `aria-disabled` rather than the `disabled` attribute, so the compliance-boundary explanation in their `title` stays discoverable, and an already-selected out-of-boundary row remains deselectable exactly as before. The drawer picks this up too, since both render the same `DataSourcesSelector`.

## 2. The tab strip did not scroll the opened tab into view

The strip is `overflow-x-auto`, and opening the panel straight onto a tab (`initialTab` / the `atlas:open-settings` event) only moved `activeTab` — nothing scrolled. Adding a sixth tab in second position pushed General, User Info and Admin past the right edge on a narrow window.

At a 400px viewport, dispatching `atlas:open-settings` with `{tab:'admin'}` on `main`:

```
scrollLeft: 0, selected: "Admin", selLeft: 590, selRight: 689, stripRight: 400, offscreen: true
```

The Admin panel rendered under a strip parked at the far left with no tab visibly selected. After the fix, the same probe returns `scrollLeft: 289, offscreen: false`. Arrow-key navigation was already fine — `focus()` scrolls on its own — so only the programmatic open path changed. The call is guarded because jsdom does not implement `scrollIntoView`.

## Verification

- `npx vitest run` — **790 passed / 70 files** (783 on `main` before this change, plus the 7 new tests).
- `npx eslint` clean on the three touched files.
- New `frontend/src/test/data-sources-tab-keyboard.test.jsx` covers `aria-pressed` on/off, keyboard activation, the `aria-disabled` out-of-boundary row staying focusable and inert, the selected out-of-boundary row staying deselectable, and `scrollIntoView` firing for both the opened tab and a clicked one.
- Live browser (Playwright, dev server against a local backend): the Data Sources tab now exposes **8** focusable elements instead of 3; focusing the first row and pressing <kbd>Enter</kbd> flips it to `aria-pressed="true"`; the Sources drawer renders 5 row buttons with matching state; and the tab renders unchanged visually at 1400px and scrolls correctly at 400px with no horizontal page overflow.

Also re-exercised and found working on `main`, unchanged here: the `+N more` dataset pill strip, the "Data Sources:" chat-bar shortcut routing to the new tab, selection persisting across panel close/reopen, and the unsaved-tool-changes guard still firing when the panel is closed from the Data Sources tab.

https://claude.ai/code/session_012ie5mYAwYzq9ahbq9WXAY6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### PR #890 - 2026-09-04
+- **Follow-ups to the Data Sources tab** (#889): the source rows are real toggle buttons (`aria-pressed`) instead of click-only `<div>`s, so the picker is operable from the keyboard -- promoting it into a focus-trapped modal tab had left every source unreachable by Tab, with "Enable All"/"Clear All" the only way in. Out-of-boundary rows stay focusable via `aria-disabled` so their explanation is still discoverable. The tab strip also scrolls the selected tab into view when the panel opens straight onto one; adding a sixth tab in second position had pushed General, User Info and Admin past the right edge on a narrow window, so those opened with the strip parked at the far left and no tab visibly selected.
+
 ### PR #888 - 2026-09-04
 - Move data source selection into its own second tab in Tools and Settings instead of combining it with MCP tool selection.
 
--- a/frontend/src/components/DataSourcesSelector.jsx
+++ b/frontend/src/components/DataSourcesSelector.jsx
@@ -190,15 +190,26 @@ const DataSourcesSelector = () => {
               const clickable = !outOfBoundary || isSelected
 
               return (
-                <div
+                // A real <button> with aria-pressed, not a click-only <div>:
+                // the picker is now a tab of the Tools and Settings modal,
+                // whose every other tab is fully keyboard-operable, and a
+                // plain div left the source rows unreachable by keyboard
+                // altogether -- Enable All / Clear All were the only way in.
+                // Out-of-boundary rows stay focusable via aria-disabled rather
+                // than the `disabled` attribute, so their explanatory title is
+                // still discoverable.
+                <button
+                  type="button"
                   key={selectionKey}
+                  aria-pressed={isSelected}
+                  aria-disabled={clickable ? undefined : true}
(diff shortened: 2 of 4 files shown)


garland3:
## Summary
Makes data-source rows keyboard-operable by rendering them as `<button type="button">` with `aria-pressed`/`aria-disabled`, and adds a `SettingsPanel` effect that scrolls the active tab into view when the panel opens or the tab changes. Adds a new test file covering the changed behavior.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — well demonstrated for a change this size, with before/after DOM probes for both defects and seven new tests. The one gap is that the "renders unchanged visually at 1400px" claim is narrated rather than backed by a screenshot, which matters because converting a row `<div>` to a `<button>` inherits user-agent styling.

## Notes
- Copy still says "Click to enable/disable" / "click to deselect" although rows are now keyboard-operable — use device-neutral wording (`frontend/src/components/DataSourcesSelector.jsx:171`).
- `aria-pressed` is emitted on out-of-boundary rows that cannot toggle and give no feedback on Enter; consider `aria-pressed={clickable ? isSelected : undefined}` (`frontend/src/components/DataSourcesSelector.jsx:203`).
- Test teardown assigns `Element.prototype.scrollIntoView = undefined` instead of deleting it; prefer `vi.spyOn` with restore (`frontend/src/test/data-sources-tab-keyboard.test.jsx:152`).

---

PR #892, opened 2026-09-05
feat(auth): OIDC login, confidential-client authentication, and delegated OAuth credentials (closes #891)

### LINKED ISSUE #891: “Support OIDC login, confidential-client authentication, and delegated OAuth credentials
A parallel deployment approach with the existing reverse proxy with auth separate and this new where the atlas app is the oauth app. 

This can be used with or with out the ngnix.

Atlas itself is an oauth app. This enables doing token key exchange for mCP oauth

garland3:
Closes #891.

Adds a second, opt-in deployment model alongside the existing trusted-header auth mode: **Atlas can be the OAuth application itself**, terminating login and minting delegated downstream credentials, with or without an authenticating reverse proxy in front. Header auth is untouched and stays the default.

| | Trusted-header mode (existing) | OIDC mode (new) |
| --- | --- | --- |
| Who terminates login | Reverse proxy (nginx, ALB) | Atlas |
| How Atlas learns the user | `X-User-Email` / ALB JWT | Atlas's own session cookie |
| Reverse proxy required | Yes | No |
| Selected by | Default | `FEATURE_OIDC_AUTH_ENABLED=true` |

The two coexist: a request with a live Atlas session authenticates on that basis, anything else falls through to the header path unchanged, so a mixed rollout does not break existing traffic.

## The three capabilities from the issue

**1. OIDC login via Authorization Code + PKCE.** `GET /auth/oidc/login` → IdP → `GET /auth/oidc/callback` → Atlas session. Endpoints come from OIDC discovery, with the document's own `issuer` claim checked so a redirect on the discovery URL cannot substitute another provider's endpoints. `state` and `nonce` are single-use and cleared before the code is redeemed. **The browser holds only an opaque session id** in a signed cookie; access tokens, refresh tokens and validated ID token claims stay in a server-side store inside the Atlas credential boundary.

**2. Atlas as a confidential OAuth client.** `OIDC_CLIENT_AUTH_METHOD` selects `client_secret_basic`, `client_secret_post`, or `private_key_jwt` — an RFC 7523 assertion audience-bound to the token endpoint, five-minute lifetime, fresh `jti` each time. The private key is read once, cached in memory, and only ever used to sign.

**3. Delegated downstream authorization**, as named interchangeable providers so a deployment picks one by config and a third can be added without touching call sites:
- `token_exchange` — RFC 8693 OAuth Token Exchange.
- `entra_obo` — Microsoft Entra ID On-Behalf-Of (JWT bearer grant + `requested_token_use=on_behalf_of`; a configured audience becomes `<audience>/.default` when no scope is given).

Atlas **never forwards the user's inbound token** to a downstream service — it exchanges it for a short-lived, audience-specific, minimally-scoped one, cached per (user, audience, scope) and re-minted near expiry. A token with no advertised expiry is never cached across calls. Logout discards the user's cached delegated tokens.

An MCP server configured with `auth_type: "delegated"` gets its credential this way instead of asking the user to upload one; `audience` defaults to the server URL, the canonical resource identifier in the MCP authorization spec:

```json
"protected-tools": {
  "url": "https://tools.example.gov/mcp",
  "auth_type": "delegated",
  "delegation": { "audience": "api://tools.example.gov", "scope": "tools.read" }
}
```

If delegation is off, the user has no OIDC session, or the exchange fails, the server reports as unauthenticated — exactly as an unauthenticated bearer server does today. A tool call never fails with a delegation stack trace.

## Two changes worth reviewer attention

**Session middleware moved after `AuthMiddleware`.** Starlette runs the most recently added middleware outermost, so the Globus-only registration sat *inside* `AuthMiddleware` — `request.session` was unavailable to authentication and readable only in routes that skip auth entirely. Registered outside, both HTTP and the chat WebSocket resolve identity from the login session. Globus is unaffected (its routes still see the session).

**A live OIDC session skips the `PROXY_SECRET` gate.** That gate exists to stop a caller who reaches the backend directly from spoofing an identity *header*. A session cannot be spoofed that way: the cookie is signed and its id only resolves against this process's session store. This is what lets OIDC mode run with no reverse proxy at all. Requests without a session face the gate unchanged.

## Verification

- Full backend suite: **2786 passed, 17 skipped** (`bash test/atlas_tests.sh`) — 58 of those are new.
- New tests cover discovery validation (issuer mismatch, missing endpoints, PKCE advertisement), the PKCE challenge actually being S256 of the verifier, all three client-auth methods including verifying a real signed assertion against its public key, the session store (expiry, eviction, refresh-token preservation, no token material in the status projection), both delegation providers' wire format and the token cache, the routes including open-redirect handling on `?next=` and error-code reflection, and `AuthMiddleware` resolving identity from a session.
- End-to-end against the real `atlas.main` app with OIDC enabled and only the IdP calls stubbed: login redirect carries `code_challenge_method=S256`, the callback establishes a session, `/api/config/shell` then returns 200 **with no proxy secret and no identity header**, a fresh client gets 401 on the API and a redirect to `/auth/oidc/login` on a browser path.
- `ruff check` clean on every changed file; `scripts/check-docs.sh` OK.

## Not included

No frontend surface. Login and logout are server-side redirects, so the flow works without UI changes, but a signed-in user has no in-app logout link yet — `/api/config` now exposes an `oidc_auth` feature flag so that can follow. Session state is in-process, which the docs call out (single worker, or use header mode behind a proxy that holds the session).

Docs: `docs/admin/oidc-authentication.md`, linked from the admin index and cross-referenced from `authentication.md`.

https://claude.ai/code/session_01PUT7cT32HbmBKhCXSdoqEg


garland3:
## Summary
Adds first-class OIDC authentication: discovery, three client-authentication methods, a server-side session store backing an opaque session cookie, HTTP and WebSocket identity resolution, and RFC 8693 / Entra On-Behalf-Of token delegation for MCP servers marked `auth_type: delegated`.

## Verdict
Request changes.

## Required changes
1. Reject backslash- and control-character-prefixed values in `_safe_return_to` at `atlas/routes/oidc_auth_routes.py:79` (guard at `:118`) — `?next=/\evil.example` normalizes to a protocol-relative URL in browsers, giving a post-login open redirect; prefer `urlsplit` with empty scheme and netloc, and add `/\evil.example` and `/%09/evil.example` cases to `test_login_next_parameter_cannot_be_an_open_redirect`.
2. Drive `SessionMiddleware(https_only=...)` from configuration and default it to `True` when OIDC auth is enabled, at `atlas/main.py:365` — the cookie is now the sole authenticator for both HTTP and the chat socket and is currently sent over plaintext HTTP.
3. Bind delegation to the OIDC session rather than the email string in `_find_subject_token` at `atlas/core/oidc/mcp_delegation.py:48` — in mixed mode a request authenticated only by `X-User-Email` selects and spends another user's OIDC access token; at minimum refuse delegation when `request.state.auth_source != "oidc"` and include the session id in the cache key.
4. Preserve the Globus session lifetime at `atlas/main.py:340` — Globus-only deployments now inherit the 8-hour OIDC session lifetime in place of the previous 14 days, logging existing users out early.
5. Replace the single manager-wide `asyncio.Lock` held across `provider.exchange()` at `atlas/core/oidc/delegation.py:263` with per-cache-key locking — one slow IdP exchange serialises every delegated tool call process-wide behind a 20s timeout.
6. Add WebSocket authentication tests for `_resolve_oidc_websocket_user` (`atlas/main.py:520`, guard at `:598-625`) — cover live session connects without a proxy secret, expired session rejected, feature flag off leaves the proxy-secret gate intact, and header path still wins with no session; this is the one path in the PR that both suppresses the proxy-secret check and has no coverage.
7. Test `_mint_delegated_token` at `atlas/modules/mcp_tools/mcp_user_clients.py:60-108` — assert the header the downstream transport is built with and that a token with no advertised expiry is stored with a bounded `expires_at`.

## Evidence
Evidence: 8/10 — the change is well demonstrated for its size, with a full-suite result, 58 enumerated new tests, and an end-to-end run against the real application object. The remaining gap is that every identity-provider interaction is stubbed; a redacted transcript of one login against a local Keycloak (or equivalent) would validate discovery, client auth, and callback together.

## Notes
- Session-to-identity resolution is copied three times (`atlas/core/middleware.py:57`, `atlas/main.py:520`, `atlas/routes/oidc_auth_routes.py:262`) — extract one helper in `atlas/core/oidc/session.py`.
- RP-initiated logout omits `id_token_hint` at `atlas/routes/oidc_auth_routes.py:252`, so most providers ignore the end-session request; logout is also a state-changing GET.
- Login failures redirect to a query parameter no frontend reads and misconfiguration surfaces as raw JSON (`atlas/routes/oidc_auth_routes.py:67`, `:102`).

---

garland3:
## Summary
Adds a full OIDC authentication mode (login/callback/logout routes, a session store and cookie-backed middleware, ID-token validation, and token-exchange delegation for MCP servers) behind `FEATURE_OIDC_AUTH_ENABLED`/`FEATURE_OIDC_DELEGATION_ENABLED`, alongside the existing trusted-header path. 21 files, ~+3200 lines, backend only.

## Verdict
Block — critical issues found.

## Required changes
1. Resolve the delegation subject token from the authenticated session id, not the email string, at `atlas/core/oidc/mcp_delegation.py:48` — `_find_subject_token` scans all live sessions and matches `user_id`, so in mixed mode a header-authenticated (`X-User-Email`) request selects and spends another user's OIDC access token. Refuse delegation when `request.state.auth_source != "oidc"`, and include the session id in `DelegationRequest.cache_key()` so cached downstream tokens are not shared across auth sources.
2. Drive `https_only` from configuration at `atlas/main.py:365` (default true when `FEATURE_OIDC_AUTH_ENABLED`) and state `httponly=True` explicitly — the session cookie is now the sole authenticator for HTTP and the chat WebSocket and is currently sent in cleartext.
3. Require `email_verified is True` when the identity claim is `email` at `atlas/core/oidc/oidc_client.py:243`, and make the `preferred_username`/`sub` fallback opt-in — otherwise a self-registered account with a privileged user's address inherits that user's groups and stored credentials.
4. Tighten `_safe_return_to` at `atlas/routes/oidc_auth_routes.py:79` to reject `/\evil.example` and `/%09/evil.example` (validate with `urlsplit`, requiring empty scheme and netloc) — browsers normalize the backslash into a protocol-relative redirect.
5. Refresh expired access tokens before delegating, at `atlas/core/oidc/session.py:62` — `refresh_access_token`/`update_tokens` are defined but never called, so with an 8h session and a 1h IdP token every delegated exchange fails for the rest of the session. Guard the refresh with a per-session lock.
6. Only assign `access_token`/`access_token_expires_at` when supplied in `OIDCSessionStore.update_tokens` (`atlas/core/oidc/session.py:150`) — the current unconditional assignment silently clears the token when a caller updates only the refresh token, permanently disabling delegation for that session.
7. Replace the manager-wide `asyncio.Lock` in `DelegationManager.get_token` (`atlas/core/oidc/delegation.py:263`) with a per-cache-key lock — the single lock is held across a 20s IdP POST, serializing every unrelated delegated tool call in the process.

## Evidence
Evidence: 7/10 — strongly demonstrated for its size (58 new tests, full-suite pass, an end-to-end run against the real app with only IdP calls stubbed). The most important gap is the WebSocket authentication handshake in `atlas/main.py:533`, which has no test at all.

## Notes
- `SessionMiddleware` now applies `oidc_session_max_age_seconds` unconditionally, shortening Globus-only cookie lifetime from 14 days to 8h — `atlas/main.py:340`.
- `/auth/oidc/login` returns raw JSON naming internal env vars when misconfigured, unlike every other failure path in the module — `atlas/routes/oidc_auth_routes.py:102`.
- docker-compose exposes the feature flags but none of the mandatory `OIDC_*` settings, so the container starts silently in header mode — `docker-compose.yml:129`.

---

garland3:
## Summary
Adds an OIDC authentication mode — discovery, authorization-code/PKCE login, a process-local session store, session-based identity for HTTP and the chat WebSocket — plus RFC 8693 token delegation for MCP servers, behind `FEATURE_OIDC_AUTH_ENABLED` / `FEATURE_OIDC_DELEGATION_ENABLED`.

## Verdict
Block — critical issues found.

## Required changes
1. Bind delegation to the authenticating session at `atlas/core/oidc/mcp_delegation.py:48` — `_find_subject_token` scans all live sessions by email, so in mixed mode a request authenticated only by `X-User-Email` spends a different login's OIDC access token; look the session up by id, refuse delegation unless `request.state.auth_source == "oidc"`, and include the session id in `DelegationRequest.cache_key()` so cached downstream tokens cannot cross sessions or auth sources.
2. Make the session cookie `Secure` at `atlas/main.py:365` — `https_only=False` is hardcoded while the cookie is now the sole authenticator for HTTP and `/ws` and suppresses the proxy-secret gate; add an `OIDC_SESSION_COOKIE_SECURE` setting defaulting to true when OIDC is enabled, and state `httponly=True` explicitly.
3. Tighten `_safe_return_to` at `atlas/routes/oidc_auth_routes.py:79` — it accepts `/\evil.example` and `/%09/evil.example`, which browsers normalize into an off-site post-login redirect; validate with `urlsplit` requiring empty scheme and netloc, and add those payloads to the open-redirect test.
4. Require a verified email in `extract_user_identifier` at `atlas/core/oidc/oidc_client.py:243` — the `email` claim is trusted without `email_verified` and falls back to `preferred_username`/`sub`, so an account registered with a privileged user's address inherits that user's groups and stored credentials; make the fallback opt-in.
5. Wire token refresh before delegation at `atlas/core/oidc/session.py:62` — `access_token_needs_refresh` / `refresh_access_token` have no caller, so with an 8h session and a 1h IdP token every delegated exchange fails for most of the session; refresh under a per-session lock and drop the session when refresh fails.
6. Stop clobbering the Globus cookie lifetime at `atlas/main.py:340` — `max_age` is unconditionally `oidc_session_max_age_seconds`, cutting existing Globus-only deployments from Starlette's 14-day default to 8 hours (and re-signing with the OIDC secret when both are on); select lifetime and secret together.
7. Add WebSocket session-auth tests for `_resolve_oidc_websocket_user` (`atlas/main.py:533`, guard at `:611-625`) — cover live session with no proxy secret, expired session rejected, feature flag off still enforcing the proxy-secret gate, and session-vs-header precedence; this is the only path that both resolves identity and suppresses the proxy-secret check, and nothing exercises it.

## Evidence
Evidence: 8/10 — the change is large but well-demonstrated for its size, with a full-suite result, 58 new tests matching the described claims, and an end-to-end run against the real app object with the identity provider stubbed. The most valuable missing artifact is a transcript of one real login and one delegated exchange against a live provider (e.g. local Keycloak), since every IdP interaction is currently simulated.

---

garland3:
## Summary
Adds a generic OIDC authentication path — discovery, PKCE login/callback routes, an in-process session store, middleware and WebSocket identity resolution — plus token-exchange delegation so MCP servers receive audience-bound downstream tokens instead of the user's inbound token.

## Verdict
Block — critical issues found.

## Required changes
1. Resolve the delegation subject token by session id, not by email, at `atlas/core/oidc/mcp_delegation.py:48` — a request authenticated only by `X-User-Email` (or ALB/debug path) currently matches and spends a genuinely logged-in user's OIDC access token. Thread the session id from `request.state` through `mint_delegated_token_for_server`, refuse delegation unless `request.state.auth_source == "oidc"`, and include the session id in `DelegationRequest.cache_key()` so a cached downstream credential cannot be served to a different session or auth source.
2. Tighten `_safe_return_to` at `atlas/routes/oidc_auth_routes.py:79` — `/\evil.example` and `/%09/evil.example` pass the regex and browsers normalize them into protocol-relative redirects. Use `urlsplit` and require empty scheme and netloc; add both payloads to the open-redirect test table.
3. Stop hardcoding `https_only=False` at `atlas/main.py:365` — the session cookie is now the sole authenticator for HTTP and the chat WebSocket, so it must not travel over plaintext. Add an `OIDC_SESSION_COOKIE_SECURE` setting defaulting to true when OIDC is enabled, and set `httponly=True` explicitly.
4. Require `email_verified is True` when the configured identity claim is `email` at `atlas/core/oidc/oidc_client.py:243`, and make the `preferred_username`/`sub` fallback opt-in — otherwise an IdP account carrying an unverified privileged address inherits that user's authorizations.
5. Wire up the access-token refresh path (`atlas/core/oidc/session.py:58`) — `access_token_needs_refresh`, `update_tokens`, and `refresh_access_token` have no callers, so with an 8h session and a 1h token, delegation silently fails as "unauthenticated" for most of a login. Refresh before delegating, under a per-session lock, and drop the session when the refresh is rejected.
6. Only assign `access_token` and `access_token_expires_at` when supplied in `update_tokens` at `atlas/core/oidc/session.py:150` — a partial update currently wipes the subject token or its expiry (permanently disabling refresh). Add a test asserting a refresh-token-only update leaves both intact.
7. Decide session lifetime and signing secret together at `atlas/main.py:340` — a Globus-only deployment now inherits the 8h OIDC `max_age`, and enabling both flags re-signs the Globus cookie with the OIDC secret, invalidating every cookie in flight. Add a regression test for the Globus-only branch.
8. Add WebSocket authentication tests for `_resolve_oidc_websocket_user` (`atlas/main.py:533`, guard at :611-625) mirroring `TestMiddlewareOIDCSession`: live session connecting without a proxy secret, expired session refused, and feature flag off leaving the proxy-secret gate intact.

## Evidence
Evidence: 8/10 — the change is well demonstrated for its size, with two new test modules covering discovery, PKCE, all three client-auth methods, the session store, both delegation providers, and the login/callback routes, plus a full-suite run. The most important gap is that the WebSocket authentication entry point, which this PR both adds identity resolution to and lets bypass the proxy-secret check, has no test at all.

## Notes
- `DelegationManager.get_token` holds one process-wide lock across a 20s IdP round trip, serializing unrelated users and audiences — `atlas/core/oidc/delegation.py:263`.
- Delegated tokens persist to MCP token storage and survive logout, which only clears the in-memory cache — `atlas/modules/mcp_tools/mcp_user_clients.py:60`.
- `request.state.auth_source` is set only on the OIDC branch, so the obvious consumer raises `AttributeError` on header/debug requests — `atlas/core/middleware.py:142`.

---

garland3:
## Summary
Adds a first-class OIDC authentication mode to Atlas — discovery, PKCE login/callback, a server-side session store, session-based HTTP and WebSocket identity resolution, and RFC 8693 token delegation for MCP servers — alongside the existing Globus and proxy-header paths.

## Verdict
Block — critical issues found.

## Required changes
1. Resolve the delegation subject token by authenticating session id, not by email, at `atlas/core/oidc/mcp_delegation.py:48` — `_find_subject_token` scans all live sessions, so a request authenticated only by the trusted `X-User-Email` header spends a genuinely logged-in user's OIDC access token and gains their delegated authority downstream. Thread the session id from `request.state` and use `store.get(session_id)`; refuse delegation unless `getattr(request.state, "auth_source", None) == "oidc"`.
2. Add the session id (and auth source) to `DelegationRequest.cache_key()` in `atlas/core/oidc/delegation.py` — the key is currently `(user_id, audience, resource, scope)`, so a downstream credential minted for one login is served from cache to a different session or auth source for the same email.
3. Set the session cookie `Secure` at `atlas/main.py:365` — `SessionMiddleware` is registered with `https_only=False` and an implicit `httponly`, and the cookie is now the sole authenticator for HTTP and the chat WebSocket and suppresses the `PROXY_SECRET` gate. Add an `OIDC_SESSION_COOKIE_SECURE` setting defaulting to true when OIDC is enabled, pass it as `https_only`, and state `httponly=True` explicitly.
4. Wire the access-token refresh path into delegation at `atlas/core/oidc/session.py:62` — `access_token_needs_refresh`, `update_tokens`, and `refresh_access_token` have no callers, and the 8h default session outlives typical 5–60min access tokens, so delegated MCP tools silently stop working partway through every session. Check the expiry before returning a subject token, refresh under a per-session lock, and drop the session when the IdP rejects the refresh.
5. Make `update_tokens` at `atlas/core/oidc/session.py:150` treat `access_token` and `access_token_expires_at` conditionally, as it already does for `refresh_token` — a refresh-token-only update currently erases the access token, leaving a session that authenticates but can never delegate.
6. Select the session secret and `max_age` in the same branch at `atlas/main.py:340` — Globus-only deployments now inherit `OIDC_SESSION_MAX_AGE_SECONDS` (8h instead of Starlette's 14-day default), and enabling OIDC on a Globus deployment re-signs the cookie with a different key, invalidating every in-flight session. Add a regression test for the Globus-only branch.
7. Validate `?next=` with `urlsplit` (empty scheme and netloc) at `atlas/routes/oidc_auth_routes.py:79` — `_safe_return_to` accepts `/\evil.example` and `/%09/evil.example`, which browsers normalize into protocol-relative open redirects. Add both payloads to the existing test table.
8. Add WebSocket authentication tests for `atlas/main.py:533` — `_resolve_oidc_websocket_user` both establishes identity and suppresses the proxy-secret gate with no coverage. Cover live session with no proxy secret, expired session, feature flag off, and session-vs-header precedence.

## Evidence
Evidence: 8/10 — a ~3200-line auth change backed by 58 new tests, matching wire-format assertions, and an end-to-end run against the real app; adequate for its size. The most important gap is that every identity-provider interaction is stubbed, so discovery-document shape, client-auth method acceptance, and `aud` handling are undemonstrated against a real provider.

## Notes
- Delegated tokens persisted in `token_storage` survive logout; only the in-memory cache is invalidated (`atlas/modules/mcp_tools/mcp_user_clients.py:60`).
- `email` claim is trusted without checking `email_verified`, with a silent `preferred_username`/`sub` fallback (`atlas/core/oidc/oidc_client.py:243`).
- A single manager-wide lock is held across the 20s IdP exchange, serialising all delegated tool calls (`atlas/core/oidc/delegation.py:263`).

---

garland3:
Thanks — worked through the review feedback. Four findings were real defects; I verified each against the code before fixing rather than taking the report at face value.

**Fixed**

- **`delegation` never reached the exchange** (Codex P1). Confirmed: `MCPServerConfig` is a plain `BaseModel` and `MCPToolManager` rebuilds each server from `model_dump()`, so Pydantic dropped the block the docs told operators to write. Every exchange silently fell back to the server URL as audience and sent no scope — exactly the case Entra rejects. Added a typed `DelegationConfig` field, plus a test asserting a round-tripped `model_dump()` still carries audience and scope.
- **No refresh path for the subject token** (Copilot + Codex P1). Confirmed: `access_token_needs_refresh()` / `refresh_access_token()` / `update_tokens()` existed with no caller. New `atlas/core/oidc/session_refresh.py` refreshes on demand before the exchange, with one in-flight refresh per session so concurrent tool calls cannot burn a rotating refresh token. A dead, unrefreshable token now yields no delegated credential instead of being presented downstream.
- **Revocation reached only one of three caches** (Codex P1). Confirmed: a minted credential also lives in `MCPTokenStorage` and in the MCP user-client cache, so logout and the DELETE endpoint left it usable. All three are cleared now; tokens the user uploaded themselves are untouched, with a test for that boundary.
- **Session cookie was never `Secure`** (Codex P2). Now derived from the redirect URI's scheme, with `OIDC_COOKIE_SECURE` to override so local http development still works.
- **Redirect dropped the original path** (Copilot). The unauthenticated-browser redirect now carries `?next=`; the login route re-validates it against its own allowlist, so nothing is trusted across the hop.
- **CHANGELOG entry too long** (Copilot). Trimmed to the 1–2 lines AGENTS.md asks for.
- **Missing PR validation script** (Codex P1). Added `test/pr-validation/test_pr892_oidc_login_and_delegation.sh`. It starts a real minimal OIDC provider on localhost — discovery document, JWKS, an RS256-signed ID token, and an RFC 8693 token-exchange endpoint — and drives the real Atlas app against it over HTTP. 26 checks: the login redirect's PKCE parameters, the callback, an API call authenticated by the session alone with no identity header and no proxy secret, a stranger still getting 401, the exchange carrying the configured audience and scope, revocation, and logout. Nothing inside Atlas is mocked.

All 14 CodeQL alerts are also resolved at the source (open-redirect guard rewritten as a positive allowlist with parametrised tests, log-injection barriers reworked, empty excepts and the unused global removed).

**Not fixed, documented instead**

Codex's point about discovery-time 401s for delegated servers is correct and I have written it into `docs/admin/oidc-authentication.md` rather than papering over it: tool discovery runs once at startup with the process-level client, before anyone has logged in, so there is no session to delegate from. A delegated server that also requires authorization on `initialize`/`tools/list` will register no tools. Delegated servers must currently allow unauthenticated discovery and enforce authorization on tool *invocation*. Per-user lazy discovery is the right fix and is a larger change than belongs in this PR — happy to open a follow-up issue if you agree.

Full suite: 2845 passed, 17 skipped.

garland3:
## Summary
Adds OIDC authentication to Atlas — discovery, PKCE login, an in-process session store with access-token refresh — and uses the resulting session to mint per-user delegated credentials (RFC 8693 token exchange / Entra OBO) for MCP servers, with revocation on logout.

## Verdict
Block — critical issues found.

## Required changes
1. Resolve the delegation subject token by session id, not email, at `atlas/core/oidc/mcp_delegation.py:48` — `_find_session` scans all live sessions by normalized email, so a request authenticated only by `X-User-Email` selects a genuinely logged-in user's OIDC session and exchanges their access token for a downstream credential carrying that user's authority. Thread the session id from `request.state` and use `store.get(session_id)`; refuse delegation when `request.state.auth_source != "oidc"`. Add a test asserting a header-authenticated request for an email that also has a live session gets no delegated token.
2. Include the session id in `DelegationRequest.cache_key()` at `atlas/core/oidc/delegation.py:57` — the key is `(email, audience, resource, scope)`, so a cache hit replays a minted downstream credential to any later requester presenting the same email and short-circuits the subject-token lookup entirely. Keep the user id as the key prefix so `invalidate_user` still works.
3. Reject backslash- and whitespace-prefixed return paths at `atlas/routes/oidc_auth_routes.py:79` — `_SAFE_RETURN_TO` accepts `/\evil.example` and `/%09/evil.example`, which browsers normalize into a protocol-relative post-login open redirect. Validate with `urlsplit` requiring empty scheme and netloc, and add both strings to the existing open-redirect test table.
4. Make the access-token/expiry write in `OIDCSessionStore.update_tokens` conditional at `atlas/core/oidc/session.py:150` — a refresh response lacking `expires_in` sets `access_token_expires_at=None`, after which `access_token_needs_refresh()` is permanently `False` and the token is never renewed again; a response lacking `access_token` nulls the token and makes the session undelegatable. Guard both fields as `refresh_token` already is, treat a missing `access_token` as a failed refresh, and substitute a short default TTL when the provider advertises no lifetime.
5. Default `access_token_expires_at` to a short TTL when the IdP omits `expires_in` at session creation (`atlas/core/oidc/session.py:62`, `atlas/routes/oidc_auth_routes.py` callback) — otherwise such sessions never refresh at all and every delegated exchange fails once the real token dies.
6. Select the session cookie's signing secret and `max_age` in the same branch at `atlas/main.py:340` — a Globus-only deployment now inherits `oidc_session_max_age_seconds` (8h) in place of Starlette's 14-day default, and enabling OIDC alongside Globus switches the signing secret so existing Globus cookies fail verification. Add a regression test for the Globus-only branch.
7. Make `https_only`/`httponly` explicit for the session cookie at `atlas/main.py:365` and add WebSocket coverage for `_resolve_oidc_websocket_user` (`atlas/main.py:543`) — the cookie is now the sole authenticator for both HTTP and `/ws` and suppresses the proxy-secret gate, yet the WebSocket path has no test at all. Mirror `TestMiddlewareOIDCSession`: live session with no proxy secret, expired session refused, feature flag off leaving the gate intact, and session-over-header precedence.

## Evidence
Evidence: 9/10 — the change is well-demonstrated for its size, with per-capability unit tests, a stated full-suite result, and a scripted end-to-end run of the real app against a locally-hosted identity provider. The main gaps are the untested WebSocket authentication path (above) and `entra_obo`, which is verified only at the wire-format level and never against a live endpoint.

## Notes
- Hold the delegation lock per cache key rather than process-wide — `atlas/core/oidc/delegation.py:263` serializes every user's exchange behind one 20s IdP round trip.
- `_refresh_locks` grows one lock per session forever; only the vanished-session path clears it — `atlas/core/oidc/session_refresh.py:31`.
- Login failures redirect to `/?oidc_error=<code>` but nothing in the frontend reads it, so users see a silent redirect loop — `atlas/routes/oidc_auth_routes.py:112`.

---

Issue #756, opened 2026-08-09
Support the OpenAI Responses API (/v1/responses) so reasoning models can use tools


garland3:
## Summary

Add support for the OpenAI Responses API (`/v1/responses`) as a per-model endpoint, alongside the existing Chat Completions (`/v1/chat/completions`) path.

Newer OpenAI reasoning models cannot use function tools on Chat Completions at all. The GPT-5.6 family (`gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`) rejects any request that combines tools with a non-`none` reasoning effort:

```
OpenAIException - Function tools with reasoning_effort are not supported for
gpt-5.6-luna in /v1/chat/completions. To use function tools, use /v1/responses
or set reasoning_effort to 'none'.
```

This fires even though ATLAS never sends `reasoning_effort` — there are no references to the parameter anywhere in the codebase. These models default to medium effort, so the incompatibility is triggered by the default. The result is that GPT-5.6 models are unusable in tools mode and agent mode; only plain chat works.

This is not specific to ATLAS. Other clients built on Chat Completions hit the same wall:

- https://github.com/BerriAI/litellm/issues/33221
- https://github.com/danny-avila/LibreChat/issues/14231
- https://community.openai.com/t/gpt-5-6-chat-completion-reasoning-effort-bug-behavior-change/1386454

OpenAI's own guidance is that reasoning models should use Responses, and that Chat Completions is the legacy path: https://developers.openai.com/api/docs/guides/reasoning

## Current behavior

`LiteLLMCaller` is built entirely on `litellm.acompletion` (`atlas/modules/llm/litellm_caller.py`). Every model in `llmconfig.yml` is routed through Chat Completions regardless of `model_url`, and `ModelConfig` (`atlas/modules/config/models.py`) has no field for selecting a different API surface or for passing extra request-body parameters.

## Requested change

Allow a model entry to opt into the Responses API, e.g.

```yaml
  gpt-5.6-sol:
    model_url: "https://api.openai.com/v1/responses"
    model_name: "gpt-5.6-sol"
    api_key: "${OPENAI_API_KEY}"
    api_flavor: "responses"        # new; defaults to "chat_completions"
    reasoning_effort: "medium"     # new; optional, passed through
    supports_tools: true
```

Scope to consider:

- Route those models through LiteLLM's Responses support instead of `acompletion`.
[...]

PR #893, opened 2026-09-05
fix(llm): send per-model reasoning_effort so GPT-5.6 models can use tools

atilavahedian:
Addresses #756.

## What breaks

Agent mode on `gpt-5.6-luna` fails before any generation. Every tool carrying request
returns HTTP 400, the string #756 quotes:

```
Function tools with reasoning_effort are not supported for gpt-5.6-luna in /v1/chat/completions. To use function tools, use /v1/responses or set reasoning_effort to 'none'.
```

Recorded with `"status_code": 400` and `"type": "api_status_error"` on 60 of 60 requests.

## Versions and what was exercised

* Repo at `d8245b87dd3befad2c6fde3057199b9217febee2` ("feat: canvas width and orientation
  controls (#754) (#883)"), with `litellm` at the locked 1.97.0 (`uv.lock`).
* Path exercised: agent mode, non streaming. `AgenticLoop._call_llm` at
  `atlas/application/chat/agent/agentic_loop.py:398` calls
  `LiteLLMCaller.call_with_tools` (`atlas/modules/llm/litellm_caller.py:1259`, the
  `_acompletion_with_retry` at `:1289`), whose kwargs come from `_get_model_kwargs`
  (`:848`).
* Payload measured: `{model, messages, tools, tool_choice: "auto", temperature: 0.7,
  max_completion_tokens: 10000}` on `POST https://api.openai.com/v1/chat/completions`.
  That is what LiteLLM 1.97.0 puts on the wire for a `gpt-5` family model: `max_tokens`
  becomes `max_completion_tokens` (`llms/openai/chat/gpt_5_transformation.py:256`) and
  `temperature` is kept rather than dropped, because the model map marks these models as
  supporting `reasoning_effort="none"` and this backend sends no effort
  (`gpt_5_transformation.py:277-282`).
* Baseline model: `gpt-5.4`. Candidate model: `gpt-5.6-luna`, the model #756 names.
* System prompt, the six tool schemas and the tool behaviours all come from this repo at that
  sha (`atlas/config/prompts/system_prompt.md`, six tools from the default stdio MCP servers
  in `atlas/config/mcp.json`). Nothing from this repo was executed on the measuring host.

## Why `gpt-5.5` is not the baseline

`_get_model_kwargs` puts `temperature` in every payload unconditionally
(`atlas/modules/llm/litellm_caller.py:861-864`), defaulting to `0.7`
(`atlas/modules/config/models.py:80`). Direct probes, one 16 token request each, no tools:

| request | `gpt-5.4` | `gpt-5.6-luna` |
| --- | --- | --- |
| chat, no tools, `temperature: 0.7` | 200 | 400 `Unsupported value: 'temperature' does not support 0.7 with this model. Only the default (1) value is supported.` |
| chat, tools, `temperature: 0.7` | 200 | 400 function tools / reasoning effort |
| chat, tools, `temperature: 0.7`, `reasoning_effort: "none"` | 200 | 200 |
| `/v1/responses`, tools, `temperature: 0.7` | 200 | 400 `Unsupported parameter: 'temperature' is not supported with this model.` |
| `/v1/responses`, tools, no temperature | not probed | 200 |

So `gpt-5.6-luna` has two independent blockers for this backend, and the function tools one
is returned first and hides the temperature one.

## The Responses API route was tried first, and it does not work on its own

#756 asks for `/v1/responses`. That was the first repair candidate measured, and it failed:
sending the identical payload to `/v1/responses` clears the function tools 400 and
immediately returns `Unsupported parameter: 'temperature' is not supported with this model.`
on 60 of 60 requests.

A Responses migration is still worth doing, but it is not a drop in fix while `temperature`
is unconditional: it has to drop the sampling parameter at the same time. That is a much
larger change than this one, none of it was measured, and it is not in this PR.

## Patch

Reasoning effort becomes a per model registry setting and is forwarded from the single place
every OpenAI bound request already builds its kwargs, so the plain path, the tool path and
both streaming twins all get it:

| call path | file:line | kwargs from |
| --- | --- | --- |
| plain | `atlas/modules/llm/litellm_caller.py:1057` | `_get_model_kwargs`, `:1037` |
| with tools | `atlas/modules/llm/litellm_caller.py:1289` | `_get_model_kwargs`, `:1274` |
| plain, streaming | `atlas/modules/llm/litellm_streaming.py:98` | `_get_model_kwargs`, `:65` |
| with tools, streaming | `atlas/modules/llm/litellm_streaming.py:314` | `_get_model_kwargs`, `:280` |

* `atlas/modules/config/models.py`: `ModelConfig.reasoning_effort`, optional, default `None`.
* `atlas/modules/llm/litellm_caller.py`, in `_get_model_kwargs`:

```python
        reasoning_effort = getattr(model_config, "reasoning_effort", None)
        if reasoning_effort:
            kwargs["reasoning_effort"] = reasoning_effort
```

* `atlas/config/llmconfig.yml`: a `gpt-5.6-luna` entry with `reasoning_effort: "none"`, since
  the file ships no GPT-5.6 entry today.
* `atlas/tests/test_llm_reasoning_effort.py`: 7 unit tests, built the way
  `atlas/tests/test_llm_customer_id_header.py` builds its caller. No network.

A model that leaves the field unset sends the exact payload it sends today, so gpt-4.1,
gpt-4o, the Anthropic, Gemini, Groq and OpenRouter entries are all unchanged. The setting is
an operator choice rather than a hardcoded switch on the model name, because
`reasoning_effort: "none"` does turn reasoning off and a deployment should be able to say so
for itself.

## Before and after

Twelve case behavioural suite over agent mode, deterministic checks only: `no_api_error`,
`tool_called` (with argument subsets where the argument matters), `tool_not_called`,
`turns_at_most`, and regex or substring assertions on the final answer. The tools are the
calculator, the prompts catalog, two `structured_output_demo` tools and two
`username-override-demo` tools, faked in memory from this repo's own tool bodies. Standard
sync tier, 5 reps per case per configuration, thresholds fixed throughout.

| run | model | params | cases | reps |
| --- | --- | --- | --- | --- |
| baseline | `gpt-5.4` | `temperature: 0.7` | 12/12 pass | 60/60 |
| candidate | `gpt-5.6-luna` | `temperature: 0.7` | 0/12 | 0/60 |
| candidate, routed to `/v1/responses` | `gpt-5.6-luna` | `temperature: 0.7` | 0/12 | 0/60 |
| patched | `gpt-5.6-luna` | `temperature: 0.7` plus `reasoning_effort: "none"` | 12/12 pass | 119/120 |

All twelve cases regressed at one sided Fisher exact p = 0.00397 each. No case was flaky and
none was contested. Every one of the 60 candidate reps failed on the function tools 400
above, and every one of the 60 responses route reps failed on the temperature 400.

The patched configuration restored 12 of 12 broken cases across 10 reps each (5 screening
plus 5 verification) and broke 0 previously passing cases. The single non passing rep out of
120 is a brittle assertion in the measuring suite, not a model failure: the model answered
"does **not** have permission" and the suite's regex did not match across the markdown
emphasis. The suite was deliberately left unchanged rather than loosened after the fact.

Zero tokens were billed on either unpatched candidate run, because the request is rejected
before generation, so the break costs nothing to reproduce.

## Limitations

1. This repo's own test suite was not run locally, and the 7 added tests were not executed
   here. The change was written against the pinned sha. No code from this repo ran on the
   measuring host.
2. One call path of four was measured: agent mode, non streaming. The other three build
   their kwargs through the same `_get_model_kwargs`, so both the 400 and the fix transfer by
   inspection, but they were not run.
3. The fix was verified through a separate chat/completions transport carrying the
   reconstructed payload, not through `LiteLLMCaller` itself. What is established is that the
   same tool schemas, the same system prompt and the same model return 400 without
   `reasoning_effort: "none"` and 200 with it. The forwarding code is new and was not run
   against the live API.
4. `gpt-5.6-sol` and `gpt-5.6-terra` were not measured here, although #756 names them and the
   error text is the same. The YAML entry added is for `gpt-5.6-luna` only; the same line
   applies to the other two.
5. The unconditional `temperature` at `_get_model_kwargs` is left as it is. For a GPT-5.6
   model configured with `reasoning_effort: "none"` it is accepted again, which the 120
   patched reps show directly since they all carry `temperature: 0.7`. Whether some other
   model needs a `supports_temperature` style flag was not measured, so none is shipped.
6. Four of the ten default MCP servers are in the suite, contributing six tools. The other
   six need file generation, MCP sampling, UI round trips or a separate HTTP process. None of
   them changes any payload field the 400 depends on.
7. `reasoning_effort: "none"` turns reasoning off. Twelve tool routing and error reporting
   cases are unchanged by it at 119 of 120 reps, which says nothing about deployments that
   chose a reasoning model for its reasoning.

## Evidence

Run records, the probe table, the case list, the recorded 400s and the cost breakdown:
https://gist.github.com/atilavahedian/e38f64cc8d3d26214a85e330f8056924

Measured with upshift (https://github.com/Mechanism-world/upshift). Total measurement cost
$0.64.

### DIFF
--- a/atlas/config/llmconfig.yml
+++ b/atlas/config/llmconfig.yml
@@ -44,6 +44,25 @@ models:
       calling, and up to 1M tokens of context. It is ideal for classification,
       autocompletion, and fast lightweight tasks where latency matters more than
       peak reasoning quality. Context window: 1M tokens. Output limit: 32K tokens.
+  gpt-5.6-luna:
+    model_url: "https://api.openai.com/v1/chat/completions"
+    model_name: "gpt-5.6-luna"
+    api_key: "${OPENAI_API_KEY}"
+    compliance_level: "External"
+    supports_tools: true
+    # Required, not optional. Without it every tool-carrying request to this
+    # model is rejected with HTTP 400: "Function tools with reasoning_effort are
+    # not supported for gpt-5.6-luna in /v1/chat/completions. To use function
+    # tools, use /v1/responses or set reasoning_effort to 'none'." The same
+    # setting also makes the model accept the non-default temperature this
+    # backend always sends. The other GPT-5.6 models (gpt-5.6-sol,
+    # gpt-5.6-terra) return the same error and need the same line.
+    reasoning_effort: "none"
+    model_card: >
+      GPT-5.6 Luna is the low-cost member of OpenAI's GPT-5.6 reasoning family.
+      Reasoning effort is pinned to "none" here because function tools are
+      rejected on /v1/chat/completions at any other effort level; raise it only
+      if this deployment moves tool calls to /v1/responses.
   groq-gpt-oss-120b:
     model_url: "https://api.groq.com/openai/v1"
     model_name: "openai/gpt-oss-120b"
--- a/atlas/config/llmconfig.yml
+++ b/atlas/config/llmconfig.yml
@@ -57,12 +57,22 @@ models:
     # setting also makes the model accept the non-default temperature this
     # backend always sends. The other GPT-5.6 models (gpt-5.6-sol,
     # gpt-5.6-terra) return the same error and need the same line.
+    #
+    # The other route that 400 offers -- moving tool calls to /v1/responses --
+    # is not available to this entry. ATLAS builds every request through
+    # LiteLLMCaller._get_model_kwargs against the chat/completions URL above,
+    # and /v1/responses rejects the unconditional `temperature` that method
+    # always sends ("Unsupported parameter: 'temperature' is not supported with
+    # this model."). Raise the effort here only if this deployment has moved
+    # tool calls to /v1/responses and stopped sending `temperature`.
     reasoning_effort: "none"
     model_card: >
-      GPT-5.6 Luna is the low-cost member of OpenAI's GPT-5.6 reasoning family.
-      Reasoning effort is pinned to "none" here because function tools are
-      rejected on /v1/chat/completions at any other effort level; raise it only
-      if this deployment moves tool calls to /v1/responses.
+      GPT-5.6 Luna is the low-cost, low-latency member of OpenAI's GPT-5.6
+      family and supports tool/function calling. Reasoning is switched off for
+      this model in ATLAS: replies come back without the extended reasoning pass
+      the family is otherwise capable of, which makes them faster and cheaper.
+      Best for: quick tool-using chat and lightweight tasks. Choose a reasoning
+      model instead for work that needs step-by-step analysis.
   groq-gpt-oss-120b:
     model_url: "https://api.groq.com/openai/v1"
     model_name: "openai/gpt-oss-120b"
--- a/atlas/modules/config/models.py
+++ b/atlas/modules/config/models.py
@@ -15,7 +15,7 @@ Dependency direction: ``config_loader`` -> ``settings`` -> ``models``.
 
 import os
 import re
-from typing import ClassVar, Dict, List, Literal, Optional
+from typing import ClassVar, Dict, List, Literal, Optional, get_args
 
 from pydantic import BaseModel, Field, field_validator, model_validator
 
@@ -70,6 +70,16 @@ def resolve_env_var(value: Optional[str], required: bool = True) -> Optional[str
     return value
 
 
+# Reasoning-effort levels LiteLLM will carry on a chat/completions request.
+# Mirrors ``litellm.types.llms.openai.REASONING_EFFORT`` (litellm 1.97.0, the
+# version pinned in uv.lock). Written out here rather than imported so that
+# loading configuration does not depend on a LiteLLM internal type, and so an
+# unusable value is rejected when the config file is read instead of arriving as
+# a provider HTTP 400 in the middle of a user's chat.
+ReasoningEffort = Literal["none", "minimal", "low", "medium", "high", "xhigh"]
+REASONING_EFFORT_VALUES: tuple = get_args(ReasoningEffort)
+
+
 class ModelConfig(BaseModel):
     """Configuration for a single LLM model."""
     model_name: str
@@ -105,13 +115,15 @@ class ModelConfig(BaseModel):
     # Whether this model supports tool/function calling.
     # When false, tools are stripped from requests and the user is warned.
     supports_tools: bool = True
-    # Reasoning effort to send with every request to this model, e.g. "none",
-    # "low", "medium", "high". Left unset for models that have no reasoning
-    # control, which is every model shipped before the GPT-5.6 family: the key
-    # is then omitted entirely and the payload is byte-identical to today's.
-    # Required for OpenAI's GPT-5.6 models, which reject function tools on
-    # /v1/chat/completions unless reasoning effort is explicitly "none".
-    reasoning_effort: Optional[str] = None
+    # Reasoning effort to send with every request to this model. One of
+    # REASONING_EFFORT_VALUES ("none", "minimal", "low", "medium", "high",
+    # "xhigh"); anything else is rejected when the config file is loaded.
+    # Left unset for models that have no reasoning control, which is every model
+    # shipped before the GPT-5.6 family: the key is then omitted entirely and the
+    # payload is byte-identical to today's. Required for OpenAI's GPT-5.6 models,
+    # which reject function tools on /v1/chat/completions unless reasoning effort
+    # is explicitly "none".
+    reasoning_effort: Optional[ReasoningEffort] = None
     # Rich model card text shown in the UI info panel (markdown allowed).
     # Provides details like context window, training info, strengths, etc.
     model_card: Optional[str] = None
@@ -131,6 +143,41 @@ class ModelConfig(BaseModel):
     # applies when pass_user_as_customer_id is true and the value actually ends
     # with the suffix; otherwise the value is sent unchanged.
     customer_id_strip_suffix: Optional[str] = None
+
+    @field_validator('reasoning_effort', mode='before')
+    @classmethod
+    def validate_reasoning_effort(cls, v):
+        """Reject an unusable reasoning effort while the config file is loading.
+
+        Without this the provider is the first thing to notice: "meduim",
+        "None" and " none" all load without complaint and then come back as an
+        HTTP 400 on a user's first message. Surrounding whitespace is trimmed
+        and a blank value is treated as unset, which is what an operator who
+        clears the field in YAML means.
+        """
+        if v is None:
+            return None
+        if not isinstance(v, str):
+            raise ValueError(
+                f"reasoning_effort must be a string, got {type(v).__name__}. "
+                f"Valid values: {', '.join(REASONING_EFFORT_VALUES)}; omit the key to send none."
+            )
+        normalized = v.strip()
+        if not normalized:
+            return None
+        if normalized not in REASONING_EFFORT_VALUES:
+            hint = ""
+            if normalized.lower() == "none":
+                hint = (
+                    ' Note that "none" must be lowercase and quoted: it means "send '
+                    'reasoning_effort: none", which turns reasoning off. To send no '
+                    "reasoning_effort at all, omit the key."
+                )
+            raise ValueError(
+                f"reasoning_effort must be one of {', '.join(REASONING_EFFORT_VALUES)}, "
+                f"or omitted; got {v!r}.{hint}"
+            )
+        return normalized
 
 
 class LLMConfig(BaseModel):
--- a/atlas/modules/config/models.py
+++ b/atlas/modules/config/models.py
@@ -105,6 +105,13 @@ class ModelConfig(BaseModel):
     # Whether this model supports tool/function calling.
     # When false, tools are stripped from requests and the user is warned.
     supports_tools: bool = True
+    # Reasoning effort to send with every request to this model, e.g. "none",
+    # "low", "medium", "high". Left unset for models that have no reasoning
+    # control, which is every model shipped before the GPT-5.6 family: the key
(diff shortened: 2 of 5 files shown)


garland3:
## Summary
Adds an optional per-model `reasoning_effort` setting to `ModelConfig`, forwards it from `_get_model_kwargs` to LiteLLM, and pins `"none"` on the shipped `the reviewer-5.6-luna` entry so tool-carrying requests stop returning HTTP 400. Includes a new unit-test file.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is thoroughly demonstrated for its size — reproduction, isolation of the two blockers, before/after behavioural runs, and unit tests. The main gap is that the added tests were never executed anywhere and the fix was verified through a reconstructed payload rather than through `LiteLLMCaller` itself; a single `pytest atlas/tests/test_llm_reasoning_effort.py` run or CI-green confirmation would close it.

## Notes
- Constrain `reasoning_effort` to a `Literal[...]` or add a validator — `atlas/modules/config/models.py:114` accepts `"None"`, `" none"`, or `"meduim"` at load and only fails as a provider 400 mid-chat.
- Document the new key in `docs/admin/llm-config.md` alongside `supports_vision` / `supports_tools`, including that the 5.6 family requires `"none"` for tool calls.
- `atlas/config/llmconfig.yml:61` — the `model_card` is user-facing; the endpoint-routing rationale belongs in the YAML comment, with the card stating plainly that reasoning is disabled here.

---

atilavahedian:
Thanks for the review. All three nits are addressed in `dc5686f`, and the evidence gap is
closed: the tests now run, including on the PR as you reviewed it.

## 1. `reasoning_effort` is validated at load

`atlas/modules/config/models.py`. The field is now typed and validated instead of being a
bare `Optional[str]`:

```python
ReasoningEffort = Literal["none", "minimal", "low", "medium", "high", "xhigh"]
REASONING_EFFORT_VALUES: tuple = get_args(ReasoningEffort)
```

```python
    reasoning_effort: Optional[ReasoningEffort] = None

    @field_validator('reasoning_effort', mode='before')
    @classmethod
    def validate_reasoning_effort(cls, v):
        ...
```

The value set is `litellm.types.llms.openai.REASONING_EFFORT` at
`litellm` 1.97.0 — the version `uv.lock` pins, and the library that actually puts the field
on the wire. It is written out in `models.py` rather than imported so that loading
configuration does not depend on a LiteLLM internal type.

Behaviour on your three examples, all at `ModelConfig` construction, i.e. while
`llmconfig.yml` is being read:

| value in YAML | before | now |
| --- | --- | --- |
| `"meduim"` | loads, HTTP 400 on the first chat | `ValidationError` naming the value and all six legal levels |
| `None` (unquoted — YAML reads it as the string `"None"`) | loads, HTTP 400 | `ValidationError`, with a note that `"none"` must be lowercase and quoted and that omitting the key is how you send nothing |
| `" none"` | loads, HTTP 400 | trimmed to `"none"` and accepted — the stray space never reaches the provider |

Two deliberate non-errors: whitespace is trimmed rather than rejected, and a blank value is
treated as unset (`None`), which is what an operator who clears the field in YAML means.
That keeps the existing `test_empty_string_is_treated_as_unset` true. A non-string value is
rejected with a message that lists the legal levels.

The failure surfaces through `LLMConfig` too, so a bad `llmconfig.yml` fails when the file
is loaded rather than per-model at chat time.

**The test that proves it** is
`atlas/tests/test_llm_reasoning_effort.py::TestReasoningEffortIsValidatedAtLoad::test_a_misspelled_level_is_rejected`
— it asserts `ValidationError` on `"meduim"` and that the message contains both the rejected
value and every legal level. The rest of that class covers the same ground from the other
sides:

* `test_every_documented_level_is_accepted` (parametrised over all six levels)
* `test_the_string_None_is_rejected_and_explained`
* `test_a_wrong_case_level_is_rejected` (`"NONE"`)
* `test_an_unknown_level_is_rejected` (`"off"`)
* `test_a_non_string_level_is_rejected`
* `test_surrounding_whitespace_is_trimmed` (`" none "` → `"none"`)
* `test_a_blank_value_is_unset_not_an_error`
* `test_a_bad_level_fails_the_whole_llmconfig`
* `test_a_rejected_level_never_reaches_the_payload` — the end-to-end point: with the
  validator in place, `_get_model_kwargs` cannot be reached with a value the provider will
  reject, because the caller cannot be constructed.

The file goes from 7 tests to 22.

## 2. The key is documented

`docs/admin/llm-config.md`:

* A bullet in **Configuration Fields Explained**, next to `supports_vision`, giving the six
  accepted values, saying that an invalid value is rejected at config load rather than on the
  first chat, that omitting the key leaves the payload unchanged, and that OpenAI's GPT-5.6
  family requires `"none"` for tool calls. `supports_tools` turned out to have no bullet at
  all in that list (only an appearance in a YAML example), so it got one too — say the word
  if you would rather that were a separate change.
* A new **Reasoning Effort** section in the style of the LiteLLM Customer ID Header section:
  a YAML example, the accepted values and the load-time failure, the fact that the setting is
  sent on all four call paths because they share `_get_model_kwargs`, and a
  **GPT-5.6 requires `"none"` for tool calls** subsection carrying the verbatim 400 and the
  note that routing tool calls to `/v1/responses` is not a drop-in replacement while
  `temperature` is unconditional.

## 3. `llmconfig.yml` — rationale in the comment, plain language in the card

The endpoint-routing reasoning moved out of `model_card` and into the YAML comment above
`reasoning_effort`, which now also records why the `/v1/responses` route the error text
suggests is not available to this entry. The card is now about the model, and states the
consequence plainly:

> GPT-5.6 Luna is the low-cost, low-latency member of OpenAI's GPT-5.6 family and supports
> tool/function calling. Reasoning is switched off for this model in ATLAS: replies come back
> without the extended reasoning pass the family is otherwise capable of, which makes them
> faster and cheaper. Best for: quick tool-using chat and lightweight tasks. Choose a
> reasoning model instead for work that needs step-by-step analysis.

## 4. The evidence gap — what was run, and where

You were right that nothing had been executed. It has been now.

**Environment.** A fresh clone of `sandialabs/atlas-ui-3` on macOS 27.0 (arm64), the PR
branch checked out, dependencies installed with the repo's own lockfile
(`uv sync --frozen --group dev`): Python 3.12.13, pytest 9.1.1, pydantic 2.12.5, litellm
1.97.0 — the pinned version the PR description reasons about. No network access during the
tests, no API keys present.

**Results.**

| command | result |
| --- | --- |
| `pytest atlas/tests/test_llm_reasoning_effort.py` at `387f175` (the PR as you reviewed it) | **7 passed** |
| `pytest atlas/tests/test_llm_reasoning_effort.py` at `dc5686f` | **22 passed** |
| `pytest` over the LLM + config neighbourhood (`test_llm_reasoning_effort`, `test_llm_customer_id_header`, `test_llm_env_expansion`, `test_llm_user_api_key`, `test_llm_retry`, `test_llm_anthropic_history`, `test_config_manager`, `test_config_manager_paths`, `test_streaming_token_flow`) | **186 passed** |
| `ruff check` on the four changed files | clean |

**The full suite, three times, same machine, same venv:**

| tree | result |
| --- | --- |
| `d8245b8` (this PR's base) | 183 failed, 2541 passed, 19 skipped |
| `387f175` (PR head) | 183 failed, 2548 passed, 19 skipped |
| `dc5686f` (with this commit) | 183 failed, 2563 passed, 19 skipped |

The same 183 fail in all three trees, so neither the PR nor this commit adds a failure; the
passing count moves by exactly the 7 and then 15 tests added. Those 183 are environment-bound
rather than code failures on this host — the telemetry, feedback, chat-history and file-route
suites return `503` with no backing store configured, and `test_capture_store.py` fails on
the macOS `/var` → `/private/var` symlink in a `path.resolve()` containment assertion. I have
the full list if it is useful, but they are equally red on your base commit. Repo-wide
`ruff check` reports 274 findings on base and branch alike; untouched.

**What is still not run.** Nothing here talked to the OpenAI API, so limitation 3 in the PR
description stands unchanged: the forwarding code has still not been exercised against the
live API through `LiteLLMCaller` itself. This commit adds load-time validation, documentation
and card text — for a valid value the request payload is byte-identical to `387f175`, and the
two tests that pin the payload (`test_effort_is_sent_when_configured`,
`test_effort_does_not_disturb_the_other_kwargs`) still pass. CI on your side would be the
thing that closes the rest.

garland3:
## Summary
Adds a validated `reasoning_effort` field to `ModelConfig`, forwards it from the shared `_get_model_kwargs` on all LiteLLM call paths, pins `reasoning_effort: "none"` on the shipped `the reviewer-5.6-luna` entry, and documents the setting. Fixes a provider 400 on the reviewer-5.6 models when function tools are used.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size, with a reproduced 400, an isolation probe table, a 12-case behavioural run and executed unit/lint suites. The one missing artifact is a single end-to-end request through the changed `_get_model_kwargs` path against the live model; the streaming paths are argued from shared code rather than exercised.

## Notes
- Docs and the field comment say a bad value "raises a configuration error when ATLAS loads `llmconfig.yml`", but `ConfigManager.llm_config` (`atlas/modules/config/config_loader.py:193-205`) catches it and falls back to `LLMConfig(models={})` — one typo silently drops every model. Reconcile code and docs, ideally by failing loudly or per-model.
- `test_shipped_gpt_56_entry_pins_effort_none` (`atlas/tests/test_llm_reasoning_effort.py:84`) asserts on a specific key in the deployed YAML; a rule over all `the reviewer-5.6*` + `supports_tools` entries would be both more robust and cover `sol`/`terra`.
- The non-string validation message at `atlas/modules/config/models.py:166` says "omit the key to send none", but `none` is itself a legal wire value — reword to "send no reasoning_effort at all".

---

PR #897, opened 2026-09-07
fix(config): re-raise a model validation error instead of loading zero models

atilavahedian:
Follow-up to #893, addressing the first note in the review that landed on it.

#893 documented that an invalid `reasoning_effort` "raises a configuration error when ATLAS loads `llmconfig.yml`" (`docs/admin/llm-config.md`, and the field comment at `atlas/modules/config/models.py:118-125`). It does not. `ConfigManager.llm_config` builds the whole file with one `LLMConfig(**data)` inside a catch-all (`atlas/modules/config/config_loader.py:189-213`) that falls back to `LLMConfig(models={})`, so one typo in one entry loads zero models and leaves only an ERROR log line.

This re-raises `ValidationError` ahead of that catch-all, matching the refusal-to-start contract `app_settings` already states at `config_loader.py:172-176`.

Three tests in `atlas/tests/test_llm_reasoning_effort.py`: the typo raises, a missing config file still yields an empty config, a valid file still loads. The first fails on current `main`.

Verified in a clean container against the repo's `uv.lock`: 98 passed across `test_config_manager`, `test_config_manager_paths` and `test_llm_reasoning_effort`; `ruff check` clean. No provider call.

### DIFF
--- a/atlas/modules/config/config_loader.py
+++ b/atlas/modules/config/config_loader.py
@@ -15,6 +15,7 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional
 
 import yaml
+from pydantic import ValidationError
 
 from .models import (
     FileExtractorsConfig,
@@ -200,6 +201,13 @@ class ConfigManager:
                     self._llm_config = LLMConfig(models={})
                     logger.info("Created empty LLM config (no configuration file found)")
 
+            except ValidationError:
+                # A model entry that fails validation is an intentional
+                # refusal-to-start, the same contract app_settings above states.
+                # Falling back here would swap one bad entry for zero models and
+                # leave only a log line, and would contradict docs/admin/llm-config.md,
+                # which promises a load-time error for an invalid `reasoning_effort`.
+                raise
             except Exception as e:
                 logger.error(f"Failed to parse LLM configuration: {e}", exc_info=True)
                 self._llm_config = LLMConfig(models={})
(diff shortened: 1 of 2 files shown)


garland3:
## Summary
Makes `ConfigManager.llm_config` re-raise pydantic `ValidationError` instead of falling back to an empty `LLMConfig`, so a typo'd model entry (e.g. `reasoning_effort: meduim`) fails at config load rather than silently yielding zero models. Adds three tests and a docs note.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well-demonstrated for its size — the new failing-typo test is stated to fail on current `main`, and verification covers 98 tests plus lint in a clean container. No missing artifact of consequence.

## Notes
- The re-raise path logs nothing before raising, dropping the old `Failed to parse LLM configuration` line; log the resolved `file_paths` before `raise` at `atlas/modules/config/config_loader.py:204` so operators know which `llmconfig.yml` was read.
- `list_models` and `export_config` read `llm_config` unguarded, so a typo now tracebacks out of an operator CLI while `atlas config validate` reports it cleanly — `atlas/modules/config/cli.py:45`.
- The pre-existing `except Exception` fallback in front of which the new clause sits has no test, and the new tests monkeypatch the private `_load_file_with_error_handling` rather than writing a real YAML file — `atlas/tests/test_llm_reasoning_effort.py:170`.

---

garland3:
Update to the changelog required.

garland3:
Superseded by replacement PR #900, which includes the validation fix, operator-path logging, real YAML regression coverage, and changelog alignment.

PR #899, opened 2026-09-07
feat(mcp): interactive OAuth 2.1 authorization for MCP servers (closes #898)

### LINKED ISSUE #898: MCP OAuth 2.1: no interactive authorization flow — /api/mcp/auth/<server>/oauth/start is advertised but does not exist
## Summary

Atlas can *store* an OAuth access token for an MCP server, but it can never *obtain* one. There is no interactive OAuth 2.1 authorization-code flow: no FastMCP `OAuth` client auth, no Dynamic Client Registration, and no browser-facing routes to start the flow or receive the callback. For any server configured with `auth_type: "oauth"`, Atlas simply assumes a token is already sitting in its store.

The gap is not merely missing — it is already advertised to users. `atlas/modules/mcp_tools/mcp_execution.py` raises `AuthenticationRequiredException` with

```python
oauth_start_url = f"/api/mcp/auth/{server_name}/oauth/start"
```

(lines 212–215, 229, 417, 430), which `atlas/application/chat/utilities/tool_executor.py:745,759` forwards to the frontend. **That route does not exist.** `atlas/routes/mcp_auth_routes.py` registers only `GET /status`, `POST /{server_name}/token`, and `DELETE /{server_name}/token`. A user who follows the "connect" affordance for an OAuth server gets a 404.

So today the only way to use an OAuth-protected MCP server is for the user to obtain a bearer token out-of-band and paste it into `POST /api/mcp/auth/{server}/token`. Where the authorization server relies on Dynamic Client Registration there is no realistic way for a user to do that by hand at all.

## Reproduction

Any remote MCP server that implements the MCP authorization spec will do. Such a server rejects an unauthenticated `initialize` with a `401` and an RFC 9728 challenge:

```
$ curl -s -i -X POST https://<mcp-server>/mcp \
    -H 'Content-Type: application/json' \
    -H 'Accept: application/json, text/event-stream' \
    -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"probe","version":"0"}}}'

HTTP/2 401
www-authenticate: Bearer resource_metadata="https://<mcp-server>/.well-known/oauth-protected-resource"
```

Following that challenge yields the standard discovery chain — RFC 9728 protected-resource metadata naming one or more `authorization_servers`, then RFC 8414 authorization-server metadata exposing `authorization_endpoint`, `token_endpoint`, a `registration_endpoint` for RFC 7591 Dynamic Client Registration, `code_challenge_methods_supported: ["S256"]`, and public-client auth (`token_endpoint_auth_methods_supported: ["none"]`). This chain was verified live against a public hosted MCP server on 2026-09-07; it is the shape the spec mandates, not a vendor quirk.

To reproduce in Atlas, add such a server to the MCP config:

```json
"example-remote": {
  "url": "https://<mcp-server>/mcp",
  "auth_type": "oauth",
  "groups": ["users"],
  "description": "Remote MCP server requiring OAuth 2.1"
}
```

Then call any tool on it as a logged-in user. `_requires_user_auth()` returns true, `_get_user_client()` returns `None` (no stored token), and the chat surfaces an auth-required error pointing at `/api/mcp/auth/example-remote/oauth/start` → 404. There is no path forward for the user.

## Current state of the code

There is real scaffolding already in place; this issue is about finishing it, not starting from scratch.

Present:
- `OAuthConfig` in `atlas/modules/config/models.py:143` (`scopes`, `client_name`, `callback_port`) with an `oauth_config` field on `MCPServerConfig` (line 171) and `auth_type: "oauth"` accepted at line 169.
- Encrypted per-user token storage: `atlas/modules/mcp_tools/token_storage.py` (Fernet, PBKDF2, key gated by `MCP_TOKEN_ENCRYPTION_KEY` with a ≥32-char/placeholder check), including `update_oauth_tokens()` (line 535) which already writes `token_type="oauth_access"` plus a refresh token.
- `AuthenticationRequiredException.oauth_start_url` (`token_storage.py:119`) and its propagation to the frontend.
- `_requires_user_auth()` already lists `"oauth"` (`atlas/modules/mcp_tools/mcp_user_clients.py:47-52`).

Missing:
- No route serving `/api/mcp/auth/{server_name}/oauth/start`, and no callback route.
- `oauth_config` is **never read** by any code path (grep: only its declaration in `models.py`).
- `update_oauth_tokens()` is **never called** by any code path.
- `mcp_user_clients.py:238-270` treats `oauth` identically to `bearer`: it passes the stored token string as `auth=<token>` to the FastMCP `Client`. It never constructs `fastmcp.client.auth.OAuth`, so no discovery, no DCR, no PKCE, and no automatic refresh ever happens.
- No refresh-on-401 path: an expired `oauth_access` token just fails the tool call.
- `docs/admin/mcp-server-authentication.md` documents only `api_key`/`jwt`/`bearer`/`none`; `oauth` is undocumented.

## Not Globus OAuth

This is distinct from the existing Globus integration (`atlas/core/globus_auth.py`, `atlas/routes/globus_auth_routes.py`, `docs/admin/globus-auth-integration-2026-02-24.md`). Globus is a *single, statically pre-registered* identity/transfer provider: one `client_id`/`client_secret` from settings, a fixed `/auth/globus/callback`, and tokens tied to Globus services. The gap here is *per-MCP-server, dynamically registered* OAuth: an arbitrary number of third-party resource servers discovered at runtime from a `www-authenticate` challenge, each with its own authorization server, its own DCR-issued client credentials, and its own per-user tokens. Globus's route module is a good pattern to copy for CSRF state, session handling, and redirect hygiene — but the two flows should stay separate, and Globus should not be extended to cover MCP servers.

Related: #96 ("Oauth2.1 flows enable") raised OAuth 2.1 generally; the manual-JWT-upload half of that issue shipped. This issue scopes the remaining interactive-authorization half concretely. Also note #885 (FastMCP 4 evaluation) — whichever major version we land on, the client auth API should be checked against it.

## Expected behavior

1. An admin configures a server with `auth_type: "oauth"` (optionally with `oauth_config.scopes` / `client_name`).
2. `GET /api/mcp/auth/status` reports the server as requiring auth and not yet authenticated, and the Tools/auth UI offers a "Connect" action.
3. The user clicks Connect → `GET /api/mcp/auth/{server}/oauth/start` → Atlas discovers the authorization server from the MCP endpoint (401 `www-authenticate` / `.well-known/oauth-protected-resource` → `.well-known/oauth-authorization-server`), performs DCR if needed, and redirects the browser to the provider's authorization endpoint with PKCE.
4. The provider redirects back to Atlas's own callback (`/api/mcp/auth/{server}/oauth/callback`), which validates state, exchanges the code, and stores the access + refresh tokens encrypted per user via `update_oauth_tokens()`.
5. Subsequent tool calls transparently use the token, and refresh it when it expires.
6. Disconnect (`DELETE /api/mcp/auth/{server}/token`) removes tokens, invalidates cached clients, and ideally calls the provider's `revocation_endpoint` when advertised.

## Likely FastMCP integration points

- `fastmcp.client.auth.OAuth(mcp_url=..., scopes=..., client_name=..., token_storage=...)` handles discovery, DCR, PKCE, exchange and refresh. Pinned range today: `fastmcp[tasks]>=3.4.3,<4.0.0` (`pyproject.toml:27`).
- **Key design constraint:** `OAuth`'s stock flow spins up a *local* callback HTTP server on `127.0.0.1:<callback_port>` and shells out to `webbrowser.open()`. That is a desktop-client assumption and is wrong for Atlas, which is a multi-user server where the browser is on a different machine. We need Atlas's own server-side redirect/callback routes to be the `redirect_uri`, which means either supplying a custom redirect handler to `OAuth`, or driving the underlying `httpx`/`mcp` OAuth provider pieces directly and keeping FastMCP only for the authenticated `Client`. Whichever way, `mcp_user_clients.py:238-270` must branch on `auth_type == "oauth"` instead of falling through to the bearer-string path.
- `token_storage` on `OAuth` expects an `AsyncKeyValue` backend. Atlas already has an encrypted per-user store; the cleanest bridge is an adapter over `token_storage.py` keyed by `{user_email}:{server_name}` so tokens stay in Atlas's existing Fernet-encrypted store rather than a second on-disk cache. Do **not** introduce a per-process disk store that ignores user identity.
- DCR client credentials (`client_id`, and `client_secret` when the AS issues one) are per-server, not per-user, and need their own persisted home so we don't re-register on every restart or every user.

## Storage and security requirements

- Access + refresh tokens stored per `(user_email, server_name)`, encrypted at rest with the existing Fernet/`MCP_TOKEN_ENCRYPTION_KEY` machinery. No new plaintext token file.
- No token, authorization code, refresh token, or client secret in logs — route everything user- or server-derived through `sanitize_for_logging` as the existing MCP auth routes do.
- CSRF: a per-attempt `state` bound to the session (mirror `globus_auth_routes.py`'s `request.session[...]` state check), single-use, with a TTL.
- PKCE `S256` required; reject a callback whose `state` is unknown, already used, or belongs to a different user.
- Authorization on both routes: the user must be in the server's `groups` — reuse `get_authorized_servers(current_user, is_user_in_group)` exactly as `upload_token` does, and 403 otherwise. `server_name` must be validated against configured servers before any path interpolation.
- Open-redirect safety: the post-callback redirect back into the app must be a fixed internal path or a strictly validated allowlist, never attacker-controlled — same posture as the Globus callback's allowlisted error codes.
- The `redirect_uri` registered via DCR must be derived from a trusted configured base URL, not from the inbound `Host` header.
- Only permit the flow for HTTP/SSE transports (`_is_http_server`), consistent with the existing per-user auth restriction.
- On revocation/disconnect, invalidate cached FastMCP clients via `_invalidate_user_client` (already done for the manual-token path).

## Acceptance criteria

- [ ] `GET /api/mcp/auth/{server_name}/oauth/start` exists, is group-authorized, and 302s an authenticated user to the provider's authorization endpoint with PKCE and a session-bound `state`.
- [ ] `GET /api/mcp/auth/{server_name}/oauth/callback` validates `state` and user, exchanges the code, and persists access + refresh tokens through the existing encrypted per-user store.
- [ ] Authorization-server discovery works from the MCP URL alone, including the 401 `www-authenticate: resource_metadata=...` path.
- [ ] Dynamic Client Registration (RFC 7591) is performed when the AS advertises a `registration_endpoint` and no client credentials are cached; credentials are persisted per server and reused across restarts.
- [ ] `mcp_user_clients.py` builds an OAuth-aware FastMCP client for `auth_type: "oauth"` rather than passing the raw token as a bearer string.
- [ ] Expired access tokens are refreshed automatically; a refresh failure surfaces as a re-auth prompt rather than an opaque tool error.
- [ ] `oauth_config.scopes` and `client_name` are actually read and applied.
- [ ] `GET /api/mcp/auth/status` reports OAuth servers correctly (authenticated / expired / has refresh token), and the frontend surfaces a working Connect/Disconnect affordance; the `oauth_start_url` already returned by `AuthenticationRequiredException` resolves to a real route.
- [ ] `DELETE /api/mcp/auth/{server_name}/token` clears OAuth tokens, invalidates cached clients, and calls the provider's `revocation_endpoint` when advertised.
- [ ] Backend tests alongside `atlas/tests/test_mcp_client_auth.py` and `atlas/tests/test_token_storage.py` cover: discovery from a 401 challenge, DCR, state mismatch → rejection, cross-user callback → rejection, unauthorized server → 403, token persistence and refresh, and no secrets in logs. Suite run via `bash test/atlas_tests.sh`.
- [ ] `docs/admin/mcp-server-authentication.md` documents `auth_type: "oauth"`, `oauth_config`, the required redirect URI / base-URL setting, and how it differs from Globus.
- [ ] Manual verification against a real remote MCP server that requires OAuth 2.1 with DCR: a fresh user completes the browser flow and successfully lists and calls tools.

## Out of scope

- Changing or extending Globus OAuth.
- Admin-level / shared (non-per-user) OAuth credentials.
- The FastMCP 4 upgrade (#885), beyond keeping the client-auth API choice compatible.

garland3:
Closes #898.

## The problem

Atlas could *store* an OAuth access token for an MCP server but had no way to *obtain* one. Worse, it advertised a route that did not exist: a tool call against an `auth_type: oauth` server raised `AuthenticationRequiredException` carrying `/api/mcp/auth/<server>/oauth/start` (`mcp_execution.py:212-215,229,417,430`), the frontend surfaced that as a connect affordance, and the route answered **404**. `OAuthConfig` and `token_storage.update_oauth_tokens()` both existed and were referenced by nothing.

The only way to use such a server was to acquire a bearer token out of band and paste it into the upload endpoint — impossible where the provider relies on Dynamic Client Registration.

## What this does

Implements the flow end to end, server-side.

**Discovery.** An unauthenticated probe of the MCP endpoint reads the RFC 9728 `WWW-Authenticate: Bearer resource_metadata="..."` challenge, falling back to the well-known paths when a server publishes none. The protected-resource document names the authorization server; its RFC 8414 metadata supplies the endpoints. The document's `issuer` must match the issuer it was fetched for (RFC 8414 §3.3), so a redirect cannot substitute another provider. Results are cached with a TTL.

**Dynamic client registration (RFC 7591)** when the authorization server advertises a `registration_endpoint` and Atlas holds no credentials. Atlas registers as a public client using PKCE — once per MCP server, not per user — and persists the registration encrypted, so restarts reuse it rather than accumulating orphaned clients. A pre-registered `client_id` in `oauth_config` skips registration entirely for providers without DCR.

**Authorization Code + PKCE (S256)** with a single-use `state` bound to both the server and the user who started the flow, held in the browser session with a ten-minute TTL and a cap on concurrent attempts. An RFC 8707 `resource` indicator is sent when the protected-resource document publishes one, so the issued token is audience-bound.

**Per-user tokens** stored through the existing Fernet-encrypted store. An expired access token is refreshed silently from its refresh token before the user is ever asked to re-authorize; a refresh that fails degrades to "re-authorize", never to a failed tool call.

**Disconnect** revokes at the provider's `revocation_endpoint` when advertised, gated on the token having been issued by this flow — a hand-uploaded bearer token is never sent to a revocation endpoint. Local disconnect still succeeds when the provider is unreachable.

**Frontend.** The Tools panel auth indicator now covers `oauth` servers: the key icon starts the browser flow instead of opening the token modal. The callback returns with `?mcp_auth_server=&mcp_auth_success|mcp_auth_error`, which the panel surfaces as a banner, refreshes status from, and strips from the URL.

## Why not `fastmcp.client.auth.OAuth`

That helper binds a callback listener on `127.0.0.1` and calls `webbrowser.open()` — a single-user desktop assumption. Atlas is a multi-user server whose users browse from other machines, so the redirect target has to be an Atlas route. The resulting access token is handed to FastMCP as a bearer credential; this code owns discovery, registration, refresh and revocation.

## Security notes

- The `redirect_uri` comes from `MCP_OAUTH_REDIRECT_BASE_URL` (falling back to `BACKEND_PUBLIC_URL`), **never** from the inbound `Host` header.
- Every endpoint in the flow must be `https://` (loopback `http://` tolerated for local dev), *including URLs read out of discovery documents* — a compromised MCP server cannot downgrade the flow to cleartext.
- Provider-supplied error strings are mapped through an allowlist before being echoed into a redirect, following the pattern established in `oidc_auth_routes.py`.
- Both routes enforce group authorization via `get_authorized_servers`, and reject a callback whose state belongs to a different user or names a different server.
- Client registrations, including any issued `client_secret`, are encrypted with the same key as the tokens under a distinct salt.
- No token, authorization code, refresh token or client secret is logged.

## New settings

| Variable | Purpose |
|---|---|
| `MCP_OAUTH_REDIRECT_BASE_URL` | Public base URL for the callback. Falls back to `BACKEND_PUBLIC_URL`. |
| `MCP_OAUTH_SESSION_SECRET` | Only needed when neither OIDC nor Globus login is enabled — those already install session middleware and their secret is reused. |

The flow needs a browser session, which previously existed only under OIDC/Globus login. Both routes report `session_unavailable` rather than raising a 500 when no session middleware is installed.

## Not Globus, not OIDC login, not delegation

Four distinct things, configured independently, documented in `docs/admin/mcp-server-authentication.md`. Globus and OIDC login are single statically-registered providers for logging in *to Atlas*. `delegated` exchanges an existing OIDC token for a downstream credential with no browser interaction. This is per-MCP-server: N runtime-discovered resource servers, each with its own authorization server, DCR-issued credentials, and per-user tokens.

Existing auth modes are untouched — `api_key`, `jwt`, `bearer` and `delegated` behave exactly as before and the token-upload endpoints are unchanged.

## Verification

- **Backend suite:** `bash test/atlas_tests.sh` → **2952 passed, 17 skipped** (was 2845 before; 85 new tests here plus what landed on main).
- **Frontend:** `npx vitest run` → **819 passed, 72 files**. `npm run lint` → **0 errors** (4 pre-existing warnings, none in touched files).
- **PR validation:** `bash test/pr-validation/test_pr898_mcp_oauth_flow.sh` → **all 47 checks passed.** It stands up a real OAuth-protected MCP deployment on localhost — an endpoint that 401s with a genuine `resource_metadata` challenge, its protected-resource document, and an authorization server with DCR plus authorize/token/revoke endpoints that actually verify the PKCE `code_verifier` against the challenge and reject an unregistered `redirect_uri` — then drives the real Atlas app through its own routes: discovery, registration, the authorize redirect, a forged-state rejection, consent, the callback, encrypted storage, per-user isolation, using the token against the MCP server, silent refresh, and disconnect with revocation. Nothing inside Atlas is mocked.

New tests cover challenge parsing, well-known URL derivation, issuer-mismatch and plaintext-endpoint rejection, DCR body shape and error handling, PKCE and resource indicators on the token call, refresh-token carry-forward and rotation, revocation failure modes, the encrypted client store (round-trip, ciphertext, key rotation, issuer/server keying, secret expiry), route authorization (403/400), single-use state, cross-user and cross-server callback rejection, state expiry, error allowlisting, and the disconnect revocation gate.


garland3:
## Summary
Adds a browser-redirect OAuth flow for remote MCP servers: RFC 9728/8414 discovery, dynamic client registration with an encrypted client store, PKCE authorization and token refresh/revocation routes, plus Tools-panel UI for connecting and disconnecting servers.

## Verdict
Request changes.

## Required changes
1. Derive `_cookie_secure` from the MCP OAuth redirect base URL as well as `oidc_redirect_uri` at `atlas/main.py:387` — in the new header-auth deployment this PR enables, `oidc_redirect_uri` is unset, so the session cookie carrying the PKCE `code_verifier` and `state` is issued without `Secure` over https.
2. Constrain the `resource_metadata` URL from the `WWW-Authenticate` challenge to the configured MCP endpoint's origin, and disable `follow_redirects`, at `atlas/modules/mcp_tools/mcp_oauth.py:271` — as written a malicious MCP server can drive a server-side fetch to any host and name an attacker-controlled authorization endpoint for the user's consent redirect.
3. Surface the OAuth outcome outside the Tools panel body at `frontend/src/components/ToolsPanel.jsx:160-181` — persist the parsed notice (e.g. `sessionStorage`) or render it from an always-mounted component, since the effect strips `mcp_auth_*` from the URL on a page load where the panel is closed by default, so the default post-consent path shows no success or error at all.
4. Move the pending-authorization record (verifier, redirect_uri, user, server) into a server-side store keyed by an opaque handle at `atlas/routes/mcp_auth_routes.py:405`, or encrypt it with the existing Fernet key — Starlette sessions are signed but plaintext, so the PKCE verifier is readable from the cookie.
5. Percent-encode `server_name` and `code` in `_oauth_error_redirect` at `atlas/routes/mcp_auth_routes.py:328` — the value is unvalidated on the `session_unavailable` and `missing_params` branches and can inject extra query parameters into the SPA landing URL.
6. Map redirect error codes to user-facing sentences at `frontend/src/components/ToolsPanel.jsx:786-788` — raw codes like `discovery_failed` and `invalid_state` give the user no idea whether to retry or contact an administrator.
7. Add a component test for the banner and the `oauth` auth-status indicator branch in `ToolsPanel.jsx` — both new rendered states are currently exercised only at the hook level.

## Evidence
Evidence: 8/10. The backend is strongly demonstrated (85 unit tests plus a 47-check end-to-end run against a real OAuth-protected MCP deployment); the missing artifact is any visual evidence of the Tools-panel states — key icon, success banner, error banner.

## Notes
- Accept a trailing-slash difference when comparing the metadata `issuer` — `atlas/modules/mcp_tools/mcp_oauth.py:279`.
- Cache discovery failures and scope the discovery lock per-URL so one unreachable provider does not stall or repeatedly re-pay timeouts for others — `atlas/modules/mcp_tools/mcp_oauth.py:398`.
- Make `nonce`/`scope` optional in `build_authorize_url` instead of stripping empty params afterwards — `atlas/modules/mcp_tools/mcp_oauth_service.py:210-236`.

---

garland3:
Thanks — worked through all of the review feedback. Summary of what changed and why, plus the two items I did not act on.

**Validate every discovery redirect hop** (Codex P1 — the best catch here). Metadata fetches used `follow_redirects=True`, so `validate_endpoint_url` only ever saw the *first* URL. A compromised MCP server could answer an https request with a redirect to plaintext or to an internal address and defeat the transport restriction entirely. Redirects are now followed manually, bounded at 5 hops, with relative locations resolved and every destination validated before it is fetched. Three tests: a redirect to `http://` is refused, an https redirect is followed, a redirect loop terminates.

**Serialize first-time dynamic client registration** (Codex P2). Confirmed the race — two users starting authorization concurrently both miss `store.get()`, both register, and the second `put()` overwrites the first, so the first user's callback exchanges its code under a `client_id` the provider never issued it to. Now guarded by a per-`(server, issuer)` lock with a re-check inside. Test drives 5 concurrent `_resolve_client` calls and asserts exactly one registration.

**Serialize token refresh** (Codex P1). Same shape, worse blast radius: parallel tool calls against one expired token each presented the same refresh token, and a rotating provider accepts one and rejects the rest — so calls that could have used the newly stored token instead reported "re-authorize". Now a per-`(user, server)` lock, re-reading storage after acquiring it so the losers return the winner's token. Test runs 5 concurrent refreshes and asserts one provider call and five usable results. This also added a short-circuit test proving a still-valid token never hits the provider.

**Encode the outcome redirect query string** (Copilot). Fixed — server names are operator-supplied and one containing `&` or `=` would split into extra parameters. Both redirect sites now go through a single `urlencode`d helper, with a test using a deliberately hostile server name.

**Show the OAuth outcome banner** (Codex P2). Correct, and it was worse than described: the callback is a full-page navigation, so the app reloads with every panel closed and `ToolsPanel` — which was reading the query parameters — is never mounted to see them. `App.jsx` now consumes the parameters, stashes the result, and opens the Tools tab; the panel renders it from there and clears it so a reopen does not replay it. Writing the component test also surfaced a second bug: the banner markup sat *inside* the server-list branch, so it was invisible whenever the list was empty — exactly the state a failed authorization can leave you in. Moved above that branch.

**CHANGELOG length** (Copilot). Shortened to the 1–2 lines AGENTS.md asks for.

**Backend tests in the validation script** (Copilot). Added; `test/pr-validation/README.md` step 8 does document it and the other scripts do it.

Not acted on:

- **CodeQL `py/log-injection`** (14 alerts across the new modules): these are the project's known false positives. Every interpolated value goes through `sanitize_for_logging`, which the query does not model. I did fix all the *other* CodeQL findings on this PR — unused imports, side effects inside `assert`, unclosed file handles, and implicit/explicit return mixing.
- **`chatgpt-codex-connector`'s "add the required Unreleased changelog entry"**: stale. It reviewed `74ecb92`, and the entry landed in `2b0f8fe`.

Verification after all of the above: backend `2961 passed, 17 skipped`; frontend `822 passed` across 72 files; `npm run lint` 0 errors (4 pre-existing warnings, none in touched files); `bash test/pr-validation/test_pr898_mcp_oauth_flow.sh` all 47 checks passed plus the backend unit tests.

garland3:
## Summary
Adds a full browser-based OAuth flow for MCP servers: RFC 8414/9728 discovery, dynamic client registration with an encrypted client store, PKCE authorization via Atlas-hosted start/callback routes, silent token refresh, and Tools-panel UI for connect/disconnect with success and error banners.

## Verdict
Request changes.

## Required changes
1. Constrain the discovery fetch at `atlas/modules/mcp_tools/mcp_oauth.py:271` — require the `resource_metadata` URL from the MCP server's `WWW-Authenticate` header (and every redirect hop) to share the configured MCP endpoint's origin, and stop accepting loopback URLs read out of discovery documents; today a hostile endpoint drives server-side GETs to arbitrary internal hosts and can name an attacker-controlled authorization server.
2. Broaden the failure handling in `start_oauth` and `oauth_callback` at `atlas/routes/mcp_auth_routes.py:390` to catch non-`MCPOAuthError` exceptions and redirect with a generic error code — reachable `ValueError`/`OSError` (e.g. a string `client_secret_expires_at` in `register_client`) currently strand the browser on a raw 500 page.
3. Normalize both sides of the issuer comparison at `atlas/modules/mcp_tools/mcp_oauth.py:279` — only the requested issuer is `rstrip("/")`ed, so a provider publishing a trailing-slash issuer can never complete discovery.
4. Widen the revocation guard at `atlas/routes/mcp_auth_routes.py:264` to `except Exception` — the token is already deleted by that point, so a client-store `OSError`/`ValueError` turns a successful disconnect into a 500 and leaves the UI stuck.
5. Cache negative discovery results with a short TTL and key `_MetadataCache`'s lock per URL (`atlas/modules/mcp_tools/mcp_oauth.py:377,398`) — an unreachable provider re-pays ~70s of sequential timeouts on every tool call and its single global lock blocks discovery for healthy servers.
6. Keep the PKCE `code_verifier` out of the session cookie at `atlas/routes/mcp_auth_routes.py:405` — store an opaque handle and hold the verifier server-side (or encrypt the pending record), since Starlette sessions are signed but readable.
7. Derive the session cookie's `Secure` flag from `mcp_oauth_redirect_base_url`/`backend_public_url` as well as `oidc_redirect_uri` at `atlas/main.py:383` — the new header-auth deployment this PR enables has no OIDC redirect URI, so the cookie holding `state` and the verifier ships without `Secure` over https.

## Evidence
Evidence: 8/10. The backend is strongly demonstrated for a change of this size (85+ unit tests plus a 47-check end-to-end script against a real OAuth-protected MCP deployment); the gap is the user-facing half, where the new key icon and success/error banners are shown only by component tests — a screenshot of each new Tools-panel state, and one of the post-consent landing, is the missing artifact.

---

garland3:
## Summary
Adds browser-based OAuth 2.1 support for MCP servers: RFC 9728/8414 discovery, dynamic client registration, PKCE authorization and callback routes, an encrypted client store with refresh and revocation, plus Tools-panel UI for connecting and disconnecting.

## Verdict
Request changes.

## Required changes
1. Normalize both sides of the issuer comparison at `atlas/modules/mcp_tools/mcp_oauth.py:279` (`advertised.rstrip("/") != issuer.rstrip("/")`) — a provider publishing a trailing-slash issuer currently fails every discovery candidate and the server becomes permanently unusable with only `discovery_failed`.
2. Use the non-registering client lookup in the refresh path at `atlas/modules/mcp_tools/mcp_oauth_service.py:186` and return "re-authorization required" on a redirect-URI mismatch — re-registering mid-refresh rebinds the client and invalidates every user's stored refresh token at once.
3. Constrain the `resource_metadata` URL (and each redirect hop) fetched at `atlas/modules/mcp_tools/mcp_oauth.py:288` to the configured MCP endpoint's origin, and stop honoring the loopback/plaintext exemption for URLs read out of remote documents — today a hostile or MITM'd MCP server steers a server-side GET to internal addresses.
4. Derive `_cookie_secure` from `mcp_oauth_redirect_base_url` (falling back to `backend_public_url`) at `atlas/main.py:383` — in the new header-auth deployment shape `oidc_redirect_uri` is unset, so the session cookie carrying the PKCE verifier and state ships without `Secure`.
5. Coerce `client_secret_expires_at` defensively in `register_client` (`mcp_oauth.py:456`) and widen `start_oauth` / `oauth_callback` (`atlas/routes/mcp_auth_routes.py:390`) to catch `Exception` and redirect with a generic error code — a non-numeric provider value or an `OSError` currently leaves the browser on a raw 500 mid-flow.
6. Broaden the revocation guard at `atlas/routes/mcp_auth_routes.py:264` to `except Exception` — store construction failures return 500 after the token has already been deleted, so a successful disconnect reports failure.
7. Cache negative discovery results with a short TTL and key the metadata-cache lock per MCP URL (`mcp_oauth.py:377,398`) — one unreachable provider re-pays a ~70 s timeout chain per call and blocks discovery for healthy servers.

## Evidence
Evidence: 8/10 — the backend is unusually well demonstrated for a change this size (85+ new unit tests, concurrency tests, and a 47-check end-to-end run against a real OAuth-protected fixture). The gap is visual: no screenshot or recording of the new Tools-panel states or of the browser consent round-trip as a user experiences it.

## Notes
- Map redirect error codes to user-facing sentences and add a retry affordance on the error banner — `frontend/src/components/ToolsPanel.jsx:749-788`.
- `OAuthConfig` is extended but never validated; the service re-parses the raw dict field by field — `atlas/modules/config/models.py:197`.
- `AuthorizationServerMetadata.supports_refresh()` has no callers while `refresh_token` is requested unconditionally — `atlas/modules/mcp_tools/mcp_oauth.py:174`.

---

garland3:
Worked through the review. Pushed as `3f228cf`. Taking the required changes in order, including the two that were already done.

**1 & 2 — Constrain the discovery fetch / stop accepting loopback from discovery documents.** Agreed, and this was the real hole. My earlier fix validated the *scheme* of every redirect hop but not its *destination*, so a hostile MCP endpoint could still name any `resource_metadata` URL and have Atlas fetch it. RFC 9728 publishes that document on the resource's own origin, so it is now pinned there — for the challenge URL and for every hop — and an off-origin challenge is ignored with a warning rather than followed. Separately, `validate_endpoint_url` grew an explicit `allow_loopback` policy: loopback is permitted only when the MCP server is *itself* on loopback (the local-dev case the allowance exists for), so a remote server can no longer name `http://127.0.0.1:…` and borrow Atlas's position inside the network. Tests assert the off-origin challenge is never fetched, the off-origin redirect is never fetched, and a remote server naming a loopback authorization server fails discovery.

**Cookie `Secure` flag.** Correct and a good catch — this is precisely the deployment the PR enables. `_cookie_secure` now considers `mcp_oauth_redirect_base_url` and `backend_public_url` alongside `oidc_redirect_uri`.

**PKCE verifier in the session cookie.** Agreed. The pending record now lives in a server-side, TTL-bounded, single-use store keyed by the `state`; the cookie carries only the state values, which the provider already sees as a query parameter. The session remains the browser binding, so CSRF protection is unchanged — and this also closed a hole I had not noticed: a state present in the store but *not* in the presenting browser's session used to be consumed, letting a third party burn someone else's pending record. Now it is left alone. A test decodes the actual session cookie and asserts the verifier is absent and the state present.

**Broaden failure handling in `start_oauth` / `oauth_callback`.** Correct — a string `client_secret_expires_at` really does reach `float()`. Both routes now also catch non-`MCPOAuthError` and redirect with a generic code; tests inject `ValueError` and `OSError`.

**Widen the revocation guard to `except Exception`.** Correct — the token is already deleted at that point, so a store error was turning a successful disconnect into a 500. Fixed, with a test.

**Normalize both sides of the issuer comparison.** Correct: only the requested issuer was `rstrip`ed, so a provider publishing a trailing-slash issuer could never complete discovery. Both sides now normalized, non-string issuers rejected explicitly, and a genuinely different issuer still rejected.

**Cache negative discovery results and key the lock per URL.** Correct on both counts — one dead provider re-paid its whole timeout sequence per tool call, and its global lock stalled healthy servers. 60s failure TTL, per-URL locks. The isolation test starts a stalled discovery and resolves a healthy server while it is still in flight.

**Map error codes to user-facing sentences.** Done — each code now maps to a sentence saying whether to retry or to ask an administrator, with a fallback for unrecognized codes.

**Component tests for the banner and the `oauth` indicator branch.** Added: connect action offered when unauthenticated, disconnect once authenticated, expired treated as needing reconnection, clicking connect calls `startOAuth`, plus three banner message cases. (Making these work needed `vi.hoisted` — the suite's existing `vi.mock` factory can't close over a `let`.)

**Make `nonce`/`scope` optional in `build_authorize_url`.** Done, and the `_drop_empty_params` hack is gone.

Two items were already addressed before this review landed, so no change:

- **Percent-encode `server_name` and `code` in `_oauth_error_redirect`** — done in `f07d9b9`. Both redirect sites go through one `urlencode`d helper, `_oauth_error_redirect` included, so the `session_unavailable` and `missing_params` branches are covered. There's a test using a `odd&name=x` server name.
- **Surface the outcome outside the Tools panel body** — done in `f07d9b9`, and for exactly the reason given: the panel is unmounted when the callback lands. `App.jsx` consumes the parameters, stashes the result, and opens the Tools tab. Writing that also caught a second bug — the banner markup sat inside the server-list branch, invisible when the list is empty, which is the state a failed authorization leaves you in.

On the evidence gap: fair, the visual states were the weak spot. They now have real component tests, but I have not produced screenshots — that needs a browser against a running instance and I did not want to sit on the fix for it. Happy to add them if you want that artifact before merge.

Verification: backend `2982 passed, 17 skipped`; frontend `828 passed` across 72 files; `npm run lint` 0 errors (4 pre-existing warnings, none in touched files); `bash test/pr-validation/test_pr898_mcp_oauth_flow.sh` all 47 checks passed plus the backend unit tests.

garland3:
Third round — pushed as `f8a5d1e`. This review crossed with `3f228cf`, so most of it was already in flight; item 2 was new and was a real bug.

**2 — Use the non-registering lookup in the refresh path. Correct, and the most serious thing found on this PR.** I had added `_existing_client` for revocation but left refresh on the registering lookup, so a missing or stale registration would register a *new* client mid-refresh — and since the new `client_id` owns none of the existing refresh tokens, one user's expired token would silently invalidate **every** user's credential for that server. Refresh now uses the non-registering lookup and returns "re-authorization required" both when no registration is held and on a redirect-URI mismatch, exactly as suggested: re-registering to fix one user would break all the others. Three tests pin it, two of which fail loudly (`AssertionError("must not register during refresh")`) if the registering lookup ever comes back.

**5 — Coerce `client_secret_expires_at` defensively.** Agreed the broad `except Exception` I added was treating the symptom. Now handled where it is parsed: a non-numeric value logs a warning and is treated as non-expiring. The route-level catch stays as a backstop.

**Notes — `supports_refresh()` has no callers.** Right, it was dead. Now used: the `refresh_token` grant is only requested at registration when the server advertises it, or advertises nothing (an empty list means "did not say", not "refuses"). Tests cover both directions.

**Notes — retry affordance on the error banner.** Added, but deliberately not on every error: it appears for failures where retrying can help (declined consent, expired state, transient provider errors) and not for `discovery_failed` or `session_unavailable`, which will fail identically every time and where the message points at an administrator instead.

Already in `3f228cf`, pushed before this review posted — no further change:

- **1** issuer normalized on both sides, non-string issuers rejected.
- **3** `resource_metadata` and every redirect hop pinned to the MCP endpoint's origin; the loopback exemption no longer applies to URLs from remote documents.
- **4** `_cookie_secure` derives from `mcp_oauth_redirect_base_url` / `backend_public_url`.
- **6** revocation guard widened to `except Exception`.
- **7** 60s negative discovery cache and per-URL locks.

One note I don't think holds: **"`OAuthConfig` is extended but never validated; the service re-parses the raw dict field by field."** It *is* validated. `MCPToolManager` builds `servers_config` as `{name: server.model_dump() for ...}` from `MCPConfig` (`client.py:125,138`), so `oauth_config` has already been through Pydantic by the time the service sees it — what the service reads is a dumped `OAuthConfig`, not unvalidated JSON. The field-by-field accessors are defensive because `servers_config` is a plain dict everywhere else in the manager; re-parsing into the model here would be inconsistent with how every other server field is read. Happy to change it if you'd still prefer the model, but I don't want to claim a fix for something that isn't broken.

Still outstanding and acknowledged: **no screenshots of the Tools-panel states.** That is the one gap I haven't closed — it needs a browser against a running instance. Say the word and I'll do it before merge.

Verification: backend `2988 passed, 17 skipped`; frontend `831 passed`; `npm run lint` 0 errors (4 pre-existing warnings, none in touched files); `bash test/pr-validation/test_pr898_mcp_oauth_flow.sh` all 47 checks passed plus the backend unit tests.

garland3:
## Summary
Adds OAuth 2.0 authorization-code support for remote MCP servers: discovery, dynamic client registration, PKCE, encrypted per-user token/client storage, connect/disconnect routes, and Tools-panel UI for the connect flow.

## Verdict
Request changes.

## Required changes
1. Apply the loopback/private-address check to every scheme in `validate_endpoint_url` at `atlas/modules/mcp_tools/mcp_oauth.py:104` — the function returns early for any `https://` URL, so a hostile MCP server can name `https://127.0.0.1` or `https://10.0.0.5` as its authorization server and Atlas will probe internal hosts. Widen `_is_loopback` to reject loopback, link-local, private and reserved ranges (including `0.0.0.0`, `::ffff:127.0.0.1`, integer-encoded forms), re-check per redirect hop, and add tests for an `https` loopback and an `https` private-range issuer.
2. Cap response size and check content type in `_fetch_json` at `atlas/modules/mcp_tools/mcp_oauth.py:296`, and in `register_client` and `_post_token_endpoint` — provider-controlled bodies are read into memory unbounded, so one click on "connect" can exhaust a worker.
3. Route the connect action through the existing unsaved-changes guard at `frontend/src/components/ToolsPanel.jsx:945` before `startOAuth` navigates — the full-page redirect bypasses `closeGuardRef` and silently discards pending tool selections (same for the banner's "Try again").
4. Fix `test_a_success_clears_a_cached_failure` in `atlas/tests/test_mcp_oauth.py` to exercise failure-cache expiry rather than clearing the cache, and repoint the clock patch in `test_expired_pending_state_is_rejected` (`atlas/tests/test_mcp_oauth_routes.py`) at the module the store actually reads — both currently pass without testing their stated behavior.
5. Bound or evict `_registration_locks` and `_refresh_locks` at `atlas/modules/mcp_tools/mcp_oauth_service.py:44` — the per-user/per-server dicts are module-global and never pruned, so they grow for the process lifetime.
6. Document the single-worker constraint in `atlas/modules/mcp_tools/oauth_pending_store.py` and `docs/admin/mcp-server-authentication.md` — the PKCE verifier, discovery cache and registration locks are process-local, so a multi-worker deployment fails `invalid_state` intermittently with no diagnostic.

## Evidence
Evidence: 8/10. The backend is strongly demonstrated — 85+ new unit tests plus an end-to-end PR-validation script that drives a real OAuth-protected deployment with nothing inside Atlas mocked — but the user-facing half is proven only by component tests; the missing artifact is screenshots (or a recording) of the new Tools-panel states: key icon, connected shield, success banner, and a retryable error banner.

## Notes
- Connect gives no spinner or disabled state during multi-second discovery, inviting double-clicks — `frontend/src/hooks/useServerAuthStatus.js:118`.
- An expired-but-refreshable server renders as fully disconnected — `frontend/src/components/ToolsPanel.jsx:934`.
- Error copy surfaces a server-side environment variable name to end users — `frontend/src/components/ToolsPanel.jsx:65`.

---

garland3:
Two follow-ups pushed (`a11f152`, `72aaa9a`), both from things I found rather than new review items.

**A test that was not testing anything.** CodeQL flagged an unused local in `test_one_stalled_server_does_not_block_another` — and it was right, in a way that mattered. I had defined a stalling handler and then passed a *different*, non-stalling handler to the transport, so nothing was ever in flight: the test asserted only that two sequential discoveries both succeed, and would have passed just as happily with the global lock it was supposed to catch. It now hangs the unhealthy provider until the test releases it and waits for that request to be genuinely in flight first.

I checked that the fixed test can actually fail: pointing the per-URL lock map at a single shared lock makes it time out, and it passes again on revert. The per-URL lock claim from the last round is now backed by a test that would notice if it regressed.

Also from CodeQL on the same pass: exact origin comparisons instead of substring containment in the discovery tests, and the validation fixture's authorize endpoint now urlencodes its redirect rather than interpolating a `state` that could carry CR/LF into a response header.

**An honest note in the docs about what origin-pinning does not cover.** Pinning the `resource_metadata` URL closes that hole, but the authorization server legitimately lives on a *different* origin from the resource — that is the normal shape — so its URL cannot be pinned. A hostile MCP server can therefore still cause one credential-free `GET` to `https://<host-it-names>/.well-known/oauth-authorization-server`, whose response must parse as valid AS metadata to go further. That request does still reach the named host from inside the deployment's network. I would rather write that down in the security notes than let the previous round's "constrained discovery" read as more complete than it is; closing it properly needs an egress allowlist, which is a deployment control rather than something this PR should invent.

Verification: backend `2988 passed, 17 skipped`, and the same **in reverse collection order** (I ran `ATLAS_TEST_ORDER=reverse` locally, since I had added a lot of tests with module-level caches and lock maps since the last green CI); frontend `831 passed`; lint 0 errors; validation script all 47 checks plus backend unit tests.

The screenshots remain the one open item — still happy to produce them on request.

garland3:
Fourth round — pushed as `5a3541d`. Item 1 was a real hole and my previous round's claim was wrong, so leading with that.

**1 — The address check never ran on https. Correct, and my last round overstated what was fixed.** `validate_endpoint_url` returned early for *any* `https://` URL, so the loopback restriction I added only ever applied to the `http://` branch. `https://127.0.0.1`, `https://10.0.0.5` and `https://169.254.169.254/latest/meta-data/` all sailed through — which is exactly the internal probe origin-pinning was meant to stop, reachable through the one URL that legitimately cannot be origin-pinned (the authorization server). Now the check runs for every scheme and covers loopback, private, link-local, reserved, unspecified and multicast ranges, IPv4-mapped IPv6, and the integer/hex encodings (`2130706433`, `0x7f000001`) that a string-prefix allowlist misses — parsing via `ipaddress` rather than by hand. Fourteen parametrized rejection cases plus non-`http` schemes, and two positive cases so it cannot pass by rejecting everything.

I did *not* resolve DNS names: that only moves the problem to rebinding and adds a resolver dependency in the request path. The literal-address case is what a hostile document actually uses; the residual DNS surface is now written down in the admin docs alongside the authorization-server note.

**2 — Unbounded response bodies.** Agreed. Discovery, registration and the token endpoint now stream against a 512 KiB ceiling, reject a mismatched content type (accepting `+json` suffixes), and do not trust `Content-Length` — there's a test with a body that lies about its length.

**4 — Two more tests that did not test their stated behavior.** Both confirmed, and this is the second round running where that has been the finding, which I take as a signal about how I was writing them. The failure-cache test cleared the cache instead of letting the TTL lapse; it now asserts no re-fetch *inside* the TTL, recovery after it, and that success replaces the cached failure. The pending-expiry test patched the clock in the routes module after expiry had moved into the store; it now patches what the store reads and asserts the record exists first, so it cannot pass for the wrong reason.

**5 — Unbounded lock maps.** Fixed: entries not currently held are evicted past a ceiling. A held lock is never evicted, so mutual exclusion for an in-flight caller is unaffected.

**6 — Single-worker constraint.** Documented in both places, and worth calling out as the sharpest operational point in this review: the failure mode is intermittent `invalid_state` at roughly `(workers-1)/workers` of attempts, indistinguishable in the logs from a genuine CSRF rejection. The docs now say what a multi-worker deployment would need (sticky sessions, or a shared encrypted store).

**Notes — error copy naming a server-side env var.** Fixed; end users now get "ask an administrator to finish configuring it" without `MCP_OAUTH_SESSION_SECRET`.

Two items from this round I have **not** done, so they don't get lost:

- **3 — connect bypasses the unsaved-changes guard.** Real: the full-page redirect discards pending tool selections without warning, and the same applies to "Try again". Routing it through `closeGuardRef` is the right fix; I'd rather do it deliberately than rush it in alongside a security change.
- **Notes — no spinner during discovery, and expired-but-refreshable rendering as disconnected.** Both fair; the second is arguably the more misleading of the two, since that server will silently work again on the next tool call.

Happy to take those next, along with the screenshots.

Verification: backend `3010 passed, 17 skipped`, and the same in reverse collection order; frontend `831 passed`; lint 0 errors; validation script all 47 checks plus backend unit tests.

garland3:
## Summary
Adds end-to-end OAuth support for MCP servers: RFC 8414/9728 discovery, dynamic client registration, PKCE authorization via new Atlas redirect routes, an encrypted client store and in-process pending-authorization store, token refresh in the client factory, and a connect/disconnect affordance with status banners in the Tools panel.

## Verdict
Request changes.

## Required changes
1. Apply the loopback/private-address policy to `https` URLs, not just `http`, at `atlas/modules/mcp_tools/mcp_oauth.py:104` — a remote MCP server can currently publish `https://169.254.169.254` or an RFC 1918 address as its authorization server and Atlas will fetch it from inside the deployment network.
2. Replace the string-based host matching in `_is_loopback` (`mcp_oauth.py:104`) with `ipaddress.ip_address()` checks rejecting loopback, private, link-local, reserved, unspecified and IPv4-mapped IPv6 forms, and re-apply the check on each redirect hop in `_fetch_json` — `0.0.0.0`, `::ffff:127.0.0.1` and `2130706433` bypass the current check.
3. Cap and content-type-check provider response bodies before parsing in `_fetch_json`, `register_client` and `_post_token_endpoint` (`mcp_oauth.py:296` and the two later call sites) — an unbounded `response.json()` on a provider-controlled body lets one connect click OOM a worker.
4. Carry the `client_id` used at `/oauth/start` in the `PendingAuthorization` record and have `complete_authorization` (`atlas/modules/mcp_tools/mcp_oauth_service.py:314`) use exactly that client instead of `_resolve_client` — if the registration expires or is reset mid-flow, the code is redeemed under a client the provider never issued it to.
5. Route the connect and "Try again" actions through the panel's existing dirty-state guard at `frontend/src/components/ToolsPanel.jsx:945` and `:812` — `startOAuth` does a full-page navigation and silently discards unsaved tool selections.
6. Bound the module-global `_registration_locks` and `_refresh_locks` at `mcp_oauth_service.py:44` (reference-count and drop unheld locks, or use a capped LRU) — they accumulate one lock per distinct user/server pair for the process lifetime.
7. Fix `TestDiscoveryFailureCaching.test_a_success_clears_a_cached_failure` and `TestOAuthCallback.test_expired_pending_state_is_rejected` — the first calls `clear_metadata_cache()` so the failure-TTL path is never executed; the second patches `mcp_auth_routes.time.time` while the expiry decision reads `time` from `oauth_pending_store`. Also add a test through the enclosing client-creation path in `mcp_user_clients.py:294`, since deleting the new refresh branch leaves the suite green.

## Evidence
Evidence: 8/10. The backend is unusually well demonstrated — ~85 new unit tests exercising the security properties by name plus a 47-check end-to-end validation script — but the user-facing half is shown only in code; the most valuable missing artifact is a screenshot of the error banner rendered with an empty server list, the exact regression the banner fix targets.

## Notes
- Rewrite the `session_unavailable` copy to drop the `MCP_OAUTH_SESSION_SECRET` variable name from user-facing text — `ToolsPanel.jsx:69`.
- Distinguish "expired but auto-refreshable" from "never connected" in the auth-status icon — `ToolsPanel.jsx:934`.
- Document `MCP_OAUTH_SESSION_SECRET` and the single-worker/sticky-session requirement of the in-process pending store — `docs/admin/mcp-server-authentication.md:1`.

---

garland3:
Picked up the two UI items I'd deferred an hour ago rather than leaving them open — pushed as `a307bdd`.

**3 — Connect bypassed the unsaved-changes guard.** Fixed. `ToolsPanel` already has exactly the right mechanism for this: `handleCloseAttempt(onCancelled, afterClose)`, which is how navigating to the marketplace defers its action until the user has answered the dialog. Starting OAuth is the same shape — a navigation away that would discard staged selections — so it now goes through the same path, and the redirect only fires once the user has answered. The banner's "Try again" routes through it too. Test: clicking connect with a staged tool selection raises the dialog and does *not* call `startOAuth`.

**Note — expired-but-refreshable rendered as disconnected.** Fixed, and I agree this was the more misleading of the two notes. The indicator keyed on `is_expired` alone, so a server holding a perfectly good refresh token showed the yellow "connect" state and invited the user through a browser flow they did not need — Atlas renews it silently on the next tool call. Expiry now only means disconnected when there is no refresh token, and the connected tooltip says access renews automatically. Two tests, one for each side of that.

Still open from that round: **no spinner during discovery**. Real, but it's a different shape of change (the hook would need to expose pending state and the button a disabled/spinner rendering) and I'd rather not tack it onto a commit that is otherwise about correctness. Same for the screenshots.

Verification: backend `3010 passed, 17 skipped`; frontend `833 passed`; lint 0 errors.

garland3:
## Summary
Adds end-to-end OAuth support for MCP servers: discovery (protected-resource and authorization-server metadata), dynamic client registration, PKCE authorization start/callback routes, encrypted client and pending-authorization stores, silent token refresh, revocation on disconnect, and the Tools-panel connect/disconnect UI with result banners.

## Verdict
Request changes.

## Required changes
1. Carry `client_id` (and `issuer`) in `PendingAuthorization` and redeem the code with that identity at `atlas/modules/mcp_tools/mcp_oauth_service.py:297` — `complete_authorization` currently calls the *registering* `_resolve_client`, so a lapsed/reset registration triggers a fresh DCR inside the callback, redeems the code under a `client_id` the provider never issued it to, and rebinds the server for all other users. Fail with a re-authorize error instead, mirroring `_refresh_locked`.
2. Apply `_is_internal_address` to every URL regardless of scheme and of `allow_loopback` at `atlas/modules/mcp_tools/mcp_oauth.py:288` — the `https` early return in `validate_endpoint_url` lets a remote server redirect discovery to `https://169.254.169.254` or an RFC 1918 host.
3. Resolve the authorization-server hostname and reject internal resolved addresses at `atlas/modules/mcp_tools/mcp_oauth.py:118` — a DNS name always passes the literal-IP check, so `authorization_servers: ["https://name-resolving-to-169.254.169.254"]` yields an unauthenticated internal fetch. If pinning is out of scope here, state the DNS gap explicitly in `docs/admin/mcp-server-authentication.md`.
4. Percent-encode the server-name segment in `redirect_uri_for` (`mcp_oauth_service.py:142`) and in `oauth_start_url` (`atlas/routes/mcp_auth_routes.py:112`) — the frontend uses `encodeURIComponent`, so a name containing `#`, `?` or a space registers a redirect URI that never matches the callback.
5. Make `build_authorize_url` keyword-only after `authorization_endpoint` at `atlas/core/oidc/oidc_client.py:63` — moving `scope`/`nonce` to the end silently rebinds any positional caller instead of raising.
6. Bound total discovery time with a shared deadline in `_MetadataCache._discover` (`mcp_oauth.py:700`) — sequential 10s timeouts across candidates can hold a request, and the per-URL lock, for ~2 minutes against a slow provider.
7. Add a test that exercises the silent-refresh call site in the user-client factory at `atlas/modules/mcp_tools/mcp_user_clients.py:302` (expired `oauth_access` + patched `refresh_stored_token` → client returned; refresh returns None → None) — deleting that line currently leaves the suite green.

## Evidence
Evidence: 8/10. The backend is unusually well demonstrated for a change this size (~165 targeted tests plus a 47-check end-to-end validation script against a real OAuth-protected deployment); the main gap is visual evidence of the new Tools-panel states, most importantly the error banner rendered with an empty server list.

## Notes
- Cookie state cap of 5 vs. a 2048-entry server-side store makes a sixth concurrent flow fail as `invalid_state` — `atlas/routes/mcp_auth_routes.py:377`.
- Log "state absent from session" and "state expired" distinctly from "state unknown" — `atlas/routes/mcp_auth_routes.py:451`.
- Show a disabled/spinner state while `/oauth/start` runs discovery, and use `role="alert"` for the error banner — `frontend/src/components/ToolsPanel.jsx:527,812`.

---

garland3:
Fifth round — pushed as `8da8b1f`.

**Item 1 (client identity in the callback) was real and is the same class of bug as round 3's refresh finding.** `complete_authorization` now carries `client_id` and `issuer` on the pending record, resolves through the non-registering `_existing_client`, and raises "connect again" when the stored registration no longer matches — no DCR can happen inside a callback. Three tests, including one that asserts `exchange_authorization_code` is never reached.

**Item 2 — no-op, already fixed in `5a3541d`.** `validate_endpoint_url` runs the `_is_internal_address` check for *every* scheme before the `https` return (mcp_oauth.py:174-184); the line number in the review is from an earlier commit. It is still skipped when `allow_loopback` is True, which is only ever the case when the MCP server is itself on loopback — that is the local-development allowance, not a discovery-document path.

**Item 3 — documented rather than closed.** New "Known limitation: discovery URLs are not DNS-resolved" section in `docs/admin/mcp-server-authentication.md`: literal internal addresses are rejected, a DNS name that resolves to one is not, the consequence is one unauthenticated GET whose body never reaches the caller, and the mitigation is egress control. Resolving-and-pinning would need connect-time address checks to be race-free; that is a larger change than this PR should carry.

**Item 4 — fixed.** `quote(server_name, safe='')` in both `redirect_uri_for` and the status route's `oauth_start_url`, with a test that reads the status payload for a server named `a b#c` and asserts both halves agree segment-for-segment.

**Item 5 — no-op.** `build_authorize_url` has been keyword-only since it was written: the signature opens with a bare `*` (oidc_client.py:60-61), so no positional caller is possible.

**Item 6 — fixed.** `DISCOVERY_TOTAL_TIMEOUT_SECONDS = 30.0` wraps `_discover` in `asyncio.wait_for`; the timeout is cached like any other discovery failure, so a slow provider is paid for once per failure TTL rather than per tool call. Two tests.

**Item 7 — fixed, and you were right that it was untested.** The existing tests called `_refresh_oauth_token` directly; nothing exercised the call site. Two new tests in `test_mcp_client_auth.py`. Verified they catch it: replacing `mcp_user_clients.py:302` with `pass` fails both.

**Notes** — cookie cap raised 5 → 16; "state absent from the session" and "state was in the session but has no pending record" now log distinctly; `role="alert"` on the error banner (`role="status"` retained for success); connect button shows a spinner, a "Contacting the authorization server..." label and `disabled` while discovery runs, with a test asserting a second click cannot start a duplicate flow.

**Verification on this tree**
- `bash test/atlas_tests.sh` — 3019 passed, 17 skipped
- `npx vitest run` — 834 passed across 72 files
- `npm run lint` — 0 errors (4 pre-existing warnings, none in touched files)
- `bash test/pr-validation/test_pr898_mcp_oauth_flow.sh` — all 50 checks passed, plus backend unit tests

Still outstanding and still acknowledged: screenshots of the Tools-panel states.

garland3:
## Summary
Adds full OAuth 2.0 support for MCP servers: discovery with origin pinning, dynamic client registration, PKCE with session-bound state, two new encrypted at-rest stores, browser redirect/callback routes, token refresh and disconnect-with-revocation, plus Tools-panel UI for the new states.

## Verdict
Request changes.

## Required changes
1. Normalize partial and octal dotted IPv4 forms in `_parse_ip` at `atlas/modules/mcp_tools/mcp_oauth.py:104` — `127.1`, `0177.0.0.1` and `0x7f.1` fail both `ipaddress.ip_address()` and `int(host, 0)`, so `validate_endpoint_url` accepts them with `allow_loopback=False` while httpx still resolves them, reopening the internal-fetch hole. Use `socket.inet_aton` (the resolver's own grammar) before the range test, and add those four literals to the rejection parametrization.
2. Apply the `MAX_RESPONSE_BYTES` ceiling to the two remaining outbound calls — `probe_resource_metadata_url` (`atlas/modules/mcp_tools/mcp_oauth.py:330`) and `revoke_token` (`:508`) use the buffering `post()` API against attacker-controlled endpoints. Switch both to `build_request` + `send(stream=True)` with `aclose()` in a `finally`; the probe needs only status and `WWW-Authenticate`, so its body need never be read.

## Evidence
Evidence: 8/10 — the backend half is unusually well demonstrated (full suite plus reverse-order re-run, a 50-check end-to-end script driving a real OAuth-protected MCP deployment with nothing mocked, and negative verification of the concurrency fixes). The one missing artifact is any rendered proof of the new user-facing Tools-panel states; screenshots of the connect, expired-refreshable and post-callback banner states would close it.

## Notes
- Disconnect dialog tells OAuth users to "re-enter" a token they never entered, and omits that revocation may sign out every device — branch the copy on `auth_type` (`frontend/src/components/ToolsPanel.jsx:1311`).
- `MCPOAuthClientStore` re-implements the salt/KDF/atomic-write logic of `token_storage.py` line for line; extract a shared encrypted-JSON store (`atlas/modules/mcp_tools/oauth_client_store.py:60-115`).
- Disconnect can hold the request for ~70s of provider I/O and silently skips revocation when the tool manager is unavailable (`atlas/modules/mcp_tools/mcp_oauth_service.py:470`, `atlas/routes/mcp_auth_routes.py:262`).

---

garland3:
## Summary
Adds end-to-end OAuth support for MCP servers: discovery and dynamic client registration, PKCE authorization via Atlas-owned browser redirect routes, encrypted at-rest client/token stores, silent refresh with per-key locking, provider-side revocation on disconnect, and Tools-panel connect/disconnect UI.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. The backend is demonstrated well beyond the usual bar — named suite results, a reverse-order re-run to catch module-global coupling, and a 50-check PR-validation script that stands up a real OAuth-protected MCP deployment with nothing internal mocked. The one missing artifact is any rendered view of the new user-facing states; the highest-value single screenshot is the error banner with an empty server list, since that is the exact regression the banner relocation fixed.

## Notes
- SSRF guard misses shorthand/octal IPv4 literals (`127.1`, `0177.0.0.1`) that `inet_aton` still expands to loopback — normalize via `socket.inet_aton` before the range test, `atlas/modules/mcp_tools/mcp_oauth.py:104`.
- `probe_resource_metadata_url` and `revoke_token` still use buffering `httpx.post()`, bypassing the `MAX_RESPONSE_BYTES` ceiling every other outbound call now honors — `atlas/modules/mcp_tools/mcp_oauth.py:330`, `:508`.
- Disconnect can block ~70s on sequential discovery + two revocation POSTs with no pending UI state; cap it with a deadline and disable the confirm button while in flight — `atlas/modules/mcp_tools/mcp_oauth_service.py:470`, `frontend/src/components/ToolsPanel.jsx:496`.

---

garland3:

PR #900, opened 2026-09-08
fix(config): preserve invalid LLM model validation errors

garland3:
## Summary

Invalid per-model `reasoning_effort` values now propagate as configuration validation errors instead of being caught by the generic LLM loader fallback and replaced with zero models. The validation log includes the searched config paths.

Regression coverage uses real temporary YAML files for invalid, missing, and valid configurations. The admin contract now documents the refusal-to-start behavior.

## Hooks output-cap deadlock (CI fix)

`build-and-test` was failing on `test_output_over_cap_is_killed_and_treated_as_error`, which asserted `"1 MB" in outcome.reason` but got `"Hook 'f' errored: timeout"`.

This was a real bug, not a flaky test. On overflow the reader killed the child and then awaited its full reap **from inside the coroutine draining that child's stdout**. `asyncio` completes `Process.wait()` only once every stdio pipe has disconnected (`BaseSubprocessTransport._try_finish`), so once the reader stopped draining, the pipe never reached EOF and the reap never finished. The hook then burned its entire 20s timeout and reported `timeout` instead of the cap. It reproduced on CI and not locally because asyncio's transport reads eagerly until flow control pauses it — the deadlock needs the `StreamReader` to be paused at the moment of overflow, which happens under a slow/loaded runner.

The reader now signals the kill without awaiting and runs on to EOF, discarding the overflow, so the child is reapable; the buffer still stops growing at the cap, so the stream is never fully buffered. Reaping is additionally bounded, because the timeout path cancels the readers and can otherwise wait on a pipe nobody will drain.

The regression test uses a fake process that models asyncio's reaping order, so it deadlocks deterministically against the old code instead of depending on pipe-buffer timing.

## Verification

- `bash test/atlas_tests.sh` — 2872 passed, 17 skipped
- New test confirmed to hang (`TimeoutError`) with the `manager.py` fix reverted, and pass with it
- `ruff check atlas/hooks/manager.py atlas/tests/test_hooks_engine.py`

Also merged `main` to clear a `CHANGELOG.md` conflict that had the PR in `DIRTY` state.

Replaces #897; closes #897 after this PR is opened.

https://claude.ai/code/session_01D8g5oLorMWqruknYEbmzMP


garland3:
## Summary
Makes `ConfigManager.llm_config` re-raise pydantic `ValidationError` instead of falling back to an empty model set, logs the searched config paths on failure, and adds three regression tests plus admin-doc and CHANGELOG updates.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. The change is well demonstrated for its size — three regression tests over real temporary YAML files, named verification commands with pass counts, lint runs, and updated docs. The one gap worth closing later is that nothing ties the property-level raise to actual application startup, which is the behavior the admin doc promises.

## Notes
- `self.app_settings` is evaluated inside the guarded block, so an environment-variable validation failure is reported as "LLM configuration validation failed; searched: []" — resolve the filename before the `try` (`atlas/modules/config/config_loader.py:192`).
- The raised error carries pydantic's `input_value`, which can include `api_key` and `extra_headers` from every configured model; consider a redacted message built from `exc.errors()` and `SecretStr` for `ModelConfig.api_key` (`atlas/modules/config/config_loader.py:205`).
- The failure is not cached, so each later access re-reads the YAML and re-logs a full traceback; also note that only `llm_config` is fail-fast while `mcp_config` and the other loaders still swallow validation errors (`atlas/modules/config/config_loader.py:188`).

---

garland3:
## Summary
Makes `ConfigManager.llm_config` re-raise pydantic `ValidationError` instead of silently falling back to an empty `LLMConfig`, logs the searched config paths on failure, and adds three regression tests plus CHANGELOG and admin-doc updates.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — the change is adequately demonstrated for its size, with tests over real temporary YAML files for the invalid, missing and valid paths and named verification commands. The one missing artifact is any observation of the new error log line (a `caplog` assertion or pasted output showing the searched paths).

## Notes
- Failure is not memoized — `self._llm_config` stays `None`, so every later access re-reads and re-parses the YAML and logs again (`atlas/modules/config/config_loader.py:188`).
- Doc says the raised error "includes the configuration search paths", but they only appear in the preceding log record — align the wording or attach them to the exception (`docs/admin/llm-config.md:298`).
- `list_models`/`show_config` now exit with a raw traceback instead of the CLI's friendly error text (`atlas/modules/config/cli.py:45`).

---

garland3:
## Summary
Makes `ConfigManager.llm_config` fail fast on a schema-invalid `llmconfig.yml` instead of silently returning zero models, and fixes a hooks deadlock where output over the 1 MB cap was reported as a timeout rather than a cap violation (drain to EOF plus a bounded reap).

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — both halves are backed by direct regression tests, including a negative control confirming the hooks test hangs with the fix reverted. The most notable gap is that no passing CI run on this head is cited, even though the hooks failure only reproduced on the loaded runner.

## Notes
- `except ValidationError: raise` leaves `self._llm_config` as `None`, so every access re-reads and re-parses the file and emits another error record — consider caching the failure (`atlas/modules/config/config_loader.py:188`).
- `ModelConfig.api_key` is a plain `str`, so a propagated pydantic error can render the raw model entry (key included) into logs; `SecretStr` would redact it (`atlas/modules/config/models.py:87`).
- Docs promise the error "includes the configuration search paths", but they only appear in the separate `logger.error` line — reword or attach them to the raised exception (`docs/admin/llm-config.md:298`).

---

garland3:
Thanks — triaged the three bot reviews together, since the notes overlap. Addressed in 0735286.

**Fixed**

- **Failure not cached** (raised in all three reviews). Confirmed: re-raising left `_llm_config` as `None`, so every access re-read and re-parsed the file and emitted another `logger.error`. The first `ValidationError` is now cached and re-raised; `reload_configs()` clears it so a corrected file is still picked up. Two regression tests — one asserts `_search_paths` is consulted exactly once across two accesses (it fails against the old code, with the duplicated error record visible in the captured log), one asserts reload recovers.
- **Doc overclaims.** Correct — `docs/admin/llm-config.md` said the raised error "includes the configuration search paths", but the paths are only in the accompanying log record and the exception is pydantic's own. Reworded to describe what actually happens rather than attaching paths to the exception, which would mean changing the exception type (see below).

**Not changed, with reasons**

- **`self.app_settings` evaluated inside the guarded block, giving `searched: []`.** This one doesn't hold against the current code: `llm_filename` and `file_paths` are both resolved *before* the `try` (`config_loader.py:189-190`), so the path list is always populated. No change needed.
- **`SecretStr` for `ModelConfig.api_key`.** The underlying concern is real — pydantic's `missing`-type errors report the *parent* dict as `input`, so a model entry missing a required field can render its `api_key` into a traceback. But `SecretStr` is cross-cutting: every consumer that passes `api_key` through to LiteLLM needs auditing for `.get_secret_value()`, and a missed one fails at request time, not load time. That's wider than this PR and deserves its own change with its own test pass. The narrower alternative — wrapping the error in a redacted `ConfigurationError` — would break the contract these tests assert (`pytest.raises(ValidationError, match="meduim")`) and the behavior the PR documents. Flagging for a follow-up issue rather than doing it here.
- **`list_models`/`show_config` exit with a raw traceback** (`cli.py:45`). Accurate, and a consequence of fail-fast being the point of this change. Better handled where the CLI formats its other errors, not by re-softening the loader.

**On the CI evidence gap** — fair, and now closed: the earlier `build-and-test` failure was the hooks deadlock, which is fixed in 7759fbb. Full local run is `2874 passed, 17 skipped`, and CI is running green-path on this head.

garland3:
## Summary
Fixes a hook-subprocess deadlock by draining pipes to EOF after the 1 MB output cap trips and bounding the child reap, and makes `ConfigManager.llm_config` fail fast on a schema-invalid `llmconfig.yml` (caching the error) instead of silently falling back to an empty model list.

## Verdict
Request changes.

## Required changes
1. Bound the post-overflow drain at `atlas/hooks/manager.py:492` — SIGKILL reaches only the direct child, so a hook that backgrounds a process holding stdout never reaches EOF and the outcome is misreported as `timeout` rather than the output cap, which is the symptom this PR set out to fix. Wrap the residual reads in a short deadline and return the prefix on expiry (killing the process group would also close the inherited fd).
2. Prevent credential material from reaching logs and error surfaces at `atlas/modules/config/config_loader.py:205` — pydantic's `ValidationError` embeds the offending input, and for `missing`-type errors that is the whole model entry including `api_key` and `extra_headers`, which the raw re-raise now exposes to any handler that formats it. Either type those fields as `SecretStr` (updating consumers to `.get_secret_value()`) or build the raised message from `exc.errors()` with `input`/`ctx` stripped.
3. Handle `ValidationError` in `list_models` and `show_config` at `atlas/modules/config/cli.py:41` — these diagnostic commands now exit with a raw traceback on exactly the bad configs they exist to explain; print the same `❌`-style message used by `validate_config` and exit non-zero.
4. Document the caching in `docs/admin/llm-config.md:298` — the first failure is retained for the life of the process, so an operator who corrects `llmconfig.yml` and retries in-process sees the identical error and concludes the fix did not work.
5. Add coverage for the "keep draining the other pipe" branch, e.g. a `_FakeProc` with output on both streams asserting both reach EOF, `overflowed is True`, and the combined length stays within the cap — the current regression test leaves stderr empty, so deleting that branch still passes.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — a full local suite run plus a deterministic fake-process regression test for the deadlock; the one missing artifact is a completed `build-and-test` run on this head, the environment where the original failure actually reproduced.

## Notes
- Re-raising the cached exception instance grows its traceback on every access — `raise ... .with_traceback(None)` at `atlas/modules/config/config_loader.py:191`.
- `_REAP_TIMEOUT_S` is charged on top of the configured hook timeout and has no test; scale it to the remaining budget and log expiry above `debug` (`atlas/hooks/manager.py:337`, `:440`).
- Fail-fast now applies to `llm_config` alone while the other loaders still swallow validation errors — state that asymmetry in the class docstring (`atlas/modules/config/config_loader.py:187`).

---

garland3:
All five required changes are in `2fb2bea`. Two of them found real problems — thanks.

**1. Bound the post-overflow drain — fixed, and you were right that this reintroduced the original symptom.** My previous commit drained to EOF assuming SIGKILL closes stdout; it only reaches the direct child, so a hook that backgrounds something holding that fd never sends EOF and the drain ran to the hook timeout — reporting `timeout` for a cap violation, exactly what this PR set out to fix. The tail drain now carries its own deadline.

One wrinkle worth recording: my first attempt bounded only the reads *after* overflow was observed, and the new test caught that it was insufficient — the sibling stream was already parked in an unbounded `read()` and never re-checked the deadline. Reads now poll (`_DRAIN_POLL_S`), so both pipes observe the deadline. Cancelling a `StreamReader.read()` consumes nothing, so polling costs no data.

I did not switch to killing the process group. It would be the more complete cleanup, but it needs `start_new_session=True` at spawn and changes signal semantics for every existing hook — worth its own change rather than riding along here.

**2. Credential material — fixed, but not the way the first option suggests.** `SecretStr` does **not** close this. For `missing`-type errors pydantic reports the raw pre-validation mapping as `input`, before any field type applies, so the key is still a plain string. Demonstrated:

```
TYPE: missing LOC: ('models', 'bad', 'model_url')
  INPUT: {'model_name': 'b', 'api_key': 'sk-SUPERSECRET'}
STR CONTAINS SECRET: True     # with api_key: SecretStr
```

So I took the second option. The message is rebuilt from `exc.errors()` — location, reason, searched paths — echoing a value only when it is a **scalar under a non-sensitive field**. That keeps the part an operator needs (the typo'd `reasoning_effort` is still named verbatim) while a credential-bearing mapping renders as `<redacted>`. It is raised as a redacted `ConfigurationError`, so the loader tests move from `ValidationError` to `ConfigurationError`; a new test asserts neither the key nor the header token appears while `model_url` and the searched path still do.

Since the exception now genuinely carries the search paths, I also reverted the doc wording I had weakened in the previous round — the original claim is true again.

**3. CLI — fixed.** `list_models` and `export_config` catch `ConfigurationError`, print the `❌` message and exit 1. Tests assert exit code and that the offending value is named. (`show_config_paths` needed nothing; it only uses `_search_paths` and never loads the config.)

**4. Caching documented — fixed.** `docs/admin/llm-config.md` now states the failure is retained for the life of the process and that the file must be corrected **and the process restarted**.

**5. Sibling-drain coverage — added.** `test_overflow_keeps_draining_the_sibling_stream_to_eof` uses a `_FakeProc` with output on both streams and asserts both reach EOF, `overflowed is True`, and the combined length stays within the cap. Plus `test_overflow_gives_up_on_a_pipe_that_never_reaches_eof` for change 1 — it failed (31s, hit the outer bound) before the fix.

**Notes:** traceback growth on re-raise fixed via `.with_traceback(None)`. Reap expiry now logs at `warning`. Class docstring states the asymmetry (`llm_config` is fail-fast because an empty model set looks like a working server with nothing to talk to; the other loaders degrade rather than disable). I left `_REAP_TIMEOUT_S` a constant rather than scaling it to remaining budget — in the timeout path the budget is already spent, so there is nothing to scale against; it is now covered by a test that shrinks both deadlines via monkeypatch.

**Verification:** `bash test/atlas_tests.sh` → **2879 passed, 17 skipped**, exit 0. `ruff check` clean on all touched files (the repo-wide 45 are pre-existing and unchanged — verified against a clean tree).

**On the CI evidence gap:** closed independently — `build-and-test` came back **SUCCESS** on `0735286`, along with all nine other checks. That is the environment where the original hooks failure reproduced.

garland3:
## Summary
Makes `ConfigManager.llm_config` fail fast on a schema-invalid `llmconfig.yml` — raising a cached, credential-redacted `ConfigurationError` that names the searched paths — and fixes a hook-subprocess deadlock by bounding the post-overflow pipe drain and reaping the killed child.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — deterministic regression tests with a stated negative control, pasted failure output, and named verification runs; the only gap is that the green `build-and-test` is cited for `0735286` rather than the head commit `2fb2bea`, which changed the timing-sensitive drain code.

## Notes
- `AppSettings` access moved outside the `try` at `atlas/modules/config/config_loader.py:241`, so a pydantic `ValidationError` from env vars escapes unredacted and uncached — the exact leak this PR closes elsewhere.
- Redaction checks only `loc[-1]` (`config_loader.py:53`), so a bad scalar under `extra_headers` (e.g. `Authorization`) is echoed verbatim; test the whole `loc` path instead.
- `atlas-chat list-models` (`atlas/atlas_chat_cli.py:253`) — the console script actually on operators' PATH — still tracebacks on a bad config; only `atlas/modules/config/cli.py` got the handler.

---

PR #903, opened 2026-09-09
feat(chat): run conversations in parallel with per-run identity (#884)

### LINKED ISSUE #884: Support multiple parallel conversation runs in one ATLAS session
## Summary

Allow a user to start work in one conversation, switch to or create another conversation, and run both conversations concurrently without cancelling either one.

Example:

1. Start a long-running agent task in conversation A.
2. Select **New Chat** and start an unrelated task in conversation B.
3. Both runs continue independently.
4. Switch between A and B to view progress, provide input, or stop either run.

This issue uses **run** rather than operating-system **process**: the requested capability is multiple independently managed chat/agent executions, regardless of whether the deployment implements them as coroutines, worker tasks, or separate processes.

## Current behavior

ATLAS has one `active_chat_task` per WebSocket connection (`atlas/main.py:631`). The frontend's New Chat flow sends stop messages before `reset_session` (`frontend/src/contexts/ChatContext.jsx:755-786`), and the backend also cancels the active task during reset (`atlas/main.py:949-960`). The WebSocket session, conversation history, stream buffer, tool state, files, and canvas are all effectively modeled as one active conversation.

Consequently, starting a new chat while work is running stops the original turn. Attempting to keep both tasks alive without further isolation would cause events and conversation history to race or appear in the wrong chat.

Two details of today's behavior matter for the design:

- A second message sent to the *same* conversation is neither queued nor rejected. #824 injects it into the running agent loop as a steering message at the next iteration boundary (`atlas/main.py:664-700`); `should_steer` (`atlas/application/chat/agent/steering.py:83-115`) gates on task liveness, channel activity, and conversation identity. This behavior ships today and should be preserved, not replaced with a queue/reject rule.
- Client disconnect takes the same cancel path as Stop: `WebSocketDisconnect` cancels the in-flight task and releases the conversation's MCP sessions (`atlas/main.py:1069`, `atlas/main.py:144-163`). No run survives a closed tab today. This is #760.

## Desired behavior

- Navigation controls which conversation is visible, not which conversation is allowed to execute.
- Each active run has a stable `run_id`, `conversation_id`, authenticated owner, status, cancellation scope, and isolated session/history state.
- Different conversations may run concurrently.
- A single conversation permits at most one history-mutating run at a time. A second message for that conversation continues to steer the running run (#824).
- Every streamed event includes `run_id` and `conversation_id`, allowing the client to route background updates to the correct conversation.
- Stop, steering, tool approval, and elicitation responses address a specific run.
- Conversation navigation must not release MCP sessions or other resources still owned by an active run.
- A run continues when the browser is closed. Reopening the conversation shows the progress the run made while the user was disconnected.
- A run that requires tool approval pauses and waits, whether or not the user is currently viewing that conversation.

Suggested run states:

```text
queued -> running -> waiting_for_input -> running -> completed
                  \-> failed
                  \-> cancelled
```

## Enablement and limits

- The feature is enabled only when server-side chat history is on (`FEATURE_CHAT_HISTORY_ENABLED`, `atlas/modules/config/settings.py:410`) **and** the user's save mode is `server` (`frontend/src/contexts/ChatContext.jsx:85`, `619-620`). Under `local` or incognito save modes, behavior is unchanged: one run at a time. A durable run whose transcript is never written to the server cannot be resumed.
- Only effective agent-mode runs may execute in the background. The orchestrator silently downgrades `agent_mode` to `False` when the model lacks tool support or no tools are selected (`atlas/application/chat/orchestrator.py:466-509`), so the gate must key off the resolved run, not the UI toggle.
- A new admin setting caps concurrent runs per user, default **5**, configured by environment variable (`MAX_CONCURRENT_RUNS_PER_USER`). No per-user concurrency setting exists today; this is net-new.
- The cap counts every run that is not in a terminal state. A run paused awaiting tool approval **counts against the cap**.
- Conversation history shows a per-conversation indicator for runs that are still active, including runs in conversations the user is not currently viewing (comparable to the ChatGPT / Claude web clients).
- Background progress is never gated on the user having the conversation open. Approval, however, does block.

### Backend

1. Introduce a first-class run record/handle keyed by `run_id` and associated with a `conversation_id` and user.
2. Replace the single `active_chat_task` slot with a conversation/run-scoped registry.
3. Give each active conversation an independent `Session`/history object and enforce a per-conversation execution lock.
4. Decouple event publishing from the currently displayed WebSocket conversation. Add `run_id` and `conversation_id` to token, agent, tool, file, canvas, completion, and error events.
5. Change cancellation and steering messages to target a `run_id`.
6. Retain MCP sessions and run-scoped resources until the owning run completes, is cancelled, or expires.
7. Detach run ownership from the WebSocket connection so a disconnect no longer cancels the run.
8. Replace the in-memory tool approval and elicitation managers with persisted pending-request records. Both are currently process-wide singletons holding an `asyncio.Future` with a hard 300-second wait (`atlas/application/chat/approval_manager.py:174-178`, `:35`, `:49`; `atlas/application/chat/elicitation_manager.py:29`; callers `atlas/application/chat/utilities/tool_executor.py:576` and `atlas/modules/mcp_tools/mcp_routing.py:246`). A run paused on approval while the browser is closed dies after five minutes and cannot be recovered across a restart, so durable approval pause is not achievable without this change.
9. Add per-user concurrency and wall-clock limits, with clear queued/throttled states.

### Frontend

1. Store transcript, streaming buffer, run status, tools, files, and canvas state per conversation rather than globally.
2. Change New Chat and conversation selection so they do not implicitly stop another conversation's run.
3. Route incoming events using `conversation_id`/`run_id`, including when the target conversation is not visible.
4. Show running, waiting, completed, failed, and cancelled indicators in conversation history.
5. Make Stop and other interactive controls operate on the selected run only.

## Acceptance criteria

- [ ] A user can start a long-running run in conversation A, create conversation B, and start a second run without cancelling A.
- [ ] Both runs can stream and execute tools concurrently without mixing messages, tool rows, files, canvas output, or completion events.
- [ ] Switching between A and B displays the latest state of each run.
- [ ] Stopping A does not stop or alter B.
- [ ] A second message to a conversation with an active run is steered into that run (#824), not started as a concurrent turn; histories never receive concurrent unordered writes.
- [ ] A run continues after the browser is closed, and reopening the conversation shows the progress it made while disconnected.
- [ ] A run paused on a required tool approval stays paused across a disconnect and can be approved after reconnect; pending approvals and elicitations survive process restart and are not subject to a fixed short timeout.
- [ ] Conversations with active runs are visibly marked in conversation history while the user is viewing a different conversation.
- [ ] Completed, failed, interrupted, and waiting-for-input states are persisted and visible after reopening a conversation.
- [ ] The feature activates only when `FEATURE_CHAT_HISTORY_ENABLED` is set, the user's save mode is `server`, and the run resolves to agent mode; otherwise existing single-run behavior is unchanged.
- [ ] Exceeding `MAX_CONCURRENT_RUNS_PER_USER` (default 5) is rejected with a clear message; approval-paused runs count toward the cap.
- [ ] Authorization and compliance context are independently preserved for every run.
- [ ] Configurable per-user concurrency, iteration, and wall-clock limits prevent unbounded unattended execution.
- [ ] Backend and frontend tests cover concurrent runs, event routing, cancellation isolation, persistence, approvals, disconnect races, and resume after browser close.
- [ ] User, administrator, configuration, architecture, and troubleshooting documentation is updated as applicable.

## Delivery phases

Durable execution is in scope, not deferred. Phase 1 may land concurrent runs while the WebSocket stays connected, but this issue is not complete until runs survive browser close, and the acceptance criteria above are written against that durable end state.

## Dependencies

- **#760** (open) covers continuation and session resume across disconnect/reconnect. Because durable runs are in scope here, it is a dependency rather than merely related work: the current disconnect path cancels the in-flight turn, so #760 must land first or be absorbed into this work. Both must share the same conversation-scoped run identity and event-routing foundation.
- Persisted tool approval and elicitation state (see Backend item 8) is a prerequisite for the approval-pause acceptance criterion.

## Related work

- `docs/planning/async-execution-architecture-2026-01-25/` contains earlier durable execution and workflow-engine analysis.
- #824 introduced agent steering, which defines how a second message to an already-running conversation is handled.

## Open decisions

- Confirm that steering (#824) remains the answer for a second message to a busy conversation, rather than the queue-or-reject rule this issue originally proposed.
- Confirm whether #760 is absorbed into this issue or lands first as separate work.
- Confirm the approval/elicitation persistence rework is in scope here rather than spun out.
- Confirm the environment variable name `MAX_CONCURRENT_RUNS_PER_USER` versus a `FEATURE_`-prefixed enable-flag plus limit pair.

garland3:
Closes #884 (Phase 1 + browser-close survival; see Scope below for what is explicitly not done).

## The problem

Navigation and execution were the same thing. The transport tracked one `active_chat_task` per WebSocket connection, so **New Chat**, opening another conversation, and closing the tab all cancelled whatever was running.

## What this does

Introduces a first-class **run**: one history-mutating execution of a conversation, with a stable `run_id`, its own `Session`, an authenticated owner, a status and its own cancellation scope. Runs live in a process-wide registry rather than a connection-owned slot, which is what lets one outlive the socket that started it.

### Backend

- `RunRegistry` enforces the two invariants from the issue: at most one non-terminal run per conversation, and at most `MAX_CONCURRENT_RUNS_PER_USER` (default 5) per user. A run paused on tool approval is **not** terminal, so it counts against the cap — the explicit decision in the issue thread.
- Every event a tracked run emits carries `run_id` and `conversation_id`.
- Stop, steering, and approval/elicitation responses address one run, ownership-checked so a client cannot reach another user's run.
- Conversation navigation and disconnect no longer release MCP sessions or cancel a run that still owns the conversation. A disconnected run is marked detached and keeps executing.
- A second message to a busy conversation still steers the running loop (#824), unchanged.
- `TOOL_APPROVAL_TIMEOUT_SECONDS` makes the hard-coded 300s approval wait configurable; `0` waits indefinitely, which is what lets an approval pause survive a disconnect. `MAX_RUN_WALL_CLOCK_SECONDS` is the backstop against unbounded unattended execution.
- New `list_runs` / `runs_snapshot` / `run_started` / `run_status` frames. `run_status` goes to every connection of the owner, so a second tab keeps its indicators current.

### Frontend

- Events are routed by `conversation_id`. An event for a conversation that is not on screen is dropped rather than spliced into the visible transcript — the "messages in the wrong chat" failure mode.
- Run state is tracked per conversation and rebuilt from `runs_snapshot` on connect, so reopening the app shows what is still working.
- The history list marks running / queued / needs-approval / failed / stopped conversations.
- New Chat no longer stops a tracked run.

## Enablement

Gated on `FEATURE_CHAT_HISTORY_ENABLED` **and** a `server` save mode **and** an agent-mode turn with tools. Every other turn keeps the previous single-run behaviour byte for byte — this is additive, not a rewrite of existing deployments' semantics.

## Scope: what is *not* in this PR

Stated plainly rather than buried, since the issue's acceptance criteria are written against a fully durable end state:

- **Runs do not survive a server restart.** The registry is in memory, as are pending approvals and elicitations. Durable runs need a persisted run store plus persisted pending-request records (issue's Backend item 8) — a substantially larger change that deserves its own review.
- **A reconnected browser does not resume a live event stream.** It sees the run in `runs_snapshot` and, on reopening the conversation, the transcript the run has saved so far — not the tokens it missed while away. That is #760.
- **Multi-process deployments track runs per process.**

All three are documented as limitations in `docs/admin/parallel-conversation-runs.md`.

## Testing

- 43 new backend tests (`test_parallel_conversation_runs.py`, `test_parallel_run_transport.py`): concurrency cap, approval-paused runs counting toward it, per-conversation locking, cancellation isolation, sticky terminal states, ownership checks, wall-clock enforcement, reaping, status listeners, the eligibility matrix, and stop/approval frame routing.
- 11 new frontend tests: event routing by conversation, background events not touching the visible transcript, run lifecycle frame handling, snapshot rebuild.
- Full suites green: **2911 backend passed** (`bash test/atlas_tests.sh`), **827 frontend passed** (`npx vitest run`), eslint clean on every changed file.

## Docs

- `docs/admin/parallel-conversation-runs.md` — enablement, the three new env vars, run states, the protocol, resource ownership, limitations, troubleshooting.
- `.env.example` and the admin README index updated.

https://claude.ai/code/session_01EVwF7pb7PeCHv5SaPpLS99


garland3:
## Summary
Adds per-conversation background runs: a process-wide run registry, eligibility gating, per-run status frames and sidebar indicators, a wall-clock sweeper, and disconnect handling so a run survives a closed browser. Concurrency is capped per user and MCP sessions are preserved while a run still owns a conversation.

## Verdict
Request changes.

## Required changes
1. Resolve `download_file` against the run's session at `atlas/main.py:1353` — tracked runs write files into `run_record.session_id`'s context, so every download of a run-produced artifact returns "File not found in session". Either echo `run_id`/`conversation_id` from the client and resolve via the registry, or move the file map to a conversation-scoped store.
2. Seed the run session's `context["files"]` from the connection session at `atlas/application/chat/runs/registry.py:296` (or attach files conversation-scoped) — files attached via `attach_file` at `atlas/main.py:1480` land on the connection session and are invisible to the run, so the model is asked to work on a file it cannot see.
3. Make the stop fallback at `atlas/main.py:1519` fire only for frames carrying neither `run_id` nor `conversation_id` — `_cancel_addressed_run` returns `False` for an already-terminal run, so a stale stop frame cancels an unrelated untracked turn on the same connection.
4. Bind elicitation responses to the requesting user in `atlas/modules/mcp_tools/mcp_routing.py:247` — store `user_email` on `ElicitationRequest` and reject mismatches as `ToolApprovalManager` already does; the indefinite-wait option turns a 5-minute race into an open-ended one.
5. Release run-owned resources on terminal transition (`atlas/main.py:1314` `finally`) — call `mcp_tool_manager.release_sessions(...)` and `chat_service.end_session(run_record.session_id)` when no other live run holds the conversation; today a detached run strands a credentialed MCP client set and a full `Session` for the process lifetime.
6. Wrap the approval wait at `atlas/application/chat/utilities/tool_executor.py:577` in `try/finally: approval_manager.cleanup_request(tool_call.id)` — cancellation (user stop or wall-clock sweep) currently leaks the pending request and its call arguments forever.
7. Record a non-`completed` outcome when the turn fails at `atlas/main.py:1314` — `handle_chat` absorbs its own errors, so a background failure is reported as `completed` and the sidebar shows no failure marker.

## Evidence
Evidence: 7/10. The backend is strongly evidenced (43 new backend and 11 frontend tests, full-suite counts, an admin doc), but the user-visible half is asserted rather than shown; the most important gap is any demonstration of the headline close-the-browser-and-return flow against a running server, plus screenshots of the new sidebar run indicators. Also add a test for `cleanup_disconnected_session` — the disconnect branches that decide whether to release MCP sessions are entirely uncovered.

## Notes
- Reopening a conversation with a live run shows a static transcript and no Stop control — seed the busy flags from the run record (`frontend/src/contexts/ChatContext.jsx:920`).
- `run_limit_exceeded` / `conversation_busy` leave the user's typed message in the transcript unsent; restore it to the composer instead (`atlas/main.py:1104`).
- Add a dedicated `FEATURE_PARALLEL_RUNS_ENABLED` kill switch; today the only lever is disabling chat history entirely (`atlas/modules/config/settings.py:537`).

---

garland3:
## Summary
Adds tracked, per-conversation background runs so an agent turn keeps executing after the user navigates away or closes the browser, with a run registry, concurrency caps, wall-clock reaping, run-tagged WebSocket frames, snapshot/approval replay on reconnect, and sidebar run indicators.

## Verdict
Request changes.

## Required changes
1. Seed the run's session from the connection session (or move the file map to conversation scope) at run admission — `atlas/main.py:1597`; `attach_file` writes to the connection session while the run executes in a fresh session, so the model never sees the attached file.
2. Resolve downloads against the owning run's session — `atlas/main.py:1461`; files produced by a tracked run are written into `run_record.session_id` and are unreachable from the client that requested them.
3. Make the stop/cancel fallback conditional on the frame carrying neither `run_id` nor `conversation_id`, rather than on `_cancel_addressed_run` returning `False` — `atlas/main.py:1519`; a late stop for an already-finished run currently cancels an unrelated in-flight turn on the same connection.
4. Wrap the approval wait in `try/finally: approval_manager.cleanup_request(tool_call.id)` — `atlas/application/chat/utilities/tool_executor.py:577`; cancellation (user stop, wall-clock sweep) leaves the pending request, its tool arguments and an un-awaited future in a process-wide dict with no eviction, unbounded with `TOOL_APPROVAL_TIMEOUT_SECONDS=0`.
5. Tag the error frames in `handle_chat` with their run/conversation ids (or send them through the turn callback) — `atlas/main.py:1281`; untagged errors from a background run are rendered into whichever conversation is on screen, the exact cross-chat leakage this PR sets out to remove.
6. Add `user_email` to `ElicitationRequest` and reject mismatched responders in `handle_elicitation_response` — `atlas/application/chat/elicitation_manager.py:99`, mirroring `ToolApprovalManager.handle_response`; this PR's indefinite-timeout option turns a bounded window into an open-ended one.
7. Build `surviving` from all of the user's active runs, not just this connection's — `atlas/main.py:1198`; a second tab's run on the same conversation has its MCP sessions released mid-step on disconnect.

## Evidence
Evidence: 7/10 — the backend is well covered (54 targeted tests against the stated invariants, full-suite counts and commands given), but the user-visible half of a 24-file, +2500-line change is asserted rather than shown. The most important gap is a demonstration of the headline close-the-tab-and-return flow (including the approval replay) against a running server, plus a test for `cleanup_disconnected_session`'s release-vs-preserve branch.

## Notes
- `TOOL_APPROVAL_TIMEOUT_SECONDS` ships at `300`, so approval-paused runs — the common agent case — still die five minutes after the tab closes, narrower than the changelog claims (`.env.example:196`).
- Hitting the concurrency cap or a busy conversation discards the user's typed message instead of restoring it to the composer (`frontend` chat send path).
- Run indicators carry no accessible status semantics and are not announced to assistive tech.

---

garland3:
### Review findings: all addressed as of `1feec1a`

The bots re-ran their full review after each push, so several findings appear more than once and some are anchored to superseded commits. Every one was verified against the code before fixing; here is where each landed.

| Finding | Status | Where |
|---|---|---|
| Events published through the shared `WebSocketEventPublisher` were untagged | Fixed | `9f7349c` — tagging moved to `WebSocketConnectionAdapter.send_json`, driven by a contextvar bound inside each run's task |
| First agent turn in a new chat could never be a tracked run (no `conversation_id` yet) | Fixed | `9f7349c` — server mints the id at admission and reports it on `run_started`; the client adopts it |
| A finished run never released its `Session` or MCP resources | Fixed | `9f7349c` + `1feec1a` — `_release_finished_run()`; the run's single-use session is deleted, not just deactivated |
| Handled domain errors recorded the run as `completed` | Fixed | `9f7349c` — each handled branch records its type; the run terminates as `failed` |
| Pending approval requests dropped for a non-visible conversation | Fixed | `9f7349c` — the frame is kept on the run and replayed on `restore_conversation` / `list_runs` |
| No Stop control for a run rediscovered after reconnect | Fixed | `9f7349c` — `isAgentRunning` is derived from the run snapshot too |
| Wall-clock expiry said "cancelled" but recorded `failed` | Fixed | `9f7349c` — kept `failed` (cancelled means a person stopped it) and corrected the log, settings and docs wording |
| Missing `CHANGELOG.md` entry | Fixed | `9f7349c` |
| Missing `Last updated:` line on the evergreen admin guide | Fixed | `9f7349c` |

**One suggestion I did not take.** Both bots proposed rendering the approval/elicitation prompt for a conversation the user is *not* viewing. A modal about a chat that isn't on screen, without the surrounding context, is its own kind of wrong — so the history row is marked "Needs approval" instead and the server replays the request when that conversation is opened. The frame is deferred, not lost, which is the substance of the concern.

Verification on the current head: **2919 backend tests** (`bash test/atlas_tests.sh`), **828 frontend tests** (`npx vitest run`), `scripts/check-docs.sh` OK, eslint clean (the 4 remaining warnings are pre-existing in untouched files), all 10 CI checks green.

The scope limits in the PR description still stand and are documented in `docs/admin/parallel-conversation-runs.md`: runs are in-memory, so they do not survive a **server restart**, and a reconnected browser sees saved progress rather than a resumed live stream (#760).

garland3:

garland3:
## Summary
Adds server-side parallel conversation runs: a `RunRegistry` that owns run admission, lifecycle and per-user concurrency limits, run-id/conversation-id tagging of WebSocket frames, runs that survive tab disconnect, a wall-clock sweeper, and sidebar run indicators. Supporting config knobs (`TOOL_APPROVAL_TIMEOUT_SECONDS`, `MAX_RUN_WALL_CLOCK_SECONDS`), docs and tests ship with it.

## Verdict
Request changes.

## Required changes
1. Bind elicitation responses to their owner at `atlas/application/chat/elicitation_manager.py:99` — add `user_email` to `ElicitationRequest`, populate it in `create_elicitation_request` (`mcp_routing.py:~230`), and reject mismatches from the `elicitation_response` branch (`atlas/main.py:1694`), mirroring the fail-closed check in `ToolApprovalManager.handle_response`; today any authenticated user with an elicitation id can inject data into another user's tool execution, and the indefinite wait keeps it answerable for the process lifetime.
2. Resolve `download_file` against the run's session at `atlas/main.py:1479` — echo `run_id`/`conversation_id` on the frame and look the run up via the registry (falling back to the connection session), or move the file map to a conversation-scoped store; otherwise every download of a file produced by a background run returns "File not found in session".
3. Seed the run session's `context["files"]` from the connection session at run admission (`atlas/main.py:1615`, `RunRegistry.start`) — files attached via `attach_file` land on the connection session and are invisible to the tracked run that needs them.
4. Compute the MCP release decision from `run_registry.active_for_conversation(conv_id, user_email)` at `atlas/main.py:1188` instead of the connection-filtered `connection_run_ids` — closing a second tab currently tears down MCP sessions out from under a live run started in another tab.
5. Wrap the approval wait at `atlas/application/chat/utilities/tool_executor.py:577` in `try/finally: approval_manager.cleanup_request(tool_call.id)` — cancellation (user Stop or the sweeper) skips both existing cleanups, leaking the request and its `filtered_args` permanently, with no timeout path at all when `TOOL_APPROVAL_TIMEOUT_SECONDS=0`.
6. Tag or route `handle_chat`'s error frames through `turn_update_callback` at `atlas/main.py:1299` — raw `websocket.send_json` calls carry no run ids, so the client renders a background run's rate-limit/timeout errors into whichever conversation is on screen.
7. Decide the stop/`agent_control` fallback from the frame rather than the outcome at `atlas/main.py:1519` and `:1674` — `_cancel_addressed_run` returns `False` both for "no run named" and "named run already terminal", so a stale stop frame cancels an unrelated untracked turn.

## Evidence
Evidence: 7/10 — the backend half is strongly evidenced (54 targeted tests, full-suite counts, docs, changelog, config samples), but the user-visible half is asserted rather than shown. The most important missing artifact is a demonstration of the headline close-the-tab-and-return flow end to end, followed by a screenshot of the new sidebar run states.

## Notes
- Add a `try/except` or `asyncio.shield` around `_release_finished_run` (`atlas/main.py:1447`) so a second cancellation cannot skip session cleanup.
- Two independent tagging implementations (`tag_run_event` at `atlas/main.py:170` vs `stamp_with_current_run` at `atlas/application/chat/runs/context.py:52`) disagree on copy-vs-mutate and will drift; collapse to one authority.
- Reject or warn on `TOOL_APPROVAL_TIMEOUT_SECONDS=0` combined with `MAX_RUN_WALL_CLOCK_SECONDS=0` (`atlas/modules/config/settings.py:576`) — together they leave a detached run with no expiry.

---

garland3:
### Review findings addressed in `b137386`

All 7 required changes and 2 of the 3 notes. Each was verified against the code at `b0e60c6` before fixing (the line numbers in the review had drifted past the merge commit).

| # | Finding | Where |
|---|---|---|
| 1 | Elicitation responses not bound to an owner | `elicitation_manager.py` — `ElicitationRequest.user_email`, populated through `_ElicitationRoutingContext` → `mcp_routing.py`, fail-closed reject in `handle_elicitation_response`, responder passed from `main.py` |
| 2 | `download_file` resolved only against the connection session | `main.py` — `_download_session_candidates`: connection session first, then the requesting user's runs (terminal included); `run_id`/`conversation_id` echoed on the reply |
| 3 | Attached files invisible to a tracked run | `main.py` — `_seed_run_session_files` at run admission, copying (not sharing) the file map |
| 4 | MCP release decision taken from `connection_run_ids` | `main.py` — `cleanup_disconnected_session` now asks `registry.active_for_conversation` |
| 5 | Approval request leaked on cancellation | `tool_executor.py` — wait wrapped in `try/finally`; the `TimeoutError` branch's now-redundant cleanup dropped |
| 6 | `handle_chat` error frames untagged | `main.py` — all 12 frames (10 error branches + the cancellation flush) routed through `turn_update_callback` |
| 7 | Stop fallback decided from the outcome, not the frame | `main.py` — `_cancel_addressed_run` returns whether the frame *addressed* a run; the existing test encoding the old semantics is updated, and its own comment ("nothing was *cancelled*") was the bug |

Notes: run cleanup is `asyncio.shield`ed so a second cancellation cannot skip session release, and `AppSettings` warns when `TOOL_APPROVAL_TIMEOUT_SECONDS=0` meets `MAX_RUN_WALL_CLOCK_SECONDS=0`. **Deferred:** collapsing `tag_run_event` and `stamp_with_current_run` into one authority — a real drift risk, but a refactor of working code that is better reviewed on its own than folded into a security fix; filing it as a follow-up rather than silently dropping it.

#### Verification

- **20 new backend tests.** The approval-leak test drives the real gate with `TOOL_APPROVAL_TIMEOUT_SECONDS=0` and cancels mid-wait; I confirmed it **fails without the fix** (temporarily reverted the `try/finally`) so it is not vacuous. Download coverage includes the cross-user case — widening the search must not widen access.
- **Full backend suite: 3103 passed, 17 skipped** (`bash test/atlas_tests.sh`, `PYTHONPATH` set to the worktree, import path verified to resolve there and not to the main checkout).
- `ruff` clean on every file in the diff. The one remaining `I001` in `atlas/main.py` is pre-existing on the base — verified by linting `git show HEAD:atlas/main.py` — and the import block is untouched by this diff.

On the evidence gap: the backend half is covered above, but the reviewer is right that the headline close-the-tab-and-return flow is still asserted rather than demonstrated. That is a manual browser run I have not done, so I am not claiming it.

garland3:
Filed the deferred note as #915 (collapse `tag_run_event` / `stamp_with_current_run` into one authority) so it is tracked rather than dropped. The copy-vs-mutate split turns out to be load-bearing — `stamp_with_current_run` mutates deliberately because the publisher builds each frame fresh and copying every token event would be wasteful — so the collapse needs to preserve that, which is the other reason it did not belong in this push.

garland3:

## Summary
Introduces a first-class "run" abstraction for chat execution: a process-wide `RunRegistry` tracks one non-terminal run per conversation (capped per user), so runs survive navigation, new-chat, and socket disconnect, with stop/steering/approval messages addressed by `run_id` and ownership-checked.

## Verdict
Looks good.

## Evidence
No evidence score was produced for this change; the diff ships four backend test modules and one frontend test module covering the registry, transport, tool-approval cleanup, and elicitation paths, which is proportionate demonstration for a change of this size. The main gap worth noting is an end-to-end check that a run genuinely survives a browser close and reattaches on reconnect.

---

garland3:
## Summary
Adds a run registry so an agent turn becomes a tracked background run that survives the socket that started it: per-user concurrency caps, wall-clock enforcement, run-tagged WebSocket frames, approval/elicitation replay after reconnect, MCP-session preservation on disconnect, and sidebar run-state indicators.

## Verdict
Request changes.

## Required changes
1. Keep a finished run's file map reachable at `atlas/main.py:1314` (`_release_finished_run`) — it calls `session_repository.delete(session_id)` on every terminal transition, so the widened download search at `atlas/main.py:1461` can never resolve a completed run's session and the "File not found in session" bug it was added to fix still occurs for the common case (download after the run completes). Either defer the session delete until the run record is reaped, or copy `context["files"]` onto the connection session before deleting; extend the test to assert a successful `handle_download_file` after termination rather than only candidate membership.
2. Bind elicitation and approval requests to an owner at creation at `atlas/modules/mcp_tools/mcp_execution.py:745` — the ownership check is skipped whenever `user_email` is empty, which happens for any server config with no `groups`, letting any authenticated user holding the request id answer another user's elicitation (indefinitely when `TOOL_APPROVAL_TIMEOUT_SECONDS=0`). Reject creation without a non-empty `user_email`, or make the legacy sentinel `None` and reject `""` on response.
3. Mint the server-side `conversation_id` only on the eligible path at `atlas/main.py:1204` — the guard is `is_agent_turn`, not `turn_is_eligible_for_background_run(...)`, so ineligible turns (no tools selected, local/incognito save mode, history disabled) are now persisted under an id the client never receives.
4. Send the `run_started` frame before `asyncio.create_task(handle_chat_guarded())` at `atlas/main.py:1519` — the `await` on the send is a suspension point, so the run's first tagged frames can reach a client that has not yet adopted the minted id and are dropped by the routing guard at `frontend/src/handlers/chat/websocketHandlers.js:407`.
5. Add tests for the disconnect release-vs-preserve branch in `cleanup_disconnected_session` (`atlas/main.py:1188`) — the new `connection_run_ids` parameter is always `None` in `test_websocket_disconnect_resilience.py`, so both branches of the PR's central behavioral change are uncovered, including the regression already fixed once in `b137386`.
6. Restrict the download candidate scan when the frame addresses a run at `atlas/main.py:1479` — file maps are keyed by filename, so a name collision across the user's own runs returns the wrong artifact silently, and a miss costs up to ~51 sequential session lookups on the receive loop.
7. Wrap the approval `wait_for_response` at `atlas/application/chat/utilities/tool_executor.py:577` in `try/finally` — cancellation (user Stop or the wall-clock sweep) otherwise leaves the pending request and its `filtered_args` in the manager for the process lifetime.

## Evidence
Evidence: 7/10. The backend is unusually well evidenced for a change this size — four new test modules, named full-suite counts with commands, and one test explicitly shown to fail with the fix reverted — but the user-visible half is asserted rather than shown; the most important missing artifact is a demonstration of the headline flow (start a run, close the tab, reopen and see it still running with its saved progress) against a live server.

## Notes
- Failed-run tooltip renders raw internal error tokens such as `malformed_tool_call` — `frontend/src/components/Sidebar.jsx:66`.
- Run indicators carry no status semantics for assistive technology, so the "Needs approval" transition is silent for screen readers — `frontend/src/components/Sidebar.jsx:60`.
- Two divergent implementations of run tagging (copy vs. mutate) will drift — `atlas/main.py:170` and `atlas/application/chat/runs/context.py:52`.

---

PR #904, opened 2026-09-09
build(deps): hold fastmcp below 4.0 and document why

garland3:
## Why

Dependabot #895 widened the `fastmcp` constraint to `>=3.4.3,<5.0.0`. That bump cannot pass CI: it fails 7 tests in `atlas/tests/test_sampling_integration.py` with

```
fastmcp.exceptions.ToolError: Error calling tool 'summarize_text': 'Context' object has no attribute 'sample'
```

This is not a rename. FastMCP 4 implements MCP protocol `2026-07-28`, which removed **server-initiated requests** (SEP-2322). `Context.sample()` is deleted from the server-authoring API, and reaching the wire directly through the SDK session raises:

```
Cannot send 'sampling/createMessage': this transport context has no back-channel for server-initiated requests.
```

There is no client-side knob to negotiate a handshake-era protocol version — neither `fastmcp.Client` nor `mcp.ClientSession` exposes a protocol-version pin — so there is no drop-in migration. FastMCP's own `server/providers/proxy.py` documents the same constraint: "server-initiated sampling is not part of FastMCP's server-authoring API ... this path only ever runs when both legs of the proxy speak the handshake era."

Atlas depends on server-initiated sampling in `atlas/mcp/sampling_demo/main.py` (11 call sites), `atlas/mcp/tool_planner/main.py`, and the sampling-routing plumbing in `MCPToolManager`. Raising the cap would silently break a shipped feature, so the major bump needs a redesign of those servers first, not a constraint change.

## What this PR does

- Documents the `<4.0.0` cap in `pyproject.toml` so the reason is next to the pin.
- Adds a Dependabot `ignore` rule for `fastmcp` major updates on the root pip ecosystem, so the same unmergeable PR is not reopened weekly.
- Links #905 (the migration off server-initiated sampling) from both, so the cap has a tracked off-ramp.

No dependency versions change. #895 is closed.

## Scope of the ignore rule

Deliberately root-only. `mocks/mcp-http-mock` has its own ecosystem block and was already widened to `<5.0.0` by merged #896 — it uses no `Context` and no sampling, so 4.x is safe there. `mocks/wormhole-mcp-mock` and `atlas/mcp/code-executor-v2` use only `ctx.session_id` and have no Dependabot entry at all.

## Reproduction

Against `fastmcp==4.0.3` in a clean venv:

```python
# srv.py
from fastmcp import FastMCP, Context
mcp = FastMCP("demo")

@mcp.tool
async def summarize_text(text: str, ctx: Context) -> str:
    from mcp import types as mcp_types
    result = await ctx.session.create_message(
        messages=[mcp_types.SamplingMessage(
            role="user",
            content=mcp_types.TextContent(type="text", text=text))],
        max_tokens=512,
        related_request_id=ctx.origin_request_id,
    )
    return result.content.text

if __name__ == "__main__":
    mcp.run()
```

```python
# cli.py
import asyncio, sys
from fastmcp import Client
from fastmcp.client.transports import StdioTransport
from mcp.types import CreateMessageResult, TextContent

async def h(messages, params=None, context=None):
    return CreateMessageResult(role="assistant",
                               content=TextContent(type="text", text="SUMMARY OK"),
                               model="t")

async def main():
    c = Client(StdioTransport(command=sys.executable, args=["srv.py"]), sampling_handler=h)
    async with c:
        print(await c.call_tool("summarize_text", {"text": "hello world"}))

asyncio.run(main())
```

`python cli.py` raises the `NoBackChannelError` quoted above. Also confirm `"sample" not in dir(fastmcp.Context)`.

## Verification

`PYTHONPATH=$(pwd) .venv/bin/python -m pytest atlas/tests/test_sampling_integration.py -q` → `11 passed` on the current pin.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,6 +24,15 @@ updates:
         update-types:
           - "minor"
           - "patch"
+    ignore:
+      # FastMCP 4 implements MCP 2026-07-28, which removed server-initiated
+      # requests (SEP-2322). That drops Context.sample(), which the
+      # sampling_demo and tool_planner MCP servers depend on. Re-evaluate once
+      # those servers no longer rely on server-initiated sampling.
+      # Tracked in https://github.com/sandialabs/atlas-ui-3/issues/905
+      - dependency-name: "fastmcp"
+        update-types:
+          - "version-update:semver-major"
 
   # Python dependencies (file-extractor-mock)
   - package-ecosystem: "pip"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### PR #904 - 2026-09-08
+- Hold `fastmcp` below 4.0 and record why: FastMCP 4 removes server-initiated sampling, which the sampling_demo and tool_planner MCP servers require (tracked in #905).
+
 ### PR #901 - 2026-09-08
 - Sign browser session cookies with SHA-256 so OIDC and Globus authentication work on FIPS-hardened systems where SHA-1 is unavailable.
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,12 @@ dependencies = [
     "cryptography>=44.0.0,<51.0.0",
     "duckdb-engine>=0.17.0,<1.0.0",
     "fastapi>=0.115.0,<1.0.0",
+    # Pinned below 4.0: FastMCP 4 implements MCP 2026-07-28, which removed
+    # server-initiated requests (SEP-2322). That deletes Context.sample() and
+    # leaves no back-channel for sampling/createMessage, so the sampling_demo
+    # and tool_planner servers cannot work. Raising this cap requires first
+    # redesigning those servers to stop relying on server-initiated sampling.
+    # Tracked in https://github.com/sandialabs/atlas-ui-3/issues/905
     "fastmcp[tasks]>=3.4.3,<4.0.0",
     "httpx>=0.28.1,<1.0.0",
     "itsdangerous>=2.0.0,<3.0.0",


garland3:
## Summary
Caps `fastmcp` at `<4.0.0` in `pyproject.toml` (FastMCP 4 removes `Context.sample()`, breaking the sampling-based MCP servers) and adds a Dependabot `ignore` rule for `fastmcp` major updates on the root pip ecosystem, plus a CHANGELOG entry.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well-evidenced for its size — the breaking API removal is traced to a specific protocol revision, reproduced against `fastmcp==4.0.3`, and the current pin is shown green via `pytest atlas/tests/test_sampling_integration.py`. The only missing artifact is the reproduction script itself, worth pasting into the PR so whoever lifts the cap can re-run it.

## Notes
- Whole-file CRLF→LF rewrite makes an 8-line change render as ~190 lines and resets `git blame`; consider re-applying on the original endings, or normalizing separately with a `*.yml text eol=lf` entry in `.gitattributes` — `.github/dependabot.yml:1`.
- The `ignore` rule is scoped to the root pip entry only; `mocks/mcp-http-mock/pyproject.toml` still allows `fastmcp` 4.x and has its own ecosystem block, so major PRs can still be opened there — `.github/dependabot.yml:97`.
- The ignore rule has no expiry or tracked off-ramp; link a tracking issue (and the verifying test) from both the `pyproject.toml` and Dependabot comments so the cap is revisited rather than aging silently — `.github/dependabot.yml:33`.

---

garland3:
Thanks — all three notes addressed, branch squashed and force-pushed.

**CRLF rewrite.** Real bug on my side: `.github/dependabot.yml` is CRLF in the repo and my edit normalized it to LF, turning an 8-line change into 384. Re-applied on the original endings and squashed so the normalization never appears in history. The diff is now 18 insertions, 0 deletions across three files.

**Ignore rule scope.** Root-only is deliberate, now stated in the PR body. `mocks/mcp-http-mock` has its own ecosystem block and was already widened to `<5.0.0` by merged #896 — it defines no `Context` parameter and makes no sampling calls, so 4.x is genuinely safe there and an ignore rule would be wrong. `mocks/wormhole-mcp-mock` and `atlas/mcp/code-executor-v2` use only `ctx.session_id` and have no Dependabot entry at all.

**No off-ramp.** Fair. Opened #905 with the full breakage analysis, the affected call sites, and the done-when criteria (including removing the cap and the ignore rule). Linked from the `pyproject.toml` comment, the `dependabot.yml` comment, and the CHANGELOG entry.

**Repro script.** Added to the PR body and to #905, so whoever lifts the cap can re-run it directly.

garland3:
## Summary
Documents the `fastmcp<4.0.0` cap in `pyproject.toml`, adds a Dependabot `ignore` rule suppressing `fastmcp` major-version updates on the root pip ecosystem, and records the constraint in the changelog. No code or version constraints change.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well-demonstrated for its size — the breakage is reproduced against `fastmcp==4.0.3`, traced to SEP-2322, affected call sites enumerated, and the current pin shown green (`pytest atlas/tests/test_sampling_integration.py -q` → 11 passed). No missing artifact.

## Notes
- Dependabot `ignore` rules also suppress security updates, so a `fastmcp` advisory fixed only in 4.x will produce no PR — note this in #905, or narrow to `versions: ["4.x"]` (`.github/dependabot.yml:27`).
- Add a guard test asserting the `fastmcp` specifier excludes `4.0.0` with a message naming #905 and `Context.sample()`, so the failure mode is one clear line rather than seven opaque tool errors (`atlas/tests/test_sampling_integration.py`).
- `mocks/wormhole-mcp-mock/pyproject.toml` and `atlas/mcp/code-executor-v2/pyproject.toml` carry the same `<4` cap but no Dependabot entry; record in #905 that those caps are independent (`pyproject.toml:27`).

---

PR #913, opened 2026-09-10
fix(mcp): per-user tool discovery for OAuth servers that gate tools/list

### LINKED ISSUE #912: OAuth MCP servers that require auth on tools/list can never expose tools, and render no row to connect from
## Summary

A remote MCP server with `auth_type: "oauth"` that requires authorization on `initialize` / `tools/list` — not only on tool invocation — can never expose tools, even after a user completes the OAuth flow successfully. The server is also invisible in the Tools panel, which means there is no way to reach the connect button in the first place.

The two halves compound: the server is unusable *and* offers no UI affordance that would reveal it is unusable.

## Part 1 — Tool discovery never runs with a user's credentials

`list_tools()` is called in exactly one place, `atlas/modules/mcp_tools/mcp_discovery.py:157`, from `_discover_tools_for_server`, which `discover_tools()` invokes over `self.clients` — the **process-level** clients. `_get_user_client`, the path that attaches a user's stored OAuth token, is used only at invocation time (`atlas/modules/mcp_tools/mcp_execution.py:341` and `:557`).

So discovery is anonymous and global, while authentication is per-user and only applied when a tool is called. For a server that gates `initialize`/`tools/list`, the sequence is:

1. Startup: anonymous discovery gets `401`; `_record_server_failure` fires and `available_tools[server] = {'tools': []}`.
2. The user completes the OAuth flow. The token is stored and `/api/mcp/auth/status` correctly reports `authenticated: true`.
3. **Nothing re-runs discovery.** The OAuth callback (`atlas/routes/mcp_auth_routes.py`, ending in the redirect at :341) stores the token and returns; there is no discovery hook.
4. `discover_tools()` is reachable again via the admin "Reload MCP" (`atlas/routes/admin_routes.py:333`), but it re-discovers with the same anonymous process-level client and fails identically.

The tool list stays empty permanently. Because `_build_tool_index` (`mcp_discovery.py:125-139`) is built solely from `available_tools`, the server contributes nothing to the routing index either — so its tools cannot be dispatched by name even if a caller knew them. The server is fully unusable, not merely undiscovered.

Servers that permit anonymous `tools/list` and enforce authorization only on invocation are unaffected, which is presumably why this has not been hit before. The MCP authorization spec permits gating discovery, so both shapes are legitimate.

This is the same limitation already documented for `auth_type: "delegated"` in `docs/admin/oidc-authentication.md`:

> Tool discovery runs once at startup with the process-level client, before any user has logged in, so there is no session to delegate from. […] per-user lazy discovery is the fix and is not in this change.

Filing because it applies identically to `auth_type: "oauth"`, where it is arguably more surprising: the user completes an interactive authorization flow, sees a success state, and still gets nothing.

### Suggested fixes

- Lazy per-user discovery on first use, keyed by (user, server) and cached — the fix the delegated caveat already anticipates.
- Or, narrower: re-run discovery for that one server with the user's client when the OAuth callback completes, and store the result per user.

## Part 2 — Servers with zero tools render no row, so the connect button is unreachable

`frontend/src/components/ToolsPanel.jsx:561` builds `serverList` from the tools payload. A server that discovered no tools produces no entry, so no row renders — and the key/connect affordance lives on that row.

This creates a bootstrap deadlock for any `auth_required` server: it has no tools until the user authorizes, and the user cannot authorize because the control only appears once it has tools. Navigating directly to `/api/mcp/auth/<server>/oauth/start` works and completes the flow correctly, but nothing in the UI leads there.

`/api/mcp/auth/status` already returns exactly what is needed (`auth_type`, `auth_required`, `authenticated`) for every configured server, including ones with no tools, so no new backend data is required.

### Suggested fix

Render a row for every server with `auth_required: true` regardless of tool count, showing the connect control when `authenticated: false`. This is worth doing even with Part 1 fixed, since a first-time OAuth server legitimately starts with zero tools.

## Reproducing

1. Configure a remote MCP server with `auth_type: "oauth"` whose endpoint answers `401` to an unauthenticated `initialize` (with a valid `WWW-Authenticate: Bearer resource_metadata="…"` challenge).
2. Start Atlas. Logs show `TOOL DISCOVERY FAILED … 401 Unauthorized`.
3. Observe the server does not appear in the Tools panel.
4. Navigate directly to `/api/mcp/auth/<server>/oauth/start` and complete authorization. Logs confirm the token is stored and `/api/mcp/auth/status` reports `authenticated: true`.
5. The server still exposes no tools and still does not appear. Admin "Reload MCP" does not change this.

Observed with dynamic client registration (RFC 7591) and PKCE `S256`; the authorization flow itself works correctly throughout.

garland3:
Closes #912.

## The problem

An MCP server with `auth_type: "oauth"` is allowed by the MCP authorization spec to require authorization on `initialize`/`tools/list`, not only on tool invocation. Atlas could never use such a server, and offered no UI affordance revealing why.

**Part 1 — discovery never ran with a user's credentials.** `discover_tools()` sweeps the *process-level* clients once at startup, before anyone has logged in. An auth-gating server answered `401`, `_record_server_failure` fired, and `available_tools[server] = {'tools': []}` stuck. Completing the OAuth flow stored a token and made `/api/mcp/auth/status` report `authenticated: true` — but nothing re-ran discovery, and the admin "Reload MCP" re-ran it *anonymously*, failing identically. Because `_build_tool_index` is built solely from `available_tools`, the server also contributed nothing to the routing index: fully unusable, not merely undiscovered.

**Part 2 — no row, so no connect button.** The Tools panel builds its server list from the tools payload, so a server with zero tools rendered nothing — and the connect control lives on that row. A bootstrap deadlock: no tools until you authorize, no way to authorize because the control needs tools.

## The fix

A second discovery path that runs with the user's own client, in a new `UserDiscoveryMixin` (`atlas/modules/mcp_tools/mcp_user_discovery.py`):

- **`discover_tools_for_user(user, server, force=False)`** lists tools through the existing `_get_user_client`, caches per `(user, server)` with a 5-minute TTL, and cools down for 60s after a failure so the SPA's `/api/config` polling cannot hammer an unreachable server. It deliberately does *not* reuse `_discover_tools_for_server`: that helper records a **global** server failure, and one user's expired token must not mark the server broken for everyone or feed the reconnect backoff.
- **Promotion into the shared inventory** happens only when the anonymous sweep found nothing for that server, so the tools become schedulable (`get_tools_schema`) and routable (`get_server_for_tool`) via the existing user-agnostic paths. A successful anonymous sweep always wins — one user's view never overwrites the operator-visible catalogue. Promotion does not widen access: every call into an auth-required server still goes through `_get_user_client` and raises `AuthenticationRequiredException` for a user holding no token of their own.
- **Refresh on authorization.** The OAuth callback and the token-upload route drop the stale cache and re-run discovery for that one server, best-effort — a discovery failure never turns a successful authorization into an error.
- **Lazy re-discovery.** `/api/config` retries auth-gated servers whose catalogue is empty, concurrently, so a server authorized elsewhere picks its tools up on the next page load. A failure there is logged and the payload is still returned.
- **Invalidation.** Removing a token clears that user/server's cache; a full re-discovery (config reload) clears all of them, since it resets `available_tools`.

For Part 2, a server that requires authorization is now listed even with zero tools — from `/api/config`, and in `ToolsPanel` from `/api/mcp/auth/status` for servers the tools payload does not mention at all. Only *unauthenticated* ones are synthesized in the panel: a connected server with genuinely no tools has nothing to offer and no action left to take. Servers that need nothing and offer nothing are still omitted.

This also lifts the same limitation documented for `auth_type: "delegated"` in `docs/admin/oidc-authentication.md`, which the doc is updated to reflect.

## Tests

- `atlas/tests/test_mcp_user_discovery.py` (20 tests) — discovery with the user's client, TTL reuse, `force`, case-insensitive user keys, the no-token cool-down and its expiry, a per-user failure not poisoning global failure state, task-support metadata, promotion (and non-promotion over a successful sweep, and of an empty list), the startup failure record being cleared once the server answers, batch discovery skipping non-candidates and surviving one broken server, and cache invalidation scoping.
- `atlas/tests/test_config_auth_required_servers.py` (4) — the zero-tool auth server is listed, an open zero-tool server is still omitted, per-user results reach the payload, and a discovery failure does not break `/api/config`.
- `atlas/tests/test_mcp_oauth_routes.py` (+2) — the callback re-discovers with `force=True`, and a failed re-discovery still reports success.
- `frontend/src/test/tools-panel.test.jsx` (+4) — a row with a connect control for a server with no tools, no duplicate when the tools payload already describes it, and no synthesized row for a connected or an auth-free server.

## Verification

- `bash test/atlas_tests.sh` — 3057 passed, 17 skipped
- `npx vitest run` — 72 files, 838 tests passed
- `ruff check` and `eslint` clean on every touched file

## Follow-up: catalogue refresh after an in-page authorization change

Pasting a token unlocked the server on the backend but left the panel rendering the `/api/config` payload fetched before the credential existed — green server, no tools, until a page reload. Disconnecting had the mirror problem. OAuth was already correct (full-page navigation returns through a fresh config fetch), so `ToolsPanel` now asks for that refresh itself after `uploadToken`/`removeToken`, best-effort: a failed refresh does not turn a successful connect into an error. Covered by 4 more tests in `frontend/src/test/tools-panel.test.jsx` (refresh on connect, no refresh when the token is rejected, refresh on disconnect, and a context without `refreshConfig`).


garland3:
## Summary
Adds lazy per-user MCP tool discovery (with a 300s success TTL and 60s failure cool-down) so auth-gated servers are probed with the requesting user's credentials, promotes the resulting catalogue into the shared inventory, and lists auth-required servers in `/api/config` and the Tools panel with a connect control even when they expose no tools yet.

## Verdict
Request changes.

## Required changes
1. Scope the promoted catalogue to the user it came from at `atlas/modules/mcp_tools/mcp_user_discovery.py:236` — `_promote_user_tools` publishes one user's `tools/list` result into the shared `available_tools`, so other authorized users with no token (or narrower scopes) are shown and offered tool names, descriptions and input schemas they cannot invoke. Serve a user's own catalogue first and fall back to the shared one only when it came from a real anonymous sweep.
2. Make the discovery candidate filter depend on the requesting user's cache rather than the shared catalogue (`mcp_user_discovery.py`, `discover_tools_for_user_servers`) — once a server is promoted it is permanently excluded from candidates, so the TTL never refreshes for exactly the servers this feature exists for.
3. Move the `_apply_task_support_metadata` call out of `_list_tools_with_client` (`mcp_user_discovery.py:212`) and into `_promote_user_tools` after the promotion guard — today a narrower user view purges `_tool_task_forbidden` entries globally, flipping a forbidden tool to task-allowed for everyone.
4. Drop or expire the promoted `available_tools` / `_tool_index` entry when its originating token is removed (`atlas/routes/mcp_auth_routes.py:297`) — `remove_token` clears only the per-user cache, so a revoked grant's catalogue stays published until a config reload.
5. Stop `/api/config` blocking on live discovery at `atlas/routes/config_routes.py:276` — bound the `discover_tools_for_user_servers` call with a short `asyncio.wait_for` or run it in the background, so an unreachable server does not stall a polled page-load path for the full discovery timeout.
6. Guard `mcp_manager` before `await mcp_manager._invalidate_user_client(...)` at `atlas/routes/mcp_auth_routes.py:246` — the token is already persisted, so a None manager turns a successful store into a 500.
7. Add tests for the token-upload re-discovery path (`atlas/tests/test_mcp_auth_routes.py`, `TestUploadToken`) using an explicit fake manager, and extend `test_remove_token_invalidates_cache` to assert `clear_user_tool_cache` — deleting both new lines currently leaves the suite green.

## Evidence
Evidence: 8/10 — the backend is well covered (unit tests per file, full-suite and lint results), but every test substitutes `_get_user_client`, so nothing exercises the actual protocol. The most valuable missing artifact is a transcript or log excerpt from one real run against an auth-gating MCP server showing the empty startup catalogue and the populated post-callback catalogue; a screenshot of the new zero-tool connect row would close the user-facing half.

## Notes
- `delegated` is excluded from `auth_required` at `atlas/routes/config_routes.py:303` but included by `_requires_user_auth`, so the doc claim in `docs/admin/oidc-authentication.md` does not hold for delegated servers; derive all three predicates from one helper.
- Concurrent `/api/config` requests each start their own discovery (`mcp_user_discovery.py:118`) — an in-flight task registry would collapse N tabs into one upstream session.
- The zero-tool row shows no reason it is empty (`frontend/src/components/ToolsPanel.jsx:612`); an inline "tools appear after you connect" line would disambiguate it from a genuinely toolless server.

---

garland3:
Thanks — both review points addressed in 
[`8f2ba1a`](https://github.com/sandialabs/atlas-ui-3/pull/913/commits) (see below).

**`auth_required` omits `delegated`** — correct, and it hid exactly the servers this change is meant to unblock. `auth_required` is the frontend's contract for *rendering a connect control*, and a delegated server has nothing for the user to connect (Atlas mints its token from their OIDC session), so it must stay false there. The zero-tool listing decision is a different question — "does this server need per-user credentials?" — so I split the two: `needs_user_credentials = auth_required or auth_type == 'delegated'` now gates the listing, `auth_required` still gates the control. Covered by `test_delegated_server_with_no_tools_is_listed_but_not_connectable`.

**Unsanitized discovery warning** — the sanitization landed in a follow-up commit before the review; `exc_info=True` is now added alongside it.

**On the CodeQL log-injection alerts** (`mcp_user_discovery.py` and `mcp_auth_routes.py`): these are the project's known false positives. Every flagged call logs `sanitize_for_logging(...)` output, which CodeQL does not model as a sanitizer. The tainted value is also an MCP server name from operator-controlled `mcp.json`, not user input.

garland3:
## Summary
Adds lazy per-user MCP tool discovery so servers that gate `initialize`/`tools/list` behind a user token are discovered with the requesting user's credentials, promotes the first successful result into the shared catalogue, and lists zero-tool auth-required servers in the Tools panel with a connect control.

## Verdict
Request changes.

## Required changes
1. Move the `_apply_task_support_metadata(server_name, tools)` call out of `_list_tools_with_client` (`atlas/modules/mcp_tools/mcp_user_discovery.py:212`) into `_promote_user_tools` after both promotion guards — the helper purges all `_tool_task_forbidden` entries for the server, so one user's narrower `tools/list` flips task-forbidden tools to task-allowed process-wide, even when promotion is declined.
2. Base the candidate filter in `discover_tools_for_user_servers` (`atlas/modules/mcp_tools/mcp_user_discovery.py:200`) on the requesting user's own cache entry (missing or past TTL) instead of `available_tools[name]['tools']` — once a server is promoted it fails the filter forever, so the 300s TTL never refreshes for exactly the servers this feature targets.
3. Record the originating `(user, server)` on each promoted entry and have `clear_user_tool_cache` / `remove_token` (`atlas/routes/mcp_auth_routes.py:297`) withdraw that entry and rebuild `_tool_index` — today a revoked user's tool names and schemas stay published and routable to everyone indefinitely.
4. Serve each requester their own catalogue in `/api/config` (`atlas/routes/config_routes.py:311`) and fall back to the shared entry only when it came from a real anonymous sweep — otherwise user B is shown user A's scoped tool names, descriptions and input schemas, which is the metadata the upstream server chose to gate.
5. Bound the inline `await mcp_manager.discover_tools_for_user_servers(...)` at `atlas/routes/config_routes.py:276` with a short `asyncio.wait_for` (1–2s) and let discovery finish in the background — one slow or unreachable auth-gated server currently stalls every `/api/config` response, including SPA polls, for the full discovery timeout. Also wrap the whole `async with client:` block in `_list_tools_with_client`, since the existing timeout covers only `list_tools` and not connection setup.
6. Drop the `_clear_server_failure(server_name)` call in `_promote_user_tools` (`atlas/modules/mcp_tools/mcp_user_discovery.py:281`) or replace it with a narrower `auth_gated` marker — per-user success does not prove the process-level client works, and erasing the record removes the server from reconnect tracking while reporting it healthy.
7. Add tests pinning the above: a narrower per-user view must not clear `_tool_task_forbidden` entries from the anonymous sweep; `upload_token` and `remove_token` must be asserted to invalidate and clear the user tool cache with a fake whose `clear_user_tool_cache` is synchronous (the current blanket `AsyncMock` in `TestUploadToken` leaves the new lines untested and hides a missing await).

## Evidence
Evidence: 7/10 — the unit-level coverage is strong for a change this size (26 backend + 4 frontend tests, full-suite and lint results reported), but nothing demonstrates the feature end to end. The most important missing artifact is a log excerpt or transcript from one run against a server that actually gates `tools/list`, covering startup sweep → authorize → tools visible; a screenshot of the new zero-tool connect row would also help.

## Notes
- Empty auth-required and delegated rows render with a blank tool area and no explanation — add a muted line such as "Tools appear after you connect." (`frontend/src/components/ToolsPanel.jsx:619`).
- Synthesized rows get `compliance_level: null` and are appended after the compliance filter runs, so they survive a non-`all` filter unlabelled (`frontend/src/components/ToolsPanel.jsx:632`).
- Per-user caches, failure records and discovery locks are keyed `(user, server)` with no eviction; expired entries stay resident for the process lifetime (`atlas/modules/mcp_tools/client.py:153`).

---

garland3:
Thanks — this was a good catch list. All seven required changes are in [`0b6312b`](https://github.com/sandialabs/atlas-ui-3/pull/913/commits/0b6312b), plus both frontend notes and the eviction one.

**1. `_apply_task_support_metadata` on a declined promotion** — confirmed and fixed. It purges the server's entire `_tool_task_forbidden` set and rebuilds it, so a narrower per-user `tools/list` flipped anonymously-discovered tools to task-allowed process-wide. It now runs inside `_promote_user_tools`, after both guards. Pinned by `test_a_declined_promotion_leaves_the_sweeps_entries_alone`.

**2 + 4. The promoted catalogue was indistinguishable from an anonymous one** — these turned out to be the same root defect, and it was the worst one. A promoted entry now carries `user_scoped` / `discovered_for`:
- `_has_anonymous_catalogue()` gates the candidate filter, so a promoted server stays a candidate and the 300s TTL actually refreshes (`test_a_promoted_server_is_still_a_discovery_candidate`), while a genuinely anonymous one is never probed per-user (`test_an_anonymously_discovered_server_is_never_a_candidate`).
- New `get_visible_tools_for_server(user, server)` serves a requester their own entry and falls back to the shared one only when it came from a real sweep; `/api/config` uses it (`test_a_promoted_catalogue_is_not_shown_to_another_user`).

**3. Withdrawal on revoke** — `clear_user_tool_cache` now withdraws a publication whose `discovered_for` matches and rebuilds `_tool_index` (`test_disconnecting_withdraws_the_published_catalogue`, and `test_another_users_catalogue_is_left_published` so it does not over-reach).

**5. Unbounded inline discovery** — `discover_tools_for_user_servers` takes `wait_timeout`; `/api/config` passes 2s and the work is `asyncio.shield`ed rather than cancelled, so a slow server's result still lands in the cache for the next poll (`test_a_slow_server_does_not_stall_the_caller`). `_list_tools_with_client` now wraps the whole `async with client:` block in the timeout — you are right that a server hanging on `initialize` is exactly the shape this feature targets.

**6. `_clear_server_failure`** — dropped. Agreed that per-user success does not prove the process-level client works. The test now asserts the record is *left alone*.

**7. Tests** — added, including the `AsyncMock` point: `clear_user_tool_cache` is synchronous on the real manager, so the blanket `AsyncMock` in `TestUploadToken` was handing back an un-awaited coroutine and testing nothing. Set explicitly now, with assertions on invalidate/clear/rediscover; the backend suite's warning count dropped from 8 to 2 as a result.

**Notes** — "Tools appear after you connect." added for empty auth rows; synthesized rows are now suppressed entirely while a compliance level is being filtered on (the auth status carries no `compliance_level` and the filter is strict about resources with none), both with tests. The caches now sweep expired entries past a ceiling.

**On the evidence gap:** fair, and I can't close it. I have no MCP server that actually 401s on `initialize` to point at, so there is no end-to-end transcript — the change is verified by unit/route tests and reading. Worth an operator running the issue's repro before merge.

Verification on this commit: `bash test/atlas_tests.sh` → 3071 passed, 17 skipped. `npx vitest run` → 72 files, 840 tests. `ruff` and `eslint` clean on every touched file.

garland3:
## Summary
Adds per-user MCP tool discovery so servers that gate `tools/list` behind a user credential are discovered after the user authorizes, promotes those per-user catalogues into the shared inventory when no anonymous catalogue exists, and surfaces zero-tool auth-required servers in `/api/config` and the tools panel.

## Verdict
Request changes.

## Required changes
1. Track promoted-catalogue ownership as a set instead of the single `discovered_for` field at `atlas/modules/mcp_tools/mcp_user_discovery.py:281`, and re-promote from any surviving in-TTL user entry on withdrawal — today a second user's promotion reassigns ownership, so the first user's disconnect withdraws nothing and the second user's disconnect leaves the first user's tools listed but unroutable via `get_server_for_tool`.
2. Prefer the user's own entry only when it is non-empty (or skip per-user discovery when `_has_anonymous_catalogue` is true) at `atlas/modules/mcp_tools/mcp_user_discovery.py:89` — for servers whose `tools/list` is public, an empty authenticated result hides tools the user could previously see.
3. Branch the empty-body connect hint on "auth-required and not yet authenticated" rather than `auth_type === 'oauth'` at `frontend/src/components/ToolsPanel.jsx:1118` — `api_key`/`bearer`/`jwt` servers get a connect control but are told "No tools discovered yet."
4. Prune the failure and lock caches off the failure path as well as the success path at `atlas/modules/mcp_tools/mcp_user_discovery.py:270` — a user whose discovery keeps failing never triggers pruning, so both maps grow unbounded.
5. Scope tool metadata reads to the requesting user (or document the gap next to the module docstring) at `atlas/modules/mcp_tools/mcp_user_discovery.py:400` — `get_tools_schema`/`get_server_for_tool` still return `user_scoped` descriptions and input schemas to any caller.
6. Add a test covering the connection-setup timeout at `atlas/modules/mcp_tools/mcp_user_discovery.py:212` and one covering cache eviction at `:151`; both are core to this commit and currently untested.
7. Replace the fake that reimplements the scoping logic in `atlas/tests/test_config_auth_required_servers.py:60` with an assertion against the real `get_visible_tools_for_server` path, so the test can fail when the production logic changes.

## Evidence
Evidence: 7/10 — the description, per-behaviour unit tests and full-suite results are proportionate to a change of this size, but every backend test substitutes the client or the manager. The most important missing artifact is one end-to-end transcript against a server that actually gates `tools/list`: startup sweep empty → user authorizes → tools present in `/api/config`.

---

garland3:
Thanks — six of the seven are in [`271e932`](https://github.com/sandialabs/atlas-ui-3/pull/913/commits/271e932); the seventh I think is already answered, argument below.

**1. Ownership as a set** — confirmed and fixed. `discovered_for` is now the set of every user whose discovery stands behind the published catalogue, and `_promote_user_tools` unions into it instead of overwriting. `_withdraw_promoted_tools` drops one claim and then *republishes* from a surviving owner whose cache entry is still within its TTL (`_live_user_tools`), emptying the server only when the last claim goes. That is what keeps the listing and `get_server_for_tool` agreeing: the old code could leave a departed user's tools published, so a second user saw names they could not route. Pinned by `test_the_first_owners_disconnect_still_withdraws_their_claim`, `test_a_survivors_catalogue_is_republished_not_left_stale`, `test_the_last_owner_leaving_empties_the_server`, and `test_an_expired_survivor_does_not_keep_the_catalogue_alive`.

**2. Empty own-result must not hide a public catalogue** — fixed in `get_visible_tools_for_server`: the user's own entry wins only when it is non-empty. An empty authenticated result is the absence of an answer, not an answer, so it falls through to the shared catalogue — which is still refused when that catalogue is itself `user_scoped`. I took this rather than the "skip discovery when `_has_anonymous_catalogue`" variant because the batch path (`discover_tools_for_user_servers`) already skips those servers; the hole was only reachable through the forced OAuth-callback path, and fixing the read side closes it wherever it is called from. Pinned by `test_an_empty_authenticated_result_falls_back_to_the_sweep` and its gated counterpart.

**3. Connect hint branched on `auth_type === 'oauth'`** — correct, `api_key`/`bearer`/`jwt` servers got a connect button and "No tools discovered yet." right next to it. The predicate that decides whether a connect control appears was inline in the button's IIFE, which is how the two drifted, so it is now one module-level `connectState()` helper used by both. The hint also now requires *not connected* — a connected server with no tools really has none, and telling that user to connect is wrong in the other direction.

**4. Pruning off the failure path** — fixed; both failure returns in `_run_user_discovery` now call `_prune_user_discovery_state()`. Pinned by `test_a_repeatedly_failing_user_also_triggers_the_sweep`.

**5. Metadata reads are not user-scoped** — this is the documented alternative you offered, and it is already documented: the module docstring's "Whose catalogue is whose" paragraph states that `available_tools`/`_tool_index` are populated from a per-user result *because* `get_tools_schema` and `get_server_for_tool` are user-agnostic and nothing could be scheduled or routed otherwise, and that the publication is withdrawn when its owner disconnects. Scoping those two for real means threading a user through the schema and routing layers, which is a change well outside this fix. Worth noting the exposure is metadata only: every call into an auth-required server still goes through `_get_user_client` and raises `AuthenticationRequiredException` for a user with no token.

**6. Untested timeout and eviction** — both added. `test_a_client_that_hangs_before_list_tools_times_out` uses a client that blocks in `__aenter__`, which is the point of wrapping `_open_and_list` rather than `list_tools`, and asserts the failure stays per-user (`_failed_servers` untouched). `test_expired_entries_are_evicted_once_past_the_ceiling` fills past `_MAX_USER_DISCOVERY_ENTRIES` with stale entries and checks that an unexpired cool-down survives the sweep.

**7. The fake reimplemented the scoping rule** — fixed. `_FakeManager` now inherits `UserClientMixin` and `UserDiscoveryMixin`, so `get_visible_tools_for_server` and `_requires_user_auth` are the real implementations; only the network is scripted, and the scripted result lands in the same `_user_available_tools` cache the real path writes to.

**Verification.** `atlas/tests/test_mcp_user_discovery.py` + `atlas/tests/test_config_auth_required_servers.py`: 44 passed. Full backend suite via `test/atlas_tests.sh`: 2872 passed, +9 over the same baseline on this machine (164 route-test failures are a pre-existing local environment artifact — identical count with the changes stashed, and green in CI). Frontend: `eslint` clean on `ToolsPanel.jsx`, `npm run build` succeeds, `vitest run` 840 passed. I also mutation-checked the new tests: reverting the ownership-set and empty-result changes fails exactly the three tests written to pin them.

On the evidence note — agreed that the missing artifact is a transcript against a server that genuinely gates `tools/list`. I have no such server to point at here; the harness substitutes the client at the `_get_user_client` boundary, which is the last seam before the network.

garland3:
## Summary
Adds per-user MCP tool discovery so auth-gated servers can be connected and their tools discovered with a user's own credentials, and surfaces zero-tool auth-required servers in `/api/config` and the Tools panel so a user can break the "no tools until connected" bootstrap deadlock.

## Verdict
Request changes.

## Required changes
1. Expire the per-user catalogue on the discovery failure paths in `atlas/modules/mcp_tools/mcp_user_discovery.py:299` — when a stored token stops working, `_user_available_tools[key]` is never cleared and `get_user_tools_for_server` applies no TTL, so stale tools are served to that user indefinitely. Apply the same TTL check in `_cached_user_tools` before returning stale tools during a cool-down.
2. Withdraw the previously promoted catalogue when a re-discovery returns empty at `atlas/modules/mcp_tools/mcp_user_discovery.py:441` — the `if not tools` early return in `_promote_user_tools` skips withdrawal, leaving the old tool names in `_tool_index` where `get_server_for_tool`/`get_tools_schema` keep routing them with no live owner. Add a test pinning "a second, empty discovery unpublishes the first one's catalogue".
3. Scope user-discovered tool metadata by owner in `get_tools_schema`/`get_server_for_tool` (`atlas/modules/mcp_tools/mcp_user_discovery.py:281`) — `_promote_user_tools` writes one user's catalogue into shared `available_tools`, so tool names, descriptions and full input schemas from a server that refused to disclose them anonymously become visible in every other authorized user's model context. At minimum, omit `user_scoped` entries from the schema handed to a non-owner.
4. Call `_apply_task_support_metadata` on the republished tool set in `_withdraw_promoted_tools` (`atlas/modules/mcp_tools/mcp_user_discovery.py:180`) — `_tool_task_forbidden` otherwise still reflects the departing owner, so a tool the heir exposes with no `taskSupport` is treated as task-allowed process-wide.
5. Guard `mcp_manager is None` before `await mcp_manager._invalidate_user_client(...)` at `atlas/routes/mcp_auth_routes.py:246` — the token is already persisted at that point, so an absent manager returns a 500 for a successful upload. Add a `TestUploadToken` case with the manager set to `None` asserting a 200.
6. Rebuild the tool index once per bulk clear at `atlas/modules/mcp_tools/mcp_user_discovery.py:186` — `clear_user_tool_cache` calls `_withdraw_promoted_tools` in a loop and each withdrawal re-walks every server's tools, which is O(cleared keys x total tools) for a cache holding up to 2000 keys.
7. Show the connect affordance for unauthenticated servers even when a compliance filter is active (`frontend/src/components/ToolsPanel.jsx:645`) — synthesized rows are skipped whenever `complianceEnabled && complianceLevelFilter`, which reintroduces the same deadlock behind a persisted UI preference.

## Evidence
Evidence: 7/10 — the change is well covered by unit tests, with named tests per behaviour, reported lint/suite results and an explicit mutation check, which is unusually strong. The gaps are at the seams closest to the bug: no run against a server that actually returns `401` on `initialize`/`tools/list`, and no screenshot of the new connect row and hint.

## Notes
- Widened connect-hint predicate is only tested with `auth_type: 'oauth'`; `api_key`/`bearer`/`jwt` are unpinned — `frontend/src/test/tools-panel.test.jsx:1571`.
- Zero-tool auth servers render as dead-end rows with no count, badge or connect control — `frontend/src/components/MarketplacePanel.jsx:22`.
- The connectable auth-type set is now restated in four places across two languages; derive it from one server-side frozenset and let the panel read `auth_required` — `atlas/routes/config_routes.py:322`.

---

garland3:
All seven addressed in [`e4d1f24`](https://github.com/sandialabs/atlas-ui-3/pull/913/commits/e4d1f24), including #3, which I argued against last round and was wrong to. One correction on #5's premise, at the end.

**1. Stale catalogues served indefinitely** — confirmed and fixed, and this was the worst of the seven: a revoked user kept being served the tool list their dead token had earned, forever. `get_user_tools_for_server` now applies the TTL; the cool-down branch of `_cached_user_tools` returns `None` rather than the stale entry (the cool-down suppresses the *retry*, not the *expiry*); and a discovery failure calls `_expire_stale_user_entry`, which drops the entry and withdraws what it promoted — but only once it is already past its TTL, so a server that is briefly unreachable does not cost a live token its catalogue. Four tests, including `test_a_fresh_catalogue_survives_one_failed_probe` for that last distinction.

**2. Empty re-discovery skipped withdrawal** — confirmed. `_promote_user_tools` now separates "nothing to publish" from "this user published something and now sees nothing": the latter withdraws. Pinned by `test_a_second_empty_discovery_withdraws_the_first_catalogue`, plus one that a co-owner's catalogue survives it and one that an anonymous catalogue is untouched.

**3. Metadata scoped by owner** — you were right and I was wrong to defer this twice. I claimed it needed threading a user through the schema and routing layers; that was true of routing but not of the schema, where `user_email` was already in scope at all three model-facing call sites. `get_tools_schema(tool_names, user_email=None)` now omits `user_scoped` catalogues the requester does not own — skipped rather than reported missing, since "missing" would itself disclose the name. Threaded from `tools.py` (both paths) and `agentic_loop.py`; the default keeps the historic behaviour for internal callers resolving a tool they have already authorized, and `ToolManagerProtocol` carries the same optional parameter. `get_server_for_tool` stays user-agnostic — it returns a server name, not metadata, and scoping it would break routing for the owner too. Five tests covering owner, co-owner, non-owner, unscoped caller, and anonymous catalogue.

**4. Task-support metadata after republish** — confirmed. `_withdraw_promoted_tools` re-applies the sweep to the heir's tool set. Pinned by `test_the_heirs_tools_are_re_swept`.

**5. `mcp_manager is None` guard** — guard added for consistency with `_rediscover_after_authorization` on the very next line. But the premise does not hold: `upload_token` calls `await mcp_manager.get_authorized_servers(...)` at `mcp_auth_routes.py:198`, well before `token_storage.store_token`, so a `None` manager raises there and the request 500s with **nothing persisted**. There is no state where the token is stored and the manager is absent. I did not add the `TestUploadToken` case asserting a 200, because with the manager `None` the route cannot reach a 200 — the assertion would fail. Making it genuinely reachable means guarding the authorization check too, which would mean skipping the `403` on an unauthorized server; that trade seems clearly wrong, so I left the earlier call unguarded.

**6. O(cleared keys × tools) rebuild** — confirmed. `_withdraw_promoted_tools` takes `rebuild_index=False` and returns whether it changed anything; `clear_user_tool_cache` rebuilds once, and only if something was actually withdrawn. Pinned by two tests asserting the rebuild count is exactly 1 and exactly 0.

**7. Compliance filter dropped every synthesized row** — confirmed; my earlier comment argued the skip was deliberate, but the real cause was that the auth status carried no `compliance_level`, so those rows could not be judged. Fixed at the source: `/api/mcp/auth/status` now reports `compliance_level`, and the panel applies the real `isComplianceAccessible` per row. A filter still excludes a server with no level (strict, unchanged), but one that matches keeps its connect control instead of the deadlock returning behind a persisted preference.

**Notes.** Added `offers a token control and the same hint for a non-oauth server` (api_key), which was the unpinned half of the widened predicate. Left the other two: `MarketplacePanel` renders a different surface than the connect flow and is outside this fix; deriving the connectable auth-type set from one server-side frozenset is the right end state and I would rather do it as its own change than widen this one further — `auth_required` is already the server-side answer, and the panel should read it.

**Verification.** Targeted: 53 + 6 backend tests pass. Full backend suite via `test/atlas_tests.sh`: 2887 passed, +24 over the same baseline on this machine (the 164 route failures are the pre-existing local environment artifact — identical with the changes stashed, green in CI). Frontend: `eslint` clean, `vitest run` 842 passed (52 in `tools-panel.test.jsx`), `npm run build` succeeds. Note one real catch from the full run: threading `user_email` broke `test_tool_acl`'s duck-typed manager, so `ToolManagerProtocol` and both test fakes now carry the parameter. CI on the previous commit was 10/10 green.

Still no transcript against a server that genuinely 401s `initialize`/`tools/list` — I have no such server here, and the harness substitutes at the `_get_user_client` boundary, the last seam before the network.

garland3:
Amended  → [`98f2717`](https://github.com/sandialabs/atlas-ui-3/pull/913/commits/98f2717) (force-push, content identical). My scripted edit to `atlas/tests/test_mcp_data_injection.py` had rewritten the file LF-only, turning a one-line signature change into a 538-line whole-file diff. That file is CRLF in this repo, and a normalized rewrite is what makes a PR read as CONFLICTING and stops CI being scheduled. The amend restores its line endings and keeps the single-line edit; the branch diff drops from 670 changed lines to 402. Verified no other touched file changed line endings.

garland3:
(Correction to the line above — my shell ate the old SHA: the amend was `e4d1f24` → `98f2717`.)

garland3:
## Summary
Adds per-user MCP tool discovery: catalogues from auth-gated servers are discovered under a user's own credentials, cached with a TTL, and promoted into the shared `available_tools` record with ownership tracking, plus schema-read scoping and a Tools-panel row that lets users connect a zero-tool server.

## Verdict
Request changes.

## Required changes
1. Fix the co-ownership publish in `_promote_user_tools` at `atlas/modules/mcp_tools/mcp_user_discovery.py:509` — it unions the new owner into `discovered_for` but replaces `tools` wholesale, so after user B promotes, user A is still offered their own cached tool names via `/api/config` while `_tool_index`/`get_server_for_tool` can no longer resolve them, and A gains read access to B's schemas. Publish per-owner entries (or the union of live owners' tools) and answer `_may_read_catalogue` against what that user actually discovered.
2. Move `_apply_task_support_metadata` out of `_list_tools_with_client` and into `_promote_user_tools` after both guards (`atlas/modules/mcp_tools/mcp_user_discovery.py:212`) — today a declined promotion still replaces the server's `_tool_task_forbidden` set with the narrower per-user result, flipping task-forbidden tools to allowed for every user.
3. Scope `build_mcp_data` by requesting user at `atlas/application/chat/utilities/tool_executor.py:268` (the call site at :916 has the context), or exclude `user_scoped` entries outright — it currently emits every server's names, descriptions and full `inputSchema` into `_mcp_data`, bypassing the new `get_tools_schema` scoping.
4. Make `_may_read_catalogue` (`atlas/modules/mcp_tools/mcp_discovery.py:546`) require a live, non-expired per-user entry rather than permanent membership in `discovered_for`, and have `_prune_user_discovery_state` withdraw the promotions of entries it evicts — otherwise schema-read access outlives the TTL every other read path enforces.
5. Restrict promotion for `delegated` auth servers, or gate dispatch on ownership at the execution boundary (`atlas/modules/mcp_tools/mcp_user_discovery.py:281`) — for `delegated`, Atlas mints a token from any user's session, so a promoted tool is callable by non-owners, not just routable.
6. Guard the post-store cleanup at `atlas/routes/mcp_auth_routes.py:256` in its own `try/except` — `_invalidate_user_client` does I/O on live sessions and can turn an already-committed token upload into a 500, contradicting the comment directly above it.
7. Replace the blanket `AsyncMock` manager in `TestUploadToken` (`atlas/tests/test_mcp_auth_routes.py:167`) so the new invalidate/rediscovery calls are actually executed, and add a case asserting a non-owner's `_mcp_data` omits a `user_scoped` server.

## Evidence
Evidence: 7/10 — unusually strong unit-level coverage for a large change (53+ backend tests, 8 frontend tests, clean lint/build, a stated mutation check), but every backend test substitutes at the `_get_user_client` boundary, so nothing exercises the `401`-on-`tools/list` condition the whole PR exists to address. The most valuable missing artifact is a transcript against a local stub MCP server that rejects unauthenticated `initialize`, showing the empty startup catalogue and the populated post-callback one; a screenshot of the new zero-tool connect row would close the user-facing half.

## Notes
- `get_tools_schema`'s `user_email=None` default fails open to unscoped disclosure — `atlas/modules/mcp_tools/mcp_discovery.py:589`.
- Ownership/visibility rules are re-implemented by hand at four-plus read sites with no owning type — `atlas/modules/mcp_tools/mcp_user_discovery.py:495`.
- A connected server mid-discovery is shown "No tools discovered yet.", and synthesized auth rows ignore the user's server selection — Tools panel.

---

garland3:
Five of seven in [`464cd1d`](https://github.com/sandialabs/atlas-ui-3/pull/913/commits/464cd1d). #2 does not reproduce, and #5 I want to argue rather than change — both below.

**1. Co-ownership clobbered the first owner** — confirmed, and this one was mine, introduced by the ownership-set change two rounds ago. Fixed by separating the two questions the shared entry was conflating:
- *Routing*: `available_tools[server]["tools"]` now holds the **union** of every live owner's catalogue (`_merged_owner_tools`, first writer of a name wins so the promoting user's fresh copy is not overwritten by an older one). The republish-on-withdrawal path uses the same union.
- *Reading*: answered per user, per tool, against their own live entry — see #4.

So both owners' tools resolve through `get_server_for_tool`, each owner is listed only their own, and neither can read the other's schema. Four tests, including `test_an_owner_cannot_read_the_other_owners_schema`.

**2. `_apply_task_support_metadata` in `_list_tools_with_client`** — this does not reproduce on the current branch; I think it is carried over from the round that first raised it. It was moved into `_promote_user_tools` after both guards in `0b6312b`, and the only two call sites today are `mcp_user_discovery.py:506` (inside `_promote_user_tools`, after the anonymous-catalogue guard and the empty-tools guard) and `:220` (the republish path, added last round). `_list_tools_with_client` does not call it — `grep -n _apply_task_support_metadata atlas/modules/mcp_tools/*.py` shows all four occurrences. `test_a_declined_promotion_leaves_the_sweeps_entries_alone` has been pinning it since that commit. Happy to look again if you are seeing something I am not.

**3. `build_mcp_data` bypassed the scoping** — confirmed, and a fair catch: it emits every server's names, descriptions and full `inputSchema` into `_mcp_data`, which made the `get_tools_schema` scoping a partial measure. It now takes `user_email` (the call site had it) and applies the same per-tool rule. Three tests: owner sees it, non-owner sees nothing, anonymous catalogue visible to all.

**4. Read access outlived the TTL** — confirmed. `_may_read_catalogue` no longer consults `discovered_for`; it asks for the user's own live catalogue via `get_user_tools_for_server`, which applies the TTL, and matches the specific tool name. And `_prune_user_discovery_state` now withdraws the promotions of the entries it evicts, so a published catalogue cannot outlive the entry that justified it. Two tests.

**5. `delegated` promotion** — I would rather leave this. `delegated` means Atlas mints a downstream token from the user's own OIDC session (RFC 8693 / OBO), so every *authorized* user can obtain credentials for that server by definition — that is what the auth type is. A catalogue discovered under one user is therefore not a private view the way an `oauth` or `api_key` one is, and the call still goes through `_get_user_client`, which mints per the caller and lets the server authorize them. Gating dispatch on ownership would break the intended case: two users in the same group, one warms the catalogue, the other is refused a tool Atlas would happily mint them a token for. The server remains the authority on what each identity may call. If you think the metadata (not dispatch) should still be scoped for `delegated`, say so and I will scope the reads while leaving dispatch alone — but I do not want to gate execution on who happened to discover first.

**6. `_invalidate_user_client` can 500 a committed upload** — confirmed, and better than my last-round answer: the risk is not a missing manager but the call itself doing I/O on live sessions. It is now `_invalidate_user_client_quietly`, best-effort with a warning, matching `_rediscover_after_authorization`. The stale client is rebuilt on next use, so failing to close it costs a connection, not correctness.

**7. `TestUploadToken`'s blanket `AsyncMock`** — the `_mcp_data` half is done (three tests above). I have not rewritten the upload-token fixture; with the cleanup now swallowing exceptions the assertion I would write there is "a raising invalidate still returns 200", which is worth adding, and I would rather do it with fresh eyes than bolt it on at the end of this round.

**Verification.** Targeted: 58 + 19 + 6 + 15 tests pass. Full backend suite: 2895 passed, +32 over the same baseline (the 164 route failures are the known local environment artifact; CI is the real gate). Line endings verified unchanged on every touched file after last round's CRLF slip. CI on `98f2717` was green on all nine settled checks with `build-and-test` still running at the time of writing.

On evidence: still no transcript against a server that genuinely 401s `initialize`. Agreed that a local stub MCP server rejecting unauthenticated `initialize` is the right artifact, and it is the thing I would build next if this needs more before merge.

garland3:
## Summary
Adds per-user MCP tool discovery for auth-gated servers: catalogues are discovered under an individual user's credentials, cached with a TTL, promoted into the shared inventory, and scoped on read so one user's gated catalogue is not disclosed to another. `/api/config` gains a bounded inline discovery wait, and the Tools panel synthesizes rows for servers the user has not connected yet.

## Verdict
Request changes.

## Required changes
1. Pass the merged catalogue to `_apply_task_support_metadata` in `_promote_user_tools` at `atlas/modules/mcp_tools/mcp_user_discovery.py:542` — compute `merged` first and sweep against it, as `_withdraw_promoted_tools` already does; today a second owner's promotion purges every `(server, *)` entry from `_tool_task_forbidden` and re-adds only that owner's tools, so co-owners' tools that omit or forbid `taskSupport` become task-dispatchable process-wide. Add a two-owner test asserting the first owner's forbidden entry survives.
2. Resolve the returned tool object from the requester's own catalogue in `get_tools_schema` at `atlas/modules/mcp_tools/mcp_discovery.py:660` — `_build_tool_index` keeps one object per tool name, so when two owners expose a same-named tool with different schemas, a user is handed the other owner's description and `inputSchema`, which is the disclosure this scoping exists to prevent.
3. Stop failing open when the session carries no user at `atlas/application/chat/utilities/tool_executor.py:936` — `inject_context_into_args` forwards a missing `user_email` as `None`, the documented "unscoped" value, so `user_scoped` catalogues are injected into `_mcp_data` on a model-facing path. Skip `user_scoped` entries when the user is unknown and reserve `None` for genuinely internal callers; pin it with a test.
4. Treat a manager without `_may_read_catalogue` as "cannot verify, exclude `user_scoped`" rather than "no restriction" in `build_mcp_data` at `atlas/application/chat/utilities/tool_executor.py:273`, and expose the check as a public, `ToolManagerProtocol`-declared method so the application layer stops duck-typing a private mixin method.
5. Use `_invalidate_user_client_quietly` in `oauth_callback` at `atlas/routes/mcp_auth_routes.py:619` — the raw call can raise after the token is already stored, turning a successful authorization into a 500 instead of the success redirect. Add a callback test whose fake manager raises from `_invalidate_user_client`.
6. Add a test that the chat call sites actually thread `user_email` into `get_tools_schema` (`atlas/application/chat/modes/tools.py:121`, `:265`, `atlas/application/chat/agent/agentic_loop.py:141`) — the parameter is optional at every layer, so deleting it from all three sites leaves the suite green while restoring the disclosure. Consider making it keyword-only with an explicit `UNSCOPED` sentinel for internal callers.
7. Reuse the in-flight discovery task per `(user, server-set)` in `discover_tools_for_user_servers` at `atlas/modules/mcp_tools/mcp_user_discovery.py:441` — every SPA `/api/config` poll currently spawns another shielded task that blocks on the per-user lock while a gated server hangs for the full discovery timeout.

## Evidence
Evidence: 7/10 — unusually strong unit-level coverage for a change of this size, with behaviour-named tests and a stated mutation check. The one artifact still missing is an end-to-end run against a server that actually rejects the anonymous `initialize`/`tools/list` (a local stub would do), plus a before/after screenshot of the zero-tool Tools-panel row; every backend test substitutes at the `_get_user_client` boundary, so the `401` path is asserted only against fakes.

## Notes
- Zero-tool rows tell users "No tools discovered yet." while discovery is still outstanding, and `delegated` rows offer no recovery at all — `frontend/src/components/ToolsPanel.jsx:1140`, `atlas/routes/config_routes.py:305`.
- `_promote_user_tools` logs `len(tools)` but publishes `merged`, so the operator log understates what is routable — `atlas/modules/mcp_tools/mcp_user_discovery.py:565`.
- The connectable/needs-credentials auth-type sets are restated in four hand-maintained places across two languages — `atlas/routes/config_routes.py:322`.

---

garland3:
Four of seven in [`1ddb751`](https://github.com/sandialabs/atlas-ui-3/pull/913/commits/1ddb751). All four were regressions from my own union change last round — this round found real damage and I'd rather say that plainly than dress it up.

**1. Task-support swept against the wrong set** — confirmed. `_apply_task_support_metadata` purges every `(server, *)` entry and rebuilds, so sweeping this promotion's slice while publishing the merged set made a co-owner's task-forbidden tools dispatchable process-wide. It now computes `merged` first and sweeps that, matching what `_withdraw_promoted_tools` already did. Two-owner test asserting the first owner's forbidden entry survives, as asked.

**2. `get_tools_schema` handed back the index's object** — confirmed, and this was the sharper one: the scoping decided *whether* the caller may read, then returned whichever object `_build_tool_index` happened to keep for that name. With two owners exposing a same-named tool, the reader got the other owner's description and `inputSchema` — precisely the disclosure the check exists to prevent. `_own_tool_object` now resolves from the requester's own catalogue. Pinned by `test_a_same_named_tool_resolves_per_owner`, which fails on the old code with each user seeing the other's description.

**3. `build_mcp_data` failed open on a missing user** — confirmed. `inject_context_into_args` forwards `session_context.get("user_email")`, which can be `None`, and `None` was documented as "unscoped". A session with no user is not an internal caller. It now fails **closed**: `user_scoped` entries are omitted when the user is unknown. Two tests.

**4. Duck-typed private method** — half done. `build_mcp_data` now treats a manager with no `_may_read_catalogue` as "cannot verify → exclude `user_scoped`", so the unverifiable case is safe (test included). I have **not** promoted it to a public `ToolManagerProtocol` method; that is the right shape and I'd rather land it as its own change than append an interface rename to this one.

**5. `oauth_callback` had the same 500** — confirmed, and I should have caught it last round when I fixed the identical bug in the upload route: it is the other route that stores a token. Now uses `_invalidate_user_client_quietly`.

**6. Nothing pins the threading** — you're right, and it is the most uncomfortable item here: `user_email` is optional at every layer, so deleting it from all three call sites leaves the suite green and silently restores the disclosure. I have **not** added that test or the keyword-only `UNSCOPED` sentinel. It is the first thing I would do next.

**7. In-flight task reuse in `discover_tools_for_user_servers`** — **not done.** Agreed on the diagnosis: every poll spawns another shielded task that then blocks on the per-user lock. The per-`(user, server)` lock and TTL bound the *connections*, so this is wasted tasks rather than a connection storm, which is why I am leaving it rather than reworking task lifetime late in a session.

**Verification.** Targeted: 61 + 21 + 15 tests pass. Full suite: 2900 passed, +37 over the same baseline. Line endings verified unchanged. CI on `464cd1d`: nine checks green, `build-and-test` still running as I write this.

**Where this stands.** Four rounds, and each one has found something real in the previous round's fix — including two rounds running where my own change introduced the bug the next round caught. That is a useful signal in itself: the ownership/visibility rules are being reimplemented at each read site by hand, which is what your standing note says, and it is why these keep slipping through. I think the honest next step is not another round of patches but the owning type that note asks for, with items 6 and 7 and the public protocol method folded into it. Flagging for a human to decide, since I can keep taking rounds but the defect rate suggests the design, not the diffs.

garland3:
## Summary
Adds per-user, lazy MCP tool discovery for auth-gated servers: user-scoped catalogues are cached, promoted into the shared inventory, and withdrawn on disconnect, with the model-facing schema and `_mcp_data` read paths newly scoped to the requesting user and a Tools-panel row for servers that need a connection.

## Verdict
Request changes.

## Required changes
1. Use `_invalidate_user_client_quietly(...)` in `remove_token` at `atlas/routes/mcp_auth_routes.py:329` and move `_drop_user_tool_cache(...)` out of its failure path — today a raise returns 500 for a disconnect that succeeded and leaves the revoked catalogue published and routable.
2. Make `get_tools_schema` fail closed on an absent user at `atlas/modules/mcp_tools/mcp_discovery.py:624` (explicit `UNSCOPED` sentinel for the genuinely internal callers), so a `None` `user_email` from `modes/tools.py` cannot hand a gated server's descriptions and input schemas to the model — `build_mcp_data` already refuses the same request.
3. Scope `get_available_tools()` at `atlas/modules/mcp_tools/mcp_discovery.py:535` through `_may_read_catalogue`, and thread the user from its callers — it still lists `server_tool` names from every promoted user-scoped entry to any caller.
4. Resolve the emitted tool object from the requester's own catalogue in `build_mcp_data` at `atlas/application/chat/utilities/tool_executor.py:307`, mirroring `_own_tool_object` — two owners of a same-named tool currently cross-disclose description and `inputSchema`.
5. Discard the result of an in-flight `_run_user_discovery` whose `(user, server)` was revoked while it ran (generation counter bumped by `clear_user_tool_cache`, `atlas/modules/mcp_tools/mcp_user_discovery.py:135`) — a slow discovery started before disconnect re-publishes the withdrawn catalogue.
6. Add tests that pin `user_email` actually reaching the scoping APIs from `modes/tools.py:121,:265`, `agent/agentic_loop.py:141` and `safe_get_tools_schema`; deleting the argument at all call sites currently leaves the suite green.
7. Confirm the trust model for `delegated` servers at `atlas/modules/mcp_tools/mcp_user_discovery.py:509` — `_get_user_client` mints a token from the caller's own session, so a promoted tool is callable by a non-owner, which the module docstring's containment claim does not cover. Exclude `delegated` from promotion or correct the claim.

## Evidence
Evidence: 7/10 — unusually thorough at the unit level for a change this size (per-commit suite, lint and mutation-check results), but nothing exercises the real failure this PR exists to fix. The most important missing artifact is one run against a stub MCP server that returns `401` on `initialize` until a token is presented, showing an empty startup catalogue and a populated one after authorization; UI screenshots of the auth-required row before and after connecting are the second gap.

## Notes
- `_may_read_catalogue` returns "allow" for a blank `server_name` — `atlas/modules/mcp_tools/mcp_discovery.py:568`.
- Catalogue scoping is re-derived per tool, making `build_mcp_data` O(n²) on the chat hot path — `atlas/application/chat/utilities/tool_executor.py:298`.
- `delegated` zero-tool rows render "No tools discovered yet." with no connect control, a dead end for the user — `frontend/src/components/ToolsPanel.jsx:1157`.

---

garland3:
Thanks — worked through the list. Addressed in `d8e2da4`:

1. **`remove_token`** — the token is already deleted by that point, so `_drop_user_tool_cache` now runs first and unconditionally and the client invalidation goes through `_invalidate_user_client_quietly`. A cleanup failure no longer returns 500 for a disconnect that happened, nor leaves the revoked catalogue routable. Test: `test_a_failing_client_invalidation_still_disconnects`.
2. **`get_tools_schema` fails closed** — `None`/omitted is an unknown user and withholds every `user_scoped` catalogue, matching `build_mcp_data`. The three internal callers that resolve an already-authorized tool pass the new `mcp_discovery.UNSCOPED` sentinel explicitly. The existing test that pinned the permissive default was rewritten to pin both halves.
4. **`build_mcp_data` cross-disclosure** — now resolves through `_own_tool_object`, as `get_tools_schema` already did. Test: `TestBuildMcpDataEmitsTheRequestersOwnToolObject`.
5. **In-flight discovery vs. revocation** — generation counter bumped by `clear_user_tool_cache` for every key it covers (including keys with no cache entry, which is the in-flight case); a stale run drops its result rather than republishing. Test: `test_a_revoked_run_neither_caches_nor_promotes`.
6. **Threading tests** — `test_the_requesting_user_reaches_the_schema_scoping_api` (+ non-streaming) pin the argument at the `modes/tools.py` call sites; deleting it now fails.
7. **`delegated` trust model** — the docstring's blanket "raises `AuthenticationRequiredException`" did not hold there, so it is corrected rather than the behaviour changed: `_get_user_client` mints a credential from the caller's *own* session, so promotion discloses the tool's existence and schema to other users of the deployment, not the owner's access, which is what `delegated` means.

Not changed, deliberately:

3. **`get_available_tools()`** — it has no non-test callers in `atlas/`, and threading a user through the `ToolManagerProtocol` signature for a name-only listing with no live consumer is churn without a reachable leak. Happy to do it if you'd rather close it off pre-emptively.
- The **O(n²)** note and the **`delegated` zero-tool row** UX dead end are real but are follow-ups rather than fixes to this change; say the word and I'll open issues.
- `_may_read_catalogue` returning allow for a blank `server_name`: a blank name matches no entry, so there is no catalogue to disclose.

Verified: `bash test/atlas_tests.sh` — 2910 passed, matching the baseline pass set plus the 10 new tests (this checkout has 164 pre-existing local env failures in `test_telemetry_routes.py`, identical with and without this diff; CI is the authority and was green on the previous commit). `ruff check` clean on every touched file.

On the evidence gap: agreed that an end-to-end run against a stub server that 401s `initialize` until a token is presented is the artifact this PR most wants. I have not built one.

garland3:
## Summary
This PR adds per-user MCP tool discovery so that servers which gate `initialize`/`tools/list` behind user credentials can publish a per-user catalogue, and fixes the bootstrap deadlock where an unauthenticated server never rendered a connect affordance in the Tools panel.

## Verdict
Request changes.

## Required changes
1. Refresh or serve the stale entry instead of silently dropping the tool at `atlas/modules/mcp_tools/mcp_discovery.py:698` — after the 300s TTL lapses, a user-selected tool on a gated server passes ACL but is `continue`d out of the schema with no `missing` entry and no log, so the turn runs as if the tool were never selected.
2. Enforce entry expiry independently of the failure cool-down at `atlas/modules/mcp_tools/mcp_user_discovery.py:400` — `_cached_user_tools` short-circuits before `_expire_stale_user_entry` ever runs, leaving an expired catalogue published and routable for the cool-down window.
3. Make catalogue withdrawal fail closed on disconnect at `atlas/routes/mcp_auth_routes.py:63` — `_drop_user_tool_cache` swallows every exception at debug level, so a failed `clear_user_tool_cache` returns 200 "disconnected" while the tool names and schemas obtained under the revoked token stay published.
4. Bump the discovery generation for every key matching the clear predicate, including lock-map-only keys, at `atlas/modules/mcp_tools/mcp_user_discovery.py:135` — a discovery started before `clear_user_tool_cache` is never marked stale and re-promotes the catalogue after the user disconnects.
5. Collapse repeated polls onto one in-flight task at `atlas/modules/mcp_tools/mcp_user_discovery.py:485` — each `/api/config` request spawns a shielded, uncancelled task per user/server, so a hanging MCP server lets tasks and their retained clients accumulate without bound.
6. Replace the `UNSCOPED = "__atlas_unscoped__"` string sentinel at `atlas/modules/mcp_tools/mcp_discovery.py:34` with a non-string sentinel compared by identity — it currently shares a value space with header-derived `user_email`, so any path influencing that string disables every scoping check.
7. Clear the server's task-support entries when the last promotion is withdrawn (`mcp_user_discovery.py`, `_withdraw_promoted_tools`) — emptying a server today leaves stale task-support metadata behind.

## Evidence
Evidence: 7/10. Unit coverage is unusually thorough for a change this size, but nothing exercises a server that actually gates `initialize`/`tools/list` (all backend tests substitute at the client boundary), and there are no screenshots of the user-facing panel states this PR introduces.

## Notes
- Expose one public, protocol-declared visibility API instead of `getattr`-ing `_may_read_catalogue`/`_own_tool_object` from the application layer — `atlas/application/chat/utilities/tool_executor.py:280`.
- `get_available_tools()` still enumerates every promoted user-scoped tool name to any caller — `atlas/modules/mcp_tools/mcp_discovery.py:535`.
- "No tools discovered yet." is shown while discovery is still running, and a `delegated` zero-tool row is a dead end — `tools-panel`.

---

garland3:
Second round addressed in `f88ed33`.

**Fixed**

1. **Silent drop at TTL lapse** — a tool withheld because its per-user catalogue is not the requester's (expired, or another user's) is now counted and logged at info, so a turn that runs without a selected tool is diagnosable. The *caller* still is not told which tool, since that would itself disclose it. A synchronous refresh is not available from `get_tools_schema`; the lazy re-discovery on the next `/api/config` is what repopulates it.
2. **Expiry deferred by the cool-down** — `_cached_user_tools` now expires a past-TTL entry before the cool-down can answer for it. Test: `test_a_stale_catalogue_is_withdrawn_even_while_cooling_down`.
3. **Disconnect fails closed** — `_drop_user_tool_cache(..., required=True)` on the disconnect path: a failed withdrawal is reported rather than hidden behind a 200. The cached-client drop stays quiet — it discloses nothing, and failing it must not report a disconnect that happened as an error (that was the previous round's finding 1). Tests for both halves.
6. **String sentinel** — good catch, that was mine from the last round. `UNSCOPED` is now a sentinel object compared by identity. Test asserts that `"UNSCOPED"` and the old literal `"__atlas_unscoped__"` as `user_email` both get scoped normally.
7. **Task-support metadata on full withdrawal** — cleared when the last promotion goes.

**Already done in `d8e2da4`** (item 4): `clear_user_tool_cache` bumps the generation for every key matching the predicate taken from the **lock map**, which is exactly the lock-map-only, in-flight case — `mcp_user_discovery.py:180`. Covered by `test_a_revoked_run_neither_caches_nor_promotes`, which holds a discovery open across the clear and asserts nothing is cached or promoted.

**Not done**

5. **Collapsing polls onto one in-flight task** — `discover_tools_for_user` already serialises per `(user, server)` behind a lock, so repeated polls share one `tools/list` rather than opening a connection each. The residue is that `discover_tools_for_user_servers` awaits shielded tasks that outlive a cancelled request, bounded by the discovery timeout. That is real but is a change to the concurrency structure rather than a fix to what this PR broke, and I would rather not reshape it unverified this late in the review. Happy to do it here if you would rather it not ship as is.

Also noted, not changed: the `getattr`-based visibility API (a protocol-declared method is the right shape, but it widens the surface this PR touches), `get_available_tools()` (still no non-test callers), and the two panel-copy issues.

Verified: `bash test/atlas_tests.sh` — **2914 passed**, matching the baseline pass set plus the 14 tests added across this and the previous round; this checkout has 164 pre-existing local env failures in `test_telemetry_routes.py`, identical with the diff stashed. `ruff check` clean on every file this PR touches (the 45 repo-wide ruff errors are also identical with the diff stashed).

The evidence gap stands and I have not closed it: no test drives a server that actually 401s `initialize`/`tools/list` — they all substitute at the client boundary — and there are no panel screenshots.

garland3:
## Summary
Adds per-user, lazy MCP tool discovery so servers that gate `initialize`/`tools/list` behind a credential can populate their catalogue after the user connects, and scopes catalogue reads (`get_tools_schema`, `build_mcp_data`, `/api/config`) to the requesting user. Also adds withdrawal of a promoted catalogue on disconnect and new Tools-panel rows for zero-tool auth-required servers.

## Verdict
Request changes.

## Required changes
1. Make a failed catalogue withdrawal recoverable in `remove_token` at `atlas/routes/mcp_auth_routes.py:355` — the token is deleted first, so a raise from `_drop_user_tool_cache(..., required=True)` returns 500, skips OAuth provider revocation entirely, and every retry short-circuits on the 404 branch, leaving the catalogue published and the provider token live with no API path to clear either. Withdraw before deleting the token (a withdrawal for a still-valid token is republished on the next discovery), and let the 404 branch still attempt the withdrawal.
2. Do not fail the disconnect flow for a withdrawal error at `atlas/routes/mcp_auth_routes.py:355` and `frontend/src/components/ToolsPanel.jsx:541` — the user sees "Internal server error while removing token", the dialog stays open, the row still reads as connected, and the second attempt errors differently. Return 200 with an explicit degraded field and have the panel close, refresh the catalogue, and show a warning.
3. Return without raising when the manager has no `clear_user_tool_cache` at `atlas/routes/mcp_auth_routes.py:74` — the `getattr` probe implies duck-typed managers are supported, but for those every disconnect is a permanent 500 after the token is already gone.
4. Guard the early "no token" failure exit with the generation check at `atlas/modules/mcp_tools/mcp_user_discovery.py:299` — a discovery started before `clear_user_tool_cache` can otherwise record a failure entry for a credential that was already revoked.
5. Collapse duplicate in-flight discoveries in `discover_tools_for_user_servers` at `atlas/modules/mcp_tools/mcp_user_discovery.py:495` — each `/api/config` poll creates a new shielded task that is never cancelled, so a server that hangs accumulates tasks and retained client objects for the whole discovery timeout, per tab.
6. Add a test asserting the agent loop forwards the session user to `get_tools_schema` (`atlas/application/chat/agent/agentic_loop.py:141`) — the parameter defaults to `None`, every existing loop test patches `safe_get_tools_schema` out, so deleting the argument reopens the cross-user schema disclosure with the suite green.
7. Scope or explicitly document `get_available_tools()` at `atlas/modules/mcp_tools/mcp_discovery.py:554` — it is a declared `ToolManagerProtocol` method that enumerates every promoted `user_scoped` tool name with no user parameter and no `_may_read_catalogue` check, so any future caller inherits the disclosure silently.

## Evidence
Evidence: 7/10. Unit and route coverage is strong and the mutation check on the ownership changes is a genuinely useful artifact, but every backend test substitutes at `_get_user_client`, so the 401-on-`tools/list` condition the whole change exists to handle is never exercised across the protocol boundary; a local stub server returning 401 until a bearer token is presented, with a transcript of the empty-startup-then-populated sequence, is the single most valuable missing artifact.

## Notes
- `_may_read_catalogue` returns `True` for a falsy server name before the `user_scoped` check — `atlas/modules/mcp_tools/mcp_discovery.py:588`.
- Ownership rules are re-derived at eight read sites across two mutually-dependent mixins; a `ServerCatalogue` value type would put the invariant in one place — `atlas/modules/mcp_tools/mcp_user_discovery.py`.
- A `delegated` zero-tool row shows "No tools discovered yet." with no control and no explanation that it resolves on its own — `frontend/src/components/ToolsPanel.jsx:1157`.

---

PR #914, opened 2026-09-10
fix(mcp): deliver tool-returned images to the LLM, not just the canvas

### LINKED ISSUE #909: MCP tool-returned images go to the canvas but never to the LLM: agent gets {"results": {}} for screenshots
## Summary

Images returned by MCP tools (e.g. a Fusion 360 viewport screenshot from `fusion360_fusion_mcp_read`) are routed to the canvas as a frontend artifact and are **never** shown to the LLM. The model receives only the textual portion of the tool result — which, for a tool whose entire payload *is* the image, normalizes to `{"results": {}}`.

This is especially damaging in agent mode: the agent asks for a screenshot to check its own work, gets an empty dict back, correctly reports "I can't directly inspect the returned image here", and then continues editing the model blind. In the observed session it re-ran a build script it had already run, ending up with two overlapping rocket bodies in the same part file, because it had no way to see the current state.

Related to #907 — both share the underlying gap that nothing ever populates `image_b64` for files that were not just uploaded by the client — but the root cause and the fix here are distinct.

## Evidence

From a single agent turn (agent mode, `gpt-5.4`, which has `supports_vision: true`):

```
01:33:46.104  Streaming LLM call with tools: 6 messages, 135138 chars    <- before the screenshot tool
01:33:48.118  Successfully called fusion_mcp_read on fusion360
01:33:48.134  Sent 1 artifact(s) from tool fusion360_fusion_mcp_read to frontend
01:33:48.153  File uploaded successfully: category=generated, size=245975 bytes, content_type=image/png
              (.../generated/<ts>_<hash>_mcp_image_0.png)
01:33:48.154  Streaming LLM call with tools: 8 messages, 135250 chars    <- after
```

The screenshot arrived intact: a real 246 KB PNG, correctly typed and stored. But the prompt grew by exactly **112 characters** across the tool call. A 246 KB PNG is roughly 328,000 base64 characters. Those 112 chars are the `{"results": {}}` tool message plus its role / `tool_call_id` overhead — the image contributed nothing.

The tool result the model saw, verbatim from the chat export:

```json
{ "results": {} }
```

and its next assistant message:

> I can't directly inspect the returned image here, so I'm going to query the model state and then refine the rocket...

## Root cause

Three places, none of which has a path to the LLM:

1. **`atlas/modules/mcp_tools/mcp_result_processor.py` (~L211-270)** — `_extract_v2_components` pulls MCP `ImageContent` blocks out of the tool result and converts each into a *frontend artifact* (`mcp_image_<n>.png`, `viewer: "image"`, `open_canvas: true`). Separately, `_normalize_mcp_tool_result` (~L40) builds the LLM-facing text, and the module docstring states the intent explicitly: *"No base64 payloads are inlined into the textual result to avoid prompt bloat."* The Fusion tool returns no structured content besides the image, so `results` normalizes to `{}`.

2. **`atlas/application/chat/utilities/file_processor.py` (`ingest_v2_artifacts`, ~L724-793)** — the artifact is uploaded and added to session context with only `key` / `content_type` / `size` / `source`. It never receives `image_b64` / `image_mime_type`, which are exactly the keys `message_builder.build_messages` (~L190-199) looks for when assembling vision content blocks. `ingest_tool_files` (~L548) has the same shape and the same omission.

3. **Nothing re-hydrates it on a later turn.** `handle_session_files` only sets `image_b64` from freshly uploaded client bytes (see #907), so the screenshot does not become visible on the following turn either — at best it appears as a name-only manifest line `- mcp_image_0.png`.

Confirming there is no path anywhere in the loop: grepping `image_b64|image_url|ImageContent` across `atlas/application/chat/agent/` and `atlas/application/chat/modes/` returns no matches.

## Suggested fix

Note a provider constraint that rules out the obvious approach: the **OpenAI chat-completions API does not accept image content inside a `role: "tool"` message**, so the tool result itself cannot simply be widened to carry the image.

The portable pattern is to inject the image as a synthetic **user** message immediately following the tool result — e.g. text `Image returned by <tool_name>:` plus an `image_url` data-URI block, reusing the existing `_build_multimodal_user_message` helper in `message_builder.py`.

Guards this needs:

- Gate on `model_supports_vision`; when false, keep current behavior and ideally add a short textual note to the tool result so the model knows an image exists but is not visible to it.
- Cap how many tool-returned images are retained in context. At ~330 KB of base64 each, keeping every screenshot across a 10-step agent loop would blow the context window and the bill. A "most recent N only" policy (dropping older tool images back to manifest entries on later iterations) matches how the existing `_MAX_VISION_IMAGES_PER_REQUEST` cap behaves.
- Reuse the existing size / MIME validation already present in `_extract_v2_components` (there is already a MIME allowlist there).

Anthropic's Messages API *does* permit images inside `tool_result` blocks, so a provider-aware implementation could place the image in the tool message directly on that path and fall back to the synthetic-user-message form elsewhere.

## Impact

Any MCP tool whose value is visual is currently a no-op from the model's perspective: CAD screenshots, chart/plot renderers, browser-automation screenshots, diagram generators. The user sees the image on the canvas and reasonably assumes the model did too, which makes the subsequent conversation confusing — the model behaves as though it is guessing, because it is.

garland3:
## Summary

MCP tool-returned images reached the frontend canvas but never the LLM. A tool whose payload is an image (Fusion 360 viewport screenshot, chart render, browser-automation capture) normalized to `{"results": {}}` for the model: the images lived only in `ToolResult.artifacts`, which drive canvas display, while the `role: "tool"` message carried just the text. The agent then reported it "can't directly inspect the returned image" and kept working blind -- in the session that surfaced this, it re-ran a build script it had already run because it had no way to see the current state.

Closes #909.

## Why a synthetic user message

The OpenAI chat-completions API rejects image content inside a `role: "tool"` message, so the tool result itself cannot be widened. The portable pattern -- and what OpenAI's own guidance recommends for tool-generated images -- is to append a synthetic **user** message right after the step's tool results carrying the images as `image_url` data-URI blocks. LiteLLM translates that block shape to each provider's native format (Anthropic image block, Gemini `inline_data`, Bedrock image), and it is the exact shape `message_builder._build_multimodal_user_message` already emits for user uploads, so the provider paths are exercised in production today.

## What changed

- New `atlas/application/chat/utilities/tool_image_context.py`: `ToolImageInjector` plus pure helpers (`extract_llm_ready_images`, `build_tool_image_message`, `model_supports_vision`).
- Wired into all three tool-execution paths that append tool results for the LLM:
  - the agentic loop (`agentic_loop.py`, after each step's results),
  - tools mode's streaming continuation loop (`modes/tools.py`, before the next round / synthesis),
  - the tools-mode synthesis workflow (`tool_executor.execute_tools_workflow`, new optional `image_injector` param).

## Guards

- **Vision gating**: injection only when the model has `supports_vision: true` in `llmconfig.yml` (same config the orchestrator uses for user uploads). Non-vision models instead get a short note appended to the tool result telling the model an image exists but cannot be shown -- the observed failure was the model guessing why it could not see the screenshot.
- **Rolling recency cap**: at most 6 tool images live in the transcript; when a new one pushes past the cap the oldest injected message is demoted in place to a one-line note (ordering untouched, `role` sequence preserved).
- **Size budgets**: 5 MB base64 per image, 12 MB aggregate across the turn -- headroom under Bedrock's ~20 MB whole-request ceiling.
- **Validation at injection time**: MIME allowlist (PNG/JPEG/GIF/WebP/BMP; SVG excluded even though the canvas accepts it), base64 decode validation, per-image size cap. Structured `artifacts` entries reach this module without `mcp_result_processor`'s screening, so re-validation matters.
- **Turn lifetime**: images live in the turn's working transcript only. Conversation history is text-only and `handle_session_files` clears stale vision payloads each turn, so nothing is persisted to the conversation store and a later turn sees the tool digest + session file listing, not the pixels.
- **Fail-open**: any injection failure logs and degrades to the previous text-only behavior; the tool-calling turn is never broken.

## Testing

- New `atlas/tests/test_tool_image_context.py` (29 tests): extraction/validation (MIME allowlist, SVG exclusion, extension inference, invalid/oversized base64, newline-chunked base64, garbage input), message shape, vision gating + unseen-image note, rolling cap demotion, size-budget eviction, and end-to-end wiring through `AgenticLoop` and `execute_tools_workflow`.
- `./test/run_tests.sh backend`: 3053 passed, 17 skipped.
- `ruff check atlas/`: no new findings (pre-existing 46 on main unchanged; none in touched files).
- `bash scripts/check-docs.sh`: OK (no orphans/broken links).

Docs updated in `docs/developer/mcp-tool-outputs.md` (new "How the model sees tool-returned images" section); CHANGELOG entry added.

No frontend changes: the canvas path already worked, and injected images exist only in the server-side working transcript.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### PR #910 - 2026-09-10
+- **MCP tool-returned images now reach the LLM, not just the canvas** (closes #909): a tool's ImageContent blocks were routed to the canvas while the model saw only the normalized text result -- `{"results": {}}` for a screenshot-only tool -- so agent-mode work on visual outputs (CAD screenshots, charts, browser captures) ran blind. Images are now attached to the live transcript as a synthetic user message right after the step's tool results (the OpenAI-style API rejects image content inside `role: "tool"` messages, so this is the portable form across providers via LiteLLM). Gated on the model's `supports_vision` setting (unsupported models get a note explaining the image exists but cannot be shown); capped at 6 live images with 5 MB/12 MB per-image/aggregate base64 budgets that demote the oldest images to notes; re-validated for MIME allowlist and base64 integrity at injection time. Applies to agent mode and tools mode; injection is fail-open. See `docs/developer/mcp-tool-outputs.md`.
+
 ### PR #904 - 2026-09-08
 - Hold `fastmcp` below 4.0 and record why: FastMCP 4 removes server-initiated sampling, which the sampling_demo and tool_planner MCP servers require (tracked in #905).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
-### PR #910 - 2026-09-10
-- **MCP tool-returned images now reach the LLM, not just the canvas** (closes #909): a tool's ImageContent blocks were routed to the canvas while the model saw only the normalized text result -- `{"results": {}}` for a screenshot-only tool -- so agent-mode work on visual outputs (CAD screenshots, charts, browser captures) ran blind. Images are now attached to the live transcript as a synthetic user message right after the step's tool results (the OpenAI-style API rejects image content inside `role: "tool"` messages, so this is the portable form across providers via LiteLLM). Gated on the model's `supports_vision` setting (unsupported models get a note explaining the image exists but cannot be shown); capped at 6 live images with 5 MB/12 MB per-image/aggregate base64 budgets that demote the oldest images to notes; re-validated for MIME allowlist and base64 integrity at injection time. Applies to agent mode and tools mode; injection is fail-open. See `docs/developer/mcp-tool-outputs.md`.
+### PR #914 - 2026-09-10
+- **MCP tool-returned images now reach the LLM, not just the canvas** (closes #909): screenshot-only tools no longer normalize to `{"results": {}}` for the model; images are attached after the step's tool results as a synthetic user message, gated on `supports_vision`, with rolling 6-image and per-image/aggregate base64 budgets and MIME/base64 re-validation. Details and limits in `docs/developer/mcp-tool-outputs.md`.
 
 ### PR #904 - 2026-09-08
 - Hold `fastmcp` below 4.0 and record why: FastMCP 4 removes server-initiated sampling, which the sampling_demo and tool_planner MCP servers require (tracked in #905).
--- a/atlas/application/chat/agent/agentic_loop.py
+++ b/atlas/application/chat/agent/agentic_loop.py
@@ -32,6 +32,7 @@ from ..utilities import error_handler, tool_executor
 from ..utilities.dropped_calls import publish_dropped_call_warning
 from ..utilities.search_tool_selection import with_search_tool
 from ..utilities.tool_history import ToolCallRecorder
+from ..utilities.tool_image_context import ToolImageInjector, model_supports_vision
 from .protocols import AgentContext, AgentEvent, AgentEventHandler, AgentLoopProtocol, AgentResult
 from .steering import SteeringChannel
 from .streaming_final_answer import stream_final_answer
@@ -228,6 +229,14 @@ class AgenticLoop(AgentLoopProtocol):
         # would let one turn inherit another's spent sleep budget.
         turn_sleep_budget: Dict[str, Any] = {}
 
+        # Tool-returned images reach the LLM as a synthetic user message
+        # appended after each step's tool results (issue #909). Gated on the
+        # model's configured vision support; one injector per turn tracks the
+        # rolling most-recent-N cap across the loop's steps.
+        image_injector = ToolImageInjector(
+            enabled=model_supports_vision(self.config_manager, model),
+        )
+
         while steps < max_steps:
             steps += 1
 
@@ -354,6 +363,17 @@ class AgenticLoop(AgentLoopProtocol):
                     "content": result.content,
                     "tool_call_id": result.tool_call_id,
                 })
+
+            # Issue #909: the artifacts on these results (e.g. a screenshot)
+            # went to the canvas but never to the model. Attach the images to
+            # the transcript now so the next LLM call can actually see them.
+            # _to_tool_call_dict normalizes both attribute objects and plain
+            # dicts, so the id -> name map works for either shape.
+            tool_names = {
+                normalized["id"]: normalized["function"]["name"]
+                for normalized in (_to_tool_call_dict(tc) for tc in tool_calls)
+            }
+            image_injector.after_tool_results(messages, results, tool_names=tool_names)
 
             await event_handler(AgentEvent(
                 type="agent_tool_results", payload={"results": results},
--- a/atlas/application/chat/modes/tools.py
+++ b/atlas/application/chat/modes/tools.py
@@ -29,6 +29,7 @@ from ..utilities.citation_publishing import attach_citations, publish_citations
 from ..utilities.dropped_calls import publish_dropped_call_warning
 from ..utilities.search_tool_selection import with_search_tool
 from ..utilities.tool_history import ToolCallRecorder
+from ..utilities.tool_image_context import ToolImageInjector, model_supports_vision
 from .streaming_helpers import stream_and_accumulate
 
 logger = logging.getLogger(__name__)
@@ -192,6 +193,11 @@ class ToolsModeRunner:
                 config_manager=self.config_manager,
                 skip_approval=self.skip_approval,
                 user_email=user_email,
+                # Issue #909: tool-returned images reach the synthesis call
+                # as a synthetic user message when the model supports vision.
+                image_injector=ToolImageInjector(
+                    enabled=model_supports_vision(self.config_manager, model),
+                ),
             )
         except BaseException:
             # A Stop / disconnect during tool execution would otherwise discard
@@ -382,6 +388,11 @@ class ToolsModeRunner:
         current_response = final_llm_response
         executed_signatures: set = set()
         extra_round = 0
+        # Issue #909: one injector per turn tracks the rolling most-recent-N
+        # cap across continuation rounds.
+        image_injector = ToolImageInjector(
+            enabled=model_supports_vision(self.config_manager, model),
+        )
 
         try:
             while True:
@@ -441,6 +452,16 @@ class ToolsModeRunner:
                         "content": content,
                         "tool_call_id": tc_id,
                     })
+
+                # Issue #909: attach tool-returned images to the transcript so
+                # the next continuation round (or synthesis) can see them.
+                image_injector.after_tool_results(
+                    messages, results,
+                    tool_names={
+                        self._tool_call_id(tc): self._tool_call_signature(tc)[0]
+                        for tc in fresh
+                    },
+                )
 
                 if self.artifact_processor:
                     await self.artifact_processor(session, results, effective_callback)
--- a/atlas/application/chat/modes/tools.py
+++ b/atlas/application/chat/modes/tools.py
@@ -567,10 +567,19 @@ class ToolsModeRunner:
             except Exception:
                 pass  # Best-effort UI notification; synthesis proceeds regardless
 
-        # Build synthesis messages
+        # Build synthesis messages. Only plain-string user messages count as
+        # the question: multimodal user turns (inline image/PDF blocks from
+        # build_messages, or the synthetic tool-image message from issue
+        # #909) carry a list of content blocks, and the prompt provider's
+        # ``user_question.strip()`` would raise on those, silently dropping
+        # the configured synthesis prompt.
         user_question = ""
         for m in reversed(messages):
-            if m.get("role") == "user" and m.get("content"):
+            if (
+                m.get("role") == "user"
+                and isinstance(m.get("content"), str)
+                and m.get("content")
+            ):
                 user_question = m["content"]
                 break
 
--- a/atlas/application/chat/utilities/tool_executor.py
+++ b/atlas/application/chat/utilities/tool_executor.py
@@ -28,6 +28,7 @@ from atlas.modules.mcp_tools.token_storage import AuthenticationRequiredExceptio
 
 from ..approval_manager import get_approval_manager
 from .event_notifier import _sanitize_filename_value  # reuse same filename sanitizer for UI args
+from .tool_image_context import ToolImageInjector
 
 logger = logging.getLogger(__name__)
 
@@ -128,11 +129,17 @@ async def execute_tools_workflow(
     config_manager=None,
     skip_approval: bool = False,
     user_email: Optional[str] = None,
+    image_injector: Optional["ToolImageInjector"] = None,
 ) -> tuple[str, List[ToolResult]]:
     """
     Execute the complete tools workflow: calls -> results -> synthesis.
 
     Pure function that coordinates tool execution without maintaining state.
+
+    ``image_injector`` (issue #909) receives the step's tool results right
+    after their messages are appended, so tool-returned images reach the
+    synthesis call as a synthetic user message when the model supports
+    vision.
     """
     logger.debug("Entering execute_tools_workflow")
     # Add assistant message with tool calls
@@ -159,6 +166,18 @@ async def execute_tools_workflow(
             "content": result.content,
             "tool_call_id": result.tool_call_id
         })
+
+    if image_injector is not None:
+        tool_names: Dict[Any, str] = {}
+        for tc in (llm_response.tool_calls or []):
+            try:
+                tool_names[tc.id] = tc.function.name
+            except AttributeError:
+                fn = tc.get("function") if isinstance(tc, dict) else None
+                tool_names[tc.get("id") if isinstance(tc, dict) else None] = (
+                    fn.get("name", "") if isinstance(fn, dict) else ""
+                )
+        image_injector.after_tool_results(messages, tool_results, tool_names=tool_names)
 
     # Determine if synthesis is needed
     final_response = await handle_synthesis_decision(
--- a/atlas/application/chat/utilities/tool_executor.py
+++ b/atlas/application/chat/utilities/tool_executor.py
@@ -1054,10 +1054,19 @@ async def synthesize_tool_results(
 
     Pure function that coordinates LLM call for synthesis.
     """
-    # Extract latest user question (walk backwards)
+    # Extract latest user question (walk backwards). Only plain-string user
+    # messages count: multimodal user turns (inline image/PDF blocks from
+    # build_messages, or the synthetic tool-image message from issue #909)
+    # carry a list of content blocks, and the prompt provider's
+    # ``user_question.strip()`` would raise on those, silently dropping the
+    # configured synthesis prompt.
     user_question = ""
     for m in reversed(messages):
-        if m.get("role") == "user" and m.get("content"):
+        if (
+            m.get("role") == "user"
+            and isinstance(m.get("content"), str)
+            and m.get("content")
+        ):
             user_question = m["content"]
             break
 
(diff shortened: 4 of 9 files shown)


garland3:
## Summary
Adds a `tool_image_context` utility that extracts image artifacts from MCP tool results and injects them as a synthetic multimodal user message so vision models actually see tool-returned images, with a non-vision fallback note, per-image/aggregate size budgets, and a rolling six-image cap. Wired into three call paths (agentic loop, tools mode, tool executor) with unit, integration, and an end-to-end harness.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size: 25 unit tests, integration assertions that the injected message reaches the next LLM call, and an in-process harness driving a real FastMCP server with a byte-for-byte base64 check. The one missing artifact is a before/after transcript against a live vision provider confirming the claimed block-shape translation.

## Notes
- Vision-enabled path silently drops unusable images (SVG/TIFF, invalid or oversized base64, budget-dropped) with no explanatory note, reproducing the original blind-model symptom — `atlas/application/chat/utilities/tool_image_context.py:283`.
- Aggregate byte budget is evaluated before `_enforce_count_cap()` demotes older images, so a fresh screenshot can be rejected in favor of six stale ones — `atlas/application/chat/utilities/tool_image_context.py:299`.
- Tool-call id→name mapping is re-implemented three ways; the `tool_executor` variant can register a `None -> ""` entry, and injected intro text has a double space — `atlas/application/chat/utilities/tool_executor.py:170`, `tool_image_context.py:196`.

---

garland3:
Addressed the review feedback in 0466fed:

- **JSON-corrupting note (Copilot, tool_image_context.py:371)**: the vision-off note is now a separate `role: "system"` message appended after the tool results -- the tool result content is never modified. This matches the position where the tools-mode synthesis paths already append their files-manifest system messages. Covered by `test_vision_disabled_appends_separate_note_message`.
- **Synthesis prompt silently dropped (Codex P2)**: confirmed the root cause -- `prompt_provider.get_tool_synthesis_prompt` calls `user_question.strip()`, which raised on list content and was swallowed by its safeguard, killing the synthesis prompt. Fixed both lookup sites (`synthesize_tool_results`, `_stream_synthesis`) to require plain-string user content; this also protects the pre-existing case of a user-upload multimodal message. Covered by `test_list_content_user_messages_skipped_for_question`.
- **Wholesale demotion of an over-cap batch (Codex P2)**: a single result carrying more than 6 images now keeps its newest 6 up front, and the rolling cap trims the oldest message's image blocks individually instead of dropping the entire message (4-then-4 keeps the 6 newest as [2, 4]). Covered by `test_batch_over_cap_keeps_newest_images` and `test_batch_boundary_demotes_individually_not_wholesale`.
- **Size budget rejecting before evicting (Codex P2)**: the aggregate budget now demotes the oldest live images individually (newest wins) until the new image fits, instead of rejecting it. Covered by `test_budget_evicts_old_images_instead_of_rejecting_new`.
- **CHANGELOG length (Copilot)**: condensed to 2 lines per AGENTS.md, and the PR number corrected to #914.
- **End-to-end validation script (Codex P1)**: added in 5412148 before this review pass finished -- `test/pr-validation/test_pr914_tool_images_reach_llm.sh` runs a real in-process FastMCP server whose tool returns a real PNG ImageContent and pushes it through the real `MCPToolManager` normalization/extraction and injection hook (`bash test/run_pr_validation.sh 914`).

One comment not actioned: removing provider names from the module docstring (`Anthropic image block, Gemini inline_data, Bedrock image`). The OPSEC convention in AGENTS.md governs attribution of AI-generated work (no model/provider sign-offs or branding of the producing tool); naming provider API formats in technical reference comments is established practice in this repo -- e.g. `file_processor.py` ("Claude on Bedrock caps PDFs at 100 pages"), `message_builder.py` ("follows Anthropic's guidance"), `docs/developer/canvas-renderers.md`. These references describe wire formats, not provenance. Happy to neutralize if maintainers prefer a stricter reading.

garland3:
## Summary
Adds a `ToolImageInjector` that lifts image artifacts out of `role: "tool"` results and re-injects them as a synthetic multimodal user message (with a `role: "system"` explanatory note for non-vision models), wired into the agentic loop, `execute_tools_workflow`, and the tools-mode continuation loop, with per-image/per-turn/aggregate base64 caps and rolling demotion.

## Verdict
Request changes.

## Required changes
1. Make an explicit MIME authoritative in `_infer_image_mime` at `atlas/application/chat/utilities/tool_image_context.py:196` — reject when `mime`/`mime_type` is present but outside the allowlist instead of falling through to the filename extension, and consult `_EXT_TO_IMAGE_MIME` only when no MIME field exists; today `{"name": "shot.png", "mime": "image/svg+xml"}` ships as `image/png`, defeating the documented SVG exclusion and getting the whole turn rejected by the provider on a magic-byte mismatch.
2. Stop interpolating server-derived strings into the injected messages at `tool_image_context.py:192` and `:429` — restrict the tool label to names from the client's own `tool_call_id -> name` map, drop the `_tool_name_from_artifact` description fallback in favor of the constant `"a tool"`, and truncate/character-class any name that is shown; MCP-server-controlled prose currently lands in `role: "user"` and `role: "system"` blocks.
3. Guarantee forward progress in `_demote_oldest_images` at `tool_image_context.py:311` — when no `image_url` block is found for the oldest entry, pop it from `self._live` before breaking, and bound the callers at `_inject:299` and `_enforce_count_cap:305` by the initial `len(self._live)`; otherwise both `while` loops spin forever once anything else trims content blocks from the transcript.
4. Store the injected image count as structured state in the `self._live` entry at `tool_image_context.py:441` and delete `_count_injected_images` — the count is currently recovered by string-splitting the intro prose on `" returned "`, so a copy edit or a tool name containing that substring yields "0 image(s) ... were removed from context".
5. Emit an explanatory note when a vision-capable model's images are all dropped (`tool_image_context.py:283`) — MIME/base64/size/budget rejections currently leave the model with `{"results": {}}` and no explanation, reproducing the exact blind-model symptom this PR fixes.
6. Add a regression test for the tools-mode continuation path at `atlas/application/chat/modes/tools.py:456` (in `test_tools_mode_iteration.py`): round 1 returns an image artifact, round 2 repeats a call; assert exactly one synthetic user message with an `image_url` block lands after the round-1 tool messages and that its intro names the real tool.
7. Build the `tool_call_id -> name` map with one shared helper used by `agentic_loop.py:371`, `modes/tools.py:459` and `tool_executor.py:170`, skipping falsy ids — the inline variant at `tool_executor.py:170` inserts `tool_names[None] = ""`, which swallows the artifact-name fallback for results with no `tool_call_id`.

## Evidence
Evidence: 8/10 — the change is well demonstrated for its size, with 25 targeted unit tests, integration assertions on the next LLM call for two of the three paths, and an end-to-end harness driving a real FastMCP PNG through the real normalization and injection hooks. The one missing artifact is any observation of a live round-trip: no transcript excerpt or post-translation request body showing a real vision model actually receiving the injected block for a non-the reviewer provider.

## Notes
- Vision-capability lookup is copied verbatim from `Orchestrator._model_supports_vision`; extract one shared accessor — `tool_image_context.py:89`.
- Injected notes promise the file "remains in the session files", which artifact ingestion does not guarantee — `tool_image_context.py:66`.
- Full `b64decode(validate=True)` runs on every candidate before the count/budget trim, and the intro text has a double space — `tool_image_context.py:164`, `:195`.

---

garland3:

## Summary
Adds a `tool_image_context` module that extracts LLM-ready images from tool artifacts and injects them as a synthetic multimodal user message, wired into the agentic loop, tools mode, and `execute_tools_workflow`, with vision gating, a rolling six-image cap, and an aggregate size budget.

## Verdict
Request changes.

## Required changes
1. Make a present MIME field authoritative in `_infer_image_mime` at `atlas/application/chat/utilities/tool_image_context.py:196` — a declared `image/svg+xml` on a `.png`-named artifact currently falls through to the extension and ships SVG bytes labelled `image/png`, defeating the allowlist and failing the whole provider turn. Consult `_EXT_TO_IMAGE_MIME` only when no MIME field is present, and add a test for `{"name": "shot.png", "mime": "image/svg+xml"}` asserting the artifact is skipped.
2. Stop skipping list-typed content in the backwards question scan at `atlas/application/chat/utilities/tool_executor.py:1060` and `atlas/application/chat/modes/tools.py:571` — genuine multimodal user turns are skipped along with the injected one, so synthesis answers a question from an earlier turn. Extract text blocks from list content, and mark the injected message with an explicit sentinel to skip on instead.
3. Do not interpolate server-controlled artifact `description` text into the injected `user`/`system` messages at `atlas/application/chat/utilities/tool_image_context.py:192` — prefer the client-side `tool_call_id -> name` map, and when no name is known use a fixed literal rather than `_tool_name_from_artifact`.
4. Guarantee forward progress in `_demote_oldest_images` at `atlas/application/chat/utilities/tool_image_context.py:311` — pop the stale entry on the no-`image_url`-block path and bound both caller `while` loops by the initial `len(self._live)`, since a zero return with no pop is an unbounded CPU-spinning loop inside the request handler.
5. Emit a note to the model when image artifacts were present but every one was rejected, at `atlas/application/chat/utilities/tool_image_context.py:283` — the vision-capable path currently returns silently, leaving the empty-results symptom this PR set out to fix.
6. Add a path-level test for the tools-mode continuation wiring at `atlas/application/chat/modes/tools.py:456` — round 1 returns an image artifact, round 2 repeats a call, asserting exactly one synthetic user message with an `image_url` block lands after the round-1 tool messages.

## Evidence
Evidence: 8/10 — well demonstrated for the change's size, with a dedicated 25-test module plus a PR-validation harness that drives a real MCP server and asserts the transcript base64 decodes byte-for-byte to the tool's PNG. The one missing artifact is any observation of a live provider round-trip confirming the injected block survives translation and the model actually sees the image.

## Notes
- `model_supports_vision` duplicates `Orchestrator._model_supports_vision` verbatim — `atlas/application/chat/utilities/tool_image_context.py:89`.
- Demotion count is recovered by parsing the module's own prose rather than tracked structurally — `atlas/application/chat/utilities/tool_image_context.py:441`.
- Every candidate image is fully base64-decoded before the count/budget caps trim it — `atlas/application/chat/utilities/tool_image_context.py:164`.

---

garland3:
Review follow-up summary (all points addressed or dispositioned):

**Copilot**
- Non-vision fallback mutating tool-message JSON: fixed in 0466fed -- the note is now a separate `role: "system"` message after the tool results, and the tool result string is never modified (regression tests assert the JSON stays parseable).
- CHANGELOG length: condensed to the 1-2 line convention in 0466fed (and the entry renumbered #914).

**Codex**
- P1 validation script: `test/pr-validation/test_pr914_tool_images_reach_llm.sh` + `test/pr-validation/fixtures/pr914/harness.py` run a real in-process FastMCP server whose tool returns real `ImageContent`, push it through `MCPToolManager` normalization and the injector, and assert the transcript bytes match the tool's PNG (added in 5412148).
- P2 synthesis question lookup: `synthesize_tool_results()` and `_stream_synthesis()` now count only plain-string user messages as the question, so the synthetic multimodal message can no longer silently drop the synthesis prompt (0466fed).
- P2 count cap: images are demoted individually, so a single tool returning more than the cap keeps its newest cap-sized slice instead of losing the whole message (0466fed).
- P2 size budget: the oldest live images are demoted (newest-wins) before a new image is rejected (0466fed).
- P1 vendor names in the docstring: left as-is. The text describes provider wire-format behavior (the same technical references `message_builder.py`, `file_processor.py`, and `litellm_caller.py` already make); the OPSEC rule concerns attributing generated work, and neither the commits nor the PR attribute authorship to any AI tool.

CI is green on the head commit.

garland3:
## Summary
Adds a `ToolImageInjector` that extracts images from MCP tool results and injects them into the working transcript as a synthetic multimodal message (with a text-only note for non-vision models), wired into the agentic loop, tools-mode continuation, and `execute_tools_workflow`, under per-image/per-turn/aggregate size caps.

## Verdict
Request changes.

## Required changes
1. Pop the entry on the no-`image_url`-block path in `_demote_oldest_images` (`atlas/application/chat/utilities/tool_image_context.py:361`) and bound both caller loops (`_inject` budget loop, `_enforce_count_cap`) by the initial `len(self._live)` — today a zero-progress return leaves both loop conditions unchanged, spinning the async request handler at 100% CPU with no exception to catch.
2. Make a present MIME field decisive in `_infer_image_mime` (`atlas/application/chat/utilities/tool_image_context.py:112`): return `None` when `mime`/`mime_type` is set but outside `_LLM_READY_IMAGE_MIME_TYPES`, and consult `_EXT_TO_IMAGE_MIME` only when no MIME field exists — `{"name": "shot.png", "mime": "image/svg+xml"}` is currently relabelled `image/png`, defeating the documented SVG exclusion and failing the provider request on the magic-byte mismatch.
3. Extract text blocks from list-typed user content in both backwards scans (`atlas/application/chat/utilities/tool_executor.py:1067` and `atlas/application/chat/modes/tools.py:580`) instead of requiring `isinstance(content, str)`, and identify the injected message by an explicit sentinel key set in `build_tool_image_message` — the current guard also skips genuine multimodal user turns, so synthesis answers an earlier turn's question whenever the user uploaded a file.
4. Restrict the tool label derived from server data at `atlas/application/chat/utilities/tool_image_context.py:220` — prefer the client-built `tool_call_id -> name` map, fall back to the literal `"a tool"`, and if server-derived names must be shown, truncate to ~64 chars of `[A-Za-z0-9_.-]` — unbounded MCP-server prose currently lands verbatim inside `role: "user"` and `role: "system"` messages.
5. Store `"count": len(kept)` on the live entry and format `_DEMOTED_NOTE` from it, deleting `_count_injected_images` (`atlas/application/chat/utilities/tool_image_context.py:440`) — recovering the count by splitting the module's own prose on `" returned "` yields `0 image(s)` after any copy edit or on a tool label containing that phrase.
6. Append a note when image artifacts were present but every one was rejected (`atlas/application/chat/utilities/tool_image_context.py:334`) — the vision-capable path currently returns silently, leaving the model with `{"results": {}}` and no explanation, which is the exact symptom this change exists to remove.
7. Add tests for the three gaps above: `{"name": "shot.png", "mime": "image/svg+xml"}` is skipped; a live entry whose `image_url` blocks were stripped externally terminates rather than hangs on the next over-cap injection; and the tools-mode continuation loop in `_run_extra_rounds` (`atlas/application/chat/modes/tools.py:456`) injects exactly one image message with the real tool name.

## Evidence
Evidence: 8/10 — a dedicated 29-test module, path-level integration tests for two of the three wiring points, and an end-to-end harness driving a real FastMCP server through normalization and asserting byte-for-byte base64 round-trip make this well-demonstrated for a backend change of this size. The one gap is any observation of a real provider round-trip: nothing exercises the translation layer, so the central claim that this block shape becomes native image input is argued by analogy to user uploads rather than shown.

---

garland3:
Addressed the AGENT-REVIEW-BOT-3 pass in 6e69414, with one deviation:

1. Bounded demotion (required #1): a live entry whose blocks were stripped externally is retired to the note-and-pop path, and both caller loops break on zero progress -- no spin possible.
2. Decisive MIME (required #2): an explicit `mime` field outside the allowlist now returns None; the extension fallback applies only when no MIME field exists. `{"name": "shot.png", "mime": "image/svg+xml"}` is skipped (regression test added).
3. Question lookup (required #3): both scans now extract text blocks from genuine multimodal user turns and skip injected messages structurally via a new `is_tool_image_message()` check. Deviation: instead of a sentinel key set in `build_tool_image_message` (extra keys on message dicts flow through `_sanitize_messages` -- which only strips empty `tool_calls` -- into provider requests, and strict APIs reject unknown message fields), the injected/demoted messages are recognized by the module's automated-note prefix, which tools and users do not emit. This also fixes the demoted-note variant: after the cap empties a message, its string note previously still qualified as the question.
4. Tool label bounding (required #4): labels from the id->name map and the artifact-description fallback are sanitized to a 64-char filename-like subset before entering any `role: "user"`/`"system"` message.
5. Stored count (required #5): the live entry carries `"count": len(kept)`; `_count_injected_images` prose parsing deleted.
6. Rejection note (required #6): a step whose images are all budget-rejected appends the explanatory note.
7. Tests (required #7): SVG-with-PNG-name skipped; stripped-blocks desync terminates with the stale entry retired; tools-mode continuation round injects exactly one image message with the real tool name; plus budget-rejection note, label sanitization, and demoted-note-not-the-question coverage. 37 tests, all passing; ruff clean; the end-to-end FastMCP harness still passes.

The provider round-trip observation gap in the evidence section stands -- exercising a real provider's image translation needs live credentials, so it remains argued by parity with the user-upload path (`message_builder._build_multimodal_user_message`), which runs the same block shape through LiteLLM in production today.

PR #924, opened 2026-09-11
refactor(runs): collapse run-event tagging into one authority (#915)

### LINKED ISSUE #915: Collapse the two run-event tagging implementations into one authority
Follow-up from the review of #903 (parallel conversation runs, #884). Raised as a note there and deliberately deferred: it is a refactor of working code, and folding it into a set of security/correctness fixes would have made both harder to review.

## The problem

Two functions stamp `run_id`/`conversation_id` onto an outbound frame, and they disagree on more than style:

| | `tag_run_event` (`atlas/main.py:165`) | `stamp_with_current_run` (`atlas/application/chat/runs/context.py:51`) |
|---|---|---|
| Mutation | copies (`dict(message)`) | mutates in place |
| Source of ids | passed explicitly | read from the `_current_run` contextvar |
| No run in scope | n/a — caller has ids | returns the frame untouched |

Both correctly let an existing id win, which is the rule that actually matters for routing. But the copy-vs-mutate split is the kind of difference that drifts silently: a change to one (say, overwriting a stale `conversation_id`, or normalising a field) does not reach the other, and the two are reached by different paths — the turn callback versus the shared event publisher — so a divergence would show up as events being routed correctly from one code path and not the other. That failure mode looks like "messages in the wrong chat", which is the exact bug the run ids exist to prevent.

The mutation difference is also load-bearing as written: `stamp_with_current_run` documents that it mutates because the publisher builds each frame fresh and copying every token event would be wasteful. So the fix is not simply "make both copy".

## Suggested shape

One tagging function with an explicit copy/in-place choice, with `stamp_with_current_run` reduced to resolving the contextvar and delegating. That keeps the hot-path allocation behaviour while leaving a single place where the tagging rule lives.

## Acceptance

- One function implements the stamping rule; the other delegates.
- The existing tagging tests in `atlas/tests/test_parallel_run_transport.py` still pass unchanged (they cover "existing ids win", non-dict passthrough, and no-mutation-of-the-original).
- A test asserts the hot path still does not copy, so the reason for the split is not lost in the collapse.

garland3:
Closes #915.

## Problem

Two functions stamped `run_id`/`conversation_id` onto outbound frames, each with its own copy of the "existing ids win" rule:

- `tag_run_event` (`atlas/main.py`) — copies, ids passed explicitly, used by the turn callback.
- `stamp_with_current_run` (`atlas/application/chat/runs/context.py`) — mutates in place, ids read from the `_current_run` contextvar, used by the shared `WebSocketEventPublisher`.

The rule being written twice is the risk: a change to one (overwriting a stale `conversation_id`, normalising a field) does not reach the other, and the two are reached by different paths — so a drift surfaces as events routed correctly from one path and not the other. That failure mode is "messages in the wrong chat", the exact bug the run ids exist to prevent.

The copy/mutate split is load-bearing, though: the publisher builds each frame fresh per send, so copying every token event would be pure waste. "Make both copy" is not the fix.

## Change

`tag_event(data, run_id, conversation_id, *, copy=True)` in `runs/context.py` is now the single authority for the stamping rule. The copy/in-place choice is an explicit argument rather than an implicit difference between two functions.

- `tag_run_event` delegates with `copy=True` (it is handed an event it does not own).
- `stamp_with_current_run` resolves the contextvar and delegates with `copy=False` (hot path, unchanged allocation behaviour).

No behaviour change on either path.

## Tests

- The existing tagging tests in `atlas/tests/test_parallel_run_transport.py` pass unchanged (existing-ids-win, non-dict passthrough, no-mutation-of-the-original).
- Added `test_publisher_hot_path_stamps_in_place_without_copying` — asserts `stamp_with_current_run` returns the *same object*, so the reason for the split cannot be quietly lost in a future cleanup.
- Added `test_both_tagging_entry_points_share_one_rule`.

Full backend suite: 3251 passed, 17 skipped.

https://claude.ai/code/session_01GVmeURf3k3pEUQXiqfrjkM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### PR #924 - 2026-09-11
+- The two run-event tagging implementations (`tag_run_event` on the transport callback and `stamp_with_current_run` on the shared publisher) now share one authority, `tag_event`, with the copy-vs-in-place choice made explicit; the publisher hot path still stamps in place and is covered by a test (closes #915).
+
 ### PR #920 - 2026-09-10
 - Transient LLM failures (rate limits, timeouts, 5xx) are now retried with exponential backoff on the streaming path (previously none) as well as the non-streaming path, with the retry count configurable via `LLM_MAX_RETRIES` (default 5) and a cumulative backoff cap via `LLM_RETRY_MAX_WAIT_SECONDS` (default 300s = 5 minutes); once a streaming response has yielded a token, failures surface instead of retrying (closes #919).
 
--- a/atlas/application/chat/runs/context.py
+++ b/atlas/application/chat/runs/context.py
@@ -48,17 +48,49 @@ def clear_current_run() -> None:
     _current_run.set(None)
 
 
+def tag_event(
+    data: Dict[str, Any],
+    run_id: str,
+    conversation_id: str,
+    *,
+    copy: bool = True,
+) -> Dict[str, Any]:
+    """Stamp an outbound event with the run that produced it (issue #884).
+
+    The single authority for the tagging rule; every caller goes through here
+    so the rule cannot drift between code paths (issue #915).
+
+    With several conversations executing at once, a bare event is ambiguous:
+    the client cannot tell whether a token, tool row, file, or completion
+    belongs to the conversation on screen or to one running in the background.
+    Every event a tracked run emits carries both ids so the client can route it
+    -- and, crucially, so it can *drop* events for a conversation it is not
+    displaying instead of splicing them into the visible transcript.
+
+    Existing ids are never overwritten: a producer deeper in the pipeline that
+    already knows its own conversation is more authoritative than the run
+    envelope.
+
+    ``copy`` picks between the two call sites' needs. The turn callback is
+    handed an event it does not own, so it copies. The publisher builds each
+    frame fresh per send and stamps in place, because copying every token event
+    would be wasteful for no benefit.
+    """
+    if not isinstance(data, dict):
+        return data
+    tagged = dict(data) if copy else data
+    tagged.setdefault("run_id", run_id)
+    tagged.setdefault("conversation_id", conversation_id)
+    return tagged
+
+
 def stamp_with_current_run(data: Dict[str, Any]) -> Dict[str, Any]:
     """Add run_id/conversation_id to an outgoing frame, if a run is executing.
 
-    Mutates and returns the frame: it is built fresh per send by the publisher,
-    and copying every token event would be wasteful for no benefit. Existing
-    values win -- a producer that knows its own conversation is more
-    authoritative than the ambient run.
+    Resolves the ambient run and delegates the stamping rule to
+    :func:`tag_event`, in place -- see its ``copy`` note for why.
     """
     context = _current_run.get()
-    if context is None or not isinstance(data, dict):
+    if context is None:
         return data
-    data.setdefault("run_id", context.run_id)
-    data.setdefault("conversation_id", context.conversation_id)
-    return data
+    return tag_event(data, context.run_id, context.conversation_id, copy=False)
--- a/atlas/application/chat/runs/context.py
+++ b/atlas/application/chat/runs/context.py
@@ -18,7 +18,7 @@ adapter stamps outgoing frames from it.
 from __future__ import annotations
 
 from contextvars import ContextVar
-from typing import Any, Dict, NamedTuple, Optional
+from typing import Any, NamedTuple, Optional
 
 
 class RunContext(NamedTuple):
--- a/atlas/application/chat/runs/context.py
+++ b/atlas/application/chat/runs/context.py
@@ -49,12 +49,12 @@ def clear_current_run() -> None:
 
 
 def tag_event(
-    data: Dict[str, Any],
+    data: Any,
     run_id: str,
     conversation_id: str,
     *,
     copy: bool = True,
-) -> Dict[str, Any]:
+) -> Any:
     """Stamp an outbound event with the run that produced it (issue #884).
 
     The single authority for the tagging rule; every caller goes through here
@@ -71,6 +71,10 @@ def tag_event(
     already knows its own conversation is more authoritative than the run
     envelope.
 
+    Anything that is not a dict is returned untouched -- the publisher and the
+    turn callback both forward whatever a producer handed them, and a non-dict
+    frame has nowhere to put the ids.
+
     ``copy`` picks between the two call sites' needs. The turn callback is
     handed an event it does not own, so it copies. The publisher builds each
     frame fresh per send and stamps in place, because copying every token event
@@ -84,7 +88,7 @@ def tag_event(
     return tagged
 
 
-def stamp_with_current_run(data: Dict[str, Any]) -> Dict[str, Any]:
+def stamp_with_current_run(data: Any) -> Any:
     """Add run_id/conversation_id to an outgoing frame, if a run is executing.
 
     Resolves the ambient run and delegates the stamping rule to
--- a/atlas/main.py
+++ b/atlas/main.py
@@ -47,7 +47,7 @@ from atlas.application.chat.runs import (
     RunStatus,
     get_run_registry,
 )
-from atlas.application.chat.runs.context import set_current_run
+from atlas.application.chat.runs.context import set_current_run, tag_event
 from atlas.application.chat.runs.eligibility import turn_is_eligible_for_background_run
 from atlas.application.chat.service import UNSET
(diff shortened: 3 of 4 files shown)


garland3:
## Summary
Consolidates two duplicated event-tagging implementations into a single `tag_event` rule in `atlas/application/chat/runs/context.py`, with `tag_run_event` and `stamp_with_current_run` delegating to it. No behaviour change is claimed.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size: both original call sites are named, the copy-vs-in-place distinction is pinned by an identity assertion, and a full backend suite result is reported. The one gap is breadth, not presence: the equivalence test covers a single input shape.

## Notes
- Parametrise `test_both_tagging_entry_points_share_one_rule` (`atlas/tests/test_parallel_run_transport.py:228`) over pre-existing `run_id`/`conversation_id` and non-dict inputs — the current single case would pass even if the two rules diverged.
- Cover the relocated non-dict guard at its new call site: assert `stamp_with_current_run("not-a-dict")` passes through while a run is active (`atlas/application/chat/runs/context.py:93`).
- `tag_event` returns non-dict input unchanged, which its `Dict[str, Any] -> Dict[str, Any]` annotation and docstring do not reflect (`atlas/application/chat/runs/context.py:54`).

---

garland3:
Thanks — all three notes addressed in 3rd commit.

- **Parametrised the equivalence test.** `test_both_tagging_entry_points_share_one_rule` now runs over the shapes the rule actually distinguishes: a bare event, a producer-supplied `conversation_id`, both ids pre-set, `"not-a-dict"`, and `None`. Each case asserts `tag_run_event` *and* `stamp_with_current_run` against the same `tag_event` result, so a divergence can no longer hide in an input the single happy-path case never reached.
- **Non-dict guard at its new call site** is covered by that same parametrisation: the non-dict cases run with a run bound in the contextvar, so they exercise `stamp_with_current_run`'s delegation rather than its `context is None` early return.
- **Typing.** `tag_event` and `stamp_with_current_run` are now `Any -> Any` instead of `Dict[str, Any] -> Dict[str, Any]`, and the docstring states the passthrough explicitly.

On the Copilot note: the CHANGELOG entry landed in the second commit, after that review was taken on `916e8f5`.

Full backend suite: 3255 passed, 17 skipped.

Also verified in the running app rather than only in tests — two agent-mode conversations executing at once, each showing a green "Running" badge in the conversation history sidebar, with each transcript receiving only its own events.

garland3:
## Summary
Consolidates the run/conversation id stamping rule into a single `tag_event` authority in `atlas/application/chat/runs/context.py`, with `stamp_with_current_run` and `tag_run_event` delegating to it, plus parametrised tests and a CHANGELOG entry.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well-demonstrated for its size — the copy vs. in-place invariants are pinned by tests and full-suite counts are reported before and after; the only gap is that the claimed two-conversation runtime check has no screenshot or reproduction steps attached.

## Notes
- Widen `tag_run_event` to `Any -> Any` at `atlas/main.py:165` — it inherits the non-dict passthrough and the new tests already call it with `"not-a-dict"` and `None`.
- Strengthen `test_both_tagging_entry_points_share_one_rule` at `atlas/tests/test_parallel_run_transport.py:257` — it derives `expected` from `tag_event`, so a regression inside `tag_event` (e.g. `setdefault` → assignment) moves the oracle in lockstep; carry literal expected values in the parametrisation instead.
- Drop the now-unused `Dict` import at `atlas/application/chat/runs/context.py:21`.

---

garland3:
## Verified in the running app

Two agent-mode conversations executing **at the same time**, both showing a green `Running` badge in the conversation history sidebar, with each transcript receiving only its own run's events — which is exactly the routing the collapsed `tag_event` rule is responsible for.

Against the final branch code (all three commits):

Two conversations running in parallel in the history sidebar

An earlier capture of the same state, taken as the second run started:

Both conversations showing Running as the second run starts

<details>
<summary>How it was run</summary>

`python main.py` on port 8011 with a dedicated `CHAT_HISTORY_DB_URL` (the repo default duckdb file was locked by another local Atlas instance, which silently disables chat history and therefore disables background runs entirely). Driven in Chromium: model `groq-gpt-oss-120b`, agent mode ON, auto-approve ON, save mode **Saved to Server**, incognito OFF — all four are required by `turn_is_eligible_for_background_run`.

The only identity visible in the screenshots is the local dev placeholder `test@test.com`; no PII.

The images live on an orphan branch `assets/pr924-screenshots` so they stay out of this PR's diff. Happy to delete that branch once the PR is merged.
</details>

garland3:
Second round addressed in 4th commit (`atlas/main.py`, `runs/context.py`, tests).

- **Literal test oracles.** Good catch — deriving `expected` from `tag_event` meant a regression inside it (`setdefault` becoming assignment) would move the oracle in lockstep and all three entry points would still agree. Each parametrised case now carries its expected value written out, and `tag_event` itself is asserted against that literal alongside the two delegating entry points.
- **`tag_run_event` widened to `Any -> Any`**, with the docstring noting it inherits the passthrough.
- **Dropped the unused `Dict` import.**

On the evidence note: the two-conversation runtime check now has screenshots and reproduction steps attached, in the comment above.

Full backend suite: 3255 passed, 17 skipped.

garland3:
## Summary
Collapses two duplicated run/conversation event-tagging implementations into a single `tag_event` rule in `atlas/application/chat/runs/context.py`, with the publisher hot path stamping in place and the run-event path copying. Behaviour-preserving refactor across one module, one call site, a test file, and a CHANGELOG entry.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — a parametrised equivalence test pins all three entry points against literal oracles, the copy/in-place split has an identity assertion, full-suite counts are reported before and after, and parallel-conversation routing was exercised in the running app with screenshots. No missing artifact of consequence.

## Notes
- No test asserts `tag_event`'s copying half directly (`tagged is not original`); the `==`-based parametrised test cannot distinguish copy from in-place — `atlas/tests/test_parallel_run_transport.py:257`.
- A producer-supplied `conversation_id` that differs from the ambient run's is silently kept by `setdefault` with no log or counter — `atlas/application/chat/runs/context.py:85`.
- Widening the signatures to `Any -> Any` drops the only machine-checked contract at the transport boundary; a `TypeVar` would keep the passthrough honest without losing the signal — `atlas/application/chat/runs/context.py:51`.

---

garland3:
## Summary
Collapses run/conversation event tagging into a single authority in `atlas/application/chat/runs/context.py`, with a `copy` flag selecting the copying path (`tag_event`/`tag_run_event`) versus the in-place publisher hot path (`stamp_with_current_run`). Behaviour-preserving refactor plus new equivalence tests.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10 — proportionate for a medium behaviour-preserving refactor: parametrised equivalence tests across all three entry points, an identity assertion pinning the in-place invariant, before/after full-suite counts, and a two-conversation runtime check. The only gap is that the runtime screenshots are hot-linked from an orphan branch the author plans to delete; re-upload them onto a PR comment so they survive merge.

## Notes
- Restore a static contract with `T = TypeVar("T"); def tag_event(data: T, ...) -> T` instead of `Any -> Any` — `atlas/application/chat/runs/context.py:51`.
- Mirror the in-place identity test for `copy=True` using an input with both ids pre-set, so a lost copy would fail — `atlas/tests/test_parallel_run_transport.py:209`.
- Log at warning level when a producer-supplied `conversation_id`/`run_id` differs from the ambient run's, so a misrouted event is detectable — `atlas/application/chat/runs/context.py:85`.

---

garland3:
Post-merge follow-up on the last review.

**Screenshot hosting** — good catch, and it reverses my earlier offer: I will **not** delete `assets/pr924-screenshots`. `gh` cannot upload image assets to a comment (that path needs the web UI), so the orphan branch is what keeps the images alive. Deleting it would break both images above. The branch holds two PNGs and nothing else, is not in any merge path, and costs nothing to keep. If someone would rather have them attached natively, drag the files into a new comment via the web UI and the branch can then go.

**The two code notes arrived after merge**, so rather than reopen: filed as a follow-up issue so they are not lost.
- `T = TypeVar("T"); def tag_event(data: T, ...) -> T` restores a static contract that `Any -> Any` gave up. Agreed — `Any` was the blunt fix for the non-dict passthrough and a TypeVar says the same thing without discarding the type.
- Mirroring the identity assertion for `copy=True` (input with both ids pre-set, assert the result is *not* the same object) is the missing half of that pair: today a lost copy on the callback path would not fail anything.

The third note — warning when a producer-supplied id disagrees with the ambient run — is the most interesting of the three and is in the issue too, though it is a behaviour change rather than a nit: it would make the misroute this whole mechanism exists to prevent actually observable.

PR #930, opened 2026-09-13
atlas_search is only available when the user selected it (#921)

### LINKED ISSUE #921: atlas-search tool is used even when it has not been turned on
behavior. Clear all tools, select some mcp-tool that is a search like tool from another mcp server provider, then prompt with "use search find the paid leave policy" and it invokes atlas-search even though it is not on. 

expected: If atlas-search is not enabled then it should not be used.

garland3:
Closes #921. Also ships two small UX changes requested alongside it.

## Problem

Selecting a data source implied `atlas_search` (#862): the backend added it to the LLM tool schema "as if the user had ticked it". A user who cleared every tool still got `atlas_search` invoked (the sources rode along and the implication put the tool back), and a user with another provider's search tool could see the built-in tool called over their own pick -- exactly the report in #921: "use search find the paid leave policy" invoked `atlas_search` even though it was never turned on.

## What changed

**Search tool is explicit (closes #921)**
- `ChatOrchestrator._resolve_search_tool` becomes `_check_data_sources_reachable`: it no longer adds any tool. A source selection stays the ceiling on what `atlas_search` may read; the genuinely stranded case -- sources plus other tools, search tool not ticked -- now publishes a warning telling the user how to get their sources read. The check runs inside the tool-running branches, *after* tool authorization in the tools branch, so stripping `atlas_search` by ACL cannot strand the sources silently.
- No warning for turns that route to RAG mode (no tools selected, `only_rag`), which read the sources themselves; none for auto-expanded source lists (RAG toggle on, none hand-picked -- those sources were never deliberately chosen, so the chat frame carries `data_sources_auto` and the backend stays silent); and none for turns naming the tool under its legacy spelling (`atlas_rag_query`) with the flags on. A search tool named while `FEATURE_ATLAS_RAG_TOOLS_ENABLED` is off is stranded the same way and warns.
- `agentic_loop` and tools mode build the schema from the user's selection alone; `search_tool_selection.py` is deleted.
- Agent mode with no tools still downgrades to a normal/RAG turn with an in-chat note (message reworded: "ran without agent mode").

**Agent mode with no tools: warn, don't block**
- The chat-bar guard that blocked the send entirely (unless sources were selected) is gone. The composer shows a small persistent warning banner instead, and the send goes through -- the backend downgrades the turn and says so in the transcript. Blocking took away the user's typed message to make a point the banner can make without it. The banner (and the agent-run flag driving Stop-button state) are gated on `agentModeAvailable` as well as `agentModeEnabled`, since the enabled flag is persisted and can outlive a config where the feature is off.

| Composer warning with agent mode on and no tools | Sources stranded by a tools turn with no search tool |
| --- | --- |
| banner | warning |

**Tools tab: Save and Close**
- The footer button is now "Save and Close" in both the standalone modal and the Tools tab of the combined panel: it commits the staged selections and dismisses the panel. Embedded, it routes through a new `onSaveAndClose` path that still asks about an unsaved prompt draft elsewhere (the tools dirty-guard is skipped deliberately -- the commit that just ran is the reason the panel is closing, and the parent's copy of the dirty flag only updates in an effect). The in-place "Tool selections saved" confirmation is gone since saving always dismisses; the prompt-draft branch, which can leave the panel open, toasts the save instead.

## Reviewer notes
- The design record for the reversed #862 decision: `docs/developer/design-notes/search-only-when-selected-2026-09-13.md`.
- The implied-tool tests were inverted, not deleted: `TestAgenticLoopSearchIsATool` and the tools-mode iteration test now assert the model saw exactly the user's selection; `test_data_sources_reachability.py` pins the warning and its silences (auto-expanded sources, legacy name, flags off, no config).
- PR validation scripts for both this PR and the superseded #862 behavior pass (`run_pr_validation.sh`-style, all checks green).

## Test plan
- [x] Backend suite: `./test/run_tests.sh backend` (3283 passed, 17 skipped)
- [x] Frontend suite: `npx vitest run` (74 files / 868 tests) and `npm run lint` (0 errors)
- [x] `ruff check atlas/`: no new findings (45 pre-existing on main, unchanged)
- [x] `bash test/pr-validation/test_pr930_atlas_search_only_when_selected.sh` -- 6/6 checks pass; `test_pr862` updated for the superseded behavior and passing
- [x] Browser: agent mode on + no tools shows the warning banner; sending runs a normal/RAG turn with the in-chat downgrade note and no tool call
- [x] Browser: sources + a non-search tool + no `atlas_search` -> "Your data sources were not searched" warning fires (screenshot above)
- [x] Browser: Tools tab "Save and Close" commits and dismisses the panel; prompt-draft guard preserved for other exit paths


garland3:
## Summary
Reverses the behavior where selecting data sources implicitly added the `atlas_search` tool, so sources are only searched when the search tool is explicitly selected, and adds a "data sources were not searched" warning plus a composer banner and a "Save and Close" tools footer in its place.

## Verdict
Request changes.

## Required changes
1. Suppress the stranded-sources warning when the source list was auto-expanded rather than explicitly chosen — `atlas/application/chat/orchestrator.py:233`; with the global RAG toggle on and no individual source ticked, the client sends every reachable source id, so any non-search tool selection triggers "Your data sources were not searched." on every turn about sources the user never picked.
2. Gate the composer banner on `agentModeAvailable && agentModeEnabled` at `frontend/src/components/ChatArea.jsx:975` — the removed send-guard checked availability, so a persisted enabled flag on a deployment without agent mode now warns about a mode that is not in play.
3. Evaluate `_check_data_sources_reachable` against the post-authorization tool list on the tools path (`atlas/application/chat/orchestrator.py:479`) — an unauthorized `atlas_search` currently silences the warning while the turn runs without reading the sources.
4. Add a routing-level test that drives the warning through `execute()` with a real `config_manager` (`atlas/tests/test_data_sources_reachability.py:1`) — every new test calls the private helper directly, so dropping the `only_rag` keyword or moving the call below the agent-mode guard would pass the suite.
5. Reword the warning's escape hatch out of internal vocabulary at `atlas/application/chat/orchestrator.py:236` — "plain RAG" maps to no label the user can click.

## Evidence
Evidence: 7/10 — unusually well demonstrated for its size (inverted backend tests, a new reachability suite, updated frontend suites, a design note and a test plan). The main gap is user-facing: the linked screenshot URL is not a real attachment path, and the central artifact — a capture of the in-chat "were not searched" warning as the user sees it — is missing.

## Notes
- "Save and Close" can commit and then fail to close when a prompt draft is dirty, with the saved-confirmation removed — `frontend/src/components/SettingsPanel.jsx:137`.
- Footer reads inconsistently: Cancel stays disabled while the always-enabled primary button may save nothing — `frontend/src/components/ToolsPanel.jsx:1310`.
- The non-string tool-name filter is now inlined at three call sites with no named home or test — `atlas/application/chat/agent/agentic_loop.py:135`.

---

garland3:
Thanks for the review -- addressed in 5950a5f8, finding by finding:

**"PR validation suite calling the removed method" (critical)**: real -- `test_pr862_search_explicit_tool_call.sh` checks 3-4 asserted the implied-tool behavior this PR removes, so `run_pr_validation.sh all` would have gone red. They now assert the current behavior (only the user's selection reaches the schema; stranded sources warn) with a comment crediting #930 for the supersession. The `[Unreleased]` changelog finding was from the first commit; the entry is in place since e6c4dc1a.

**"Feature-flagged search can suppress warnings while silently skipping selected sources" (moderate)**: real and fixed. A turn that *named* `atlas_search` while `FEATURE_ATLAS_RAG_TOOLS_ENABLED` is off never reaches the schema, so its sources were stranded with no warning (the old code warned for exactly this combo). `_check_data_sources_reachable` now treats named-but-flag-disabled the same as unticked, and the warning text covers both reasons ("not selected, or it is turned off"). New test: `test_a_search_tool_named_while_the_flags_are_off_still_warns`.

**"ChatArea.jsx should gate the warning on agent availability" (moderate)**: real -- the enabled flag is persisted and can outlive a config where the feature is off. The banner (and a new test) now gates on `agentModeAvailable && agentModeEnabled`.

**"ChatContext.jsx can show agent-running state after a no-tool downgrade" (moderate)**: real -- `setIsAgentRunning` now requires `agentModeAvailable && agentModeEnabled && toolsToSend.length > 0`, so a downgraded turn offers normal stop handling instead of agent stop, and an unavailable-but-persisted-enabled toggle no longer shows agent-running UI. The payload intentionally still carries `agent_mode: true` with no tools so the backend's downgrade note explains the turn in the transcript (verified in the browser); the flag only drives Stop-button/indicator state.

**"SettingsPanel.jsx can discard unsaved General settings" (critical)**: the new tools-tab close path has the same behavior as the X / backdrop / Escape paths that already existed -- none of them guard the General tab's staged settings (only the General tab's own Cancel reverts them, deliberately deferred behind the close). This PR neither introduces nor worsens that: "Save and Close" from the tools tab closes exactly as the X does. Fixing the General-tab close guard for all paths is a real gap but pre-existing and out of scope here -- happy to take it as a follow-up issue.

garland3:
## Summary
Reverses the implicit-search behavior from #862: `atlas_search` becomes an ordinary selectable tool, and the orchestrator instead warns the user when data sources are selected but no search tool will run. Includes a composer banner, a relabeled "Save and Close" tools-panel footer, and new backend/frontend test suites.

## Verdict
Request changes.

## Required changes
1. Distinguish auto-expanded data-source lists from explicit user selections before publishing the warning at `atlas/application/chat/orchestrator.py:244` — when the global RAG toggle is on with no source ticked, the client sends every reachable source id (`frontend/src/contexts/ChatContext.jsx:620`), so the warning fires on every turn about sources the user never chose. Either add a client flag (e.g. `data_sources_auto`) that suppresses the warning, or send `[]` and expand server-side.
2. Evaluate reachability against the post-authorization tool list at `atlas/application/chat/orchestrator.py:502` — `filter_authorized_tools` runs afterward and can strip `atlas_search`, leaving the turn unretrieved and silent for users whose access group lacks the built-in `atlas` server.
3. Gate the composer banner on agent availability at `frontend/src/components/ChatArea.jsx:975` — it currently renders in states where the user has no way to act on it.
4. Make "Save and Close" at `frontend/src/components/SettingsPanel.jsx:137` close the panel (or keep a confirmation) on the prompt-dirty branch — it can now commit changes and leave the panel open with the in-place "saved" confirmation removed.

## Evidence
Evidence: 7/10 — the backend and automated-test side is well demonstrated for a 24-file change, but the user-facing half is not: the single screenshot link (`assets/pr921-screenshots/...`) is not a real attachment URL and renders broken. Add a capture of the "Your data sources were not searched." warning as it appears in a chat transcript, and re-upload the existing image so it resolves.

## Notes
- Non-string `selected_tools` filtering is copy-pasted at three sites; extract a shared helper (`atlas/application/chat/agent/agentic_loop.py:135`).
- Warning text leaks the internal tool/flag name to end users (`atlas/application/chat/orchestrator.py:236`).
- User guide still instructs users to press "Save Changes" (`docs/user-guide/tools-and-settings/README.md:28`).

---

garland3:
All four required changes addressed in 636d170e:

1. **Auto-expanded sources** (required): the chat frame now sends `data_sources_auto: true` when the RAG toggle alone expanded the list to "everything I can reach" (`frontend/src/contexts/ChatContext.jsx`); `main.py` forwards it and the reachability check returns early for it, so the per-turn noise on never-chosen sources is gone. Chose the client flag over `[]` + server-side expansion because the expansion semantics ("everything I can reach") already live client-side and #855 documents an empty selection meaning exactly that for the search tool; the flag keeps that behavior untouched while removing the false warnings.

2. **Post-authorization reachability** (required change 2): the check moved inside the tool-running branches -- in the tools branch it runs *after* `filter_authorized_tools`, so a user whose groups lack the built-in atlas server (where authorization strips `atlas_search`) gets the stranded-sources warning instead of a silent skip. The agent branch checks its final list the same way. `test_data_sources_reachability.py` covers the new semantics directly.

3. **Banner gated on availability**: already fixed in the previous commit (5950a5f8) -- the banner and the agent-run flag require `agentModeAvailable && agentModeEnabled`, with a test for the unavailable-but-persisted-enabled state.

4. **Prompt-dirty branch of Save and Close**: it now toasts "Tool selections saved." before raising the prompt-discard dialog, since that branch can leave the panel open and the in-place confirmation was removed.

On the notes: the wire-list normalization is extracted into `atlas/application/chat/utilities/tool_selection.py` and used at all three sites. On the warning text: `atlas_search` stays because it is the literal label of the checkbox the user must tick to act on the warning, and the flag name is kept for the same reason the pre-#921 warning named it -- it is the one part an end user cannot fix themselves and needs to hand to an administrator. The user guide's remaining "Save Changes" mention is updated. The broken screenshot link is fixed (images are committed under `assets/pr921-screenshots/` and referenced by raw URLs), and the transcript capture of the stranded-sources warning was added as requested.

garland3:
## Summary
Replaces the implicit "selecting data sources forces the search tool" behavior with an explicit reachability check that warns when selected sources cannot be read by the turn's final tool list, adds a client-supplied `data_sources_auto` flag to suppress that warning for auto-expanded sources, and relabels the tools panel footer to "Save and Close".

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — proportionate for a 27-file change, with a dedicated reachability suite, inverted assertions in the two suites that encoded the removed implication, and frontend coverage of the banner and send paths. The main gap is that the reachability check is only exercised through the private helper, never through `execute()`, so its position relative to `filter_authorized_tools` is unpinned; one routing-level test with an ACL that strips `atlas_search` would close it.

## Notes
- Agent branch evaluates reachability pre-authorization while the docstring claims otherwise — harmless today (no filtering on that path) but will silently regress if agent-path authorization is added; pin the invariant or unify the call sites (`atlas/application/chat/orchestrator.py:532`).
- `data_sources_auto` is unverifiable client state and untested on the wire; document it in the WebSocket contract and assert the payload key on both sides (`atlas/main.py:1349`, `frontend/src/contexts/ChatContext.jsx:625`).
- "Save and Close" is now always enabled and, on the prompt-dirty branch, saves without closing; consider relabeling to "Close" when `!hasChanges` and rewording the discard dialog (`frontend/src/components/ToolsPanel.jsx:1311`, `frontend/src/components/SettingsPanel.jsx:138`).

---

garland3:
Nits from the second review addressed in fcf646b7:

- **Reachability pinned through `execute()`**: two new routing-level tests run the full orchestrator with a stub ACL -- one where authorization strips `atlas_search` (asserting the stranded-sources warning fires against the post-ACL list and tools mode receives the ACL result), one with `data_sources_auto` set end to end (no warning through the full path).
- **Docstring**: clarified that the tools branch judges the post-authorization list while the agent path does no authorization filtering today, so its selection is already final -- the invariant is now stated where the next reader will trip over it.
- **`data_sources_auto` on the wire**: asserted in the frontend provider suite (RAG toggle on with no picks sends the expanded list plus `data_sources_auto: true`; the default frame sends `false`), and the flag is documented in the WebSocket contract summary in AGENTS.md's Key APIs section.
- **"Save and Close" relabel when clean**: deliberately not taken -- the always-enabled button doubling as plain close is the behavior requested for this change, and a context-dependent label would make the footer unpredictable. The prompt-dirty branch already toasts the save, which covers the "saved without visible confirmation" gap.

garland3:
## Summary
Makes `atlas_search` an explicitly selectable tool instead of an implied one: adds a per-turn "your data sources were not searched" reachability warning, replaces the agent-mode-with-no-tools send block with a backend downgrade, and relabels the tools footer button to "Save and Close".

## Verdict
Request changes.

## Required changes
1. Make a Stop control available for downgraded agent turns at `frontend/src/components/ChatArea.jsx:1220` — gate the normal stop button on `!isAgentRunning` rather than `!agentModeEnabled`, otherwise an agent-mode turn with no tools selected streams with no way to cancel it.
2. Derive one `runsAsAgent` value at send time in `frontend/src/contexts/ChatContext.jsx:745` and use it for both the `agent_mode` payload field and `setIsAgentRunning`, so the client cannot ship `agent_mode: true` while recording the turn as non-agent (happens when a send lands before `/api/config` is applied).
3. Normalize the tool list once at the top of `ChatOrchestrator.execute` (`atlas/application/chat/orchestrator.py:522`) via `normalize_selected_tools`, so the downgrade guard, the reachability check and the schema builders all see the same list — today `selected_tools: [null]` is truthy, skips the guard, and reaches the agentic loop with an empty schema.
4. Do not let `data_sources_auto` suppress the warning when a search tool was named but stripped by authorization (`atlas/application/chat/orchestrator.py:252`) — check the named-tool case first, or only set the flag client-side when `ragEnabled && !searchToolSelected`.
5. Treat `data_sources_auto` as untrusted input at `atlas/main.py:1349` — at minimum coerce with `bool(...)` and log turns where the flag is set but the source list is not the full reachable set; ideally recompute the auto condition server-side.
6. Add a routing-level test for the agent branch in `atlas/tests/test_data_sources_reachability.py` (stub `agent_mode.run`, assert the warning fires) and a handler-level test asserting `execute` is awaited with `data_sources_auto=True` from the websocket frame — both hops are currently unpinned.
7. Rework the warning copy at `atlas/application/chat/orchestrator.py:236` to split the two causes (tool not ticked vs. deployment flag off), keep `FEATURE_ATLAS_RAG_TOOLS_ENABLED` in the administrator sentence and the log line only, and avoid repeating the full multi-sentence block on every turn of an unchanged selection.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size (dedicated backend and frontend suites, a design note, committed screenshots and a checked test plan). The main gaps are cosmetic: no capture of the new "Save and Close" footer or of the in-chat downgrade note, and the screenshot links use branch-ref raw URLs that will stop resolving after merge.

---

PR #931, opened 2026-09-13
test: pin that a stopped agent run survives a refresh (#842)

### LINKED ISSUE #842: not saving conversation if agent mode is stopped
If the agent is running, and on step 30 of 50, then I stop the agent, the conversation doesn't appear to be saved. when I refresh the browser it is gone which is not great since I want to fix something and restart the agent where it stopped. 
Using the duckdb locally.

garland3:
Closes #842.

## Status: already fixed — this PR adds the missing regression coverage

Issue #842 reports that stopping an agent mid-run (say at step 30 of 50) loses the conversation when the browser is refreshed. I could not reproduce it on current `main`.

The underlying defect was fixed by the #755 work: `ChatService.handle_chat_message` catches `asyncio.CancelledError`, calls `close_open_turn()` and then `_commit_turn()` before re-raising, so a stopped turn is persisted exactly the way a completed one is (`atlas/application/chat/service.py:621-643`).

What was *not* covered end to end: #884 later moved the Stop button onto the run registry, so `agent_control {"action":"stop"}` now cancels through `RunRegistry.cancel` rather than the connection's untracked task. The existing interrupted-turn tests cover the mode runners, the service cancel handler and the digest round-trip, but nothing exercised the registry-driven stop all the way into storage.

## What this adds

`atlas/tests/test_stopped_agent_conversation_survives.py`, two cases, both reading back out of a real DuckDB `ConversationRepository` the way the browser does after a reload:

1. An agent run 30 steps deep, stopped through `RunRegistry.cancel` — the user message, all 30 agent steps, and the `[This turn was stopped before it finished.]` terminal message are all in the store.
2. A brand-new conversation stopped on its first turn — the save mints a conversation id and announces it via the `conversation_saved` frame, which is what the frontend handler uses to set the active conversation.

No product code changes.

## Verification

- `bash test/atlas_tests.sh` — **3287 passed, 17 skipped** on this branch.
- `pytest tests/test_interrupted_turn_persistence.py` — 52 passed on unmodified `main`.
- Both new tests pass against `main` with no source changes, which is the evidence that #842 is fixed.

https://claude.ai/code/session_01HKzca6V86wvDa9sa1WVFyS


garland3:
## Summary
Adds a single test module (`atlas/tests/test_stopped_agent_conversation_survives.py`) asserting that a stopped agent run still persists its conversation and interrupted-turn marker, read back from a real DuckDB repository. No product code changes.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — adequately demonstrated for a test-only change; the missing artifact is pasted output of the two new tests run against unmodified `main`, since the PR's thesis is that they pass on the existing fix.

## Notes
- Bound the waits (`asyncio.wait_for(..., 5)` or `@pytest.mark.timeout`) — an unpropagated cancel currently hangs on `asyncio.sleep(3600)` instead of failing red; `test_stopped_agent_conversation_survives.py:99,161`.
- `assert any("step 29" == c ...)` passes even if steps 0–28 were dropped; assert the full ordered sequence and the persisted `message_type` metadata — `test_stopped_agent_conversation_survives.py:110,125`.
- Both tests bypass the transport path they claim to cover: use `get_run_registry()` / `_cancel_addressed_run` rather than a locally built `RunRegistry` and a bare `task.cancel()` — `test_stopped_agent_conversation_survives.py:88,170`.

---

garland3:
Thanks — all three notes were fair and are addressed in 72ac60ca.

**1. Unbounded waits.** Every wait on the turn task now carries a 5s deadline (`CANCEL_DEADLINE_SECONDS`), so a stop that stops propagating fails red instead of parking on the hour-long sleep. Verified by fault injection rather than by inspection — I neutered the `task.cancel()` inside `RunRegistry.cancel` and re-ran:

```
INFO main:main.py:247 Cancelling run 8b8277fb-... on user request
FAILED tests/test_stopped_agent_conversation_survives.py::test_stopping_an_agent_run_by_run_id_keeps_every_step
1 failed in 6.06s
```

Red in 6s, not a hang. (Registry restored afterwards; tree is clean.)

**2. Weak step assertion.** Correct — `any("step 29" == c)` would have passed with the middle of the run dropped. It is now the full ordered sequence with persisted metadata, plus an exact length:

```python
assert rows[1:31] == [("assistant", f"step {step}", "agent_intermediate") for step in range(30)]
...
assert len(rows) == 32
```

**3. Bypassing the transport.** Also correct, and the most useful of the three. Both tests now enter at `main._cancel_addressed_run` against the process-wide `get_run_registry()`, covering both of its resolution branches — a frame naming a `run_id`, and one naming only the `conversation_id`. The second test is no longer the untracked path: it now starts a real run record keyed by the id `main.py:1210` mints for a first agent turn, stops it by conversation id, and asserts the turn is saved under that same id (the thing the `conversation_saved` frame carries back to the client).

**Evidence the tests pass against the existing fix.** Worth being explicit, since that is the PR's whole thesis: this branch contains *no product code* — `git diff main...HEAD --stat` touches one new test file and nothing else — so every run below is a run against unmodified `main`.

```
$ pytest tests/test_stopped_agent_conversation_survives.py -q
..                                                                       [100%]
2 passed in 0.92s

$ bash test/atlas_tests.sh
=========== 3287 passed, 17 skipped, 2 warnings in 93.51s (0:01:33) ============
```

garland3:
Small correction to the command in my previous comment: it should read `git diff origin/main...HEAD --stat` (a stale local `main` ref makes the unqualified form print an unrelated diff). The claim itself is unchanged:

```
$ git diff origin/main...HEAD --stat
 .../test_stopped_agent_conversation_survives.py | 219 +++++++++++++++++++++
 1 file changed, 219 insertions(+)
```

No product code on this branch.

garland3:
## Summary
Adds a single test module (`atlas/tests/test_stopped_agent_conversation_survives.py`, +219/-0, no product code) covering that a stopped agent run — addressed either by `run_id` or by `conversation_id` — still persists its full transcript and a closed-turn marker through a DuckDB reload.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size, with pasted test and full-suite runs, a diff stat confirming no source change rides along, and a negative-control run showing the first test can actually fail. The one missing artifact is a comparable fault-injection run for the `conversation_id` branch.

## Notes
- `_cancel_addressed_run(...) is True` is vacuous on the run_id branch (it returns True for any frame carrying a `run_id`); assert `registry.get(run_id).status is RunStatus.CANCELLED` instead — `test_stopped_agent_conversation_survives.py:88`.
- Neither test asserts the registry reaches a terminal state, so a lost status transition leaving the conversation permanently busy would still pass — `test_stopped_agent_conversation_survives.py:99,170`.
- The second test omits a total message-count assertion, so a double-committed cancelled turn would go unnoticed — `test_stopped_agent_conversation_survives.py:214`.

---

PR #932, opened 2026-09-13
Agent mode defaults on; drop the start banner, keep the Agent box

### LINKED ISSUE #849: remove the special message for agent mode, make agent the default.
remove "Agent Mode Started (strategy: agentic, max steps: 30)"

Continue to show the small box saying "Agent"

Make agent mode on by default everywhere

garland3:
Closes #849.

## What

1. **The verbose start banner is gone.** `agent_start` no longer inserts "Agent Mode Started (strategy: agentic, max steps: 30)" into the transcript. It still inserts a marker, so the run's beginning stays visible -- it now renders as the small purple **Agent** box alone (no text), which is the "small box saying Agent" the issue asks to keep. Other `agent_status` rows (e.g. the max-steps notice) keep badge + copy.
2. **Agent mode is on by default.** `useAgentMode` now defaults the persisted `chatui-agent-mode-enabled` key to `true`. Browsers that never made a choice start enabled; an **existing stored preference is honored** (users who turned it off keep it off), and the toggle still works as before.
3. **Max steps in the config menu matches the admin value.** `/api/config` and `/api/config/shell` now expose `agent_max_steps`; the Max Agent Iterations slider in Tools and Settings is bounded by that ceiling (previously hardcoded 1-50 while the server honors `AGENT_MAX_STEPS`, default 10 -- the mismatch called out in the issue thread), stored values above the ceiling are clamped in the UI, and the send payload is clamped to the same ceiling (the server-side clamp remains as the authority).

Scope note: "everywhere" is taken as the web UI, per the issue thread (toggle stays, stored preference honored, config-menu placement). The Python API (`AtlasClient.chat(agent_mode=False)`) and CLI `--agent-mode` keep their opt-in defaults; flipping those is a one-line follow-up if wanted.

## Verification

- `./test/run_tests.sh backend` and `./test/run_tests.sh frontend` (876 frontend tests / 76 files, backend suite) pass; `npm run lint` 0 errors (4 pre-existing warnings, none in touched files); ruff unchanged vs main (47 pre-existing, none added).
- Live check against `agent_start.sh` on :8001: fresh browser (no stored key) shows the toggle ON; stored `false` shows OFF after reload; config cache picks up `agent_max_steps: 30`; the General tab slider shows max 30 with a "10 / 30" chip; a real agent turn (calculator tool) shows the badge-only Agent marker with no strategy/max-steps text in the transcript.

## Tests

- Updated the two handler tests that pinned the old banner text.
- New: `agent-mode-default.test.jsx` (default on, stored off/enabled honored, force-disable on feature-unavailable still works), `agent-status-marker-render.test.jsx` (badge-only vs badge+copy rendering), and a shell-endpoint test asserting `agent_max_steps` matches the full config under a overridden setting.


garland3:
## Summary
Makes agent mode default-on for browsers with no stored preference, replaces the verbose agent-start transcript banner with a badge-only marker, and exposes the server's `agent_max_steps` ceiling through `/api/config` so the Max Agent Iterations slider and the send payload are bounded by it.

## Verdict
Approve with nits.

## Evidence
Evidence: 7/10. The change is well covered by new hook, render, handler, and backend tests plus a standalone validation script, and the manual verification is specific; the one notable gap is a screenshot of the new badge-only transcript marker and the settings slider, since the visible rendering is the point of the change.

## Notes
- Clamped `maxIterations` is displayed but never persisted — `setSettings` without `saveSettings`/`hasChanges` means localStorage keeps the oversized value and revert restores it (`frontend/src/components/SettingsPanel.jsx:294`).
- Pre-config fallback ceiling of `50` disagrees with the server default of `10`, so a first-time visitor can briefly pick a step count the backend will silently reduce (`frontend/src/hooks/chat/useChatConfig.js:127`, `frontend/src/contexts/ChatContext.jsx:671`).
- `aria-label="Agent mode started"` sits on a role-less `div` and will be dropped by most screen readers; `role="status"` would expose it (`frontend/src/components/Message.jsx:542`).

---

garland3:
Addressed the review findings in 7e4ac3e2:

1. **Availability gating (P1)** — `useAgentMode` now derives the effective flag as `available && storedEnabled` instead of relying on the force-disable effect, so a deployment with `FEATURE_AGENT_MODE_AVAILABLE=false` sees agent mode off everywhere from the first render (including the initial-unavailable case the old effect missed), and the stored preference is never silently overwritten. The send payload keeps an explicit `agent.agentModeAvailable && agent.agentModeEnabled` gate at the transmission boundary. Covered by `agent-mode-payload-gating.test.jsx` (real provider, mocked hook: unavailable + enabled must send `agent_mode: false`) and updated `agent-mode-default.test.jsx` cases.
2. **Docs** — added an "Agent Mode Availability and Step Budget" section to `docs/admin/configuration.md` (kill switch, default-on + honored preference, `agent_max_steps` exposure) and an "Agent mode" section to `docs/user-guide/tools-and-settings/README.md` (default-on, Ctrl+Alt+A, slider bounded by `AGENT_MAX_STEPS`).
3. **Negative `AGENT_MAX_STEPS`** — fixed in 9ce20a3e before this review landed: the endpoints now mirror the orchestrator's clamp (`int(...) or 10`, floored at 1), with degenerate-value assertions in the shell endpoint test.
4. **Slider/payload regression tests** — `settings-panel-agent-mode.test.jsx` now covers the slider bound (max follows the ceiling, scale labels track it), the clamped stored value ("30 / 30" chip, no "50 / 30"), and the fallback to the old hardcoded 50; `agent-mode-payload-gating.test.jsx` covers the outgoing `agent_max_steps` clamp both above and below the ceiling.

Full suites re-run green: backend + 884 frontend tests (77 files), lint 0 errors, `test/run_pr_validation.sh 932` passes.

garland3:
@codex review

garland3:
## Summary
Makes agent mode default-on for browsers with no stored preference, replaces the transcript's agent-start banner with a badge-only marker, and derives the Max Agent Iterations slider ceiling and send-payload clamp from the server's `AGENT_MAX_STEPS`.

## Verdict
Request changes.

## Required changes
1. Enforce the agent-mode kill switch server-side — coerce `agent_mode` to `False` in the WebSocket chat handler/`ChatOrchestrator` when `feature_agent_mode_available` is false (`atlas/main.py:1193`, `atlas/main.py:1355`), with a backend test; today the flag is only honored client-side, and `agent_mode_available` is a cached field, so a browser with a stale config cache sends `agent_mode: true` on its first turn after an admin disables the feature.
2. Add `role="status"` alongside the `aria-label` on the agent-start marker at `frontend/src/components/Message.jsx:542` — `aria-label` on a role-less `div` is not exposed, so removing the banner text leaves screen-reader users with no signal that an agent run started.
3. Persist the clamped `maxIterations` at `frontend/src/components/SettingsPanel.jsx:293` (route it through `saveSettings`, not bare `setSettings`) — localStorage keeps the over-ceiling value, so `revertGeneralSettings` and any direct consumer of `chatui-settings` still see the un-clamped number.
4. Align the pre-config slider ceiling with the server default instead of `50` at `frontend/src/hooks/chat/useChatConfig.js:127` (and the duplicate literal at `frontend/src/contexts/ChatContext.jsx:671`), or disable the control until config resolves — otherwise a first-time visitor picks 40 from a ceiling of 50 and the backend silently clamps to 10.
5. Extract one shared helper for the step-ceiling coercion used by `atlas/routes/config_routes.py:182`, `:532` and `orchestrator._bounded_agent_steps` (`atlas/application/chat/orchestrator.py:167`) — the route copies omit the orchestrator's `try/except`, so a non-numeric `agent_max_steps` 500s `/api/config` while the chat path degrades to 10.
6. Assert the accessible name and the persisted value in the new tests (`frontend/src/test/agent-status-marker-render.test.jsx:45`, `frontend/src/test/settings-panel-agent-mode.test.jsx:124`), and cover `agentMaxStepsLimit` parsing for `null`/`0`/`-5`/`'abc'` in `config-cache-hydration.test.js`; none of the new derived-state branches are currently exercised.

## Evidence
Evidence: 7/10 — the change is well covered by new frontend and backend tests plus a validation script, and manual verification is described concretely. The notable gap is visual: a PR whose stated purpose is changing how the agent marker and slider look includes no before/after screenshot, and the existing `docs/user-guide/tools-and-settings/README.md` General-tab screenshot is now stale.

---

garland3:
Addressed the second review round in 39fed511:

1. **Server-side kill switch** — `ChatOrchestrator` now refuses `agent_mode` when `FEATURE_AGENT_MODE_AVAILABLE=false` (`_agent_mode_available()`, checked before routing), publishes the same style of downgrade note as the no-tools path, and routes the turn normally. Covered by `test_agent_mode_kill_switch_downgrades_to_plain` (stale-cache client sends the flag; server downgrades to tools mode with a warning) and `..._is_noop_when_enabled`.
2. **Accessible marker** — the badge-only row now carries `role="status"` + `aria-label="Agent mode started"`, asserted via `getByRole('status', { name: 'Agent mode started' })`.
3. **Persisted clamp** — the ceiling clamp now writes through to `chatui-settings` (only the `maxIterations` key of the stored JSON), and `Reset to Defaults` clamps to the ceiling too; the settings-panel test asserts the persisted value becomes 30.
4. **Pre-config fallback** — all fallbacks (cache absence, `0`, `-5`, `'abc'`, missing response field) now resolve to the server default `10`, with parsing cases in `config-cache-hydration.test.js`.
5. **Shared coercion** — `configured_agent_max_steps()` in `atlas/modules/config/settings.py` is the single authority (`int(... or 10)`, floored at 1) used by `_bounded_agent_steps` and both config endpoints, so a non-numeric `AGENT_MAX_STEPS` degrades to 10 instead of erroring the endpoint.
6. **Screenshot** — `docs/user-guide/tools-and-settings/screenshots/836-03-general-tab.png` regenerated from the current UI (chip now "10 / 30", scale labels track the configured ceiling); doc date bumped.

After state in the live UI (badge-only start marker, agent turn on calculator):

garland3:
After state (from the regenerated General-tab screenshot in this branch): the Max Agent Iterations slider is bounded by the admin ceiling — chip `10 / 30`, scale labels 1 / 15 / 30 instead of the old hardcoded 1 / 25 / 50:

General tab with Max Agent Iterations bounded by the configured ceiling

Transcript before/after: before, an agent run inserted a compact row reading **"Agent Mode Started (strategy: agentic, max steps: 30)"** after the purple badge; after, the same row renders as the small purple **Agent** box alone (`role="status"`, announced as "Agent mode started"). The max-steps narration now lives only in the config menu.

garland3:
## Summary
Turns agent mode on by default, replaces the verbose "Agent Mode Started" transcript banner with a badge-only marker, centralizes the `AGENT_MAX_STEPS` ceiling behind a shared `configured_agent_max_steps` helper, and adds a server-side `FEATURE_AGENT_MODE_AVAILABLE` kill switch plus a ceiling-bounded Max Agent Iterations slider.

## Verdict
Request changes.

## Required changes
1. Gate the clamp effect at `frontend/src/components/SettingsPanel.jsx:293` on config having actually resolved (e.g. leave `agentMaxStepsLimit` `null` until a config response lands, or expose `configReady`) — today it persists the pre-config fallback ceiling of `10` into `chatui-settings`, silently and irreversibly destroying a user's saved value of `30` on a deployment with `AGENT_MAX_STEPS=30`.
2. Apply the same gate to the `Math.min(defaultSettings.maxIterations, maxStepsLimit)` clamp in `handleReset` at `frontend/src/components/SettingsPanel.jsx:328` — a reset during the pre-config window writes a lower default than the same click produces a second later.
3. Route the clamp write through the existing settings owner (`updateSettings({ maxIterations })`) instead of a direct `localStorage` read-modify-write at `frontend/src/components/SettingsPanel.jsx:296-306` — the third writer leaves `ctxSettings.maxIterations` un-clamped in memory and can be reverted by a later full write from `useSettings`.
4. Move the "agent mode started" wording inside the live region at `frontend/src/components/Message.jsx:542` as visually hidden text — `role="status"` announces contents, not `aria-label`, so screen-reader users now hear only the bare word "Agent".
5. Handle degenerate ceilings in the slider scale at `frontend/src/components/SettingsPanel.jsx:670` — drop the midpoint label when it collides with an endpoint and render static copy when `maxStepsLimit === 1`, which is reachable since `configured_agent_max_steps` floors negatives at 1 ("1 1 1" under an inert slider).
6. Make `_agent_mode_available()` at `atlas/application/chat/orchestrator.py:154` default to denied when `config_manager.app_settings` is missing — a security kill switch should not fail open, and tests can supply an explicit `SimpleNamespace` instead.
7. Assert the downgrade warning's text, not just its call count, at `atlas/tests/test_orchestrator_mode_routing.py:291`, and add direct unit cases for `configured_agent_max_steps` covering `None`/`0`/`-5`/`"abc"`/missing attribute — the non-numeric case that motivated the extraction is currently unasserted anywhere.

## Evidence
Evidence: 8/10 — the change is well demonstrated for its size, with tests on each new branch, an end-to-end validation script, and specific manual-verification notes. The one notable gap is a transcript screenshot of the new badge-only start marker, which is the change's headline behavior; the only attached image shows the General tab and is linked by branch ref, so it will break once the branch is deleted.

---

garland3:
@codex review

garland3:
Addressed the third review round in 0e2b2828:

1. **Clamp gated on a resolved ceiling (SettingsPanel clamp effect)** -- `useChatConfig` now tracks `agentCeilingConfirmed`, set only when a *live* config response (shell or full, never the cache) carries a valid `agent_max_steps`. The clamp effect returns early until then, so the pre-config fallback of 10 (or a stale cache) can no longer rewrite a saved 30. The earlier stale-cache P2 covers this the same way: `configReady` was insufficient because an old cache marks ready without containing the field, so the confirmation is tied to the field itself. Covered by the new `agent ceiling confirmation` cases in `config-cache-hydration.test.js` (cache alone does not confirm; live response with the field confirms; live responses without it stay unconfirmed) and the "does not persist the pre-config fallback" SettingsPanel test.
2. **handleReset gated the same way** -- the reset default only applies the `Math.min(default, ceiling)` clamp once `agentCeilingConfirmed`; before that it writes the plain defaults. Covered by the two Reset tests (unconfirmed writes 10 against a stale cached 5; confirmed writes 5).
3. **Clamp write routed through the settings owner** -- the direct `localStorage` read-modify-write is gone; the effect now calls `updateSettings({ maxIterations: ceiling })` from ChatContext, so `ctxSettings.maxIterations` is clamped in memory too and a later full write from `useSettings` cannot resurrect the un-clamped value. The persist test asserts the owner call.
4. **Live-region wording** -- the badge-only marker drops `aria-label` (`role="status"` announces contents, not the label) and renders "Agent mode started" as visually hidden text inside the region. `agent-status-marker-render.test.jsx` now asserts the wording lives in an `.sr-only` span inside the status region.
5. **Degenerate slider scale** -- a ceiling of 1 renders static copy ("This deployment caps agent runs at a single iteration.") instead of the inert slider plus "1 1 1"; the midpoint label is dropped whenever it would collide with an endpoint (ceilings of 1-2). Both have SettingsPanel tests (ceiling 1 and ceiling 2).
6. **Kill switch fails closed** -- `_agent_mode_available()` returns `False` when `app_settings` is missing; the availability matrix test covers `config_manager=None`, missing `app_settings`, and both flag values, and the downgrade test asserts the routing end to end. The two agent-routing tests that previously ran without settings now supply an explicit `SimpleNamespace`.
7. **Warning text + helper coverage** -- `test_agent_mode_kill_switch_downgrades_to_plain` asserts the published warning contains "**Agent mode is disabled.**" / "agent mode turned off", not just its call count; `TestConfiguredAgentMaxSteps` covers `None`, missing settings, missing attribute, `0`, `-5`, `"abc"` (the non-numeric case), `"30"`, and a valid passthrough. Note: `0` falls back to the default 10 (the helper's `or 10` treats falsy as missing) -- the test pins that exact behavior.

Evidence notes: the General-tab screenshot link in the earlier comment is now pinned to the commit SHA so it survives branch deletion.

garland3:
@codex review

garland3:
## Summary
Makes agent mode the default-on preference, adds a server-side availability kill switch, publishes `agent_max_steps` through the config endpoints as a shared ceiling, and replaces the transcript "Agent Mode Started" row with a badge-only marker.

## Verdict
Request changes.

## Required changes
1. Gate the wire clamp at `frontend/src/contexts/ChatContext.jsx:675` on `config.agentCeilingConfirmed`, sending the user's raw `maxIterations` until the ceiling is confirmed — today `agentMaxStepsLimit || 10` silently downgrades a saved `30` to `10` on any send before the config response lands (first visit, cleared storage, backend without the field), and the server's `_bounded_agent_steps` is already the authority.
2. Add payload-gating cases for `agentMaxStepsLimit` of `undefined` and `0` in `frontend/src/test/agent-mode-payload-gating.test.jsx` — the fallback branch is currently unpinned in either direction.

## Evidence
Evidence: 8/10 — proportionate for a 23-file change, with new tests pinning each branch and named manual verification against a live instance. The one gap is the headline visual change: a transcript crop showing the badge-only start marker was promised but not attached.

## Notes
- `role={hasContent ? undefined : 'status'}` at `frontend/src/components/Message.jsx:548` announces the contentless start marker but not informative rows such as "Agent Max Steps Reached"; consider making the role unconditional.
- The default of `10` is duplicated across `settings.py:93`, `orchestrator.py:177`, `useChatConfig.js:129`, `ChatContext.jsx:675`, and `SettingsPanel.jsx:69` — name it once per side.
- The Max Agent Iterations slider at `frontend/src/components/SettingsPanel.jsx:672` has no `htmlFor`/`id` pairing, so it is announced as an unnamed slider.

---

garland3:
Addressed the fourth review round in 13a1b097:

1. **Payload clamp deferred until the ceiling is live** -- `sendChatMessage` now applies the `Math.min(..., agentMaxStepsLimit)` clamp only when `config.agentCeilingConfirmed`; before that it sends the raw preference (`settings.maxIterations || agent.agentMaxSteps`) and the authoritative server clamp bounds early requests. The stale-cache scenario from the finding (fallback 10 while the deployment raised `AGENT_MAX_STEPS` to 30) now transmits the user's intended 30 instead of a silently shrunken 10. Covered by "sends the raw preference before the ceiling is confirmed" in `agent-mode-payload-gating.test.jsx` (ceiling 5, preference 30, unconfirmed -> wire 30; confirmed -> clamped).
2. **Kill switch applied before run admission** -- `is_agent_turn` in the WebSocket handler is now `agent_mode && agent_mode_available(...)`, so a stale or handcrafted `agent_mode: true` on a disabled deployment never reaches `run_registry.start()` with agent semantics (no run-quota/conversation-lock admission, no steering channel, no conversation-id minting for a turn that will run as plain tools). The wire flag is left intact so the orchestrator's existing check still downgrades with the in-chat note for the user. To keep both paths on one authority, the fail-closed predicate moved to `settings.agent_mode_available()`; the orchestrator delegates to it, and `test_shared_availability_predicate_fails_closed` pins its edges. (The WS admission wiring itself is exercised by the PR validation script's config-endpoint/e2e checks; the registry's own admission semantics remain covered by `test_parallel_conversation_runs.py`.)

garland3:
@codex review

garland3:
## Summary
Makes agent mode on by default, replaces the verbose agent-start banner with a compact badge, and publishes the deployment's `AGENT_MAX_STEPS` ceiling through the config endpoints so the settings slider and the wire payload respect it, with a shared fail-closed availability predicate gating both WebSocket run admission and the orchestrator downgrade path.

## Verdict
Request changes.

## Required changes
1. Gate the wire clamp on confirmation at `frontend/src/contexts/ChatContext.jsx:675` — `Math.min(..., config.agentMaxStepsLimit || 10)` fires before any live config response, silently shrinking a stored `maxIterations: 30` to 10 on first visit or after a cache clear.
2. Disable (or loading-state) the Max Agent Iterations control while `!agentCeilingConfirmed` at `frontend/src/components/SettingsPanel.jsx:661`, and show the stored value rather than `Math.min(..., 10)` — today the panel displays a fabricated "10 / 10" ceiling on a 30-step deployment, and one drag persists the downgrade, reopening the data-loss path the write guards closed.
3. Make the clamp effect at `frontend/src/components/SettingsPanel.jsx:305` a read-modify-write over the current stored value (or make `useSettings` the sole writer of `chatui-settings`) — `updateCtxSettings` serializes a context snapshot that is stale after the panel's own `saveSettings`, reverting other keys saved in the same session.
4. Derive confirmation atomically in `frontend/src/hooks/chat/useChatConfig.js:163` — `setAgentCeilingConfirmed(true)` is one-way, so a later config response that omits `agent_max_steps` drops the limit to 10 while confirmation stays true, and the clamp then persists 10 over a saved 30.
5. Add coverage for the new admission conjunction at `atlas/main.py:1193` — a test (or an extracted `_classify_agent_turn` helper) asserting that `agent_mode: true` under `FEATURE_AGENT_MODE_AVAILABLE=false` starts no run, takes no conversation lock, and still completes as a plain tools turn.
6. Give the slider an accessible name at `frontend/src/components/SettingsPanel.jsx:658` — add `id`/`htmlFor` and `aria-valuetext`; it currently announces as an unnamed slider.
7. Make `role="status"` unconditional for `agent_status` rows at `frontend/src/components/Message.jsx:548` — only the contentless start marker is announced, while rows carrying real text such as "Agent Max Steps Reached" are not.

## Evidence
Evidence: 8/10. The change is well demonstrated for its size — new frontend suites per derived branch, backend coverage of the kill switch and the step coercion matrix, a PR validation script, and live manual verification; the notable gaps are a before/after crop of the transcript row that is the PR's headline change, and a suite re-run pinned to the head commit rather than `39fed511`.

## Notes
- `agent_mode_available()` allows a settings object missing `feature_agent_mode_available` (`atlas/modules/config/settings.py:118`) — defaulting that lookup to `False` would make the helper fail-closed on both shapes.
- The module function `agent_mode_available()` shadows the `AppSettings.agent_mode_available` property with different semantics (`atlas/modules/config/settings.py:109`); renaming one would remove the ambiguity at call sites.
- `chatui-agent-max-steps` remains a second, unclamped step budget that is transmitted whenever `maxIterations` is absent (`frontend/src/hooks/chat/useAgentMode.js:8`).

---

garland3:
Two housekeeping notes while the round-4 fixes run CI:

- **Ceiling confirmation vs. omitted fields (new P2)** -- adopted the "preserve the last confirmed limit" option: `applyConfig` now only touches `agentMaxStepsLimit`/`agentCeilingConfirmed` when a response carries a *valid* `agent_max_steps`; an absent or invalid field leaves both the last confirmed ceiling and the confirmation state untouched (and before any confirmation, the pre-config fallback of 10 stands for display only). A mixed-version pod omitting the field can therefore no longer reset a confirmed 30 to the fallback 10. The "stays unconfirmed when the live responses omit the field" case in `config-cache-hydration.test.js` still pins the never-confirmed path; pushing this with the next commit.
- **Merge from main** -- main moved (#931, #933) and #933 also touched `ChatContext.jsx`, so the branch showed CONFLICTING. Merged `origin/main` in; `ChatContext.jsx` auto-merged (transcript-export additions and the payload gate coexist), `CHANGELOG.md` resolved with both entries. Full frontend suite (908 tests) and the agent/backend routing suites pass locally on the merge commit; CI is running on 10c4ffa5.

garland3:
@codex review

garland3:
Addressed the fifth review round in de64d357:

**Stop/send controls follow the entered turn mode.** The composer's stop-vs-send choice at `ChatArea` keyed on the persisted `agentModeEnabled` toggle: with agent mode defaulting on, a no-tools turn (empty workspace restore) runs as a plain chat while the toggle stays true, so Send remained enabled during generation and Enter (which had no in-flight guard at all) could start a second untracked task against the same session. Both paths now key on `isAgentRunning` -- set at send time to the mode the turn will really run as (`agentModeAvailable && agentModeEnabled && tools non-empty`):

- A plain turn in flight shows Stop-streaming instead of Send, and `handleSubmit` refuses a second submission with a toast (mirroring the `rewindAndResubmit` guard).
- An agent run keeps Send up so the next message steers the loop (#824), and the agent Stop button still appears via `isAgentRunning || agentPendingQuestion`.
- Flipping the toggle mid-run no longer changes the controls: the entered mode governs.

Covered by the new `composer-stop-controls.test.jsx`: stop-not-send on a plain turn with the toggle on, Send + agent stop during an agent run, return to Send after completion, and Enter blocked mid-plain-turn.

garland3:
@codex review

garland3:
Addressed the sixth review round in a75a5f8e:

1. **Steering sends armed by server acknowledgement (P1)** -- `isAgentRunning` is no longer predicted at send time from the persisted toggle. It is set when the server sends `agent_start` (the acknowledgement that a real agent loop began) and continues to be cleared on every terminal event (`agent_completion`, `agent_error`, `agent_max_steps`, `response_complete`, `chat_response`, `error`). A turn the backend downgrades -- kill switch, no tools, non-tool model, stale cached availability -- never receives `agent_start`, so `isAgentRunning` stays false and the plain-turn guard blocks a second submission instead of treating it as steering. Covered by the extended `agent_start` handler tests (arms the flag; tolerant of older deps without the setter) and the existing `composer-stop-controls.test.jsx` states.

2. **Stored preference retained unclamped (P2)** -- took the "retain the unclamped preference separately" option: the clamp is now applied only where the value is consumed (the panel draft, the value chip, the slider, and the send payload in ChatContext once the ceiling is confirmed). Nothing derived from a config response is persisted to `chatui-settings` anymore, so neither a first-valid-but-stale shell value (shell 10 before full 30) nor an omitted field can rewrite the stored 30. The server remains the authoritative clamp for the payload. Tests updated: the clamp test now pins "draft + chip clamped, stored preference untouched"; the pre-config and omitted-field cases still assert nothing is persisted.

garland3:
@codex review

PR #933, opened 2026-09-13
Open chat transcript exports in a new browser tab by default

### LINKED ISSUE #908: change the download transcript .json and .txt to open by default in a new browser tab
The most common use case is to get the .txt and copy and paste to some other application. .. so opening in a browser tab for quick view would be better and follows the more common flow.

garland3:
Closes #908

## What changed

The header transcript actions for `.json` and `.txt` now **open the export in a new browser tab** instead of forcing a file download. The issue's use case: the common flow is reading and copy-pasting the `.txt` (or skimming the JSON) in a tab, not saving a file — saving is still available from the opened tab (Ctrl+S) or via Print / Save as PDF.

- **New behavior** (`frontend/src/utils/chatExport.js`): `openBlobInNewTab` opens the export blob via `window.open(url, '_blank')`. If the popup is blocked — `window.open` returns null **or throws** — it falls back to the old anchor-click download (with the same `chat-export-<timestamp>` filename) so the user still gets the transcript either way. The blob URL is revoked after a 60s delay rather than immediately, so the new tab can finish loading it.
- **`ChatContext`**: the export builders are unchanged (same metadata/header/canvas content, `exportVersion` 1.3); the trigger is now open-in-tab. `downloadChat`/`downloadChatAsText` became `openChat`/`openChatAsText` so the context API says what it does.
- **Header UI** (desktop dropdown + mobile hamburger menu): items renamed `Download as JSON` / `Download as Text` → `Open as JSON` / `Open as Text`, with an external-link icon and an "Opens in a new browser tab" tooltip. The dropdown toggle got a document icon, an "Open Chat Transcript" tooltip, and `aria-haspopup="menu"` / `aria-expanded`; the dropdown is `role="menu"` with `role="menuitem"` items. Print / Save as PDF is unchanged. The empty-chat disabled state and the dropdown's click-outside behavior are unchanged. **One deliberate copy change**: the empty-chat alert text changed from `No chat history to download` to `No chat history to open`, matching the new action.

Out of scope, unchanged: the all-conversations export in the sidebar (still a download) and the browser-side file downloads from MCP tools.

## Screenshots (checked in under `test/pr-validation/assets/pr933-screenshots/`)

- `pr933-dropdown.png` — the renamed dropdown (Open as JSON / Open as Text / Print / Save as PDF)
- `pr933-transcript-txt-tab.png` — "Open as Text" → new tab with the transcript
- `pr933-transcript-json-tab.png` — "Open as JSON" → new tab with valid JSON

## Verification

- `cd frontend && npx vitest run` — 876+ tests pass (including the new ones below); `npm run lint` — 0 errors (4 pre-existing warnings in untouched files); `npm run build` — succeeds.
- New unit tests:
  - `chatExport.test.js`: `openBlobInNewTab` opens a new tab on the blob URL and does not revoke immediately; falls back to a download anchor with the given filename when `window.open` returns null **or throws**; revokes the URL after the 60s delay (fake timers for the whole describe, so no revoke timer outlives a test), not before.
  - `header-transcript-open.test.jsx`: the desktop dropdown and mobile menu offer the open-as labels (old download labels gone), route to the context openers, carry the new-tab tooltip, and expose the menu ARIA semantics.
- New validation script: `bash test/pr-validation/test_pr933_transcript_opens_in_new_tab.sh` (label/source checks, the two vitest suites, production-bundle label check, full frontend suite; screenshots above came from a live `agent_start.sh` + `mocks/llm-mock` session).
- Manual end-to-end (real browser, mock LLM): sent a message, then:
  - Desktop dropdown → **Open as Text** → new tab renders the transcript (metadata header + `USER:` / `ASSISTANT:` blocks).
  - Desktop dropdown → **Open as JSON** → new tab renders valid JSON (`metadata`, `conversation`, `canvasContent`).
  - Mobile hamburger → **Open as Text** → new tab opens the same way.
  - No download dialog in any case.

## Known browser caveats (review follow-ups documented)

- Browsers without a JSON viewer (e.g. Safari, Firefox with the viewer disabled) may render the JSON blob as raw text or download it; the text export is the copy/paste path the issue asks about and renders everywhere.
- Ctrl+S from a blob tab saves under a browser-generated name (no `download` attribute is set for the tab view). If an explicit "Download" menu item is wanted alongside the open-as items, that is a small follow-up.

## Notes for review

- The issue title says "open by default in a new browser tab"; I read that as: the transcript actions open a tab, and saving stays possible from the tab (Ctrl+S / Print). The popup-blocked fallback preserves the old download path.
- Blob URLs are revoked 60s after opening (they must survive the new tab's load); a stray long-open tab keeps its URL alive no longer than that.


garland3:
## Summary
Changes the header chat-transcript export from a forced file download to opening the transcript in a new browser tab via a new `openBlobInNewTab` helper in `frontend/src/utils/chatExport.js`, with an anchor-download fallback when the popup is blocked.

## Verdict
Approve with nits.

## Evidence
Evidence: 7/10. The change is well described and unit-tested for both the tab and popup-blocked branches; the main gap for a user-facing UI change is that no screenshot of the rendered transcript tab or the renamed dropdown items is attached, and the fallback path was never exercised in a real browser.

## Notes
- "Open as JSON" may still download instead of rendering on browsers without a JSON viewer (Safari, Firefox with the viewer off), and without `a.download` the file gets an extension-less UUID name — `frontend/src/contexts/ChatContext.jsx:1022`.
- The popup-blocked fallback is silent: no toast, no log, and a thrown `window.open` isn't caught, so the user can see nothing happen — `frontend/src/utils/chatExport.js:24`.
- The transcript dropdown lacks `aria-haspopup`/`aria-expanded`/`role="menu"` and the "opens in a new browser tab" hint lives only in a `title`, invisible to keyboard and touch users — `frontend/src/components/Header.jsx:258`.

---

garland3:
Addressed the review feedback in 18502f4f:

**Copilot**
- Lingering test timers: the `openBlobInNewTab` describe now runs under fake timers throughout, so no 60s revoke timer outlives a test or fires after the mocks are restored.
- Empty-chat alert copy: the PR description now documents the deliberate change (`No chat history to download` → `No chat history to open`); I kept the new wording since it matches the action.

**AGENT-REVIEW-BOT-3 nits**
- ARIA: the toggle now carries `aria-haspopup="menu"` + `aria-expanded`, the dropdown is `role="menu"` with `role="menuitem"` items, locked in by a test.
- `window.open` throwing (not just returning null) now also falls back to the download, covered by a test.
- Browser caveat for JSON-in-tab (Safari / no-JSON-viewer) is documented in the PR description — kept the issue-requested behavior for both formats. A visible hint beyond the `title` tooltip: happy to add a visible "(new tab)" suffix to the labels if reviewers prefer that over the tooltip.
- Evidence gap: added `test/pr-validation/test_pr933_transcript_opens_in_new_tab.sh` plus three browser-verified screenshots (dropdown, .txt tab, .json tab) checked in under `test/pr-validation/assets/pr933-screenshots/`, and the PR description now links them. The popup-blocked fallback is unit-tested (real-browser popup blocking isn't reachable in the MCP-driven session, so it stays at unit level).

**Codex**: no actionable suggestions in the review body.

CI was already green on the first push; this push re-runs the suite.

garland3:
## Summary
Changes the chat transcript export flow to open the transcript in a new browser tab (as JSON or text) via a new `openBlobInNewTab` helper, with an anchor-download fallback when popups are blocked, plus Header menu label/icon/ARIA updates.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — the change is well-demonstrated for its size (screenshots, two vitest suites covering the popup-blocked and throw branches, a validation script, CHANGELOG). The main missing artifact is any real-browser confirmation of the popup-blocked fallback; the mobile hamburger flow also has no screenshot.

## Notes
- Set `;charset=utf-8` on the text/JSON blobs — without it the navigated tab may decode non-ASCII transcripts as mojibake (`frontend/src/contexts/ChatContext.jsx:1000`).
- The dropdown declares `role="menu"`/`menuitem` but implements no arrow-key/Escape menu contract; either implement it or drop the menu roles (`frontend/src/components/Header.jsx:275`).
- The popup-blocked fallback downloads silently with no user-visible signal, and the anchor sequence is still hand-rolled in three other call sites that could use the new helper (`frontend/src/utils/chatExport.js:22`).

---

PR #934, opened 2026-09-13
Arm steering sends from the server's agent_start; retain the stored agent-steps preference

garland3:
Follow-up to #932 (merged as c9fded95). The squash merge landed the branch at de64d357; this carries the two commits that addressed the final review round but were pushed a few minutes after the merge and so never shipped:

1. **Steering sends are armed by the server, not predicted** (P1, round 6 review): `isAgentRunning` is set when the backend sends `agent_start` -- the acknowledgement that a real agent loop began -- and cleared on every terminal event as before. At send time it is reset to false. A turn the backend downgrades (kill switch, no tools selected, non-tool model, stale cached availability) never receives `agent_start`, so the composer's plain-turn guard keeps blocking a second submission instead of treating it as steering; previously the client-predicted flag enabled steering sends while `atlas/main.py` would have started a second untracked task against the same session.
2. **The stored max-iterations preference is retained unclamped** (P2, round 6 review): the ceiling clamp is applied only where the value is consumed (panel draft, value chip, slider, and the send payload once a live response confirmed the ceiling). Nothing derived from a config response is persisted to `chatui-settings` anymore, so a stale-cache value or pods answering with different `AGENT_MAX_STEPS` values during a rolling deploy can no longer rewrite a saved preference; the server remains the authoritative payload clamp.

## Verification

- `npx vitest run`: 914 tests / 79 files pass; `npm run lint` 0 errors (4 pre-existing warnings, none in touched files).
- New/updated tests: `agent_start` handler arms the run flag (and tolerates older deps without the setter); the SettingsPanel clamp test now pins "draft + chip clamped, stored preference untouched"; `composer-stop-controls.test.jsx` covers the entered-mode controls.

Closes the two open review findings from the #932 review thread (round 6, chatgpt-codex-connector).


garland3:
@codex review

garland3:
## Summary
Moves `isAgentRunning` from a client-side prediction at send time to being armed by the server's `agent_start` event, and stops the settings panel from writing the config-derived step ceiling back into the persisted preference.

## Verdict
Request changes.

## Required changes
1. Stop clearing the agent-run flag on a steering send at `frontend/src/contexts/ChatContext.jsx:761` — a message sent into a live agent loop gets no second `agent_start`, so the flag stays false for the rest of the run, hiding the agent Stop button and causing further steering sends to be rejected with "Wait for the current response to finish before sending." Skip the reset when `isAgentRunning` is already true (or have the server acknowledge the injection). Add a regression test: `agent_start` → steering send → flag still true.
2. Remove the draft clamp at `frontend/src/components/SettingsPanel.jsx:300-303` and clamp only at the consume sites (lines 653/670/671 already do) — the clamped draft is written verbatim by `saveSettings`, so a stored `maxIterations` of 50 under a reported ceiling of 30 is silently overwritten by the next Save or by the debug-mode toggle at line 751, which is the exact data loss this PR set out to prevent.
3. Extend `frontend/src/test/settings-panel-agent-mode.test.jsx:136` to save an unrelated setting after the ceiling is confirmed and assert the stored `maxIterations` is still 50; the current test only pins the absence of a write at mount.
4. Add a test that drives a real `sendChatMessage` (e.g. in `frontend/src/test/agent-mode-payload-gating.test.jsx`) and asserts the flag is false after send and true only after `agent_start` — reverting the one-line change at `ChatContext.jsx:761` currently leaves the whole suite green.

## Evidence
Evidence: 8/10 — proportionate for a 6-file, +52/-34 change, with new unit tests mapping onto the stated claims and a full-suite run reported. The main gap is that nothing exercises a downgraded turn end to end (send → streaming → completion with no `agent_start`), which is the scenario the fix exists for.

## Notes
- Composer layout reflows when `agent_start` lands — the stop control moves and Send appears under the pointer mid-turn (`frontend/src/components/ChatArea.jsx:1145`).
- A downgraded agent turn is silent to the user while the Agent toggle still reads enabled (`frontend/src/handlers/chat/websocketHandlers.js:144`).
- The ceiling clamp is now open-coded at five sites with two independent `|| 10` fallbacks; a shared helper would keep the policy in one place (`SettingsPanel.jsx:302`, `ChatContext.jsx:681-683`).

---

garland3:
Addressed the round-7 findings: sendChatMessage no longer touches the agent-run flag at all (steering sends keep it armed -- the server acks steering with `steering_queued`, not `agent_start`), and the SettingsPanel draft-clamp effect is removed outright, so a Save of unrelated settings can never persist a config-derived value; chip/slider keep clamping at their consumption points. Suite: 914 tests / 79 files, lint 0 errors.

garland3:
@codex review

garland3:
## Summary
Moves the `isAgentRunning` flag from a client-side prediction at send time to the server's `agent_start` acknowledgement, and removes the SettingsPanel effect that rewrote the stored `maxIterations` down to the deployment ceiling, clamping only at display and payload sites instead.

## Verdict
Request changes.

## Required changes
1. Reset `isAgentRunning` at view-reset boundaries — add `setIsAgentRunning(false)` to `clearChat` and `loadSavedConversation` in `frontend/src/contexts/ChatContext.jsx:840` — with the send path no longer recomputing it and terminal events dropped for non-visible conversations, the flag sticks after New Chat, leaving Send enabled during the next plain turn.
2. Gate the step-slider clamp on `agentCeilingConfirmed` at `frontend/src/components/SettingsPanel.jsx:648,663-666` — before config arrives `maxStepsLimit` falls back to 10, so a drag plus save silently overwrites a stored preference of 50, the exact data loss this PR removes the effect to prevent.
3. Add a test driving a real `sendChatMessage` after `agent_start` asserting the flag stays armed, and a second asserting it stays false for a send with no `agent_start` (`frontend/src/test/agent-mode-payload-gating.test.jsx` already has the provider harness) — restoring the deleted setter currently leaves all 914 tests green.
4. Extend `frontend/src/test/settings-panel-agent-mode.test.jsx:121` to toggle one unrelated setting after the ceiling is confirmed and re-assert stored `maxIterations` is still 50 — the current test only pins the mount-time write.
5. Make `handleReset` at `frontend/src/components/SettingsPanel.jsx:317-327` reset to `defaultSettings` verbatim instead of persisting `Math.min(defaultSettings.maxIterations, maxStepsLimit)` — it is the one remaining path that writes a config-derived value to localStorage.

## Evidence
Evidence: 7/10 — the additive half is backed by unit tests and a full-suite run, but both behaviours this PR exists to protect are deletions, and a deletion is only evidenced by a test that fails when the line returns; the most important missing artifact is a real-provider test around `sendChatMessage` (covered above).

## Notes
- Stop control changes identity between send and `agent_start`, so a cancel click can land on Send — `frontend/src/components/ChatArea.jsx:1145,1239`.
- No indication to the user that a clamped chip ("30 / 30") differs from their saved preference — `frontend/src/components/SettingsPanel.jsx:648`.
- Clamp policy (`|| 10` vs the normalized `maxStepsLimit`) is re-derived independently in `ChatContext.jsx:682`; extract one shared helper.

---

PR #936, opened 2026-09-13
Rebuild cached per-user MCP clients when the OAuth token rotates (#935)

### LINKED ISSUE #935: Cached per-user MCP client keeps a rotated OAuth token, causing persistent 401s
## Summary

A cached per-user FastMCP client keeps the access token it was built with. When that token is later rotated by a silent OAuth refresh, the cached client is **not** invalidated, so it keeps presenting a token the provider has already retired. Every subsequent tool call on that client fails with `401 Unauthorized` until the cache entry ages out (idle TTL, default 1h) or the user re-authorizes.

Observed with a remote OAuth MCP server that issues ~1h access tokens and rotates tokens on refresh: the server connects, a few tool calls succeed, and then every later call in the same conversation returns 401.

## Root cause

`_get_user_client` (`atlas/modules/mcp_tools/mcp_user_clients.py`) validates a cached entry with:

```python
stored_token = token_storage.get_valid_token(user_email, server_name)
subtoken_unchanged = self._wormhole_client_subtokens.get(cache_key) == current_subtoken
if stored_token is not None and subtoken_unchanged:
    return self._user_clients[cache_key]
```

That asks "is there *a* valid token?", never "is it the *same* token this client was built with?". The token value is baked into the `Client`/transport at construction and nothing records it per cache entry (`_pop_user_client_entries_locked` in `mcp_user_client_cache.py` tracks only last-used time and the Wormhole subtoken).

The cache key is `(user, server, conversation_id)`, and refreshes are per `(user, server)`, so a refresh driven by one cache entry silently invalidates the credential inside all the others:

- `mcp_user_discovery` acquires its client with `conversation_id=None` (see the comment at `mcp_user_discovery.py:403`). A discovery/refresh sweep there calls `mcp_oauth_service.refresh_stored_token`, rotates the token, and leaves every conversation's cached client holding the retired one.
- Same effect across two conversations for the same user and server.

`refresh_stored_token` / `_refresh_locked` (`mcp_oauth_service.py`) store the new token but never invalidate cached clients. Re-authorization does do this correctly — `atlas/routes/mcp_auth_routes.py` calls `_invalidate_user_client` — which is why "disconnect and reconnect" appears to fix it.

Secondary issue: there is no 401-triggered refresh-and-retry anywhere in tool execution (`grep` for `401` finds hits only in `mcp_oauth.py`'s discovery challenge). So nothing self-heals; the raw transport error, including the upstream endpoint URL, is handed straight back to the model as the tool result.

## Reproduce

1. Configure an `auth_type: oauth` HTTP MCP server whose provider rotates/retires the access token on refresh, with a short access-token lifetime.
2. Authorize it and make a tool call in a conversation (client gets cached).
3. Cause a refresh from a different cache entry — wait for the user-discovery sweep, or make a call from a second conversation after the access token expires.
4. Call a tool again in the first conversation: `401 Unauthorized`, and it keeps failing.

## Suggested fix

1. Record the token value (or a hash/fingerprint of it) alongside each cache entry, and treat a mismatch the way the rotated Wormhole subtoken is already treated — rebuild the client. This is the minimal fix and mirrors existing code.
2. Have `refresh_stored_token` invalidate every cached client for `(user, server)` across all `conversation_id`s, as the re-auth route already does.
3. Add a single 401-driven retry in tool execution: on `401`, refresh, rebuild the client, retry once, and only then surface a friendly "please reconnect" message rather than the raw upstream error/URL.

(1) and (3) are somewhat independent — (1) fixes the common case, (3) covers provider-side revocation where the stored token still looks valid by the clock.

garland3:
Closes #935

## Problem

A cached per-user FastMCP client keeps the access token it was built with. When that token is later rotated by a silent OAuth refresh, the cached client is not invalidated, so it keeps presenting a token the provider has already retired. Every subsequent tool call on that client fails with `401 Unauthorized` until the cache entry ages out (idle TTL, default 1h) or the user re-authorizes.

The cache key is `(user, server, conversation_id)` while refreshes are per `(user, server)`, so a refresh driven by one cache entry (e.g. the conversationless discovery sweep) silently invalidates the credential inside all the others.

## Fix

1. **Rotation-aware client cache** (`mcp_user_clients.py`, `mcp_user_client_cache.py`): every cache entry records a fingerprint (short SHA-256 digest) of the token value its client was built with. `_get_user_client` now requires the stored credential to be the *same one the client was built with*, in addition to the existing validity and Wormhole-subtoken checks; a mismatch rebuilds the client. The token value itself is never stored, only the digest.

2. **Refresh invalidates clients** (`mcp_oauth_service.py`): after a successful `_refresh_locked`, `refresh_stored_token` evicts every cached client for `(user, server)` across all conversations via `_invalidate_user_client` -- the same thing the re-authorization route already does -- so no conversation keeps a client built on the retired token.

3. **401 self-healing** (`mcp_execution.py`, `mcp_errors.py`): `call_tool` now detects an HTTP 401 from the transport (`httpx.HTTPStatusError`, walked through the exception chain since FastMCP may re-wrap it) on a per-user-auth server and, once, evicts the cached clients for `(user, server)`, attempts a **forced** token refresh (new `force=True` path on `refresh_stored_token` that contacts the provider even when the stored token still looks valid by the clock -- it has been retired server-side regardless of `expires_at`), and retries the call once. If the retry is refused too, it raises `AuthenticationRequiredException` with a friendly "re-authorized/reconnected" message and, for OAuth servers, the reconnect start URL; `execute_tool` forwards it so tool_executor emits the `auth_required` UI event instead of stringifying a raw transport error/URL into the tool result.

## Verification

- New unit tests: `atlas/tests/test_mcp_tool_401_retry.py` (retry/recovery/friendly-error/refresh-invalidation/forced-refresh) and new rotation cases in `atlas/tests/test_mcp_client_auth.py`.
- End-to-end PR validation (`bash test/run_pr_validation.sh 936`): a real FastMCP streamable-HTTP server behind a bearer gate plus a real localhost OAuth provider that rotates refresh tokens and retires the previous access token on every refresh, driven through the production `MCPToolManager` / `mcp_oauth_service` paths. Covers: silent rotation evicts all conversation clients and the next call succeeds without any 401; provider-side revocation with a valid-by-the-clock token takes exactly one 401 then self-heals; a refused refresh surfaces the friendly error with no upstream URL. 16/16 checks pass.
- `./test/run_tests.sh backend`: full backend suite green.
- `ruff check atlas/` clean for touched files (one pre-existing import-order error in `client.py` auto-fixed on the way).

## Docs

- `docs/admin/mcp-server-authentication.md`: documents the rotation-aware client cache and the 401 self-healing behavior.
- `CHANGELOG.md` entry added.


garland3:
## Summary
Makes per-user MCP client caching rotation-aware (token fingerprint check) and adds a one-shot 401 self-heal path that forces an OAuth refresh, evicts the cached client, and retries the tool call, surfacing an `AuthenticationRequiredException` when recovery fails.

## Verdict
Request changes.

## Required changes
1. Emit a terminal tool event when the 401 path gives up — `atlas/modules/mcp_tools/mcp_execution.py:392` raises `AuthenticationRequiredException`, and `tool_executor` emits only `auth_required`, which `frontend/src/handlers/chat/websocketHandlers.js:716` does not handle, so the tool row spins in `calling` forever. Either add a `case 'auth_required'` handler or also emit `tool_error`.
2. Stop tearing down in-flight clients on routine refreshes at `atlas/modules/mcp_tools/mcp_oauth_service.py:409` — `_invalidate_user_client` closes clients and releases sessions with no in-use check, so a background refresh can kill a streaming tool call on another conversation. Drop the cache entry only (let the fingerprint check rebuild), or honour `_user_client_cache_in_use_window_seconds` as the LRU enforcer does.
3. De-duplicate forced refreshes in `atlas/modules/mcp_tools/mcp_oauth_service.py:383` — `force=True` skips the post-lock short-circuit, so N concurrent 401s cause N serialized provider rotations; against a rotating provider, refresh *k+1* invalidates the credential caller *k* just retried with, and a 401-returning server can drive this once per tool call. Pass the failing token's fingerprint and return early when the stored token no longer matches, or add a short per-`(user, server)` forced-refresh suppression window.
4. Hoist the text fallback out of the `isinstance(cur, httpx.HTTPStatusError)` block at `atlas/modules/mcp_tools/mcp_errors.py:82` — as written it can only fire for an `HTTPStatusError` that is not 401, so a 401 re-wrapped by FastMCP as a non-httpx exception never triggers any of the new recovery. Match on the message for any exception type, as `_is_session_terminated_error` does.
5. Move the `_invalidate_cached_clients` await outside the `_refresh_locks` critical section at `atlas/modules/mcp_tools/mcp_oauth_service.py:409` — serial client closes bounded by a 5s timeout each can hold the refresh lock for tens of seconds against an unresponsive upstream, blocking every other tool call for that user/server.
6. Skip the retry for auth types that cannot self-heal at `atlas/modules/mcp_tools/mcp_execution.py:423` — `_refresh_oauth_token` returns `None` for `bearer`/`api_key`/`jwt`, so the rebuilt client reuses the same credential and is guaranteed to 401 again, while the message claims a refresh that never happened and tells the user to "reconnect" with no flow to reconnect through.
7. Add unit coverage for the chain-walking and non-OAuth branches of `_is_unauthorized_error` / `_recover_user_client_after_unauthorized` (`atlas/modules/mcp_tools/mcp_errors.py:82`, `mcp_execution.py:355`) — both are currently exercised only through the direct `httpx` 401 shape.

## Evidence
Evidence: 9/10 — the change is well-demonstrated for its size, with per-behavior unit tests and an end-to-end harness driving a real rotating-token OAuth provider through the production code paths. The one meaningful gap is a scenario firing concurrent `call_tool` invocations against a revoked token and asserting exactly one new token is issued, which is also what would pin the de-duplication fix above.

## Notes
- Rotation check fails open when a stored token has an empty value: `None == None` reads as "unchanged" — `atlas/modules/mcp_tools/mcp_user_clients.py:30`.
- `execute_tool_calls` now propagates `AuthenticationRequiredException` out of a `-> List[ToolResult]` API, aborting the whole batch — `atlas/modules/mcp_tools/mcp_execution.py:907`.
- Four slightly divergent copies of the "evict cached user clients, quietly" helper now exist; consider one shared function — `atlas/modules/mcp_tools/mcp_oauth_service.py:400`.

---

garland3:
## Summary
Fixes cached MCP OAuth clients surviving a token rotation: adds 401 detection with a forced, fingerprint-deduplicated token refresh, scoped per-user client cache eviction that spares recently-used entries, and a `tool_error` event so the chat UI's tool row settles after an exhausted retry.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size, with per-branch unit tests plus a live end-to-end harness driving real rotation through production paths. The one gap worth closing later: nothing confirms the UI tool row actually leaves the `calling` state after the new `tool_error` emission (`atlas/application/chat/utilities/tool_executor.py:826`) — a handler-level assertion or a noted manual observation would cover it.

## Notes
- Substring match on `"401 unauthorized"` lets server-controlled error text force a token rotation and retry — consider gating on status code or a tighter pattern, `atlas/modules/mcp_tools/mcp_errors.py:85`.
- The 401 retry re-issues the tool call without idempotency consideration, and for `bearer`/`api_key`/`jwt` auth types it costs a round-trip that cannot succeed; short-circuit those to the auth-required message, `atlas/modules/mcp_tools/mcp_execution.py:355`.
- Cache invalidation resolves its collaborator through the app-factory global and picks a private method by `getattr` name, silently no-oping if renamed — prefer an explicit public method passed in, `atlas/modules/mcp_tools/mcp_oauth_service.py:437`.

---

garland3:
Addressed the review findings in 22c81e9f:

1. **Terminal tool event** — tool_executor now emits a terminal `tool_error` right after `auth_required` (the live UI has no `auth_required` case; the persistence recorder dedupes both onto one failed row). Covered by `test_auth_required_followed_by_tool_error_stays_one_row`.

2. **In-flight teardown on routine refreshes** — refresh-driven eviction now goes through a new `_invalidate_user_clients_for_rotation`: idle entries (beyond the same in-use window the LRU enforcer uses) are closed immediately; entries possibly in use stay cached and are rebuilt by the fingerprint check on their next acquisition. A background refresh can no longer kill a streaming call elsewhere. Covered by the reworked e2e ("in-use cache entries survive the refresh") and unit test (`test_refresh_rotates_token_and_invalidates_all_conversations`).

3. **Forced-refresh dedup** — `refresh_stored_token(force=True, expected_previous_fingerprint=...)`: concurrent 401s pass the failing token's fingerprint; after acquiring the lock, a mismatch means someone already rotated and the stored credential is returned without contacting the provider. New test fires two concurrent forced refreshes and asserts exactly one provider rotation.

4. **Text fallback hoisted** — `_is_unauthorized_error` now matches the canonical "401 Unauthorized" text on any exception type in the chain (mirroring `_is_session_terminated_error`), so a 401 re-wrapped by FastMCP as a non-httpx exception still triggers recovery. Five direct unit tests added.

5. **Eviction outside the refresh lock** — `_invalidate_cached_clients` is awaited after the lock is released.

6. **Honest messages per auth type** — OAuth servers get the refresh-and-retry wording plus the percent-encoded reconnect URL (`quote(server_name, safe='')`, matching the auth-status route); `bearer`/`api_key`/`jwt` say a refreshed credential must be provided and get no OAuth URL. The retry itself is kept for those types: it is the path that heals an admin-side credential re-upload, and it costs one call when the credential is unchanged.

7. **New coverage** — chain-walking/text matching for `_is_unauthorized_error`, the non-OAuth recovery branch (honest message, no URL), `execute_tool_calls` converting the exception to a failed `ToolResult` (batch `List[ToolResult]` contract preserved), and the persistence row dedup.

Notes addressed as well: the fingerprint check now fails closed (`None == None` can no longer read as "unchanged"), and `_invalidate_cached_clients` accepts the owning manager — a programmatic `AtlasClient` refresh evicts its own manager's caches (the web singleton remains the fallback for the module-level service).

Deliberately not changed: the "second, non-forced refresh after a refused forced refresh" stays as-is — it is a single extra *refused* provider call (the provider already rejected that refresh token, so nothing rotates), and suppressing it would mean threading recovery state through `_get_user_client`'s signature, which discovery and several callers share. It is documented in `_recover_user_client_after_unauthorized`. The four "evict quietly" helpers were also left as-is; consolidating them across `mcp_delegation`/`mcp_auth_routes`/`mcp_oauth_service` is a larger refactor than this PR warrants (noted for follow-up).

garland3:
## Summary
Fixes stale cached MCP clients after an OAuth token rotation (#935) by fingerprinting the token each cached client was built with, evicting idle clients on refresh, adding a forced-refresh-and-retry path on transport 401s, and emitting a terminal `tool_error` so an auth-failed tool row stops spinning.

## Verdict
Request changes.

## Required changes
1. Tighten the 401 text fallback at `atlas/modules/mcp_tools/mcp_errors.py:85` — match only the transport wording (e.g. anchored `Client error '401 Unauthorized' for url`) or require an `httpx.HTTPStatusError` with `status_code == 401`; today any exception whose message merely contains `401 unauthorized`, including remote-server-controlled tool error text, forces a refresh-token rotation and a tool replay.
2. Avoid re-executing a possibly-completed tool at `atlas/modules/mcp_tools/mcp_execution.py:355` — retry only when the 401 is attributable to client construction/session init (or when the stored credential fingerprint actually changed after the refresh), so a tool whose body already ran is not invoked twice.
3. Take `expected_previous_fingerprint` from the cache entry, not from storage, at `atlas/modules/mcp_tools/mcp_execution.py:463` — read `_user_client_token_fingerprints[(user, server, conversation_id)]` under the lock before eviction, falling back to the stored fingerprint; reading the store defeats the dedup in exactly the concurrent-rotation race it exists for.
4. Add a per-`(user, server)` cooldown on forced refreshes near the existing `_refresh_locks` bookkeeping (`atlas/modules/mcp_tools/mcp_execution.py:412`) — the fingerprint dedup collapses only concurrent 401s, so a persistently-401ing server yields one provider rotation per tool call, which can get the grant rate-limited or revoked.
5. Make `expected_previous_fingerprint` without `force=True` an explicit error (or apply the check unconditionally) at `atlas/modules/mcp_tools/mcp_oauth_service.py:354` — the argument is currently ignored in half its call space with no warning.
6. Add a test that drives `execute_single_tool` with a tool manager raising `AuthenticationRequiredException` and asserts the `update_callback` sees both `auth_required` and a subsequent terminal `tool_error` for the same `tool_call_id` (`atlas/application/chat/utilities/tool_executor.py:831`) — deleting the new emission currently leaves the suite green.
7. Replace the name-based lookup in `_invalidate_cached_clients` (`atlas/modules/mcp_tools/mcp_oauth_service.py:437`) with a declared public method on the manager, so a rename is a hard error rather than a silent no-op that reintroduces this bug; log eviction failures at warning, not debug.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size, with per-branch unit tests plus an end-to-end harness running a real rotating OAuth provider against production code paths. The one gap worth closing is confirmation that the chat tool row actually settles out of `calling` after an exhausted 401 (a frontend handler test or one observed screenshot).

## Notes
- Unify OAuth start-URL construction; the new path percent-encodes the server name while `mcp_execution.py:531` and `:744` interpolate it raw, so at most one form resolves.
- `delegated` servers get the non-OAuth exhausted-401 wording asking for a credential nobody can supply (`atlas/modules/mcp_tools/mcp_execution.py:410`).
- `test_fingerprint_tracks_rotations` compares against `.get()` of a never-inserted key, i.e. `None` — the assertion is tautological (`atlas/tests/test_mcp_client_auth.py:672`).

---

garland3:
Follow-ups on the inline findings (ce26c4a9):

- **Retry with no buildable client** (mcp_execution.py:388): fixed. When the forced refresh is refused and the stored token is expired, the retry's client build fails with the generic "requires authentication" signal; both retry failure shapes now produce the same auth-type-aware reconnect error through a shared `_unauthorized_recovery_error` helper, so the model always hears the 401-specific message. Pinned by `test_retry_without_buildable_client_gets_401_specific_message`.
- **Log injection (mcp_user_clients.py:505)**: the rotation-eviction debug log now goes through `sanitize_for_logging`.
- **Unused `_metadata` helper in the test** (CodeQL): removed.
- **Publish race in `_get_user_client`** (mcp_user_clients.py:415): noted, not changed. The window is: build client(token X) outside the lock → concurrent refresh publishes token Y → insert client(X). The current call would 401 with the retired token and the 401 self-heal recovers it (evict + forced refresh with the failing fingerprint + one retry), and the next acquisition fingerprint-checks and rebuilds. Closing this fully would require re-validating inside the insert lock and rebuilding under the lock; the self-heal path bounds the user-visible impact, so I left the simpler structure.
- **Pop-then-close race** (mcp_oauth_service.py): the refresh-driven path no longer uses `release_sessions_for_user_server` at all (rotation-aware eviction closes per-entry); the flagged shape is the pre-existing `_invalidate_user_client` used by the re-auth route, which this PR did not change.
- **Concurrent forced refreshes**: already addressed by `expected_previous_fingerprint` in 22c81e9f (with a test asserting one provider rotation for two concurrent 401s); the inline comment predates that commit.

PR #937, opened 2026-09-14
Fix uploaded vision images across follow-up turns

### LINKED ISSUE #907: Uploaded images reach vision models for only one turn: resubmits and follow-up questions silently lose the image
## Summary

An uploaded image is sent to a vision model **only on the exact turn it was uploaded**. On every subsequent turn — including an edit/resubmit of the same prompt and any follow-up question about the image — the image silently degrades to a plain-text filename in the files manifest. The model then answers as if it can see a file reference but no pixels, which reads to the user as "the model isn't getting my image" even though the model is correctly configured with `supports_vision: true`.

No warning is emitted in this case, so the failure is invisible from the UI.

## Reproduction

1. Upload an image and send a prompt about it (a vision-capable model selected).
2. The model sees the image correctly.
3. Either:
   - edit/resubmit that same message (e.g. after switching models), **or**
   - ask a follow-up question about the image.
4. The model no longer sees the image; it only sees a manifest line `- <filename>.jpg`.

### Observed log evidence

Same 32-char prompt, two turns, same session:

Turn 1 — image uploaded, works:

```
File uploaded successfully: category=uploads, size=94818 bytes, content_type=image/jpeg
Attached 1 image(s) and 0 PDF(s) to user message for model
Streaming LLM call with tools: 2 messages, 132080 chars, 1 tools
```

Turn 2 — same prompt resubmitted after a model switch:

```
Rewind to user message 0: removed 2 message(s), 0 remaining before new prompt
Streaming LLM call with tools: 3 messages, 5784 chars, 1 tools
```

No `File uploaded`, no `Attached 1 image(s)`, and the payload drops from ~132K chars to ~5.8K — the base64 image is gone.

The assistant's reply on turn 2 was, in effect, "I see an attached file: `<filename>.jpg` — please upload it", which is exactly the manifest-only view.

## Root cause

Vision data survives exactly one turn, and nothing re-hydrates it:

1. `atlas/application/chat/utilities/file_processor.py` (`handle_session_files`, ~L204-213) unconditionally strips `image_b64` / `image_mime_type` (and `pdf_b64` / `pdf_mime_type`) from **every** file already in session context, at the start of **every** turn.
2. The only thing that repopulates those keys is `files_map` — the raw base64 the client uploads on that turn.
3. `frontend/src/contexts/ChatContext.jsx` (`rewindAndResubmit`, ~L739) passes `{}` as `extraFiles`, so a resubmit carries no bytes. A normal follow-up message carries none either, since the composer's upload chips are cleared after send.

Once `image_b64` is gone, the `exclude_vision_images` skip in `build_files_manifest` no longer fires for that file, so it falls back into the text manifest as a name-only entry. That fallback is what makes the failure look like "the model got the file but ignored it".

Note this is *not* a config problem — the model in question had `supports_vision: true` and the vision path itself works correctly on the upload turn.

## Suggested fix

In `handle_session_files`, re-hydrate rather than only clearing: when `model_supports_vision` is true, for each existing session file whose stored `content_type` is in `_VISION_IMAGE_MIME_TYPES` but which has no `image_b64`, re-read the bytes from `file_manager` using the stored `key` and repopulate. The bytes are already in object storage, so nothing extra needs to be retained in session memory, and the existing size (`_MAX_VISION_IMAGE_B64_BYTES`) and count (`_MAX_VISION_IMAGES_PER_REQUEST`) caps still apply unchanged.

This fixes both the resubmit case and the follow-up-question case in one place, and keeps the original intent of the clearing code (no unbounded stale state carried in the session dict).

The same reasoning applies to `pdf_b64` for `supports_pdf` models, which has the identical one-turn lifetime.

Open design question worth deciding explicitly: whether a conversation's images should be re-sent on *every* subsequent turn (matching how ChatGPT/Claude behave, at a recurring token cost) or only when the turn plausibly refers to them. The current behavior — one turn only, no warning — is the one case that is clearly wrong.

## Related, separate bug found while tracing

`atlas/modules/file_storage/manager.py` (`get_content_type`, ~L42-58) has no `webp` or `bmp` entry, so those uploads are typed `application/octet-stream`. But:

- the frontend explicitly accepts them (`IMAGE_MIME_TYPES` in `frontend/src/components/ChatArea.jsx`), and
- `_LLM_READY_IMAGE_MIME_TYPES` in `file_processor.py` lists both.

Result: a `.webp` or `.bmp` upload is never sent to a vision model, and because the "this model does not support vision" warning is also gated on `_VISION_IMAGE_MIME_TYPES`, no warning fires either. Adding `'webp': 'image/webp'` and `'bmp': 'image/bmp'` to the map should be sufficient.

garland3:
## Summary
- rehydrate stored vision images from object storage on later turns
- preserve inline image blocks for follow-ups and resubmits without re-uploading bytes
- prioritize current-turn and newest historical images within count and aggregate payload limits
- make stored-image rehydration best-effort when an object is unreadable
- retain transport-attributable 401 recovery, delegated-token reminting, failed-token fingerprinting, and active-call cache protection
- keep regression coverage for multimodal follow-ups, unreadable storage, newest-image selection, and 401 classification

Fixes #907

## Verification
- Focused pytest: 48 passed (`test_vision_image_support.py`, `test_mcp_tool_401_retry.py`)
- Additional focused pytest: 66 passed including per-user client and file-manager tests
- Ruff on changed files: passed
- GitHub CI on the preceding head: all reported checks passed
- Full backend suite was attempted but the isolated worktree lacked the proxy-secret test environment, producing unrelated 503 failures; frontend lint could not start because dependencies are not installed in the worktree

The branch also contains the previously stacked MCP OAuth rotation work from PR #936; whether that work should remain stacked here or be split is the remaining scope decision for the maintainer.


garland3:
## Summary
Rehydrates previously uploaded vision images from object storage so they remain visible to the model on follow-up turns. The branch also carries the MCP OAuth token-rotation fingerprinting and 401 refresh-and-retry work (15 of 17 changed files), which the PR description does not cover.

## Verdict
Request changes.

## Required changes
1. Select the images to rehydrate *before* doing any I/O at `atlas/application/chat/utilities/file_processor.py:206` — the loop currently fetches and normalizes every image in the session on every turn, then discards everything past `_MAX_VISION_IMAGES_PER_REQUEST`, so a long session pays unbounded storage egress and decode cost per message.
2. Apply the per-request image cap on a single exit path in `file_processor.py:216` — it is enforced only inside the `if not files_map or not file_manager or not user_email:` early-return branch, so a turn that includes a new upload can exceed the cap.
3. Fan out the rehydration fetches with a bounded `asyncio.gather` at `file_processor.py:219` — the awaits are serialized inside the loop, so turn latency grows linearly in the number of session images.
4. Capture the exit status immediately in `test/pr-validation/test_pr936_cached_oauth_client_rotation.sh:44` — `print_result $?` reads the status of the preceding `if`, so the "atlas importable" check can never report a failure.
5. Narrow `_is_unauthorized_error` at `atlas/modules/mcp_tools/mcp_errors.py:88` to 401s attributable to the MCP transport — any exception in the chain whose text contains `401 unauthorized` currently drives a forced credential rotation, so server-controlled error text can churn the user's refresh token.
6. Handle the `expected_previous_fingerprint is None` case explicitly at `mcp_execution.py:~470` — when the stored record is missing or empty the de-duplication guard is skipped, so N concurrent 401s each drive a provider rotation and N-1 retries fail.
7. Add a CHANGELOG entry for this PR and correct the description — the only entry added is `### PR #936` describing the OAuth work, and the vision fix this PR is titled for is unrecorded.

## Evidence
Evidence: 5/10. The MCP auth portion has a strong live-provider harness but no recorded pass/fail output, and the vision fix — the change the description actually claims — is demonstrated only by two happy-path unit tests with a mocked file manager; the most important gap is an end-to-end demonstration of a two-turn chat where the follow-up answers from the image uploaded on turn 1, plus tests for the cap, oversize, and failed-fetch branches.

## Notes
- Reconnect instruction reaches the user with no reconnect affordance, since the live UI has no `auth_required` handler — `atlas/application/chat/utilities/tool_executor.py:825`.
- `execute_tool` raises `AuthenticationRequiredException` while `execute_tool_calls` returns a failed `ToolResult` for the same condition — `atlas/modules/mcp_tools/mcp_execution.py:1014`.
- PDF attachments are still dropped on follow-up turns while images are now preserved — `file_processor.py:209`.

---

garland3:
## Summary
This PR rehydrates previously-uploaded session images from object storage so vision models still see them on follow-up turns, and — in the same diff — adds MCP OAuth token-rotation fingerprinting with a 401 detect-and-retry path for per-user-auth servers.

## Verdict
Request changes.

## Required changes
1. Narrow `_is_unauthorized_error` at `atlas/modules/mcp_tools/mcp_errors.py:88` to transport-attributable 401s (keep the `httpx.HTTPStatusError` branch; drop or type-gate the free-text `401 unauthorized` substring match) — today any tool *result* text containing that phrase forces an OAuth rotation and re-invokes the tool a second time, duplicating side effects and churning the user's refresh token.
2. Make rehydration best-effort at `atlas/application/chat/utilities/file_processor.py:190` — wrap the `load_image` body in `try/except Exception` returning `(filename, None)` with a `logger.warning`, and pass `return_exceptions=True` to the `asyncio.gather` — so one deleted or unreadable stored image degrades to "no image" instead of aborting `handle_session_files` for the whole turn.
3. Enforce the aggregate inline budget on the rehydration path at `atlas/application/chat/utilities/file_processor.py:255` — accumulate `len(image_b64)` against `_MAX_TOTAL_INLINE_B64_BYTES` and route the early-return branch through the shared cap/warning block — since that branch currently returns before the budget check and can re-inline ~200 MB on every follow-up message.
4. Compare the fingerprint instead of skipping the guard at `atlas/modules/mcp_tools/mcp_oauth_service.py:402` — the `elif expected_previous_fingerprint is not None` condition means a `None` fingerprint (absent/empty stored record) lets N concurrent 401s each rotate with the provider, retiring the credential other retries are about to use.
5. Thread the failing cache entry's recorded fingerprint into `_recover_user_client_after_unauthorized` at `atlas/modules/mcp_tools/mcp_execution.py:463`, and skip the forced refresh when storage already holds a newer token — the common case (a stale cached client) needs only eviction and a retry, not another provider rotation.
6. Capture the import status once in `test/pr-validation/test_pr936_cached_oauth_client_rotation.sh:44` (`python -c ...; RC=$?` then `print_result $RC`) — `$?` there reflects the preceding `if`, so the "atlas importable" check can never report FAILED.
7. Add coverage for the undemonstrated branches: a stored image whose fetch raises (other images still attached), a session exceeding `_MAX_VISION_IMAGES_PER_REQUEST` with and without new uploads, an oversize image, and a negative `_is_unauthorized_error` case where the phrase arrives in tool-result text.

## Evidence
Evidence: 6/10 — the MCP auth half ships a real end-to-end harness, but its pass/fail output is never recorded, and the vision fix the PR is titled for is demonstrated only by two happy-path unit tests. The single most important missing artifact is a two-turn demonstration (upload on turn 1, image-dependent answer on turn 2) against a real vision model and storage backend.

## Notes
- TIFF/format normalization now re-runs synchronously on the event loop each turn per rehydrated image — consider `asyncio.to_thread` or persisting the converted form at upload (`file_processor.py:199`).
- Dropped rehydration images are silent, unlike the upload path's user-visible warning (`file_processor.py:197`).
- CHANGELOG omits both the rehydration behavior change and the new `webp`/`bmp` image typing (`CHANGELOG.md:9`).

---

garland3:
## Summary
Adds a `_rehydrate_vision_images` helper so images uploaded in earlier turns are re-fetched from storage and re-attached on follow-up messages, and widens the filename-to-MIME map with `webp` and `bmp`.

## Verdict
Request changes.

## Required changes
1. Isolate per-image failures in `_rehydrate_vision_images` at `atlas/application/chat/utilities/file_processor.py:190` — wrap the `load_image` body in `try/except Exception` returning `(filename, None)` and pass `return_exceptions=True` to the `gather`; the call site at line 253 sits outside the enclosing `try`, so one deleted or unreadable stored object fails the entire chat turn.
2. Enforce the aggregate inline budget on the rehydration path at `file_processor.py:251` — `_MAX_TOTAL_INLINE_B64_BYTES` is only consulted inside the `if model_supports_pdf:` block on the upload path, so a follow-up turn can rebuild up to ~200 MB of base64 and be rejected by the provider on every subsequent message.
3. Select rehydration candidates most-recent-first at `file_processor.py:180` — `candidates` slices `session_files_ctx` in insertion order, so past ten images the *newest* upload is the one dropped on the next turn; apply the same ordering to the cap at line 473.
4. Route both rehydration exit paths through one shared cap/budget block instead of the early `return updated_context` at `file_processor.py:253` — today the count cap and payload budget apply to only one of the two paths, and any future guard added below line 462 will silently do the same.
5. Charge this turn's uploads against the payload budget before admitting rehydrated images at `file_processor.py:456` — otherwise a freshly uploaded PDF is demoted to text because of old images the user is not asking about.
6. Add `test_rehydration_survives_unreadable_stored_image`: make `get_file_content` raise for one key on a `files_map=None` turn and assert the turn completes with the remaining image attached. Add one test exercising more prior-turn images than `_MAX_VISION_IMAGES_PER_REQUEST`, asserting which images are kept.

## Evidence
Evidence: 6/10. The PR states its verification commands and adds a regression test that goes end-to-end through `MessageBuilder`, but the demonstration stops at the mock boundary — the missing artifact is any showing of the round trip against a real storage backend (a two-turn session where turn 2 answers from an image uploaded on turn 1).

## Notes
- Rehydration failures are silent where the upload path calls `_publish_warning`; the user gets an answer from a model that never saw the image (`file_processor.py:197`).
- `excluded_filenames=set(files_map)` skips the stored copy when a re-upload fails this turn (`file_processor.py:464`).
- PDFs are cleared but never rehydrated, so the same bug persists for PDF attachments (`file_processor.py:245`).

---

garland3:
Shepherd update for head `c7af192c`:

- Fixed best-effort vision rehydration: per-image failures no longer abort a follow-up, fetches remain bounded, and current-turn/newest historical images are selected before I/O.
- Enforced the shared inline vision payload budget and added regression coverage for unreadable objects and newest-image selection.
- Narrowed 401 detection to transport responses, carried the failed cached-token fingerprint into recovery, reminted delegated credentials, and tracked active calls so rotation does not close a live client.
- The prior changelog review is obsolete: the current head already has the PR #937 vision entry.

All existing GitHub checks on the previous head were green. Focused local tests and changed-file Ruff pass. Remaining human decision: whether the already-stacked PR #936 OAuth rotation changes should remain in this PR or be split before merge; the OAuth documentation/validation comments are tied to that scope choice.

garland3:
## Summary
Rehydrates previously-uploaded vision images from session storage so follow-up turns can still reference them, under per-request count and aggregate inline-payload caps. Also carries MCP OAuth 401-recovery changes: narrowed unauthorized detection, a delegated-token minting branch, and active-call tracking that blocks eviction of in-use cached clients.

## Verdict
Request changes.

## Required changes
1. Build the rehydration exclusion set from the successfully uploaded refs rather than `set(files_map)` at `atlas/application/chat/utilities/file_processor.py:472` — when a re-upload of an already-stored file fails, the file is excluded from rehydration and its stale `image_b64` has already been popped, so the turn is sent with no image at all.
2. Apply the fingerprint guard to the delegated branch at `atlas/modules/mcp_tools/mcp_execution.py:493` — `_mint_delegated_token` takes no `expected_previous_fingerprint` and unconditionally stores, so N concurrent 401s produce N token exchanges and retries that present an already-overwritten credential.
3. Charge the current turn's uploads (images and `pdf_b64`) against `_MAX_TOTAL_INLINE_B64_BYTES` before rehydration, and pass only the remaining allowance into `_rehydrate_vision_images` (`file_processor.py:468`) — today stale images consume the budget first and a PDF the user just uploaded is demoted to text.
4. Surface rehydration failures and vision demotions to the user at `file_processor.py:207` and `file_processor.py:502` — emit one aggregated warning per turn via `_publish_warning`, mirroring the PDF path, so the user is not told the model analyzed an image it never received.
5. Avoid re-fetching and re-normalizing every stored image on every turn (`file_processor.py:190`) — memoize the normalized payload per session (or skip the fetch when a validated payload is already present) so an image-heavy session does not repeat up to 10 storage GETs per message.
6. Move `_normalize_vision_image_for_llm` off the event loop at `file_processor.py:201` (e.g. `asyncio.to_thread`), or persist the converted PNG at upload time — TIFF decode/re-encode of up to 20 MB currently blocks the loop for all users.
7. Add tests for the aggregate byte budget on both rehydration paths, and for the two new MCP behaviors: an in-flight call preventing eviction of its cached client (`mcp_user_clients.py:493`), and a `delegated` server taking the mint path on 401.

## Evidence
Evidence: 7/10. Four targeted regression tests cover the main rehydration branches and verification commands are stated candidly, but the round trip is only demonstrated at the mocked file-manager boundary; the most important missing artifact is a two-turn run against real storage showing the turn-2 answer depending on the turn-1 image.

## Notes
- The inline-image budget is now enforced at three independent sites (`file_processor.py:210,478,494`) and will drift; consider one fetch-then-enforce pass.
- Cache-key construction bypasses the existing `normalize_user_email` helper at `mcp_execution.py:463`.
- The changelog omits the new `webp`/`bmp` support and the change to which image is dropped at the cap (`CHANGELOG.md:9`).

---

garland3:
Follow-up on head `e4da709e`:

Addressed the fresh review findings: failed re-uploads no longer suppress the stored image, delegated 401 recovery now avoids a second mint when another call already rotated the credential, current image/PDF payloads are charged before historical rehydration, rehydration warnings are surfaced, validated inline payloads are reused via a content hash, and TIFF normalization runs off the event loop. Focused tests remain green (48 passed) and changed-file Ruff passes. CI is rerunning for this head.

garland3:
## Summary
Restores vision images on follow-up turns by retaining or re-fetching inline base64 payloads for stored session files, with new per-request image count and aggregate byte caps, and adjusts MCP per-user client caching, 401 detection, and delegated-token recovery.

## Verdict
Request changes.

## Required changes
1. Enforce the image count cap and inline byte budget on the no-upload early-return path at `atlas/application/chat/utilities/file_processor.py:287` — that branch calls `_rehydrate_vision_images` and returns before the enforcement block, so retained plus rehydrated images can reach ~2× the count cap and ~2× the 18 MB budget, and every later message in the session is rejected by the provider.
2. Compute `available_slots` and `initial_inline_b64` from every ref in `session_files_ctx` that already holds `image_b64`, not just names in `files_map` (`file_processor.py:501`) — retained images are currently uncounted and unbilled against both limits.
3. Narrow 401 detection at `atlas/modules/mcp_tools/mcp_errors.py:85` to explicitly enumerated transport exception types instead of any exception exposing `.response.status_code == 401` — a remote MCP server's own upstream 401 otherwise forces an OAuth rotation/token mint and a silent re-invocation of a possibly non-idempotent tool. Add a test asserting no rotation occurs for a non-transport exception carrying `status_code == 401`.
4. Add compare-and-store semantics to `_mint_delegated_token` (an `expected_previous_fingerprint` parameter, re-read under the storage lock) and pass `failing_fingerprint` from `atlas/modules/mcp_tools/mcp_execution.py:493` — concurrent 401s on the same (user, server) both mint today, and the second overwrites the credential the first is retrying with.
5. Emit one aggregated user-facing warning naming images demoted by the count/byte caps at `file_processor.py:497` — rehydration failures are surfaced but cap demotions are log-only, so the model answers as if it saw every attachment.
6. Add tests for the two new untested behaviors in the MCP half: in-flight calls blocking cache eviction (`mcp_user_clients.py:493`) and the delegated mint branch with its fingerprint short-circuit (`mcp_execution.py:493`).

## Evidence
Evidence: 6/10. The vision half is well described and unit-tested across its main branches, but the MCP OAuth half carries no recorded verification and there is no end-to-end demonstration of the titled fix; the most valuable missing artifact is a two-turn chat transcript or screenshot against real storage and a vision-capable model.

## Notes
- Cached inline payloads are never re-validated against storage, so a deleted or revoked image keeps reaching the provider for the session's life — `file_processor.py:277`.
- PDFs are still unconditionally dropped on follow-up turns with no warning, now visibly inconsistent with images — `file_processor.py:281`.
- Changelog entry omits `webp`/`bmp` support and the flip to keeping the newest images at the cap — `CHANGELOG.md:9`.

---

garland3:
Follow-up on head `cfc39dab`:

Closed the next review pass: the no-upload path now applies the same count/budget enforcement as upload turns and accounts for retained payloads; cap demotions are user-visible; 401 classification only accepts the explicit HTTP transport exception; and the existing identity normalizer is used for cache keys. Focused tests: 48 passed; changed-file Ruff passes. CI is rerunning.

garland3:
## Summary
Keeps uploaded images attached across follow-up turns by rehydrating inline vision payloads from storage, adds count/byte caps with user-facing warnings, and extends MCP 401 recovery to delegated tokens with in-flight-call-aware client cache eviction.

## Verdict
Request changes.

## Required changes
1. Make the delegated 401 recovery a compare-and-store at `atlas/modules/mcp_tools/mcp_execution.py:494` — add an `expected_previous_fingerprint` parameter to `_mint_delegated_token` (`mcp_user_clients.py:98`) that re-checks the stored record under the token-storage lock before `store_token`, mirroring `_refresh_oauth_token`; the current pre-read comparison happens before an `await`, so two concurrent 401s both mint and the second overwrites the first, making the retry 401 again.
2. Seed `_enforce_vision_inline_limits` with the existing `pdf_b64` total at `atlas/application/chat/utilities/file_processor.py:256` — it restarts the accumulator at 0 while rehydration (line 549) and the PDF loop (line 586) count PDFs, so old images can push a just-uploaded PDF over the budget and demote it.
3. Derive `retained_names`/`initial_inline_b64` from refs that are *not* rehydration candidates at `file_processor.py:317` — payloads with a stale `image_content_hash` are currently counted both as retained and as their replacement, which can zero out `available_slots` and leave the bad payload in the request permanently.
4. Order `retained_names` newest-first at `file_processor.py:333` to match the newest-first candidate selection in `_rehydrate_vision_images` — as written, priority order is oldest-first, so the image just rehydrated is the one demoted when the cap bites, every turn.
5. Bound image decode of user-supplied data in `_convert_tiff_to_png_b64` (`file_processor.py:103`) with an explicit `Image.MAX_IMAGE_PIXELS` and an error-on-`DecompressionBombWarning`, and gate the `asyncio.gather` fan-out in `_rehydrate_vision_images` (`file_processor.py:210`) with a semaphore — decode now runs every turn for up to 10 images, so one crafted TIFF plus empty follow-ups can exhaust backend memory.
6. Add tests for the two new uncovered behaviors: a `delegated` server taking the mint path on 401 and skipping the mint when the stored fingerprint already moved (`mcp_execution.py:493`), and an in-flight call preventing cache eviction of its client (`mcp_user_clients.py:493`).
7. Add a byte-budget test on the no-upload path asserting which images survive `_MAX_TOTAL_INLINE_B64_BYTES` and that `_publish_warning` names the demoted files (`file_processor.py:233`).

## Evidence
Evidence: 7/10 — the rehydration branches have focused regression tests and the verification notes are candid about what could not run, which is above average for a change this size. The main gap is the MCP OAuth half, which lands concurrency- and credential-sensitive code with no recorded verification of the new paths.

## Notes
- The rehydrate → cap → warn sequence is duplicated on the two exit paths and has already diverged in three ways — `file_processor.py:310` and `:539`.
- Two different warning strings describe the same outcome, and an upload turn can emit both for one message — `file_processor.py:327`, `:552`.
- Retained payloads are never revalidated against storage, so an out-of-band deletion keeps being sent to the provider for the life of the session — `file_processor.py:309`.

---

garland3:
Follow-up on head `4ca61a54` (review round 5):

- Delegated 401 recovery is now a compare-and-store: `_mint_delegated_token` carries `expected_previous_fingerprint` and stores the minted token via `store_token_if_unchanged` under the token-storage lock, mirroring `_refresh_oauth_token` — a second concurrent 401 no longer overwrites the credential the first caller retries with.
- `_enforce_vision_inline_limits` seeds its accumulator with committed `pdf_b64` bytes, so the budget demotes historical images rather than pushing a just-uploaded PDF into the PDF demotion pass.
- `retained_names`/`initial_inline_b64` on the no-upload path are derived from refs that are *not* rehydration candidates (shared `_is_rehydration_candidate` predicate), so a stale payload can no longer count as both kept and its replacement; ordering is newest-first to match the rehydration selection.
- TIFF decode is bounded by an explicit `Image.MAX_IMAGE_PIXELS` with error-on-`DecompressionBombWarning`, and the rehydration gather fan-out is bounded by a semaphore (`_REHYDRATE_DECODE_CONCURRENCY`).
- New tests: delegated mint path on 401, mint skip when the stored fingerprint moved, `store_token_if_unchanged` (store/skip/fail-closed), in-flight call spared from rotation eviction, byte-budget demotion naming the demoted files on the no-upload path, and the bounded rehydration fan-out.

Also merged `main` into the branch (CHANGELOG only) — the PR is mergeable again, and the stacked #936 work is already in `main` so the effective diff is now just the vision + 401-recovery files.

Verification: focused suites 63 + 186 passed; full backend suite 3570 passed / 17 skipped; Ruff clean on changed files. CI running for this head.

garland3:
## Summary
Keeps vision images from earlier turns and rehydrates them on follow-up messages, newest first, within count and byte caps, with a TIFF pixel guard and bounded fan-out (#907). Also hardens MCP 401 recovery: 401s are detected only at the transport layer, delegated tokens are re-minted with compare-and-store, and clients with calls still in flight are not evicted.

## Verdict
Approve with nits.

## Evidence
Evidence: 7/10. Unit tests cover the change well for its size, but nothing shows the titled fix working end to end. The most useful addition would be a two-turn transcript or screenshot: upload an image, then ask a follow-up with no attachment that the model answers correctly from the image.

## Notes
- A stale inline payload is still sent after a failed rehydration. Refs with a matching key but a mismatched content hash keep `image_b64`, so pop the payload for every rehydration candidate at `atlas/application/chat/utilities/file_processor.py:356`.
- An image that is unusable or over budget is fetched again on every follow-up turn, and a duplicate warning bubble is posted each time. Record a per-key marker so these images are skipped and warned about once (`atlas/application/chat/utilities/file_processor.py:246`, `:389`).
- The decompression-bomb guard uses `warnings.catch_warnings()` and changes the global `Image.MAX_IMAGE_PIXELS`. Neither is thread-safe under `asyncio.to_thread`, and the test leaves the limit set to 1. Check `width * height` right after `Image.open` instead (`atlas/application/chat/utilities/file_processor.py:126`).

---

PR #938, opened 2026-09-14
test(mcp): guard ImageContent context isolation

### LINKED ISSUE #261: Context corruption after first png image display
When an MCP tool returns an image via direct ImageContent objects, the first image displays correctly.

Subsequent calls to the same tool seem to corrupt the chat context history and Atlas gets very confused. Tool calls start failing and the Atlas UI sometimes displays the json tool call, rather than making the actual tool call.

Unclear why this is happening. Possibly it is due to the base64 ImageContent being added to the context and the LLM being used not being multi-modal.

garland3:
## Summary

- Verifies direct MCP `ImageContent` bytes are excluded from the normalized LLM payload.
- Verifies the same image remains available as a UI artifact with canvas display metadata.
- Confirms the behavior described in #261 on current `main` is already fixed in production code.

## Verification

- `uv run pytest atlas/tests/test_mcp_tool_result_parsing.py atlas/tests/test_structured_output.py atlas/tests/test_tool_image_context.py -q`
- `uv run ruff check atlas/tests/test_mcp_tool_result_parsing.py`

Closes #261


garland3:
## Summary
Adds a single regression test asserting that base64 bytes from a direct MCP `ImageContent` block stay out of the normalized LLM payload while still surfacing as a UI artifact (issue #261). Test-only change; no production code touched.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. Adequately demonstrated for a ~28-line test-only change — intent, linked issue, and exact pytest/ruff invocations are stated; the most useful missing artifact is a pasted pass summary (and a note that the test was confirmed to fail against the pre-fix behavior, proving the guard is live).

## Notes
- Coverage is limited to the `content[]` text-fallback branch; the `structured_content`/`data` branches (`mcp_result_processor.py:129-138`) are where base64 can actually reach `results` and remain untested — `test_mcp_tool_result_parsing.py:146`.
- `assert image_data not in repr(normalized)` is implied by the preceding exact-equality assertion, and the 12-byte fixture exercises none of the size guards — `test_mcp_tool_result_parsing.py:153`.
- Add `assert len(artifacts) == 1` before indexing, and share `MockImageContent` with `test_mcp_image_content.py:15` instead of redeclaring it — `test_mcp_tool_result_parsing.py:25`.

---

garland3:
## Summary
Adds a regression test asserting that base64 image bytes from MCP tool results stay out of the LLM payload while still surfacing as a UI artifact (#261). Test-only change.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10 — adequately demonstrated for a change this small. The one gap is any indication the new test was confirmed to fail against the pre-fix behavior, which is what separates a regression guard from an assertion of current state.

## Notes
- Guard covers only the `content[]` text fallback; `data` and `structured_content` are pinned to `None` in the fixture, so the Phase 5 branch that copies structured output wholesale into `results` is untested — `atlas/tests/test_mcp_image_content.py:213`.
- `assert image_data not in repr(normalized)` is implied by the preceding exact-equality assertion and checks `repr` rather than the `json.dumps` form actually sent upstream — `atlas/tests/test_mcp_image_content.py:222`.
- The 12-byte fixture stays below every size guard (4000-byte truncation, 8000-byte fallback), so threshold regressions can't fail this test — `atlas/tests/test_mcp_image_content.py:215`.

---

PR #939, opened 2026-09-14
Add an atlas_launch built-in tool so a conversation can launch sub-conversations (#925)

### LINKED ISSUE #925: Add an `atlas_launch` tool so a conversation can launch sub-conversations (subagents)
## Ask

Add an `atlas_launch` tool to the special/built-in Atlas server tool section (alongside the other in-process Atlas tools, e.g. `canvas`), so a conversation can launch other conversations — Atlas's own equivalent of subagents.

## Why now

#884 / #903 landed parallel conversation runs with per-run identity, and #915 collapsed the event-tagging rule into one authority. That machinery already gives us the two hard parts of subagents: a run registry that can own several concurrent runs on one connection, and run-scoped event tagging so each run's output routes back to its own conversation instead of the one on screen. What is missing is a way for the *model* to start one.

## Proposed tool contract

`atlas_launch` takes three required inputs:

| Input | Meaning |
|---|---|
| `workspace` | The workspace to run under — this is what carries the tools **and** the data sources the sub-conversation is allowed to use. One handle instead of two separate lists, so the child's capability surface is a thing the user already configures and can audit. |
| `model` | Which configured model the sub-conversation runs on. Lets a parent on an expensive model fan work out to a cheaper/faster one. |
| `prompt` | The task for the sub-conversation. |

## Things to decide during design

- **Return shape.** Does `atlas_launch` block until the child finishes and return its final answer, or return a handle immediately and let the parent poll / be notified? The run registry supports the latter; the former is simpler for the model to use. Possibly both, via a flag.
- **Where the child's transcript lives.** Its own conversation in history (visible, reopenable, consistent with how background runs already surface) seems right, ideally linked to the parent.
- **Authorization.** The child must run as the same user and be restricted to workspaces that user can already reach — `atlas_launch` must not become a way to widen tool or data-source access beyond the caller's own.
- **Recursion and resource limits.** A depth cap and a cap on concurrent children per user, or a conversation can fan out without bound. Note that eligibility for background runs is already gated (`turn_is_eligible_for_background_run`): chat history on, server save mode, agent mode, at least one tool.
- **Cancellation.** Stopping the parent should stop its children.

## Acceptance sketch

- `atlas_launch` appears in the built-in Atlas tool section and is selectable like any other tool.
- Calling it with a `workspace`, `model` and `prompt` starts a real run whose events are tagged with its own `run_id`/`conversation_id` and do not leak into the parent's transcript.
- The child is confined to the named workspace's tools and sources, and cannot exceed the caller's own access.
- Depth and concurrency limits are enforced and covered by tests.

garland3:
Closes #925.

## What

A fifth built-in tool on the `atlas` server, `atlas_launch`, lets a conversation start sub-conversations -- ATLAS's own subagents. It takes `workspace`, `model` and `prompt`, and **returns a handle immediately** (the child's `run_id` and `conversation_id`) rather than blocking for the answer, per the decision on the issue. The child's transcript is its own conversation in history.

## Design decisions

- **Handle, not answer.** Blocking would pin the parent's agent loop and context open for the child's whole lifetime, serializing what fan-out is meant to parallelize.
- **The workspace is the capability boundary.** One handle carries the child's tools *and* data sources, so what a sub-conversation may touch is something the user already configures and can audit -- not a list the model assembles.
- **Reuses #884/#915.** A launched run is just another registry entry: its own `run_id`/`conversation_id`/`session_id`, its task binds `set_current_run`, and a small `_ChildConnection` stamps the child's ids on every frame before it travels out over the parent's socket, so a child's tokens cannot splice into the visible transcript.

## Boundaries (all covered by tests)

- Workspace lookup is scoped to the caller, so another user's workspace id is "not found" -- and the refusal does not leak their workspace names.
- Model goes through `check_model_access`; `UNKNOWN` and `DENIED` produce an identical message so the tool cannot probe the restricted model list.
- The workspace's stored tool list is re-filtered through the caller's ACLs **at launch time**; a workspace with nothing authorized is refused rather than started empty.
- `ATLAS_LAUNCH_MAX_DEPTH` (default 2) and `ATLAS_LAUNCH_MAX_CHILDREN_PER_RUN` (default 3) bound recursion and fan-out; `MAX_CONCURRENT_RUNS_PER_USER` still applies underneath.
- `RunRegistry.cancel` cascades to children (and grandchildren), and the wall-clock backstop cascades the same way.
- `FEATURE_ATLAS_LAUNCH_ENABLED` is off by default and only effective with chat history and agent mode on; the gate is enforced in the schema, in tool authorization and at execution.

## Verification

- `atlas/tests/test_atlas_launch_tool.py`: 21 new tests.
- Full backend suite: 3347 passed, 17 skipped.
- `bash test/run_tests.sh backend` (repo runner, includes live CLI E2E): all passed.
- `scripts/check-docs.sh`: OK.

## Docs

- `docs/developer/design-notes/atlas-launch-sub-conversations-2026-09-13.md` (design record, incl. known edges)
- `docs/admin/configuration.md` (new env vars, cost note)
- `docs/user-guide/tools-and-settings/README.md` (end-user how-to)

https://claude.ai/code/session_0147UYLhuCqxpHVrnZ9e3VZ4


garland3:
## Summary
Adds an `atlas_launch` built-in tool that lets an agent turn start a sub-conversation as a child run, with a new run launcher module, parent/child tracking and cancellation cascade in the run registry, feature-flag gating across the schema/execution/config surfaces, and docs plus 21 new tests.

## Verdict
Request changes.

## Required changes
1. Stop routing child frames through the parent turn's `ToolCallRecorder` at `atlas/application/chat/runs/launcher.py:203` — `_ChildConnection` wraps `context["update_callback"]`, which in agent mode is the parent's recorder, so the child's `tool_start`/`tool_complete`/`tool_error` rows are persisted into the parent conversation's history. Capture the underlying `send_json` transport instead, and/or make the recorder drop payloads whose `run_id` is not its own.
2. Propagate (or refuse on) incognito state at `atlas/application/chat/runs/launcher.py:283` — `_run_child` calls `handle_chat_message(..., incognito=False)` unconditionally, so a launch from an incognito turn writes the prompt and the child's full transcript to durable history. Thread the flag through or return a `LaunchRefused`, and pin the choice with a test.
3. Enforce the fan-out cap for untracked turns at `atlas/application/chat/runs/launcher.py:193` — `_check_limits` hardcodes `in_flight = 0` when `parent_run_id` is `None`, so a parentless turn can launch unboundedly and those children are also outside the cancel cascade. Either count depth-0 launched runs against the same cap or refuse launches from untracked turns.
4. Close the fan-out race at `atlas/application/chat/runs/launcher.py:376` — `_check_limits` completes before the first `await`, so parallel `atlas_launch` calls within one agent step all admit. Perform the check-and-reserve atomically under the registry lock.
5. Stop reconfiguring the global run registry on every launch at `atlas/application/chat/runs/launcher.py:374` — per-call assignment of the process-wide concurrency limit lets one request's settings snapshot rewrite a shared bound. Configure it once at registry construction.
6. Add a test for the cancellation cascade against a child with a real attached task (`atlas/tests/test_atlas_launch_tool.py:300` only asserts status fields on task-less records) and one for the wall-clock backstop cascade at `atlas/application/chat/runs/registry.py:486`.

## Evidence
Evidence: 8/10 — proportionate for a backend change of this size, with the full suite result, named new tests, and a design note. The one gap worth closing is an artifact showing a real launch end to end (handle JSON plus the child run reaching a terminal status with the flag on); every launch demonstrated today runs against test doubles.

## Notes
- Sibling gates disagree on their default: the schema gate defaults on while `_atlas_tools_info` defaults off (`atlas/routes/config_routes.py:58`), so a caller that forgets the flag silently hides the tool.
- Child step budget is hardcoded at `LAUNCH_AGENT_MAX_STEPS = 10` (`atlas/application/chat/runs/launcher.py:52`) while the interactive equivalent is admin-tunable.
- `launch_tool_enabled` lives in the launcher rather than beside the tool definition (`atlas/application/chat/runs/launcher.py:75`), forcing function-local imports at two of its three call sites to dodge a cycle.

---

garland3:
## Summary
Adds an `atlas_launch` built-in tool that starts a sub-conversation ("child run") under a chosen workspace and returns a handle immediately, plus run-registry support for parent/child tracking, depth and fan-out caps, a cancellation cascade and a wall-clock backstop, behind a feature flag.

## Verdict
Request changes.

## Required changes
1. Hand `_ChildConnection` the raw transport instead of the tool context's `update_callback` at `atlas/application/chat/runs/launcher.py:240` — in agent mode that callback is the parent turn's `ToolCallRecorder`, so the child's `tool_start`/`tool_complete`/`tool_error` frames are recorded and persisted into the *parent* conversation's transcript. As defense in depth, make `ToolCallRecorder._record` drop payloads whose `run_id` differs from the run it was created for.
2. Move the child cascade in `RunRegistry.cancel` above the `record is None or record.is_terminal` early return at `atlas/application/chat/runs/registry.py:403` — because the tool returns a handle, the parent run normally goes terminal while children are still running, making the documented "stopping a conversation stops the ones it launched" guarantee unreachable in the common case. Also keep the parent record alive (or re-parent live children) while non-terminal children exist, so `reap_terminal` cannot orphan the subtree.
3. Thread the parent turn's incognito flag into the tool context and either propagate it to the child or refuse with `LaunchRefused` at `atlas/application/chat/runs/launcher.py:293` — today `_run_child` passes `incognito=False` unconditionally, so an incognito turn silently persists the prompt and the child's full transcript to history. Pin the chosen behavior with a test.
4. Make admission atomic by moving the depth/fan-out check into `RunRegistry.start` (passing `max_depth`/`max_children`) at `atlas/application/chat/runs/launcher.py:366` — the check precedes several awaits, so N concurrent `atlas_launch` calls in one agent step all admit even when the per-run cap is 1. Add a test using `asyncio.gather`.
5. Narrow the ACL claim in `CHANGELOG.md:9`, `docs/admin/configuration.md` and `LAUNCH_TOOL_DESCRIPTION`, or add the missing filter — `resolve_child_tools` re-filters only `selected_tools`; `selected_data_sources` are copied verbatim from the stored workspace config, so "can never reach a … data source that user could not reach themselves" is not what the code does.
6. Drop the per-launch `registry.set_max_concurrent_runs_per_user(getattr(app_settings, "max_concurrent_runs_per_user", 5))` at `atlas/application/chat/runs/launcher.py:381` and configure the limit once at registry construction — a settings object missing that attribute silently resets the deployment's process-wide limit to 5 for all users.
7. Add tests for cancellation with a child that owns a real `asyncio` task (`atlas/tests/test_atlas_launch_tool.py:300` uses task-less records, so `task.cancel()` and `_run_child`'s cancellation handling never execute) and for the wall-clock backstop cascade at `atlas/application/chat/runs/registry.py:486`, which has no test.

## Evidence
Evidence: 8/10 — 21 named tests mapped onto the stated boundaries, a full-suite result, a design note and admin/user docs; adequate for a backend change of this size. The one gap worth closing is that every launch is exercised against test doubles, so no artifact shows a single real launch reaching a terminal child conversation with the flag on.

## Notes
- Fan-out refusal text attributes other conversations' children to this one on the untracked-turn path (`atlas/application/chat/runs/launcher.py:196`).
- `_atlas_tools_info` defaults `launch_enabled=False` while `atlas_tool_schemas` defaults it `True`, so a forgotten keyword hides the tool from the UI while the model still sees it (`atlas/routes/config_routes.py:58`).
- `launch_tool_enabled` sits in the launcher rather than beside the tool, forcing function-local imports and a bare `except` that turns any import error into "feature off" (`atlas/modules/mcp_tools/mcp_discovery.py:140`).

---

garland3:
## Summary
Adds an `atlas_launch` tool that lets a run start child sub-conversations, with a new launcher module, parent/child support in the run registry (depth, fan-out, cancel cascade, wall-clock backstop), settings-based gating, docs and 21 tests.

## Verdict
Request changes.

## Required changes
1. Handle "run context present but no registry record" explicitly in `_parent_run_identity` at `atlas/application/chat/runs/launcher.py:193` — returning `(run_id, 0)` makes `children_of()` permanently empty and resets depth, so both the fan-out and recursion caps are bypassed and the children cannot be cancelled. Refuse the launch or normalize to `parent_run_id = None`; add a test that sets a current run id the registry has never seen and asserts the second launch is still refused.
2. Cascade cancellation over the full subtree in `RunRegistry.cancel` at `atlas/application/chat/runs/registry.py:415` — `children_of()` defaults to `include_terminal=False`, so a completed child hides its still-running grandchildren from the stop path. Apply the same `include_terminal=True` view to the retention guard in `reap_terminal` (`registry.py:455`) so a record is never reaped while any descendant is non-terminal, and add a three-level test with a terminal middle run and a live grandchild.
3. Stop reconfiguring the process-wide limit per request at `atlas/application/chat/runs/launcher.py:381` — `registry.set_max_concurrent_runs_per_user(...)` runs on every launch, so a settings object missing the attribute resets the bound to the default for all subsequent users. Set it once in `RunRegistry.__init__` or the singleton factory.
4. Re-filter `selected_data_sources` through the caller's current authorization at `atlas/application/chat/runs/launcher.py:346`, the way `selected_tools` already is — a saved workspace otherwise carries a revoked corpus into the child, contradicting the "can never reach a … data source its caller could not" claim in `CHANGELOG.md:9` and `docs/admin/configuration.md`. If query-time enforcement is the intended contract, narrow those claims and add a test pinning a launched run being refused at query time.
5. Add tests for the two untested enforcement points: the launch gate in `ToolAuthorizationService` and the execution-time gate in `mcp_execution.py`, asserting they agree with the schema gate for the same flag value — `atlas_tool_schemas` defaults `launch_enabled=True` while `_atlas_tools_info` defaults it `False`.
6. Reword the fan-out refusal on the untracked-turn branch at `atlas/application/chat/runs/launcher.py:196` — it counts the user's launched runs across all conversations but tells the user the limit is "per conversation".

## Evidence
Evidence: 8/10. Well-evidenced for a backend change of this size (21 targeted tests, full suite counts, design note and admin/user docs); the one gap worth closing is that every launch runs against test doubles — paste the handle JSON from one real `atlas_launch` with the flag on plus the child's final run status.

## Notes
- `_raw_transport` unwraps by attribute name up to five levels (`launcher.py:187`); prefer an explicit `isinstance(cb, ToolCallRecorder)` check so a future outbound wrapper isn't silently stripped for child frames.
- `LAUNCH_AGENT_MAX_STEPS = 10` (`launcher.py:52`) is a module constant while the interactive step budget comes from settings — consider a dedicated setting.
- `reap_terminal` is quadratic in tracked runs (`registry.py:455`), and a finished parent is retained indefinitely if a child never reaches a terminal status.

---

garland3:
## Ran it for real, with the flag on

Branch checked out, `FEATURE_ATLAS_LAUNCH_ENABLED=true`, chat history on DuckDB, real models. A workspace named **Research** was saved holding `atlas_canvas` + `calculator_evaluate`; the launching conversation had only `atlas_canvas` + `atlas_launch` ticked.

**1. The tool is in the built-in `atlas` server and selectable like any other**

tools panel

**2. The handle comes back immediately — the call does not wait for the answer**

handle

```json
{
  "run_id": "05535021-3088-445c-ab14-be134665de37",
  "conversation_id": "fc164926-984c-4ebd-9b59-1e4fcf8fbc06",
  "status": "queued",
  "workspace": "Research",
  "model": "openrouter-gpt-oss-120b",
  "tools": ["atlas_canvas", "calculator_evaluate"],
  "data_sources": [],
  "depth": 1,
  "parent_run_id": "...",
  "tools_needing_approval": []
}
```

Note `tools`: the workspace's stored list re-filtered through the caller's ACLs at launch time, and `depth: 1` under the launching run.

**3. The sub-conversation is its own conversation in history, with its own transcript**

child conversation

Its final run status was `completed`; the child's transcript read back from the API:

```
user      | Compute 17 * 23 with the calculator tool and state the result.
tool      | Tool call: calculator_evaluate
assistant | The product of 17 × 23 is **391**.
```

None of the child's tool rows appear in the launching conversation.

## What running it found (and what changed because of it)

Two things no unit test caught:

1. **The child stalled on its first tool call.** Tool approval is enforced server-side on every call, and the only thing that answers one is a browser looking at *that* conversation — which cannot be opened until the run's first save. So a launched run with tools waited forever. **Decision taken:** approving the `atlas_launch` call is the approval for the sub-conversation's own tool calls (that call names the workspace, the model and the task, and goes through the normal gate). Tools an admin pinned to mandatory approval are exempt — they keep prompting inside the child and come back in `tools_needing_approval`. Flagging this explicitly because it is a product decision, not a bug fix: say the word and I will invert it (keep approvals and surface running sub-conversations in the history list instead).
2. **The incognito refusal was unreachable in the app.** Agent mode builds its own tool `session_context` rather than spreading the session's, so the flag never arrived. Threaded through `AgentContext`, with a test that pins the wiring.

## Review rounds

All three review passes are addressed. Fixed since the first: child tool rows persisting into the parent's transcript (the child now sends past the parent's `ToolCallRecorder`, which also drops foreign `run_id`s); incognito launches refused; admission made atomic (check and insert adjacent, no await between) with an `asyncio.gather` test; cancel cascading before the terminal check *and* over terminal children, so a finished parent or a finished middle run cannot hide a live descendant; the reap guard holding a parent while any descendant lives; an unknown run id treated as untracked instead of as a phantom parent; the per-launch rewrite of the process-wide concurrency limit removed; `launch_tool_enabled` moved beside the tool definition (no more function-local imports); the schema flag defaulting off to match the config payload; and the ACL claim narrowed to what the code does — tools re-filtered at launch, data sources authorized per source when queried.

Verification: 38 tests in `test_atlas_launch_tool.py`, full backend suite **3364 passed / 17 skipped**, `test/run_tests.sh backend` green, `scripts/check-docs.sh` OK.

garland3:
## Summary
Adds an `atlas_launch` built-in tool that spawns a bounded sub-conversation as a child run, with parent/child tracking in the run registry, depth and fan-out caps, ACL re-filtering of the workspace's tools, cancellation cascade and a wall-clock backstop.

## Verdict
Request changes.

## Required changes
1. Narrow the child pre-approval at `atlas/application/chat/runs/launcher.py:285` — setting `skip_approval = True` for the whole child run turns one approval on `atlas_launch` into blanket approval for every tool in the named workspace, including tools the caller did not tick in the parent conversation. Exclude any tool that is approval-requiring in the caller's current configuration (`FORCE_TOOL_APPROVAL_GLOBALLY`, per-tool `require_approval`, `REQUIRE_TOOL_APPROVAL_BY_DEFAULT`), not only the admin-mandated subset, and surface the pre-approved tool list in the `atlas_launch` approval prompt.
2. Pass `include_terminal=True` in `enforce_wall_clock` at `atlas/application/chat/runs/registry.py:519` — a terminal middle run hides its still-running grandchildren, so the backstop leaves them spending indefinitely. Add a test with a terminal middle run and a live grandchild, mirroring the equivalent cancel-path test.
3. Re-check `record.is_terminal` at the top of the `enforce_wall_clock` loop body (`registry.py:496`) and give the cascade a status/error argument — today a cascade-cancelled child is re-cancelled, reported in `expired`, and recorded `CANCELLED` (the status reserved for "the user stopped it") rather than `FAILED` with the wall-clock message.
4. Re-filter `selected_data_sources` through per-source authorization at `launcher.py:346` and strip unauthorized ids from the returned handle — stored workspace sources are copied verbatim while tools are re-filtered, so a revoked corpus id still reaches the child and is disclosed to the caller.
5. Fix the untracked-turn fan-out fallback at `launcher.py:193` — it counts every launched run the user has anywhere against a per-run cap, so a turn that has launched nothing can be refused outright. Count only launches from untracked turns (or defer to `MAX_CONCURRENT_RUNS_PER_USER`) and reword the refusal to match the quantity counted.
6. Log at warning when `_preapprove_child_tools` returns `False` (`launcher.py:283`; return value discarded at :442) and add a test for the failure branch — today a restructured `agent_mode` silently produces a child that stalls forever on its first approval, with a green suite.
7. Add a parent/child index (`parent_run_id -> set[child_run_id]`) so `children_of` is a lookup rather than a full scan of `self._runs` — `reap_terminal` now calls `_has_live_descendants` per terminal candidate, making each reap pass O(T·R) on the event loop even in deployments that never use `atlas_launch`.

## Evidence
Evidence: 9/10 — unusually complete for a change this size, with a real end-to-end run (screenshots, handle JSON, child transcript, the negative check that child tool rows stay out of the parent), tests mapped onto each stated boundary, and full-suite numbers. The one gap worth closing is a real-app demonstration of the cancellation cascade: stop a launching conversation mid-child-tool-call, ideally with a grandchild present, and show the child's final `cancelled` status.

## Notes
- Terminal records are pinned indefinitely while any descendant stays non-terminal, with no upper bound — `registry.py:459`.
- User guide says to open the sub-conversation from the history list, but a launched run is not in history until its first save — `docs/user-guide/tools-and-settings/README.md:138`.
- `test_agent_mode_carries_the_turns_incognito_flag_to_its_tools` asserts on `inspect.getsource` text rather than behavior — `atlas/tests/test_atlas_launch_tool.py:711`.

---

garland3:
Shepherd update on the latest review:

Addressed in c36893eb:
- Child approval/elicitation requests now set `waiting_for_input` and persist a replayable pending frame, including when no socket transport is attached.
- Wall-clock expiry cascades through terminal middle runs, marks descendants `failed` with the timeout reason, and skips records already made terminal during the cascade.
- Child `{"type":"error"}` responses now mark the run failed and preserve the error.
- Workspace data-source handles are intersected with the caller-authorized discovery set, while preserving query-time authorization.
- Untracked-turn fan-out counts only depth-1 runs created by other untracked turns, rather than unrelated tracked-parent children.
- A parent-to-children index backs `children_of()` and is maintained through reaping/removal.
- Handle status reports the admitted record's actual status; failed pre-approval setup is logged.

The remaining approval-policy point is intentional: the launch call authorizes user-level tools in the selected workspace, while admin-mandated tools remain interactive and are surfaced in `tools_needing_approval`. A product decision is still needed if every user-level child tool must instead require a second approval.

Other non-blocking review ideas remain separate follow-ups: passing the owning `AppFactory` through tool context for headless isolation, creating the history row before a child starts, releasing child-owned resources explicitly, rejecting models without tool support, and replacing the fixed child step budget with a setting. None is changed here because each requires a broader lifecycle/DI design choice beyond the current review fixes.

garland3:
## Summary
Adds the `atlas_launch` built-in tool and its supporting run registry, letting a conversation spawn tracked sub-conversations with depth/fan-out caps, cancellation cascade, wall-clock backstop, and workspace-derived tool and data-source resolution.

## Verdict
Request changes.

## Required changes
1. Make child pre-approval per tool rather than per run at `atlas/application/chat/runs/launcher.py:496` — a workspace mixing one admin-gated tool with ordinary ones leaves `skip_approval` false for the whole child, so every tool stalls at an approval nobody can answer until the wall-clock cap fires. Add a test with a mixed workspace asserting the user-level tool executes while the admin-gated one still prompts.
2. Return `[]` (and log at warning) when neither RAG service is available in `resolve_child_data_sources` at `atlas/application/chat/runs/launcher.py:322` — an authorization intersection that returns the caller's list verbatim fails open, and the unfiltered ids are echoed into the handle.
3. Cover the data-source filter in tests at `atlas/tests/test_atlas_launch_tool.py:392` — `_Factory` defines neither discovery service, so the intersection, the `f"{server}:{id}"` key construction and the compliance-level argument are never executed; a revoked corpus reaching the child would not fail any test.
4. Refuse the launch (or report the full tool list in `tools_needing_approval`) when `_preapprove_child_tools` returns `False` at `atlas/application/chat/runs/launcher.py:496` — today the run is admitted with an empty `tools_needing_approval`, so the model promises a result from a child that will block on its first tool call. Add a test for that branch.
5. Bound `atlas_launch_max_depth` with `le=10` in settings, or derive the recursion limit from it, at `atlas/application/chat/runs/registry.py:504` — the hard-coded `_depth > 10` guard reports "no live descendants" for deeper chains and reaps ancestors of still-running runs.
6. Fix the sub-conversation visibility claim at `docs/user-guide/tools-and-settings/README.md:138` — the guide tells the user to open the child from history and answer its approval, but a child that pauses before its first save never appears there.
7. Give the fan-out refusal at `atlas/application/chat/runs/launcher.py:196` separate text per branch — the untracked branch counts every depth-1 run the user has anywhere but reports it as a per-conversation limit.

## Evidence
Evidence: 9/10 — unusually well demonstrated for a change this size, with a real end-to-end run, screenshots, docs and a dense test suite. The main missing artifact is a real stop of a launching conversation with a live child, showing the cancellation cascade outside test doubles.

## Notes
- Depth check runs after `create_chat_service`, the ACL filter and RAG discovery — refused launches pay full admission cost (`atlas/application/chat/runs/launcher.py:485`).
- `reap_terminal` exempts any ancestor of a non-terminal descendant from both eviction paths, so one wedged child pins its chain for the process lifetime (`atlas/application/chat/runs/registry.py:479`).
- The pause/resume frame logic is duplicated between `_ChildConnection.send_json` and `turn_update_callback`, with the settled-event set a constant in one and an inline literal in the other (`atlas/application/chat/runs/launcher.py:270`).

---

PR #942, opened 2026-09-15
feat(agent): observe launched sub-conversations

### LINKED ISSUE #926: Follow-ups on the collapsed run-event tagging authority (post-merge review notes from #924)
Three review notes landed on #924 (closing #915) after it had already merged. None is a defect in the merged code; all three tighten it.

## 1. Restore a static contract with a TypeVar

`atlas/application/chat/runs/context.py` — `tag_event` and `stamp_with_current_run` are annotated `Any -> Any`. That was the blunt fix for the non-dict passthrough: the previous `Dict[str, Any] -> Dict[str, Any]` was simply wrong, since both return non-dict input untouched. But `Any` discards the type rather than describing it.

```python
T = TypeVar("T")

def tag_event(data: T, run_id: str, conversation_id: str, *, copy: bool = True) -> T: ...
```

says the same thing and keeps the caller's type. `tag_run_event` in `atlas/main.py` should follow.

## 2. Mirror the identity assertion for the copying path

`atlas/tests/test_parallel_run_transport.py` has `test_publisher_hot_path_stamps_in_place_without_copying`, which asserts `stamp_with_current_run` returns the *same object* — that is what stops a future cleanup from quietly making the hot path allocate.

There is no assertion in the other direction. Feed `tag_run_event` an event with both ids already set (so `setdefault` changes nothing and equality alone cannot tell the two apart) and assert the result `is not` the input. As it stands, a lost copy on the turn-callback path would fail no test, even though copying there is the whole reason the `copy` flag exists.

## 3. Warn when a producer-supplied id disagrees with the ambient run

The larger of the three, and a behaviour change rather than a nit.

The rule is "existing ids win" — a producer that knows its own conversation is more authoritative than the run envelope. That is correct for routing, but it means a *disagreement* between the producer's `conversation_id` and the ambient run's is silently resolved in the producer's favour and leaves no trace.

If those two ever differ, something is already wrong: either the producer is mislabelling its output or the contextvar is bound to the wrong run. Both surface to the user as "messages in the wrong chat", which is exactly the failure #884 introduced run ids to prevent, and #915 consolidated the rule to keep unambiguous.

Logging at warning level when they differ (keeping the producer's value — do not change the routing) makes that misroute detectable instead of invisible. Worth checking first whether any legitimate producer routinely disagrees, or the log will be noise.

## Context

- #915 — the consolidation itself.
- #924 — the PR, with the full review thread and a runtime check showing two conversations executing in parallel.

### LINKED ISSUE #941: Add `atlas_get_runs` and `atlas_result` tools so a parent conversation can observe launched sub-conversations
## Summary

PR #939 adds `atlas_launch`, which lets a conversation start independent ATLAS sub-conversations and immediately receive a `run_id` / `conversation_id` handle.

The next missing piece is **observation**: the launching conversation should be able to see the runs it started and retrieve their results when they finish. This completes the basic supervisor / chief-of-staff loop without introducing a larger orchestration framework.

### `atlas_get_runs`

Return the runs launched by the current conversation, including enough state for the parent agent to decide what to do next.

Suggested output per run:

```json
{
  "run_id": "...",
  "conversation_id": "...",
  "status": "running|completed|failed|cancelled|waiting_for_approval",
  "parent_run_id": "...",
  "depth": 1,
  "started_at": "...",
  "finished_at": null
}
```

Default scope should be the current run's children rather than every run owned by the user. A future option could expose descendants or all user runs if that becomes useful.

### `atlas_result`

Take a `run_id` and return the current state plus the launched conversation's final result when available.

Example:

```json
{
  "run_id": "...",
  "conversation_id": "...",
  "status": "completed",
  "result": "..."
}
```

If still running, return status without blocking. If failed or cancelled, return that state and an appropriate error/summary if available.

## Why separate tools

Keep discovery/status and result retrieval separate:

- `atlas_get_runs` is cheap and lets the supervisor inspect several parallel runs at once.
- `atlas_result(run_id)` retrieves the potentially larger final output only for runs the parent actually needs.
- This avoids injecting every completed child's answer into the parent's context just to check whether work is done.

## Desired interaction

```text
Parent conversation
    |
    +-- atlas_launch(...) -> run A
    +-- atlas_launch(...) -> run B
    +-- atlas_launch(...) -> run C
    |
    +-- atlas_get_runs()
          A: completed
          B: running
          C: completed
    |
    +-- atlas_result(A)
    +-- atlas_result(C)
    |
    +-- synthesize / launch follow-up work
```

This makes a conversation useful as a lightweight **supervisor / chief-of-staff agent** over multiple parallel ATLAS conversations.

## Authorization / boundaries

- A caller must not be able to inspect arbitrary users' runs by guessing a `run_id`.
- `atlas_result` should only expose a run the caller is authorized to access; ideally initially restrict this to runs launched by the current run / conversation.
- Preserve the existing run/conversation identity boundaries introduced for parallel runs and #939.
- Do not block waiting for completion; these should remain handle/status based.

## Acceptance criteria

- [ ] `atlas_get_runs` is available alongside `atlas_launch` when the launch feature is enabled.
- [ ] It returns the current conversation/run's launched children and their current statuses.
- [ ] `atlas_result(run_id)` returns the child's final assistant result when complete.
- [ ] `atlas_result` reports running / failed / cancelled / waiting states cleanly without blocking.
- [ ] A run belonging to another user or outside the allowed parent scope cannot be inspected by guessing its id.
- [ ] Multiple children running concurrently can be queried without affecting or serializing them.
- [ ] Tests cover completed, running, failed/cancelled, unknown/unauthorized run ids, and multiple parallel children.

## Out of scope

- DAG/workflow orchestration.
- Automatic retries.
- Agent-to-agent messaging.
- Shared scratchpads/memory between children.
- Prefect integration.
- Automatic event injection of child results back into a parent turn.

Those can come later if actual use shows they are needed. This issue is intended to complete the smallest useful `launch -> observe -> retrieve` control loop.

Related: #925 / PR #939.

garland3:
Closes #941
Closes #926

## Summary
- add scoped atlas_get_runs and atlas_result tools for launched sub-conversations
- return live, waiting, failed, cancelled, and persisted completed-child states without blocking
- preserve event tagging types, warn on producer/ambient identity mismatches, and cover copy semantics

## Verification
- backend: 3376 passed, 17 skipped
- live E2E: 5 passed
- frontend: 914 passed
- frontend lint: 0 errors, 4 existing warnings
- frontend build: passed


garland3:
Superseded by a clean branch based on current main after #939 merged; opening the same implementation as a focused PR.

PR #943, opened 2026-09-15
feat(agent): observe launched sub-conversations

### LINKED ISSUE #926: Follow-ups on the collapsed run-event tagging authority (post-merge review notes from #924)

### LINKED ISSUE #941: Add `atlas_get_runs` and `atlas_result` tools so a parent conversation can observe launched sub-conversations

garland3:
Closes #941
Closes #926

## Summary
- add scoped atlas_get_runs and atlas_result tools for launched sub-conversations
- return live, waiting, failed, cancelled, and persisted completed-child states without blocking
- preserve event tagging types, warn on producer/ambient identity mismatches, and cover copy semantics

## Verification
- backend: 3376 passed, 17 skipped
- live E2E: 5 passed
- focused regression suite: 87 passed
- frontend: 914 passed
- frontend lint: 0 errors, 4 existing warnings
- frontend build: passed


garland3:
## Summary
Adds two non-blocking observation tools, `atlas_get_runs` and `atlas_result`, letting a parent conversation list and read the results of its direct sub-conversation runs, gated behind the existing launch feature flag, plus run/conversation id tagging diagnostics in the event transport.

## Verdict
Request changes.

## Required changes
1. Move the conversation load off the event loop in `get_child_result` (`atlas/application/chat/runs/launcher.py:617`) — `repository.get_conversation` runs synchronous DB queries and decodes the whole transcript, and `atlas/application/chat/service.py:1176` already wraps this exact call in `asyncio.to_thread` for that reason; polling makes the stall repeat.
2. Wrap the repository call and surrounding logic in `execute_observation_tool` (`atlas/application/chat/runs/launcher.py:632`) in an `except Exception` that logs and returns an unsuccessful `ToolResult`, matching `execute_launch_tool` — today a DB error escapes the tool and ends the turn.
3. Make observation dispatch exhaustive at `atlas/application/chat/runs/launcher.py:625` — pass the already-resolved name from `mcp_execution.py`, compare against the imported `GET_RUNS_TOOL_NAME`/`RESULT_TOOL_NAME`, and reject anything else instead of falling through to the result branch.
4. Demote the expected producer-override case in `tag_event` (`atlas/application/chat/runs/context.py:89`) to `debug`, or dedupe per (run_id, conversation_id) — `test_producer_supplied_conversation_id_wins` documents that path as correct, yet it can emit one WARNING per streamed token.
5. Distinguish "no result" from "result unreadable" in the `atlas_result` payload (`atlas/application/chat/runs/launcher.py:604-624`) — a completed child currently returns `result: null` whether the transcript was empty, the repository was missing, or the load failed.
6. Normalize non-string assistant content before returning it as `result` (`atlas/application/chat/runs/launcher.py:606`) and stop at the last assistant message unconditionally, so list-shaped content is not emitted as a JSON array and an empty final message does not return an earlier turn's answer.
7. Add tests for `execute_observation_tool` and the execution-time feature gate (`atlas/modules/mcp_tools/mcp_execution.py:819`) — both tool names, a refused `run_id`, non-dict `arguments`, and a flag-off dispatch asserting refusal; also assert the new names are dropped from the authorized list when the flag is off (`atlas/application/chat/policies/tool_authorization.py:92`).

## Evidence
Evidence: 8/10 — the change is well covered by unit tests added in the same diff, with suite results and a CHANGELOG entry recorded. The one gap worth closing is an artifact showing a real child actually observed while in flight (launch → get_runs → result across its lifecycle), rather than states synthesized via `registry.set_status`.

## Notes
- Runs launched from an untracked turn get a handle no observation call can honour, and a reaped terminal child is reported as "not a sub-conversation" rather than expired — `atlas/application/chat/runs/launcher.py:574`, `:589`.
- The launch-gated tool triple is spelled out literally in four places; export one `LAUNCH_GATED_TOOL_NAMES` — `atlas/modules/mcp_tools/atlas_server.py:296`.
- Tool descriptions omit the returned fields, the status vocabulary, and any pacing guidance pointing at `atlas_sleep` — `atlas/modules/mcp_tools/atlas_server.py:299-320`.

---

garland3:
Addressed the actionable review feedback in commit `eb6c2d99`: repository reads now run via `asyncio.to_thread` with fail-safe ToolResults; dispatch is exhaustive and passes the resolved tool name; result payloads distinguish pending/available/empty/unavailable/unreadable and normalize the final assistant content; terminal records are reaped before scope lookup; expected conversation-id overrides log at debug; manager-level execution/gate tests and the docs contract are included. Focused regression suite: 90 passed.

garland3:
@copilot review

garland3:
## Summary
Adds two model-facing observation tools (`atlas_get_runs`, `atlas_result`) for sub-conversations launched by `atlas_launch`, extends the launch feature gate to cover them across the schema/authorization/execution layers, and adds run-id and conversation-id mismatch diagnostics to event tagging.

## Verdict
Request changes.

## Required changes
1. Demote or dedupe the run-id mismatch warning at `atlas/application/chat/runs/context.py:89` — child frames are stamped with the child's `run_id` and tagged against the parent's, so this logs at WARNING once per streamed token for every sub-conversation, which is the feature's normal steady state (the `conversation_id` branch is already at `debug`).
2. Remove the `registry.reap_terminal()` call from `_scoped_children` at `atlas/application/chat/runs/launcher.py:586` — a read path that evicts records lets a completed child disappear between an `atlas_get_runs` listing and the `atlas_result` call that follows it; `start()` and the periodic reap already own retention.
3. Distinguish an expired/reaped child from an unknown or foreign id in `get_child_result` at `atlas/application/chat/runs/launcher.py:586` — both currently return "That run is not a sub-conversation of this conversation.", so the parent is told its own child never existed.
4. Cap the size of the returned child result at `atlas/application/chat/runs/launcher.py:606` and signal clipping through `result_status` — the final assistant message is copied verbatim into the parent's prompt on every poll, unbounded.
5. Document the `result_status` vocabulary (`pending`, `available`, `empty`, `unreadable`, `unavailable`) in `RESULT_TOOL_SCHEMA` at `atlas/modules/mcp_tools/atlas_server.py:316-321` and state which statuses are worth re-polling — the consumer is a model that was never told the enum exists.
6. Add tests asserting `result_status`/`result_error` values for the `unreadable`, `unavailable` and `empty` branches, and extend the feature-gate test at `atlas/tests/test_atlas_launch_tool.py:862` to cover `GET_RUNS_TOOL_NAME` and `RESULT_TOOL_NAME` in both directions — the new three-layer gate is only partially pinned.
7. Sanitize `data["run_id"]` and `data["conversation_id"]` with `sanitize_for_logging` before interpolating them at `atlas/application/chat/runs/context.py:89`, matching every other identifier log site in the codebase.

## Evidence
Evidence: 8/10 — proportionate for a backend-only change, with tests added in the same diff and full suite results stated. The one gap is an end-to-end demonstration of a real child being observed while in flight and after completion; current tests synthesize child state by driving the registry and a stub repository.

## Notes
- Collapse the launch-gated tool tuple into one exported `frozenset`; it is now spelled out at four sites (`atlas/modules/mcp_tools/atlas_server.py:424` and three others).
- `unavailable` conflates a failed/cancelled child with a missing repository or transcript, so a caller cannot decide retry vs. give up (`atlas/application/chat/runs/launcher.py:615`).
- A completed child whose last assistant message is empty reports `result_status: "available"` with `result: ""` (`atlas/application/chat/runs/launcher.py:624`).

---

garland3:
## Summary
Adds two observation tools (`atlas_get_runs`, `atlas_result`) alongside `atlas_launch`, letting a parent conversation list its launched sub-conversations and read a completed child's final assistant message, backed by a new `_conversation_children` index in the run registry and a three-layer launch gate.

## Verdict
Request changes.

## Required changes
1. Handle content-less final assistant messages at `atlas/application/chat/runs/launcher.py:606` — `json.dumps(None)` returns the literal string `"null"` as `result` with `result_status: "available"`, so a child ending on a tool call reports its answer as the word `null`; fall back to the last assistant message with usable text, otherwise report `empty`.
2. Remove the `registry.reap_terminal()` call from `_scoped_children` at `launcher.py:588` — a read path that evicts makes `atlas_get_runs` destructive, so a child it just listed can be gone by the time `atlas_result` reads it; retention is already owned by `start()` and the periodic reaper.
3. Distinguish an expired child from a foreign one at `launcher.py:589` — all four refusal conditions share the message "That run is not a sub-conversation of this conversation.", which tells a parent its own reaped child never existed; return a distinct expired status or wording.
4. Resolve or explicitly refuse the untracked-turn case at `launcher.py:530` and `:584` — when `run_context` is absent, `parent_conversation_id` is `None`, the child is never indexed, and `atlas_get_runs` silently returns `{"runs": []}` for a handle the parent was just told it could observe.
5. Truncate extracted text rather than the serialized JSON at `launcher.py:607` — clipping `json.dumps` output mid-document (or mid-escape) yields unparseable text flagged as usable via `available_truncated`.
6. Split the overloaded `unavailable` status at `launcher.py:598`/`:615`/`:621` — a `failed`/`cancelled` child and a completed child whose transcript is not yet persisted are reported identically, so the retryable case is guaranteed to be abandoned.
7. Add tests for the truncation branch, non-string assistant content, and registry conversation-index cleanup on reap, and assert on the `atlas_get_runs` case in the manager-dispatch test at `test_atlas_launch_tool.py` (currently asserts nothing).

## Evidence
Evidence: 7/10 — unit coverage is solid but synthesizes child state directly via `registry.start`/`set_status` and stubs the repository, so no artifact shows a real child being launched, polled in flight, and read to completion; a sample `atlas_result` payload from a real run is the main gap, and the full-suite numbers predate the head commit's registry-indexing changes.

## Notes
- Extract one `LAUNCH_GATED_TOOL_NAMES` constant — the gated tool tuple is restated at `atlas_server.py:426`, `policies/tool_authorization.py:92`, `routes/config_routes.py:77`, and `mcp_execution.py:819`.
- `atlas_launch` never mentions the observation tools, and the `result_status` vocabulary and size cap are documented only in the schema string.
- The conversation-id mismatch log at the mismatch-diagnostics path skips the sanitizer used one branch above.

---

garland3:
## Summary
Adds observation tools (`atlas_get_runs`, `atlas_result`) for launched sub-conversations, backed by a new per-conversation child index in the run registry and a `result_status` vocabulary for reporting child outcomes.

## Verdict
Request changes.

## Required changes
1. Remove the `registry.reap_terminal()` call from `_scoped_children` at `atlas/application/chat/runs/launcher.py:586` — a read path currently evicts records, so `atlas_get_runs` can delete the run the following `atlas_result` needs (and it re-sweeps the whole registry on every poll). Filter stale records at read time instead.
2. Extract text rather than serializing at `atlas/application/chat/runs/launcher.py:606` — a terminal assistant message with `content: None` (tool-call-only, a normal shape) is returned as the literal string `"null"` with `result_status: "available"`. Return `result_status: "empty"` when no text exists, and join text blocks for list-shaped content.
3. Truncate the extracted text, not the serialized document, at `atlas/application/chat/runs/launcher.py:607-610` — today a clipped JSON fragment is handed back as `available_truncated`, which the schema documents as usable.
4. Split the `unavailable` status at `atlas/application/chat/runs/launcher.py:598,613,620` — a completed child whose transcript has not yet been persisted is retryable, but shares a status with permanently failed children and a missing repository, so the caller stops polling and the result is never read.
5. Guard the `json.dumps(content)` path at `atlas/application/chat/runs/launcher.py:606` (or pass `default=str`) — non-serializable persisted content raises `TypeError` out of a function contracted to always return a status.
6. Wrap both interpolated values of the `conversation_id` mismatch log in `sanitize_for_logging` at `atlas/application/chat/runs/context.py:98` — the branch directly above sanitizes, and this field is producer-supplied, permitting forged log lines.
7. Fix `atlas/tests/test_atlas_launch_tool.py:925` — `assert tool_name == GET_RUNS_TOOL_NAME or child.run_id in result.content` is vacuously true for the `atlas_get_runs` parameter; assert on the parsed `runs` array so the scoping index is actually pinned. Also add cases for `available_truncated` and non-string content.

## Evidence
Evidence: 7/10. Unit coverage is solid for the main paths, but every test seeds registry state by hand and stubs the repository, so nothing demonstrates a real launched child being observed end to end; the reported suite numbers also predate the last two behavior-changing commits.

## Notes
- `children_of_conversation` (`atlas/application/chat/runs/registry.py:255`) takes no `user_email`; the tenant boundary lives only in the caller's post-filter.
- Runs launched from a turn with no ambient run context are silently unobservable — the launch still returns a handle (`atlas/application/chat/runs/launcher.py:530`).
- The read-side logic (`launcher.py:581-694`) would sit better in a sibling `observation.py`; `launcher.py` now owns both halves plus two tool adapters.

---

PR #586, opened 2026-05-07
Fix CodeQL sanitizer model for sanitize_for_logging

garland3:
## Summary
- switch the CodeQL extension to the Python `sanitizerModel` schema for `py/log-injection`
- model `atlas.core.log_sanitizer.sanitize_for_logging` return values as sanitized
- keep existing CodeQL config/workflow wiring intact

Closes #574

PR #946, opened 2026-09-15
feat(auth): grant admin via ADMIN_USERS or dynamic ADMIN_GROUP membership

### LINKED ISSUE #945: Support admin authorization via either ADMIN_USERS or dynamic membership in ADMIN_GROUP
## Summary
Atlas currently supports static admin authorization via `ADMIN_USERS` or `AUTH_STATIC_GROUPS`, but does not support the common deployment pattern where an operator wants to set an admin group name and let external group-membership logic determine membership.

Today, `AUTH_STATIC_GROUPS` requires explicit memberships in `group:user1,user2` format. Supplying only a group name produces a warning like:

```
Ignoring malformed AUTH_STATIC_GROUPS entry (expected 'group:user1,user2'): '<group-name>'
```

This is confusing for deployments that already have non-public group-membership logic and only want Atlas to recognize a configured `ADMIN_GROUP` without restating individual users in public-facing config.

## Current behavior
From the current docs and code:

- `ADMIN_USERS` is sugar for a single `<ADMIN_GROUP>:` static entry.
- `AUTH_STATIC_GROUPS` only accepts explicit static membership mappings.
- When `AUTH_GROUP_CHECK_URL` is configured, it is authoritative and static config is not consulted as an additional grant.

That makes these supported:

- `ADMIN_USERS=user1,user2`
- `AUTH_STATIC_GROUPS=admin:user1,user2`

But not this operator intent:

- set `ADMIN_GROUP=<group-name>`
- rely on external/non-public group membership code to answer whether a user is in that group
- optionally also allow a small static override list of admin users

## Requested behavior
Support both of these at the same time:

1. **Static user list** for admins
   - `ADMIN_USERS=user1,user2`
2. **Named admin group**
   - `ADMIN_GROUP=<group-name>`
   - Atlas should ask the configured group-membership mechanism whether the user belongs to that group

In other words, admin access should be granted if either condition is true:

- user is explicitly listed in `ADMIN_USERS`, or
- user is a member of `ADMIN_GROUP` according to the existing group-membership check path

## Why this helps
This matches real deployment needs:

- some environments want a small emergency/manual admin allowlist
- some environments already manage group membership outside this repository
- operators should not have to encode a whole group as static `group:user1,user2` entries just to make admin work
- Atlas should cleanly support both static and dynamic membership sources

## Possible implementation direction
One reasonable approach:

- Keep `ADMIN_USERS` as a static override/grant.
- Keep `AUTH_STATIC_GROUPS` for explicit static group membership mappings.
- For admin checks, evaluate:
  1. explicit static admin grant via `ADMIN_USERS`
  2. otherwise normal membership check against `ADMIN_GROUP`
- Preserve current behavior for non-admin groups unless explicitly expanded.

This would let deployments use:

```env
ADMIN_GROUP=<group-name>
ADMIN_USERS=user1,user2
```

without forcing `AUTH_STATIC_GROUPS=<group-name>:...`.

## Acceptance criteria
- Atlas accepts a custom `ADMIN_GROUP` name without requiring static member enumeration.
- `ADMIN_USERS` remains supported as a manual override.
- Admin authorization succeeds when either static admin user membership or dynamic membership in `ADMIN_GROUP` is true.
- `AUTH_STATIC_GROUPS` continues to require `group:user1,user2` syntax for explicit static mappings.
- Docs clearly distinguish:
  - static membership mapping
  - admin user override list
  - dynamic group-membership checks

## Notes
This issue intentionally avoids naming any private group or exposing deployment-specific membership logic. It is about supporting the configuration pattern in the public codebase.

garland3:
Closes #945.

## Problem

Admin was reachable in exactly two ways, both of which require restating individual users in public-facing config: `ADMIN_USERS=...`, or `AUTH_STATIC_GROUPS=<admin-group>:user1,user2`. An operator who already has group-membership logic outside this repository and just wants to point Atlas at a group name had no supported path — setting `AUTH_STATIC_GROUPS=<group-name>` produced `Ignoring malformed AUTH_STATIC_GROUPS entry (expected 'group:user1,user2')`.

## Change

Admin access is granted when **either** condition holds:

1. the identity is listed in `ADMIN_USERS`, or
2. the normal membership check says the identity is in `ADMIN_GROUP` (the external authorizer when one is configured, otherwise the static table / debug mock table).

So this now works with no member enumeration anywhere:

```env
ADMIN_GROUP=my-org-atlas-admins
ADMIN_USERS=alice@example.org   # optional override list
```

The substantive behavior change is that `ADMIN_USERS` moves from being a *fallback* (only consulted when no `AUTH_GROUP_CHECK_URL` is set) to being an *override*: it is checked before the external call and grants the configured admin group even when the authorizer denies, errors, or is unreachable. That is the point of an emergency allowlist — it has to keep working when the authorization service is the thing that broke. It is narrowly scoped: it only ever grants, only for `ADMIN_GROUP`, and only for identities the operator typed out by hand.

Everything else is unchanged:

- `AUTH_STATIC_GROUPS` still requires `group:user1,user2`, still yields to a configured authorizer, and still fails closed when `AUTH_GROUP_CHECK_URL` is set without a usable API key. Its malformed-entry warning now points at `ADMIN_GROUP` for the dynamic case.
- Non-admin groups are untouched.
- `ADMIN_USERS` is still unioned into the static table, so the no-authorizer path and the "no authorization source configured" startup warning behave as before.

`admin_user_set` is a new normalized, cached property on `AppSettings`, kept separate from `static_group_members` precisely because the two are consulted under different rules.

## Docs

`docs/admin/authentication.md` gains an "Admin Authorization: `ADMIN_GROUP` + `ADMIN_USERS`" section with a table distinguishing the three mechanisms (dynamic group, admin override list, explicit static mappings) and stating which survive a configured external authorizer. `.env.example` documents `ADMIN_GROUP` and warns against putting a bare group name in `AUTH_STATIC_GROUPS`.

## Tests

New coverage in `atlas/tests/test_core_auth.py`: dynamic `ADMIN_GROUP` resolution through the authorizer (allow and deny, no static config at all); `ADMIN_USERS` granting a renamed admin group against a denying authorizer, case/whitespace-tolerantly, while an unlisted identity is still denied and the override does not leak to other groups; either source sufficing independently; the override surviving a transport failure while every other group still fails closed; and `admin_user_set` normalization. The two pre-existing "external authorizer wins" tests now assert that invariant via `AUTH_STATIC_GROUPS` (where it still holds unchanged) rather than via `ADMIN_USERS`.

Full backend suite: 3373 passed, 17 skipped.


garland3:
## Summary
Makes `ADMIN_USERS` an override that grants the `ADMIN_GROUP` even when an external authorizer is configured and denies, adds a normalized `admin_user_set` on `AppSettings`, and updates tests, `.env.example`, CHANGELOG and the authentication docs.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10 — the behavior change is stated plainly and each decision path has a named test, with a full backend suite pass reported. The one gap is any end-to-end demonstration that `ADMIN_GROUP` travels from `.env` through settings into a live admin-gated request; worth adding but not required here.

## Notes
- Log the override grant (identity + group + `ADMIN_USERS` as source) before `return True` at `atlas/core/auth.py:117`, mirroring the adjacent `SKIP_AUTHORIZATION_CHECKS` warning — otherwise break-glass admin actions are indistinguishable from real group grants in logs.
- Add an upgrade note to `CHANGELOG.md:9` and `docs/admin/authentication.md:273`: deployments with both `ADMIN_USERS` and `AUTH_GROUP_CHECK_URL` set had inert, possibly stale entries that now grant admin; also retitle the stale "Static Configuration: `ADMIN_USERS` / `AUTH_STATIC_GROUPS`" section at `docs/admin/authentication.md:305`.
- Pin the URL-set-without-API-key case in `atlas/tests/test_core_auth.py:560` — its docstring claims `ADMIN_USERS` is covered there, but no test asserts the override's behavior in that fail-closed state.

---

garland3:
## Summary
Makes `ADMIN_USERS` an authorizer-independent override for `ADMIN_GROUP` membership rather than a no-authorizer fallback, adds a cached normalized identity set in settings, and ships unit tests plus an end-to-end validation script.

## Verdict
Request changes.

## Required changes
1. Capture pytest's exit status before the pipe (or `set -o pipefail`) at `test/pr-validation/test_pr946_admin_group_auth.sh:255` — `print_result $?` currently reads `tail`'s status, so step 7 reports PASSED even when every test in `atlas/tests/test_core_auth.py` fails.
2. Log a warning on the override grant before `return True` at `atlas/core/auth.py:121`, recording identity, group, and `ADMIN_USERS` as the source — the adjacent `SKIP_AUTHORIZATION_CHECKS` bypass logs for the same reason, and without it an emergency grant is indistinguishable from real group membership after an incident.
3. Emit a startup warning when `admin_user_set` is non-empty and `auth_group_check_url` is set (`atlas/modules/config/settings.py:391`), and call the upgrade action out as its own line in `CHANGELOG.md:10` — previously inert `ADMIN_USERS` entries, including de-provisioned staff, become live admin grants on restart with no signal.
4. Add a test in `atlas/tests/test_core_auth.py` for `AUTH_GROUP_CHECK_URL` set without `AUTH_GROUP_CHECK_API_KEY`, asserting the `ADMIN_USERS` identity is granted and all other identities/groups still fail closed — `test_static_config_ignored_when_authorizer_url_set_without_api_key:560` claims this is "covered above" but no test exercises that branch with the override.
5. Add `monkeypatch.delenv("ADMIN_USERS", raising=False)` (and `AUTH_STATIC_GROUPS`) to `_external_authorizer_env` at `atlas/tests/test_core_auth.py:154` — the authorizer-is-authoritative tests now depend on ambient config, including values loaded from `../.env`.
6. Use `normalized_user` / `normalized_group` in the debug mock-table branch at `atlas/core/auth.py:168` — debug mode still compares raw values, so the documented case/whitespace tolerance holds on only part of the function.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size, with one named unit test per decision path and a real end-to-end script against a live HTTP authorizer; the missing artifacts are the URL-without-API-key case and any sample of the audit signal on the override path (items 4 and 2 above).

---

garland3:
Addressed the Codex review in the latest push:

- **Changelog** — already landed in `c97f7997`, pushed just after the commit Codex reviewed.
- **PR validation script** — added `test/pr-validation/test_pr946_admin_group_auth.sh` (+ `fixtures/pr946/`). It is genuinely end-to-end: it drives `GET /admin/banners` on the real app over HTTP *and* starts a real local HTTP authorization service, so the outbound membership POST is actual network I/O rather than a mock. Cases: dynamic `ADMIN_GROUP` membership with nothing enumerated anywhere (authorizer "yes" → 200, "no" → 403); the `ADMIN_USERS` override against a denying authorizer; the same override against a *closed port* (the emergency case); the override's scoping to `ADMIN_GROUP` only; and `AUTH_STATIC_GROUPS` keeping both its `group:user1,user2` syntax and its deference to the authorizer. 10 passed, 0 failed.
- **Date stamp** — `docs/admin/authentication.md` now reads `Last updated: 2026-09-15`.

For the human reviewer, the one judgement call worth a second opinion: `ADMIN_USERS` moves from *fallback* to *override*, i.e. it now grants `ADMIN_GROUP` even when `AUTH_GROUP_CHECK_URL` is configured and denies. An earlier review (the `URL`-without-`API_KEY` fail-closed test) deliberately made static admin fail closed in that situation. I kept that behaviour for `AUTH_STATIC_GROUPS` and carved out only `ADMIN_USERS`, on the reading that an "emergency/manual admin allowlist" is worthless if it stops working exactly when the authorization service is the thing that broke. If you'd rather `ADMIN_USERS` keep yielding to a healthy authorizer, the new branch in `auth.py` can be gated on `not auth_url` — say the word and I'll flip it.

garland3:
Addressed all six items from the AGENT-REVIEW-BOT-3 review in `f3c4746a`. All were real — thanks.

1. **`print_result $?` read `tail`'s status** (`test_pr946_admin_group_auth.sh` step 7) — the script now captures pytest's own exit status into a variable before echoing. It genuinely would have reported PASSED through a fully red suite.
2. **Audit log on the override** — a grant that actually overrides a configured authorization service now logs a warning naming the identity, the group and `ADMIN_USERS` as the source, mirroring the `SKIP_AUTHORIZATION_CHECKS` warning beside it. With *no* authorizer configured it logs at info instead: there `ADMIN_USERS` is simply how admins are set up, and a warning on every admin request would train operators to ignore the line. Both paths are now asserted.
3. **Startup warning + upgrade note** — `AppSettings` warns when `ADMIN_USERS` and `AUTH_GROUP_CHECK_URL` are both set, naming the count and the admin group. The `CHANGELOG.md` entry carries the upgrade action as its own bullet, and `docs/admin/authentication.md` carries it as a blockquote: previously-inert entries, de-provisioned staff included, become live grants on the restart that picks up this version.
4. **`URL` without `API_KEY`, with the override in play** — new test asserts the `ADMIN_USERS` identity is granted there while every other identity and group (including a statically-mapped admin) still fails closed. The stale "covered above" docstring is corrected.
5. **Ambient config in `_external_authorizer_env`** — now clears `ADMIN_USERS` and `AUTH_STATIC_GROUPS`. Given this repo's history with `.env`-driven test failures, that was worth fixing regardless.
6. **Raw comparisons in the debug branches** — the mock table and the `test_user` check compared raw values while everything above them compared normalized ones. Both normalize now, with a test covering it.

Verification: full backend suite **3377 passed, 17 skipped**; validation script **10/10**; `scripts/check-docs.sh` OK.

On the earlier review's evidence note about end-to-end coverage — that is what `test/pr-validation/test_pr946_admin_group_auth.sh` now does: `ADMIN_GROUP` travels from a fixture `.env` through settings into a live `GET /admin/banners`, against a real local HTTP authorization service.

garland3:
**On the CodeQL `py/log-injection` annotations on `atlas/core/auth.py`** — the CodeQL check itself passes; these are advisory annotations. `43d80246` routes the identity and group name through `core.log_sanitizer.sanitize_for_logging` (and gives the pre-existing `SKIP_AUTHORIZATION_CHECKS` warning in the same function the same treatment), so the code is doing the right thing. CodeQL still flags it, and while chasing why I found something worth a separate look:

`.github/codeql/extensions.yml` registers `sanitize_for_logging` as a `log-injection` sanitizer — added in #586 precisely to stop these alerts — but **it is never loaded**. `codeql-config.yml` has no `packs:` entry referencing it, there is no `qlpack.yml` making it a loadable CodeQL pack, and `security.yml` passes only `config-file`. So the model is dead config, which would explain why these alerts persist repo-wide despite call sites sanitizing correctly.

I have deliberately **not** fixed that here — it is unrelated to #945, it touches the security workflow, and getting the pack wiring wrong risks the CodeQL job for every PR. Flagging it so someone can pick it up separately; happy to open an issue if useful.

One note on the sanitization itself, since it is easy to misread: on the `ADMIN_USERS` path this is defence in depth, not the control. Normalization strips only the *edges*, so an identity carrying an interior newline (`alice@example.org\nmallory`) is not the allowlisted identity and is denied outright — there is a test for that, separate from the test that the log line is newline-free.

garland3:
## Summary
Makes admin authorization resolve dynamically against `ADMIN_GROUP` via the external authorization service, and promotes `ADMIN_USERS` from a fallback to a break-glass override that is checked first, with normalized identity matching, per-grant audit logging, a startup warning, and updated docs.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — per-branch unit tests plus an end-to-end script against a live local authorization service; the only notable gap is that the claimed results are summarized rather than shown, so pasting the validation script's stdout and the pytest summary line would close it.

## Notes
- Normalization stops at the external authorizer — `auth.py:145` posts raw `user_id`/`group_id` while the docstring and `docs/admin/authentication.md` promise case/whitespace tolerance generally; send the normalized values or scope the documented claim to locally-evaluated sources.
- The `ADMIN_USERS` grant logs on every admin-gated request (`auth.py:130`, warning with an authorizer, info without) and the "startup" warning re-fires on every `AppSettings` construction (`settings.py:1125`); de-duplicate both so the audit signal stays readable.
- Debug-mode mock table keys on `""` when `ADMIN_TEST_USER`/`TEST_USER` is set empty, so a blank or `None` identity is granted admin (`auth.py:196`); skip empty normalized keys.

---

garland3:
Thanks — two of the three nits were worth fixing, and one of them was a real bug. Fixed in `161f2ab6`.

**Blank identity granted admin in debug mode.** Correct and the sharpest of the three: an empty `ADMIN_TEST_USER`/`TEST_USER` keyed the mock table on `""`, so a blank or missing identity got admin — the one input an unauthenticated request is most likely to arrive with. Empty keys are now dropped and a blank identity is denied outright on both debug branches, with a test covering `""`, `"   "` and `None`.

**Audit line fired per request.** Also correct — admin routes are polled by the UI, so the signal was buried in its own repetition. Now throttled to once per identity and group per hour via a bounded TTL table, in the style of the ALB key cache above it. A test asserts five checks for one identity produce one line while a second identity still logs.

**Normalization stops at the external authorizer.** This one is intentional, so I took your second option and scoped the claim rather than changing the payload: the service owns its own matching rules, and rewriting the identity Atlas asks about could change its answer — especially for a deployment whose directory is case-sensitive. The docstring and `docs/admin/authentication.md` now say the tolerance covers the sources Atlas evaluates itself (the override, the static table, the `users` short-circuit, the debug mock table) and explicitly not the endpoint, and a new test pins the raw POST body.

**One I'm not taking:** de-duplicating the *startup* warning. It fires on `AppSettings` construction, which in production is startup and config reload — not a hot path. Process-global de-dup state would make the warning order-dependent across tests for no real-world gain. Happy to be overruled.

On the evidence note — here is the actual output rather than a summary:

```
$ bash test/pr-validation/test_pr946_admin_group_auth.sh
Passed: 10
Failed: 0

$ bash test/atlas_tests.sh
3382 passed, 17 skipped, 2 warnings in 92.82s

$ bash scripts/check-docs.sh
check-docs: OK (118 markdown files, 18 indexes, no orphans or broken links)
```

garland3:
## Summary
Changes `ADMIN_USERS` from a fallback consulted only when no external authorization service is configured into an override evaluated ahead of that service, scoped to `ADMIN_GROUP` and grant-only, with a throttled audit log line, a startup warning, identity normalization, and updated operator docs.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size: a named unit test per decision path, an end-to-end script driving `GET /admin/banners` against a live local authorizer, and verbatim command output. The one gap worth closing is that no rendered sample of the audit/startup log lines appears anywhere, even though post-incident forensics is the whole justification for them.

## Notes
- Audit-line throttle caches the log *decision*, not its severity, so an authorizer configured mid-TTL suppresses the warning for up to an hour — include `bool(auth_url)` in the cache key (`atlas/core/auth.py:169`).
- Debug mock table grants an empty group name when `ADMIN_GROUP` is set empty; mirror the blank-identity guard with `if not normalized_group: return False` (`atlas/core/auth.py:236`).
- `assert caplog.text == ""` / `caplog.records[0]` collect propagated records from every logger and can fail spuriously on settings warnings; filter to `r.name == "atlas.core.auth"` (`atlas/tests/test_core_auth.py`).

---

garland3:
All three correct — fixed in `5322cc33`.

- **Throttle cached the decision, not its severity.** Right, and a real hole in the audit signal: a deployment that configures an authorizer partway through a TTL would have had its first override *warnings* swallowed by an info-level entry cached before that authorizer existed. `bool(auth_url)` is now part of the key, with a test that walks exactly that transition.
- **Blank group name.** Also right, and it needed to go higher than suggested: the `test_user` branch grants the blank group before the mock table is ever reached, so the guard is hoisted above every debug branch rather than sitting next to `mock_groups.pop("", None)`. Test covers `""`, `"   "` and `None` for both mock identities.
- **`caplog` collects propagated records from every logger.** Correct, and I'd already been bitten by it twice writing these very tests — a settings startup warning failed them spuriously both times. All three assertions now filter on `r.name == "atlas.core.auth"`.

On the evidence gap: `docs/admin/authentication.md` now shows both lines as they actually render, which is more useful to an operator grepping logs after an incident than my describing them.

```
$ bash test/atlas_tests.sh
3384 passed, 17 skipped, 2 warnings in 93.28s

$ bash test/pr-validation/test_pr946_admin_group_auth.sh
Passed: 10
Failed: 0

$ bash scripts/check-docs.sh
check-docs: OK (118 markdown files, 18 indexes, no orphans or broken links)
```

I'm going to call this the last automated round unless something substantive comes up — the findings have moved from real bugs to polish, and the remaining decision on this PR is a human one: whether promoting `ADMIN_USERS` from fallback to break-glass override is the semantics you want.

garland3:
## Summary
Makes `ADMIN_USERS` a break-glass override evaluated ahead of the external authorization service, adds throttled per-grant audit logging and a startup warning when both are configured, and tightens blank identity/group handling in `is_user_in_group`.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — per-branch unit tests, an end-to-end script against a real local authorization service, and verbatim command output. The one gap is that the validation script's ten cases predate this head commit, so the blank-identity/throttle-severity fixes are covered only by unit tests.

## Notes
- `ADMIN_USERS` is now un-revocable by the authorization service; the trade-off is documented, but consider gating it behind an explicit flag or honouring it only when the authorizer actually fails — `atlas/core/auth.py:117`.
- Blank identity still matches the `users` short-circuit above the new blank guard, so `is_user_in_group("", "users")` returns `True` in production — `atlas/core/auth.py:235`.
- `.env.example:46` documents `ADMIN_USERS`/`ADMIN_GROUP` under a "consulted after AUTH_GROUP_CHECK_URL" header that the new ordering contradicts.

---

garland3:
Two of the three fixed in `4f7302d3`; the third is the open design question, not a nit.

- **Blank identity still matched the `users` short-circuit.** Correct, and the better catch of the two — my guard sat *below* it, so `is_user_in_group("", "users")` was still `True` and an identity-less request landed in the group most features check. Guard hoisted above it. Routes authenticate before asking so this is defence in depth, but "nobody" should not be a member of anything. The full suite is unchanged at 3384 passing, so nothing relied on the old behaviour.
- **`.env.example` header contradicted the new ordering.** Mine to fix — it still said "consulted after AUTH_GROUP_CHECK_URL". It now states the actual order and notes that the external service receives the raw values rather than the normalized ones.
- **Evidence gap.** Fair: the script's cases predated the last two commits. It has two more now — blank identity/group denied everywhere including `users`, and five checks for one identity producing exactly one audit line. 12 checks, all passing.

On **gating the override behind a flag, or honouring it only when the authorizer actually fails**: that is the core design question of this PR rather than a nit, and I don't think an automated round should decide it. "Only when the authorizer fails" specifically has a sharp edge — an authorizer that is reachable and answering `false` because a group sync broke is the likeliest way to lock every admin out, and that case looks healthy from here. Flagging it for whoever reviews this; the change is small either way and I'll implement whichever they want.

```
$ bash test/atlas_tests.sh
3384 passed, 17 skipped, 2 warnings in 92.16s

$ bash test/pr-validation/test_pr946_admin_group_auth.sh
Passed: 12
Failed: 0
```

CI was green on `5322cc33` (all checks passing) before this push.

garland3:
## Summary
Reworks `is_user_in_group` so `ADMIN_GROUP` is resolved dynamically, `ADMIN_USERS` acts as an unconditional break-glass override ahead of the external authorizer, identities and groups are normalized at the boundary, and blank identity/group is denied — with a throttled audit log line, a startup warning, docs, and a CHANGELOG upgrade note.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well demonstrated — per-branch unit tests plus an end-to-end validation script against a real local authorization service, with verbatim output pasted. The only gap worth closing is an HTTP-level case for a request with an empty `X-User-Email` on an ordinary `users`-gated route, which is the input the latest commit was written for.

## Notes
- Docs and `.env.example` still describe `ADMIN_USERS` as "consulted after `AUTH_GROUP_CHECK_URL`", which no longer matches the code — `docs/admin/authentication.md`, `.env.example`.
- Blank-identity guard sits inside the no-authorizer branch, so a configured authorizer is still POSTed `{"user_id": ""}`; hoist it above the dispatch — `atlas/core/auth.py:235`.
- Negative case of the startup-warning test passes `auth_group_check_url=None` under the field name, which the `validation_alias` drops, so it reads the ambient environment — `atlas/tests/test_core_auth.py`.

---

PR #949, opened 2026-09-16
feat: add ATLAS launch option discovery

### LINKED ISSUE #944: Add an ATLAS self-query tool for pre-launch workspace and LLM discovery
## Summary
There needs to be a new ATLAS tool that lets the AI self-query what launch options are available before launch, so it knows how to construct a valid launch.

## Background
A recent PR got the atlas launch tool working. The remaining gap is pre-launch discovery: the AI should not have to guess valid launch arguments or rely on hardcoded assumptions. Before launch, it needs a way to ask the system what workspaces and LLM/model options are currently available.

## Requested behavior
- Add a dedicated ATLAS tool for pre-launch discovery
- The tool should let the AI query available workspaces
- The tool should let the AI query available LLMs/providers/models
- The tool should be usable as a self-query step before launch so the AI can choose or validate launch arguments from real available options
- Launch should depend on this discovery step instead of assuming values
- If discovery fails or returns no valid choices, ATLAS should block launch and return a clear error
- Surface discovery/loading/failure states clearly in the UI or API

## Why
The launch arguments require these values. A dedicated self-query tool makes the launch flow reliable because the AI can discover the current valid options before attempting launch. That keeps the launch tool as the source of truth and avoids invalid launches caused by guessed parameters.

## Acceptance ideas
- A new ATLAS tool exists specifically for pre-launch self-query of launch options
- The AI can call it to list available workspaces before launch
- The AI can call it to list available LLM/provider/model choices before launch
- Launch arguments are built from or validated against the discovered values
- Launch is blocked with a clear error when discovery cannot supply required inputs
- Basic verification covers both successful discovery and failure handling

garland3:
## Summary
- add an ATLAS self-query tool for authorized launch workspaces and provider/model choices
- require discovery-backed launch arguments and fail clearly when discovery is unavailable or empty
- cover successful discovery and fail-closed launch behavior

Closes #944

## Verification
- pytest -q atlas/tests/test_atlas_launch_tool.py atlas/tests/test_atlas_consolidated_tools.py
- ./test/run_tests.sh backend (backend suite passed; live E2E phase exceeded local timeout)
- cd frontend && npm test -- --run
- cd frontend && npm run lint
- cd frontend && npm run build


garland3:
## Summary
Adds a new built-in `atlas_discover_launch_options` tool and makes a successful discovery call a hard precondition for `atlas_launch`, threading the discovered options through the per-tool-call session context.

## Verdict
Block — critical issues found.

## Required changes
1. Add `DISCOVER_LAUNCH_OPTIONS_TOOL_NAME` to the launch gate in `ToolAuthorizationService.filter_authorized_tools` (`atlas/application/chat/policies/tool_authorization.py:82`) — built-in `atlas_*` tools with no explicit gate fall through to the MCP-server prefix branch and are dropped, so the discovery tool never reaches the model and every `atlas_launch` is refused.
2. Make the discovery state survive an agent-mode turn — hoist the `session_context` literal at `atlas/application/chat/runs/agentic_loop.py:322` above the `while` loop (or move `launch_discovery` into the per-turn scratchpad alongside `TURN_BUDGET_KEY`) — today it is rebuilt every step, so discovery in step N is invisible to a launch in step N+1.
3. Handle discovery and launch emitted in the same step at `atlas/application/chat/utilities/tool_executor.py:84-95` — `asyncio.gather` can run the launch before discovery populates the shared dict, refusing nondeterministically; re-run discovery when the cache is empty instead of refusing.
4. Ensure `atlas_launch` cannot be selected without discovery (`atlas/application/chat/runs/launcher.py:577`) — emit the discovery schema whenever launch is requested, or have `launch_sub_conversation` self-discover; otherwise a user following the guide's "tick `atlas_launch`" instruction hits an unrecoverable loop, as do saved conversations with persisted tool selections.
5. Make the discovery payload and the validator agree at `atlas/application/chat/runs/launcher.py:243` — `_require_discovered_option` accepts only `name`, while discovery publishes `provider` and `model`; accept those forms or stop emitting them.
6. Add a test that drives discovery and then launch through the execution entry point against one shared `session_context`, plus coverage of the two mismatch refusals in `_require_discovered_option` (`atlas/application/chat/runs/launcher.py:233-246`) — the current suite pre-seeds `_DISCOVERY` and would stay green if the membership check were made permissive.
7. Update the launch section of `docs/user-guide/tools-and-settings/README.md:127-137` and add the tool constant to `frontend/src/constants/atlasTools.js`, with a CHANGELOG upgrade note for deployments already running `FEATURE_ATLAS_LAUNCH_ENABLED`.

## Evidence
Evidence: 7/10 — three new unit tests plus a stated backend run give a proportionate baseline for a backend-only change, but the live E2E phase is reported as timing out and nothing exercises the tool end to end. The single most valuable addition is one test that calls the tool by name through `execute_launch_discovery_tool` and then launches with the same context object.

## Notes
- Discovery serializes one uncached auth round-trip per group-scoped model; parallelize `filter_authorized_models` (`atlas/application/chat/runs/launcher.py:172`).
- Refusals from `_require_discovered_option` emit no log or metric, leaving no signal for "launches stopped working after upgrade" (`atlas/application/chat/runs/launcher.py:222-246`).
- Discovery returns the raw backing `provider`/`model` identifiers that `/api/config` deliberately withholds (`atlas/application/chat/runs/launcher.py:176`).

---

garland3:
Addressed the review findings in commits 9ff8660 and 60bd77f1: discovery state now persists in the agent loop and is cleared on refresh failure, launch requires discovery on every execution path, the owning app factory is forwarded, provider/model values derive from configured model names, workspace reads use , launch authorization/schema selection includes discovery automatically, same-step discovery runs before launch, docs/frontend constants/changelog are updated, and execution-path coverage was added. Targeted verification: 72 tests passed; changed Python files pass Ruff.

garland3:
## Summary
Adds an `atlas_discover_launch_options` built-in tool that publishes the caller's authorized workspaces and models into per-turn session state, and makes a successful discovery a hard precondition for `atlas_launch`, with the discovery tool auto-included in the authorized set whenever launch is selected.

## Verdict
Request changes.

## Required changes
1. Order discovery before launch within a step at `atlas/application/chat/utilities/tool_executor.py:84` — `execute_multiple_tools` gathers all calls concurrently, so a step containing both tools refuses the launch or not depending on scheduling; alternatively have `_require_discovered_option` fall back to running discovery itself when the shared state is empty.
2. Clear `context["launch_discovery"]` in place at the start of `discover_launch_options` (`atlas/application/chat/runs/launcher.py:145`) and repopulate only on success — every failure branch currently raises before touching state, so a failed refresh leaves stale options authorizing a later launch.
3. Reconcile the discovery payload with the validator at `atlas/application/chat/runs/launcher.py:176-186` and `:233` — entries publish `name`, `provider` and `model` but only `name` is accepted, so a caller using a returned `model` value gets a refusal that re-running discovery cannot fix. Either accept those forms or emit `name` only and say so in the tool description.
4. Add a test that starts from a context with no `launch_discovery` key, runs the discovery tool and then `atlas_launch` through the tool-execution entry point against one shared `session_context`, plus cases for the unknown-workspace and unknown-model refusals in `_require_discovered_option` (`atlas/application/chat/runs/launcher.py:233-246`).
5. Stop mutating the module-level `_DISCOVERY` dict in place at `atlas/tests/test_atlas_launch_tool.py:33-36` — build a fresh copy per test so ordering does not affect results.
6. Update the launch section of `docs/user-guide/tools-and-settings/README.md:127-137` to state that selecting `atlas_launch` also enables `atlas_discover_launch_options` and what happens when discovery fails; land or drop the claimed `frontend/src/constants/atlasTools.js` change, which is not in the diff.

## Evidence
Evidence: 7/10 — unit coverage and a named backend test run are present, but the live E2E phase is reported as having timed out and every launch test pre-seeds the discovery state rather than producing it. The single most useful missing artifact is a transcript or test of one turn that discovers and then launches through the real execution path.

## Notes
- Introduce a shared `LAUNCH_DISCOVERY_KEY` constant instead of the bare `"launch_discovery"` literal repeated across launcher, tool executor and agentic loop (`atlas/application/chat/runs/launcher.py:212`).
- Discovery runs one uncached authorization round-trip per group-scoped model and recomputes the workspace listing on every call (`atlas/application/chat/runs/launcher.py:172`).
- `ATLAS_SERVER_DESCRIPTION` now reads as a run-on with two bare `and`s (`atlas/modules/mcp_tools/atlas_server.py:325-330`); CHANGELOG entries are out of order (`CHANGELOG.md:12-24`).

---

garland3:
## Summary
Adds an `atlas_discover_launch_options` tool and makes a successful discovery a hard precondition for `atlas_launch`, wiring the pairing through tool-name normalization, the authorization filter, and the tool executor (discovery calls now run before the rest of a step).

## Verdict
Request changes.

## Required changes
1. Accept the identifiers discovery publishes, or stop publishing them — `atlas/application/chat/runs/launcher.py:233`: `model_options` emits `name`/`provider`/`model` (launcher.py:176-186) but `_require_discovered_option` matches on `name` only, so a model configured as `the reviewer/the reviewer-4o` is refused when the caller passes back the `model` value it was just handed. Either add the `model` and `provider/model` forms to `model_names`, or emit `name` only and state in `DISCOVER_LAUNCH_OPTIONS_TOOL_DESCRIPTION` that `name` is the sole valid `model` argument.
2. Make the refusal recoverable in standard tools mode — `atlas/application/chat/modes/tools.py:416`: the anti-loop guard keys on `(name, arguments)` and `atlas_discover_launch_options` takes no parameters, so the retry the refusal text demands is always dropped as a duplicate and the turn ends. Exempt the discovery tool from the signature set, or have `_require_discovered_option` self-heal by calling `discover_launch_options` when the shared state is empty (which also removes the need for the executor pre-pass and the selection-list injection).
3. Deduplicate and parallelize the authorization sweep — `atlas/core/model_access.py:120-122` via `atlas/application/chat/runs/launcher.py:172`: discovery awaits `is_model_allowed` per model, each awaiting `is_user_in_group` per group, and every group check opens a fresh `httpx.AsyncClient` (`atlas/core/auth.py:194-201`). Resolve the distinct group set once with a single `asyncio.gather` and reuse a shared client, so a 20-model deployment does not pay ~40 serial round-trips on each launch turn.
4. Route discovery calls through `asyncio.gather(..., return_exceptions=True)` at `atlas/application/chat/utilities/tool_executor.py:92` — awaiting them directly lets a pre-`try` failure in `execute_single_tool` propagate out of `execute_multiple_tools`, breaking the "callers always get a list" contract and discarding sibling results; gathering also removes the serialized latency added to every other tool in the step.
5. Add a test in `atlas/tests/test_atlas_launch_tool.py` passing `[launch_call, discovery_call]` (launch first) through `execute_multiple_tools` with one fresh `session_context` and asserting both succeed and land at their input indices — the new same-step ordering is otherwise unexercised.
6. Add a test asserting `context["launch_discovery"] == {}` after a failed refresh and that a subsequent launch with that context is refused — the stale-state clearing at `atlas/application/chat/runs/launcher.py:153-155` can be deleted today with the suite still green. Cover the unknown-workspace and unknown-model refusal branches at `launcher.py:233-246` in the same file.
7. Stop publishing raw routing identifiers — `atlas/application/chat/runs/launcher.py:176`: `provider`/`model` are split from `model_name` (e.g. `azure/prod-gpt4o-eastus2-dep3`), which `/api/config` deliberately withholds, and they are persisted into the transcript and forwarded to sub-conversations. Emit the display `name` (and `description`, if a label is needed).

## Evidence
Evidence: 7/10 — the change is well described and unit-tested at the function level, but no test or transcript drives a real discovery-then-launch turn through `execute_multiple_tools` or the agent loop, and the reported `./test/run_tests.sh backend` run is annotated as having exceeded its timeout during the live E2E phase. Re-run or scope that E2E phase and say what it covered.

---

garland3:
## Summary
Adds an `atlas_discover_launch_options` tool and makes a successful discovery a hard precondition for `atlas_launch`, with turn-scoped discovery state threaded through the agent loop, a discovery-first pre-pass in the tool executor, automatic pairing of the two tools during name normalization/authorization, and an anti-loop exemption in tools mode.

## Verdict
Request changes.

## Required changes
1. Wrap the discovery pre-pass in `atlas/application/chat/utilities/tool_executor.py:93` in the same exception protection as the main gather (`asyncio.gather(..., return_exceptions=True)`, or a `try/except` storing the exception at the index) — `execute_single_tool` does work outside its `try` (name read, `get_server`, `start_span`), so a malformed tool call raises out of `execute_multiple_tools`, aborting every sibling call and leaving the transcript without the required `tool_call_id` messages.
2. Make the discovery payload and its validator agree at `atlas/application/chat/runs/launcher.py:176` and `:233` — drop `provider`/`model` from the emitted entries (or accept them in `model_names`), since a caller that passes back the field literally named `model` is refused with advice ("run discovery again") that can never resolve the refusal.
3. Stop emitting backing routing identifiers to the model at `atlas/application/chat/runs/launcher.py:176` — `split_provider` exposes the configured `model_name` (e.g. an Azure deployment name) that `/api/config` deliberately withholds, and it is persisted in transcripts and forwarded to sub-conversations. Emitting only `name` (+ optional `description`) resolves this together with item 2.
4. Bound the repeat cost of discovery at `atlas/application/chat/modes/tools.py:448` — serve repeat no-arg calls from the existing `launch_discovery` state (or allow one exempted retry per turn), because the exemption also removes the `if not fresh: break` termination path, so a model looping on discovery burns the full round budget re-running the authorization sweep.
5. Parallelize the authorization sweep behind `atlas/application/chat/runs/launcher.py:172` — `filter_authorized_models` awaits `is_model_allowed` serially and each group check opens a fresh `httpx.AsyncClient`, so a deployment with ~20 group-scoped models adds seconds of user-visible latency to every launch turn. Resolve the distinct group set once with a shared client and memoize `(user, group)` results.
6. Add a test driving `execute_multiple_tools` with `[launch_call, discovery_call]` against one fresh `session_context` that has no `launch_discovery` key, asserting both results land at their input indices and that the launch succeeds — today the pre-pass and the `zip(remaining, gathered)` re-scatter can both be deleted with the suite green.
7. Add coverage for the enforcement branches: the unknown-workspace and unknown-model refusals at `atlas/application/chat/runs/launcher.py:233-246`, and the stale-state clearing at `:153-155` (seed valid discovery, fail a refresh, assert `context["launch_discovery"] == {}` and that a subsequent launch is refused).

## Evidence
Evidence: 6/10. The change is well described and unit-tested at the function level, but the reported `./test/run_tests.sh backend` run is annotated as having timed out in exactly the live E2E phase that would exercise a real turn; the most important missing artifact is a completed run (or a scoped launch-test run with the command and summary line shown).

## Notes
- Keep `normalize_tool_names` pure and move the launch/discovery pairing into a named policy function or data mapping — `atlas/modules/mcp_tools/atlas_server.py:376`.
- `DISCOVER_LAUNCH_OPTIONS_TOOL` is exported but has no consumer in `frontend/src`; the docs also do not mention that ticking `atlas_launch` auto-enables a second visible tool — `frontend/src/constants/atlasTools.js:18`, `docs/user-guide/tools-and-settings/README.md:128-137`.
- Unrelated deletions (`.env.otel.example`, `.gemini/settings.json`, archived screenshots) and the out-of-order `CHANGELOG.md:12-24` entries belong in a separate commit.

---

garland3:

garland3:
## Summary
Adds an `atlas_discover_launch_options` tool that lists the user's authorized workspaces and models into per-turn session state, and makes `atlas_launch` refuse unless its arguments match that state. Supporting changes: a discovery-first pre-pass in `execute_multiple_tools`, auto-injection of the discovery tool whenever `atlas_launch` is selected, an anti-loop exemption so a failed discovery can be retried in tools mode, and an unrelated change making toast actions await async handlers.

## Verdict
Request changes.

## Required changes
1. Exclude exempted discovery retries from `repeated_ids` at `atlas/application/chat/modes/tools.py:417` — the retry is re-executed but its tool message is replaced with "(skipped: identical tool call already executed this turn)", so the model never sees the discovered options and the launch keeps failing, defeating the exemption. Derive `repeated_ids` from the complement of `fresh`, and add a tools-mode test where the first discovery fails, the model repeats it, and the second round's tool message carries the real content.

## Evidence
Evidence: 6/10. Function-level unit tests exist, but the new control flow is undemonstrated: nothing exercises the discovery-first pre-pass in `execute_multiple_tools`, the unknown-workspace and unknown-model refusals, or the stale-state clearing after a failed refresh, and the launch tests pre-seed the discovery state they claim to produce. The most important missing artifact is a test driving `[launch_call, discovery_call]` through `execute_multiple_tools` against one fresh `session_context`.

## Notes
- Failing discovery stays exempt from the anti-loop guard for the whole turn and re-runs the workspace query plus a serial per-model, per-group authorization sweep each round; cap exempted retries at one per turn and memoize group checks (`atlas/application/chat/modes/tools.py:424`, `atlas/application/chat/runs/launcher.py:178`).
- Tool description, user guide and CHANGELOG say discovery returns "provider/model" choices, but `model_options` emits `name` only; also document that ticking `atlas_launch` auto-enables the discovery tool (`atlas/modules/mcp_tools/atlas_server.py:121`, `docs/user-guide/tools-and-settings/README.md:131`).
- The launch/discovery pairing and the `"launch_discovery"` key are duplicated across several modules, and `normalize_tool_names` is no longer a pure normalizer; consider having the launcher run discovery itself when state is empty so the executor and anti-loop guard can stay tool-agnostic (`atlas/application/chat/utilities/tool_executor.py:86`, `atlas/modules/mcp_tools/atlas_server.py:421`).

---

garland3:
## Summary
Adds a built-in `atlas_discover_launch_options` tool that lists the user's saved workspaces and authorized models. `atlas_launch` now refuses to run until discovery has succeeded in the same turn and the requested workspace and model appear in its results. Discovery is paired with launch automatically in schema building and authorization, and the executor, tools mode and agent loop are updated to run discovery first and keep its state for the whole turn.

## Verdict
Approve with nits.

## Evidence
Evidence: 6/10. The discovery and launch functions have their own tests. The new control flow across modules does not: executor ordering, the tools-mode retry exemption and agent-loop state carry-over are untested, and the live E2E phase timed out. The most important missing artifact is a test or transcript of one real turn that runs discovery, launches, and returns a child run handle.

## Notes
- **Tools mode hides the result of a retried discovery.** `repeated_ids` comes only from `executed_signatures`, so a discovery retry that the exemption let through is still replaced with the "skipped" note. Build it from the calls that are not in `fresh`, and add a failed-then-retried discovery test. `atlas/application/chat/modes/tools.py:417`
- **A discovery that keeps failing re-runs every round.** Each re-run repeats the serial, uncached group-authorization sweep. Allow at most one exempted retry per turn, and memoize `(user, group)` lookups within a sweep. `atlas/application/chat/modes/tools.py:424`, `atlas/application/chat/runs/launcher.py:179`
- **Tool description and docs promise "provider/model" choices.** The tool returns only model `name`s. Remove "provider/" so the model doesn't build `provider/model` strings that the launch then refuses. `atlas/modules/mcp_tools/atlas_server.py:122`

---

garland3:
Addressed the review in `90cdb507` (fix + tests), `16471794` (docs), plus two merges of `origin/main` (`de4cb91d`, `d89287fb`) that unblocked CI — the PR had been `DIRTY` since #956 landed, so no checks were being scheduled at all.

**Required change 1 — exempted discovery retries were still in `repeated_ids`.** Confirmed against current code. `repeated_ids` is now the complement of what is actually re-executed: both it and `fresh` come from one `_is_fresh` predicate, so an exempted call can never land in both. New tools-mode test `test_a_retried_discovery_reports_its_real_options_not_the_cached_note` — first discovery fails, the model repeats it, and the second round's tool message must carry the real options. Verified it fails against the old `repeated_ids` derivation with exactly the reported symptom:

```
assert '(skipped: id...ed this turn)' == '{"workspaces...: "gpt-4o"}]}'
```

**Evidence gap — the pre-pass was undemonstrated.** Added the artifact you named: `test_a_step_holding_both_calls_discovers_before_it_launches` drives `[launch, discovery]` (launch listed first) through `execute_multiple_tools` against one fresh `session_context` with nothing pre-seeding `launch_discovery`, so only the discovery-first pre-pass can make the launch succeed. Verified it fails with `discovery_indices = []`.

Writing it surfaced a real inconsistency: `execute_launch_tool` ignored the app factory the tool manager forwards on the context, while `execute_launch_discovery_tool` honors it — the two handlers could read different application instances and disagree about which workspaces exist. Fixed.

**Note — "provider/model" in the tool description, user guide and CHANGELOG.** You're right, `model_options` emits a bare `name`. Corrected the three texts rather than widening the payload: that name is exactly what `atlas_launch` takes as its `model` argument, and promising a provider the model never sees invites it to invent one. The test fixture carried the same fiction and now mirrors the real shape, with `test_discovery_publishes_workspace_ids_and_bare_model_names` pinning it. The user guide also now states that ticking `atlas_launch` auto-enables the discovery tool.

**Notes not taken, deliberately.** Capping exempted retries at one per turn and memoizing group checks, and moving the pairing into the launcher so the executor and anti-loop guard stay tool-agnostic, are both fair — but they are behavior/architecture changes beyond this PR's review round, and the second one touches the tool executor for every tool, not just these two. Better as a follow-up issue than as a late change to a PR that has been waiting on a merge conflict; happy to open one.

Verified locally on the merged tree: 3552 backend tests pass (`bash test/atlas_tests.sh`), 1040 frontend tests pass (`npx vitest run`), and `ruff check` is clean on every file this branch touches.

garland3:
Opened #969 for the two deferred notes (cap exempted discovery retries + memoize group checks; decouple the launch/discovery pairing from the tool executor), with the two tests added here named as the behavior a refactor must preserve.

garland3:
## Summary
Adds `atlas_discover_launch_options` and makes it a required step before `atlas_launch`. The launcher refuses workspaces or models that the latest discovery did not return. Discovery is also added automatically wherever launch is selected, it runs first in the executor, and its state carries across agent steps. The PR also includes an unrelated change that makes toast actions await async handlers.

## Verdict
Approve with nits.

## Evidence
Evidence: 7/10. The backend change is reasonably demonstrated, including new ordering and retry tests that fail against the old code. The biggest gap is that no test covers the unknown-workspace and unknown-model refusals in `_require_discovered_option` (`atlas/application/chat/runs/launcher.py:231`), and there is no end-to-end launch turn or E2E result.

## Notes
- **A discovery call that keeps failing is never stopped by the tools-mode anti-loop guard** (`atlas/application/chat/modes/tools.py:442`). Each round re-runs the serial model-authorization sweep until `max_extra_rounds` is used up. Allow only one exempted retry per turn (already tracked in #969).
- **The async toast action can run twice and leaves rejections unhandled** (`frontend/src/components/ui/ToastProvider.jsx:77`). Ignore clicks while the action is pending, and catch and log any rejection.
- **The discovery tool appears in the tools panel as its own checkbox, but unticking it has no effect while launch is selected** (`atlas/routes/config_routes.py:80`). Either hide it, or show it checked and disabled with a "Required by atlas_launch" hint.

---

garland3:
Added the PR validation script in `50fdcba2` — `test/pr-validation/README.md` requires one for every PR that changes backend or feature code before the PR is reviewed or merged, and #949 had none.

`test/pr-validation/test_pr949_launch_option_discovery.sh` starts a real backend with launch enabled and drives the real WebSocket chat endpoint and workspaces REST API. The model is a puppet only so the *order* of the two tool calls can be chosen deliberately (`plan=launch_only`, `plan=discover_launch`, `plan=same_step`, `plan=discover_only`); the discovery, the authorization and the launch itself are the real code paths. Assertions read `tool_complete` frames, so they check what the tools actually returned rather than the model's retelling of it.

**6/6 passed on two consecutive clean runs, exit 0.** What it covers:

- the discovery tool is advertised by `/api/config` when the flag is on, and with `FEATURE_ATLAS_LAUNCH_ENABLED=false` the atlas server still reports `canvas sleep` and neither launch tool
- `atlas_launch` with no prior discovery is refused, the refusal names the discovery tool and says the launch is *blocked*, no sub-conversation is started, nothing ran discovery behind the model's back, and the refusal actually reaches the model
- discovery then launch starts the child under the discovered workspace and hands back its run id
- one step holding `[launch, discovery]` — launch listed first — still discovers first and launches, which is the `execute_multiple_tools` pre-pass end to end
- discovery reports the saved workspace and a bare model name, with no provider
- a launch naming an undiscovered workspace, or an undiscovered model, is refused

Three harness traps worth recording, since each produced a confidently wrong result before it was found:

1. **Agent mode sends no `chat_response`.** The answer arrives as `token_stream` and the turn ends on a terminal `run_status`.
2. **A launched sub-conversation streams its own `run_started`/`run_status`/tool frames down the same socket.** The turn loop has to bind to the parent run id — otherwise a *child's* completion ends the loop before the parent's tools have run, and every check after the first successful launch fails for no real reason. This one cost the most time and looked exactly like a product bug.
3. **`start_backend` must `exec` uvicorn** so `$!` is the server, not the subshell. Killing the subshell left the old server up, and the flag-off check then read a stale config and *passed against the wrong process* — a false green.

No product defects surfaced: every refusal and every success behaved as the PR describes.

garland3:
## Summary
Adds an `atlas_discover_launch_options` tool that lists the workspaces and models a user is authorized for. `atlas_launch` is now refused unless discovery ran earlier in the same turn and returned the chosen options. Discovery is injected automatically wherever launch is selected, runs before launch inside a single step, and is exempt from the tools-mode anti-loop guard. The PR also includes an unrelated change that makes toast actions await async handlers.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. For a large backend control-flow change this is well demonstrated: unit tests, an end-to-end validation script and full-suite results are all provided. The main gap is a unit test proving the discovery-membership refusals themselves. The E2E unknown-workspace and unknown-model cases would still pass if `_require_discovered_option` were removed, because `resolve_workspace`/`resolve_model` refuse those inputs later anyway. Seed discovery with options that differ from the launch arguments, then assert the refusal says "was not returned by the latest launch discovery".

## Notes
- A discovery refresh that fails before its handler runs (approval rejected, hook denial, dispatch exception) leaves the earlier options valid for the rest of the turn. Clear `launch_discovery` on any failed discovery result. See `atlas/application/chat/runs/launcher.py:158` and `atlas/application/chat/utilities/tool_executor.py:688`.
- The discovery-first pre-pass reads only attribute-style tool calls. With dict-shaped calls, discovery ends up running concurrently with launch. See `atlas/application/chat/utilities/tool_executor.py:86`.
- The async toast action can be triggered twice while it is pending, and a rejected promise is not caught. Add a pending guard and a `catch`, or move this change to its own PR. See `frontend/src/components/ui/ToastProvider.jsx:77`.

---

PR #950, opened 2026-09-16
Make the chat composer usable one-handed on mobile and in-car

garland3:
## Problem

Three things made Atlas awkward one-handed on a phone and close to unusable from a car mount. All are in the chat composer area.

### 1. The feedback button sat on top of Send

`FeedbackButton` was pinned at a constant `bottom-32` (128px) on anything narrower than 1102px. The composer is not a fixed-height strip -- the textarea auto-resizes up to `COMPOSER_MAX_HEIGHT`, and above it stack the active-tools row, attached-file pills, and the disconnected / no-tools / agent-mode banners. Past a couple of lines the composer grows past 128px and the round feedback bubble lands on Send.

Measured on a 390x844 viewport with an eight-sentence message, the feedback button's 44x44 box was **entirely inside** Send's box -- **1936px of overlap**, send arrow completely hidden. A thumb aimed at Send opened the feedback modal.

### 2. New Chat raised a native confirm on every use

`clearChat` called `window.confirm` whenever the chat had *any* content. A native dialog is a hard two-step on a phone, cannot be styled or made touch-sized, and steals focus -- for an action users take constantly.

### 3. Touch targets

Upload / Stop / Send were 44-52px wide; the feedback button was 44px.

## Changes

- **Composer-anchored feedback button.** `ChatArea` publishes the live composer height as `--atlas-composer-height` (`hooks/useComposerHeightVar.js`, a `ResizeObserver` on the composer `<footer>`), and the button anchors with `bottom: calc(var(--atlas-composer-height, 8rem) + 1rem)`. It tracks the composer instead of guessing a constant. The hook clears the property on unmount so a stale height cannot strand anything anchored to it.
- **New Chat confirm narrowed to the one irreversible case.** The confirm now fires only when an *untracked* reply is still generating and would be cancelled outright. Everything else clears immediately and pushes a toast with an **Undo** action. Undo replays the snapshot through the same `restore_conversation` path a history load uses, so the transcript **and** the backend's context both come back. A reply belonging to a tracked background run is unaffected -- it keeps running and New Chat stays pure navigation (#884).
- **Toast actions.** `ToastProvider` toasts may now carry one inline `action: { label, onClick }`, rendered as a full-size button. This is what Undo uses.
- **Bigger touch targets.** Upload / Stop / Send are `min-w-[48px] min-h-[48px]` (Send and Stop-streaming 56px wide with a 24px icon); the feedback button goes 44 -> 56px. The toast dismiss button gets a 40px hit area.

## Verification

Playwright against two local servers -- one serving the pre-change bundle, one the fixed bundle -- at 390x844 and 1280x800, measuring both buttons' bounding boxes and driving a real New Chat.

| | before | after |
|---|---|---|
| feedback/Send overlap @390x844, long message | **1936px** | **0px** |
| feedback/Send overlap @1280x800, long message | 0px | 0px |
| native confirm dialogs on New Chat | `["Start a new chat? This will clear the current conversation from view."]` | `[]` |
| transcript cleared | false (dialog dismissed) | true |
| Undo offered | false | true |

Screenshots in the comments below.

- `npx vitest run` -- 82 files, 929 tests, all passing (was 79/915; +14 new)
- `npx eslint src` -- 0 errors (4 pre-existing warnings, untouched files)
- `npm run build` -- clean
- `bash scripts/check-docs.sh` -- OK

No PII in any test data, screenshot, or log: the only text typed is a generic routing question and a generic greeting, against a local dev server.

Design note: `docs/developer/design-notes/mobile-composer-friendliness-2026-09-16.md`.


garland3:
## Screenshots: feedback button vs. Send (390x844, long message)

Same page, same message, pre-change bundle vs. this branch. Both captured with Playwright against a local dev server; the only text typed is a generic routing question.

**Before** -- the feedback bubble is sitting *on* the Send button. The send arrow is not visible at all; the blue square on the right is Send, and the circle in the middle of it is the feedback button. Measured overlap: **1936px**, i.e. the feedback button's entire 44x44 box.



**After** -- the feedback button is anchored above the composer and the send arrow is visible again. Measured overlap: **0px**. Both controls are also bigger (feedback 44 -> 56px, Send 52 -> 64px wide).



Worth noting the "before" shot with an *empty* composer too: even at rest the bubble clips the Auto-approve toggle and the edge of Send.

garland3:
## Screenshots: New Chat

Playwright drove a real turn (generic greeting prompt, local dev server), then clicked **New Chat**.

**Before** -- a native `window.confirm` fires. Playwright captured its text:

```
["Start a new chat? This will clear the current conversation from view."]
```

A native dialog can't be screenshotted, so here is the state after it was dismissed: the transcript is still there and the user has to go back and tap New Chat again. On a phone that is two deliberate taps on small targets; in a car mount it is worse.



**After** -- no dialog. The chat clears in one tap and the Undo toast appears with a full-size button. Undo replays the snapshot through `restore_conversation`, so the backend's context comes back, not just the pixels.



## Desktop is unchanged in behavior (1280x800)

Overlap was already 0px at desktop widths and stays 0px; the button just sits above the composer now instead of at a fixed corner offset, and the controls are larger.

garland3:
## Things I deliberately did not do, for reviewers

1. **The one surviving `window.confirm` is still native.** It fires only when an untracked reply is mid-generation and would be cancelled -- genuinely irreversible, so a confirm is right there. But it could go through the in-app `useDialog().confirm` and get big touch targets like everything else. That API is async and `clearChat` is synchronous with two callers gating on its boolean return (`Header` line 119 / 196). Making it async is a contained change but not a free one, and I would rather do it as a follow-up than bolt it on here. Say the word if you want it in this PR.

2. **Toast placement on mobile.** The Undo toast renders at `top-4 right-4`; on a 390px screen it covers the header's New Chat and menu buttons for its 8-second life. It auto-dismisses and has an explicit close, but a bottom-anchored mobile placement would be better. I left it alone because it affects every toast in the app, not just this one.

3. **Scope was the composer.** The sidebar, header and settings panels have their own small controls; not touched.

## Note on `clearChat` dependencies

The Undo snapshot needs the full transcript, but taking `messages` as a `useCallback` dependency would change `clearChat`'s identity on every streamed token -- and `Header` re-registers its Ctrl+Alt+N keydown listener whenever `clearChat` changes. The snapshot reads the transcript from a ref instead, and the dependency stays `messages.length`. That is what the second commit does.

garland3:
## Review responses (all four addressed in `882093ec`)

Copilot independently flagged the same first two, so treating both as confirmed.

### P1: "Restore transient snapshots without a fake persisted ID" -- valid, fixed

Verified against `atlas/application/chat/service.py`: `handle_restore_conversation` returns `{"error": "Conversation not found"}` for any id the configured repository does not know, and for an id it *does* know it deliberately uses the stored messages over the client payload so a client cannot replay forged history into LLM context. My fabricated `undo_<ts>_<rand>` id would have drawn an error frame and no re-seed while the UI showed what looked like a successful restore. Exactly right.

I did not weaken the backend check -- the DB-canonical rule is a security property, not an obstacle. Instead Undo now carries the real id or none, and has two shapes:

- **Real `activeConversationId`**: `restore_conversation` as before; the server re-seeds from its canonical copy and the next turn has full context.
- **No id** (incognito, the default -- the conversation only ever existed in the tab): no backend round-trip at all. The transcript is restored locally and a system row is appended saying the assistant no longer has those messages in context. Undo does not pretend to be a full recovery when it isn't.

The better long-term answer is to *defer* `reset_session` until the user sends the first turn of the new chat or the undo window closes -- then incognito Undo needs no re-seed because the session was never torn down. That is a session-lifecycle change I did not want to land unreviewed; it is written up in the design note.

### P1: "Invalidate Undo before it discards the replacement chat" -- valid, fixed

Sending a turn, loading a conversation from history, or clearing again now all retire the offer and dismiss its toast (`invalidateUndoOffer`), and the action re-checks a token before running so a toast that outlives its dismissal is inert rather than destructive. Covered by four new tests, including the exact scenario described (send into the replacement chat, then tap the old Undo -- the replacement survives and no `restore_conversation` is sent).

### P1: "Add the required changelog entry" -- already done, stale

Reviewed against `50a16f1e`; the entry landed in the next commit (`c81ac9a0`, now `02124df1`) and has since been extended to cover this round. Sorry for the split -- it should have been in the first commit.

### P1: "Remove the generator attribution from the commit" -- valid, my error, fixed

AGENTS.md "OPSEC: LLM Identity Disclosure" is explicit that it overrides the agent's default trailer behavior, and I added one anyway. All four commits have been rewritten to drop the `Co-Authored-By` trailer and the PR description's generated-with line is gone; the branch was force-pushed. Author metadata is the human account. `git log origin/main..HEAD` now greps clean for provider/tool names.

### Re-verified after these changes

- `npx vitest run` -- 82 files, **934 tests**, all passing (+5 over the previous round)
- `npx eslint src` -- 0 errors
- `npm run build` -- clean
- `bash scripts/check-docs.sh` -- OK
- Playwright re-run against the rebuilt bundle at 390x844: feedback/Send overlap still **0px**

garland3:

## Summary
Replaces the New Chat confirm dialog with a silent clear plus an 8-second Undo toast (restoring either a saved conversation or an in-memory transcript for incognito), and anchors the floating feedback button to a live `--atlas-composer-height` variable so it no longer overlaps the composer on small viewports.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. The change is well demonstrated for its size — before/after mobile and desktop screenshots, a quantified overlap table, new unit tests, and full suite/lint/build results — but the screenshots predate the final rework, so the most important missing artifact is a re-captured post-Undo shot from the current bundle, ideally in incognito mode where the "assistant no longer has these in context" system row appears.

## Notes
- Move `invalidateUndoOffer()` below the `isConnected` guard — a rejected send currently destroys the offer for nothing (`frontend/src/contexts/ChatContext.jsx:621`).
- Snapshot live message objects instead of routing Undo through `buildPersistedMessage`, which erases `system` rows' `subtype`/`text` (`frontend/src/contexts/ChatContext.jsx:875`).
- Add a `.catch` to the restore chain so a failed Undo surfaces an error instead of an unhandled rejection (`frontend/src/contexts/ChatContext.jsx:935`).

---

garland3:
## Follow-up: found a bug in my own fix by driving it in a browser (`13beb4be`)

After posting the review responses I re-ran the incognito Undo path end-to-end instead of trusting the unit test, and the honesty note I had just added was rendering as a **blank row**.

`Message.jsx` renders an unrecognised system subtype from `message.content`; I had set only `message.text`. The unit test asserted on `text` and passed happily. That is the worst version of this bug -- Undo goes back to looking like a full recovery, which is precisely what the row exists to prevent.

Fixed by setting both fields, and the test now asserts on `content` (the one that actually renders) so it would catch a regression.

Verified in the browser at 390x844 against the rebuilt bundle:

```json
{
  "nativeDialogs": [],
  "clearedTranscript": true,
  "undoRestoredTranscript": true,
  "undoWarnedAboutLostContext": true,
  "toastGone": true
}
```



934 tests still passing, eslint 0 errors, build clean.

garland3:
## Found the same bug one element over (`6ff78228`)

While the CI ran I grepped for other constant bottom offsets in the same layer. The welcome-screen "Powered by ATLAS" logo (gated on `VITE_FEATURE_POWERED_BY_ATLAS`) was pinned exactly the same way -- `bottom-32 sm:bottom-36 md:bottom-40`.

Typing a long message *before* sending the first one grows the composer past those offsets and the logo gets drawn over it. Verified by building with the flag on and measuring against the composer footer at 390x844:

| | logo y | composer footer y | overlap |
|---|---|---|---|
| before | 689 | 573 (h 271) | **3888px** -- 116px inside the footer |
| after | 530 | 573 | **0px** |

It is decorative and not interactive, so unlike the feedback button it cost appearance rather than mis-taps -- but it is the same root cause and takes the same one-line anchor, and leaving it would just mean the next person rediscovers it.

934 tests still passing, eslint 0 errors, build clean (with and without the flag), `check-docs.sh` OK.

garland3:

## Summary
Replaces the blocking `window.confirm` on New Chat with an immediate clear plus an 8-second Undo toast, and anchors the feedback button and welcome-screen logo to a live `--atlas-composer-height` custom property so they no longer overlap a grown composer. Also enlarges composer touch targets and adds four test files.

## Verdict
Request changes.

## Required changes
1. Gate the Undo snapshot id on server persistence, not on `activeConversationId` being truthy, at `frontend/src/contexts/ChatContext.jsx:874` — in `local` save mode the autosaved `local_<ts>_<rand>` id is unknown to the server, so `restore_conversation` returns "Conversation not found" while the UI still presents the Undo as a complete recovery.
2. Move the canvas restore outside the id-present branch at `frontend/src/contexts/ChatContext.jsx:935` — an incognito Undo with an open canvas never calls `setCanvasContent` and silently loses the canvas content.
3. Snapshot live message objects rather than `buildPersistedMessage` output at `frontend/src/contexts/ChatContext.jsx:875` — system event rows carry `text`/`subtype`, which the persisted shape drops, so file-attach and agent-notice rows come back as empty gray rows after Undo.
4. Add a `.catch` to the restore chain at `frontend/src/contexts/ChatContext.jsx:935` and surface a failure toast — the offer token is already cleared, so a rejection leaves a half-restored view, an unhandled rejection, and no retry.
5. Move `invalidateUndoOffer()` below the `!isConnected` early return at `frontend/src/contexts/ChatContext.jsx:621` — a send that never leaves the tab currently destroys the only recovery path for an incognito transcript.
6. Add a test covering a failed/rejected Undo and one covering the 8-second expiry in `frontend/src/test/new-chat-undo.test.jsx`, and widen the snapshot fixture beyond two plain chat messages so the lossy round-trip in item 3 would be caught.

## Evidence
Evidence: 8/10 — the change is well demonstrated for its size, with before/after captures at 390x844 and 1280x800, quantified overlap (1936px to 0px), and full vitest/eslint/build results re-reported after each rework. The most useful missing artifact is a before/after capture of the welcome-screen logo fix from the head commit, since the desktop and pre-Undo shots are pinned to an earlier commit.

## Notes
- Restore spreads `msg.metadata` after `role`/`content`/`type`, letting metadata override reserved fields — `frontend/src/contexts/ChatContext.jsx:918`.
- The `calc(var(--atlas-composer-height, 8rem) + 1rem)` string is hardcoded in two consumers instead of importing the constant the hook already exports — `frontend/src/components/FeedbackButton.jsx:116`.
- Adding `toast` to `clearChat`'s dependency list reintroduces identity churn on the provider — `frontend/src/contexts/ChatContext.jsx:919`.

---

PR #951, opened 2026-09-16
fix: make tool-produced files downloadable from chat

garland3:
## The bug

A file an MCP tool produced could be downloaded from the File Manager library but not from chat.

The library fetches by S3 key, so it always worked. Chat asks for the file **by name**, and the two names disagree:

- storage sanitizes on the way in, so `Q3 Sales Report (final).csv` is stored — and keyed in `session.context["files"]` — as `Q3_Sales_Report_final_.csv`;
- the chat and canvas download controls carry the name the tool advertised (an artifact `name`, a `meta_data.output_files` entry), which is the pre-sanitization original.

`handle_download_file` did an exact-key lookup, so it missed. And `ChatContext.downloadFile` pre-filtered against the session file list, so the click was dropped in the browser and nothing was ever sent — no request, no error, no feedback. Even when a request did go out, a `file_download` error frame was only `console.error`'d.

This is not specific to one tool: any MCP server whose artifact name contains a space, parenthesis, or other non-word character hits it. Only the v2 artifact path sanitizes its session-map key; the legacy `returned_file_names` path keys by the unsanitized name and was never affected.

## Reproduction

Against a running instance, with `file_viewer`'s `display_file` pointed at a file named `Q3 Sales Report (final).csv` (driven by a scripted tool-calling model so the run is deterministic):

- `GET /api/files` lists `Q3_Sales_Report_final_.csv`, and `GET /api/files/download/<key>` returns `200` with the contents — the library path works.
- Clicking the chat/canvas download control sends **nothing** on the WebSocket (verified by instrumenting `WebSocket.prototype.send`).
- Sending the frame that control would send, `{"type":"download_file","filename":"Q3 Sales Report (final).csv"}`, answers `{"error":"File not found in session"}`.

## The fix

**Ingestion records the advertised name.** `upload_files_from_base64` stores it as `original_filename` on the file reference, alongside the sanitized key.

**One resolver reconciles the two names.** `resolve_session_file` in `file_processor` is used by both the download path (`ChatService._resolve_session_file` delegates to it) and the canvas notification, so the two views of a session cannot drift apart. Its precedence:

1. the exact key, **unless** the entry there is claimed by some *other* advertised name — this is what keeps a user's attachment (keyed by its real filename, recording no original) from being shadowed by a tool artifact advertising the same name;
2. the entry whose recorded `original_filename` is the request — when two advertised names sanitize alike only one can hold the key, so the entry sitting at `a_b.txt` may be the one advertised as `a b.txt`;
3. the exact key after all — the ordinary case, where the request is the sanitized stored name and nothing else claims it;
4. failing everything, and only for a map where nothing records an original, a sanitized-form comparison, so sessions written before originals were recorded keep working.

A name claimed by more than one entry is ambiguous and resolves to nothing rather than to whichever came first.

**Ingestion no longer lets artifacts displace each other.** `_ingest_v2_artifacts` merged with a plain `dict.update`, so `a b.txt` and `a_b.txt` — which both want the key `a_b.txt` — silently overwrote one another and the loser vanished from the session. The newcomer now takes a suffixed key (before the extension, so type sniffing by suffix still works). An artifact re-emitted under a name it already holds refreshes that entry wherever it ended up, so a tool rewriting its output neither piles up copies nor leaves two entries claiming one advertised name.

**Download frames name their conversation.** The backend searches this user's run sessions after the connection session, so an unaddressed frame could be answered by a different parallel conversation that had produced a same-named file. `downloadFile` now sends `conversation_id` and `run_id` (the pair `stopAgent` already uses).

**Input validation.** `filename` arrives from a client JSON frame, so it is validated as a non-empty string before the lookup; a malformed frame (`null`, a number, a list) gets an ordinary error reply instead of raising out of the websocket handler.

**Frontend.** `downloadFile` asks the backend instead of pre-filtering. That guard is what made the failure silent, and duplicating the sanitizer in JS would only move the drift somewhere harder to see. The `file_download` error frame now renders as a system message.

Ownership checks are untouched throughout; content comes back under the name the user was shown.

## Verification

- `atlas/tests/test_chat_download_tool_filenames.py` — new, 16 cases. Covers: the advertised name resolving (fails on `main` with `File not found in session`); the stored name still resolving; `a b.txt` / `a_b.txt` each keeping its own bytes **in both arrival orders, whether they arrive in one tool result or separate ones**; a re-emitted artifact refreshing its own entry, including after being re-keyed; a user attachment not shadowed by a tool advertising the same name; the canvas event carrying the named artifact's `s3_key`; a legacy entry with no recorded original; non-string and empty filenames; an unknown name still rejected.
- `frontend/src/test/chat-file-download.test.js` — new; saving under the advertised name, and surfacing the error.
- `bash test/atlas_tests.sh` — 3402 passed, 17 skipped.
- `npx vitest run` — 80 files, 917 tests passed.
- `npx eslint src` — 0 errors (4 pre-existing warnings, none in changed files).
- `bash scripts/check-docs.sh` — OK.
- Re-ran the browser reproduction against the final build: the chat download frame returns the file and the browser saves it as `Q3 Sales Report (final).csv`; an unknown name renders `Error: could not download nope.csv -- File not found in session` in the transcript.

## Noted, not fixed here

The canvas Download button is additionally gated by `canDownloadFile`, which rejects `isInline` files — an artifact streamed inline to the canvas stays undownloadable from that button regardless of naming. That is a separate issue from the name mismatch and is out of scope for this PR.


garland3:
## Summary
Fixes chat file downloads failing when a tool advertises a filename that differs from the sanitized name stored in session context: the backend now resolves a requested name against the session file map (exact key first, then a unique sanitized match), and the frontend drops its client-side pre-filter so misses return a visible error instead of a silent no-op.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — a deterministic before/after reproduction with the exact frames, plus backend and frontend tests and quoted suite results. The only missing artifact is a screenshot of the new download-failure system message as rendered.

## Notes
- The `len(matches) == 1` ambiguity guard has no test; two stored names collapsing to one sanitized key would not be caught by a regression — `atlas/tests/test_chat_download_tool_filenames.py:118`.
- Resolution misses log nothing and return the same "File not found in session" error for both no-match and refused-ambiguous cases, making the collision path undiagnosable — `atlas/application/chat/service.py:1046`.
- The download-failure message is a bare plain-text system line interpolating the raw backend string (including `str(e)` from the catch-all), unlike the structured `file-attach-error` shape next to it — `frontend/src/handlers/chat/websocketHandlers.js:639`.

---

garland3:
Thanks — all four review points were valid and are addressed in `8dc3b4c5`.

**Sanitization-equivalent aliases could return the wrong bytes** (raised independently by both reviewers). Agreed, and the "exactly one match" count did not detect the collision case, exactly as described. Fixed at the source instead of the lookup: `upload_files_from_base64` now records the advertised name as `original_filename` on the file reference, and `_resolve_session_file` matches that recorded original rather than re-deriving it. `a b.txt` and `a_b.txt` still collapse to one stored entry, but asking for the one that lost ingestion is now a reported miss rather than the survivor's content under the loser's label — `test_colliding_names_do_not_resolve_to_the_survivor` asserts both halves. The sanitized-form comparison survives only for entries written before originals were recorded, and only while no entry in the map records an original at all, so a map that can answer exactly is never second-guessed.

**Non-string `filename` from a malformed frame.** Fixed: `handle_download_file` validates it as a non-empty string up front and returns an ordinary `file_download` error reply. Covered for `None`, a number, a list, a dict and `""`.

**Changelog.** Added under `## [Unreleased]` as `### PR #951 - 2026-09-16`.

**AI attribution in the commit message.** Removed; the commit was amended and force-pushed, and the PR body's sign-off is gone too.

Re-ran after the changes: `bash test/atlas_tests.sh` → 3396 passed, 17 skipped; `npx vitest run` → 917 passed; `npx eslint src` → 0 errors; `bash scripts/check-docs.sh` → OK.

garland3:
## Summary
Fixes chat download failures for files whose advertised name differs from the sanitized storage key, by recording `original_filename` on ingested session file refs, resolving downloads against that alias in `_resolve_session_file`, removing the client-side pre-filter in `downloadFile`, and surfacing download errors in the transcript.

## Verdict
Request changes.

## Required changes
1. Make the legacy fallback gate per-entry at `atlas/application/chat/service.py:1078` — `any(meta.get("original_filename") ...)` is map-wide, so one newly ingested artifact disables sanitized-form matching for every pre-existing entry in that session and the original bug returns. Restrict the sanitized-form comparison to entries that themselves record no original.
2. Add a mixed-map test in `atlas/tests/test_chat_download_tool_filenames.py` — ingest two refs, strip `original_filename` from only one, and assert the stripped entry still resolves by its advertised name; the existing legacy test strips it from every ref and cannot catch the above.
3. Cover the actual frontend fix in `frontend/src/test/chat-file-download.test.js` — assert that `downloadFile('Q3 Sales Report (final).csv')` sends a `download_file` frame when `sessionFiles` contains only the sanitized name, plus the new `if (!filename) return` exit; restoring the removed guard currently fails no test.
4. Stop rendering raw backend exception text at `frontend/src/handlers/chat/websocketHandlers.js:639` — `data.error` can be the catch-all `str(e)` from `atlas/application/chat/service.py:1141`, which leaks bucket names, keys and endpoint URLs into the persisted transcript. Return a fixed user-facing string from the catch-all (detail stays in the server log) or map known errors to copy on the client.
5. Log the miss in `handle_download_file` around `atlas/application/chat/service.py:1112` with the sanitized requested name and a reason code (`no_match` / `ambiguous_alias`), since every failure mode currently returns the same string with no server-side trace.
6. Restore the original line endings on `atlas/modules/file_storage/manager.py` so the diff shows only the two intended edits — the CRLF→LF flip rewrites 621 lines, destroys `git blame` provenance for the module, and leaves the repo mixed.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size, with a before/after websocket frame reproduction, targeted tests on both sides, and quoted suite results. The only gap is a screenshot of the new download-failure message as it renders in the transcript.

## Notes
- Dead branch: `len(aliased) > 1` at `atlas/application/chat/service.py:1070` is unreachable, since the map key is a deterministic function of the recorded original.
- `original_filename` at `atlas/modules/file_storage/manager.py:257` means the pre-sanitization name here but the post-sanitization name in S3 metadata (`s3_client.py:193`) — consider renaming the session field.
- A transcript turn per failed click stacks duplicates with no dismissal (`websocketHandlers.js:639`); a transient notice may fit better.

---

garland3:
Force-pushed `7c91b82b`: the previous revision rewrote `atlas/modules/file_storage/manager.py` with LF endings, and that file is CRLF in the tree, so the whole file showed as changed. That is what put a CodeQL log-injection note on line 312 — a pre-existing `logger.error` line the normalization presented as new. Line endings restored; the diff for that file is now the intended 8 added / 3 removed lines and the alert's subject is untouched by this PR. No behavior change from the previous revision; `bash test/atlas_tests.sh` re-run: 3396 passed, 17 skipped.

garland3:
## Summary
Fixes chat downloads for tool-produced files whose advertised name differs from the sanitized S3 key, by recording `original_filename` on session file refs and resolving download requests through a new `_resolve_session_file` alias lookup. Also removes the client-side `sessionFiles` pre-filter in `downloadFile` and adds a transcript message when a download fails.

## Verdict
Request changes.

## Required changes
1. Stop rendering backend error text verbatim in the transcript at `frontend/src/handlers/chat/websocketHandlers.js:639` — `data.error` can be the catch-all `str(e)` from `atlas/application/chat/service.py:1141`, so storage exceptions carrying bucket names, object keys and endpoint URLs get written into the persisted conversation. Map known error values to client copy and render anything else as a fixed generic string; keep the exception detail in `logger.error` only.
2. Give a failed download a transient, de-duplicated surface instead of one permanent system turn per click at `frontend/src/handlers/chat/websocketHandlers.js:639` — restored conversations never rehydrate `session.context["files"]` (reset client-side at `ChatContext.jsx:874`, not restored server-side), yet `Message.jsx:420` still renders a Download button for every `meta_data.output_files` entry, so every click on a reopened conversation appends an identical undismissable error line. Repro: run a tool returning `output_files`, reload, reopen from history, click Download three times.
3. Make the legacy fallback gate per-entry at `atlas/application/chat/service.py:1079` — `any(meta.get("original_filename") ...)` is evaluated map-wide, so one newly ingested artifact disables sanitized-form matching for every pre-upgrade entry in the session, which contradicts the docstring at line 1052. Compare sanitized forms only against entries that themselves record no `original_filename`.
4. Add a mixed-map test in `atlas/tests/test_chat_download_tool_filenames.py:154` — `test_legacy_entry_without_recorded_original_still_resolves` strips `original_filename` from every ref, so it only ever exercises the gate in its false state. Seed two refs, pop the field from one, and assert both still resolve.
5. Add a frontend test covering the fix itself at `frontend/src/contexts/ChatContext.jsx:925` — reverting the pre-filter removal leaves the suite green today. Assert `sendMessage` is called with the advertised name when `sessionFiles` holds only the sanitized key, and not called for empty/undefined input.
6. Log the resolution miss in `handle_download_file` (`atlas/application/chat/service.py:1118`) with a reason code (`no_match` / `ambiguous_alias` / `legacy_gate_closed`) and a truncated, newline-stripped requested name — today three distinct failures return the same string with no server-side trace, and `atlas/application/chat/service.py:1142` formats a client-controlled filename into a log line unescaped.

## Evidence
Evidence: 9/10 — the change is well-demonstrated for its size, with a named deterministic before/after reproduction against a running instance plus targeted backend and frontend tests. The one missing artifact is a screenshot of the new download-failure message as actually rendered in the transcript, which matters given item 2 above.

## Notes
- Rename the session-ref field `original_filename` (`atlas/modules/file_storage/manager.py:257`) — the same key in S3 metadata holds the post-sanitization name, the opposite meaning.
- `if aliased:` after the single-match return and the `if not self.file_manager` guard are both unreachable (`atlas/application/chat/service.py:1069`, `:1078`) — drop them or comment them as future-proofing.
- Quote the filename and drop the bare `--` separator in the error copy (`frontend/src/handlers/chat/websocketHandlers.js:641`) — the names this PR fixes contain spaces and parentheses.

---

garland3:
@codex review

garland3:
Both P1s on `7c91b82b` were real. Fixed in `4f160f74`.

**Reverse-order collision on the exact-key branch.** Confirmed — ingesting `a_b.txt` then `a b.txt` left the second owning the key, and a request for `a_b.txt` took the exact-key branch and returned the wrong bytes. Fixed at both ends:

- **Ingestion no longer lets them displace each other.** `_ingest_v2_artifacts` merged with a plain `dict.update`, so one artifact silently overwrote the other — the loser was gone from the session entirely, which is the root of this whole family of cases. A newcomer whose key is already held by an entry recording a *different* advertised original now gets a suffixed key (before the extension). Re-uploading the same original still refreshes its own entry in place, so this does not turn a rewrite into a pile of copies.
- **The advertised name now takes precedence over a like-named key** in `_resolve_session_file`. The entry sitting at `a_b.txt` may be the one advertised as `a b.txt`, so a request for `a_b.txt` belongs to the entry that *claims* it, not the one that merely occupies it. When no entry claims the requested name — the ordinary case, where the request is the sanitized stored name — the exact key still answers.

`test_names_that_sanitize_alike_keep_their_own_files` is parametrized over both arrival orders and asserts each advertised name downloads its own bytes; it fails on `7c91b82b` for `order0`. `test_re_uploading_the_same_name_refreshes_in_place` pins the non-collision case.

**Unaddressed download frames could answer from another conversation.** Confirmed against `_download_session_candidates` in `atlas/main.py`, and removing the client guard did widen the exposure. `downloadFile` now sends `conversation_id` and `run_id` (the same pair `stopAgent` already uses), so the conversation's own session is tried ahead of the user's other retained runs instead of last.

Re-ran: `bash test/atlas_tests.sh` → 3398 passed, 17 skipped; `npx vitest run` → 917 passed; `npx eslint src` → 0 errors. Changelog entry updated to match.

garland3:
## Summary
Fixes chat downloads of tool-produced files by recording the tool-advertised filename on session file refs, resolving download requests against it, and re-keying rather than overwriting names that sanitize alike; the client now sends `conversation_id`/`run_id` and surfaces download errors in the transcript.

## Verdict
Request changes.

## Required changes
1. Treat a missing `original_filename` on the incumbent entry as "unknown", not a collision, in `_merge_without_displacing` at `atlas/application/chat/utilities/file_processor.py:820` — entries written by `handle_attach_file` (`atlas/application/chat/service.py:994`) and the legacy `returned_file_names` path lack the field, so re-ingesting the same artifact name adds `report_1.csv`, `report_2.csv`, … on every turn and then trips the `len(aliased) > 1` ambiguity check, leaving the file undownloadable.
2. Scope the legacy sanitized-form fallback per entry instead of map-wide at `atlas/application/chat/service.py:1086` — `any(meta.get("original_filename") ...)` means one newly ingested artifact closes the fallback for every pre-upgrade entry in the session, reintroducing the original bug for those files.
3. Route all four writers of `session.context["files"]` through one constructor (`file_processor.py:207`, `:578`, `:780`, `service.py:994`) so the `original_filename` + no-displacement invariant holds for user uploads and attachments, not only v2 artifacts.
4. Key `upload_files_from_base64` results so two artifacts in a single `tool_result.artifacts` list that sanitize alike do not collapse before the merge runs (`atlas/modules/file_storage/manager.py:246`) — today the first file's bytes are orphaned and unreachable.
5. Validate the tool-advertised name before storing it as `original_filename` at `atlas/modules/file_storage/manager.py:257` — cap length, strip control/bidi characters, and reject path separators and `..` segments, since the value is echoed to the browser as the save-as name and into chat copy unfiltered.
6. Return a fixed user-facing string from the download catch-all at `atlas/application/chat/service.py:1148` instead of `str(e)` — raw storage exceptions (bucket names, object keys, endpoint URLs) are now rendered and persisted in the transcript by `frontend/src/handlers/chat/websocketHandlers.js:639`.
7. Add a frontend test for `downloadFile` (`frontend/src/contexts/ChatContext.jsx:929`) asserting the frame is sent with `filename`, `conversation_id` and `run_id` — the regression this PR fixes lives entirely in untested client code today.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — a deterministic before/after reproduction against a running instance plus targeted tests on both sides; the only notable gap is a screenshot of the new download-failure message as rendered, including the repeat-click case.

## Notes
- Failed downloads append a permanent transcript turn per click with no dedup or dismissal — `frontend/src/handlers/chat/websocketHandlers.js:639`.
- Dotfile names produce a leading-underscore key on collision (`.env` → `_1.env`) — `atlas/application/chat/utilities/file_processor.py:826`.
- Download misses (`no_match` / `ambiguous_alias` / `legacy_gate_closed`) are indistinguishable in logs, and the client-controlled filename is logged unescaped — `atlas/application/chat/service.py:1125`.

---

garland3:
@codex review

garland3:
All three P1s on `4f160f74` were real — each of the three new tests fails on that commit and passes on `ae2c7ec5`.

**A tool alias could shadow a user attachment.** Correct, and it was a regression from putting the alias branch ahead of the exact key. `handle_attach_file` keys an attachment by its real filename and records no `original_filename`, so a tool artifact advertising `report final.csv` (stored at `report_final.csv`, claiming that original) answered for the user's own file sitting at the exact key.

The precedence is now: exact key **unless some other advertised name claims that entry**, then the entry claiming the requested name, then the exact key after all. That resolves all four shapes without one shadowing another — the attachment (no original recorded → exact key wins), the collision (entry at `a_b.txt` claims `a b.txt` → skip to the claimant), the ordinary sanitized-name request (nothing claims it → exact key), and the advertised-name request (alias).

**Re-emitting a collided name allocated a third key.** Correct — `a_b.txt` re-emitted after being re-keyed to `a_b_1.txt` produced `a_b_2.txt`, leaving two entries claiming one advertised name, which the resolver then rightly called ambiguous and refused. `_merge_without_displacing` now looks for an existing entry recording the same advertised original first and refreshes it in place, wherever it ended up.

**The canvas emitted the key's occupant, not the named artifact.** Correct, and also a consequence of re-keying: `notify_canvas_files_v2` indexed the session map by artifact name. Rather than patch that lookup separately, both paths now go through one `resolve_session_file` in `file_processor`, which `ChatService._resolve_session_file` delegates to — so the download and the canvas cannot disagree about which entry a displayed name means. `test_canvas_event_carries_the_re_keyed_artifact_key` asserts the emitted `s3_key` is the named artifact's.

Re-ran: `bash test/atlas_tests.sh` → 3401 passed, 17 skipped; `npx vitest run` → 917 passed; `npx eslint src` → 0 errors. Changelog entry updated. Also re-confirmed the original reproduction end to end in the browser against the final build — the chat download frame now returns the file and the browser saves it as `Q3 Sales Report (final).csv`.

garland3:
## Summary
Fixes chat download of tool-generated files by recording the tool-advertised filename on each session file ref and resolving downloads against it, adding a no-displacement merge for colliding sanitized names, scoping download frames to conversation/run, and surfacing download failures in the transcript.

## Verdict
Request changes.

## Required changes
1. Fix the wrong-bytes case in `resolve_session_file` at `atlas/application/chat/utilities/file_processor.py:838` — when a user attachment written by `handle_attach_file` (`atlas/application/chat/service.py:994`) occupies the exact key with no `original_filename`, a tool artifact re-keyed off that name is unreachable and the download returns the attachment's content under the artifact's label. Carry the entry's stored key (or a stable file id) on the download control rather than relying on name precedence.
2. Scope the legacy sanitized-form fallback per entry at `atlas/application/chat/utilities/file_processor.py:864` — the map-wide `any(... original_filename ...)` gate means one newly ingested artifact re-breaks every pre-upgrade entry in that session; compare only against entries that themselves record no `original_filename`.
3. Stop collapsing colliding refs before the merge at `atlas/modules/file_storage/manager.py:246` — two artifacts in one tool result whose names sanitize alike both upload, but the second overwrites the first in `uploaded_refs`, leaving the first orphaned in object storage and unreachable from the session.
4. Render a bounded user-facing message instead of `data.error` verbatim at `frontend/src/handlers/chat/websocketHandlers.js:641`, and have the catch-all at `atlas/application/chat/service.py:1105` return fixed copy rather than `str(e)` — storage exceptions otherwise persist bucket names, keys and endpoints into the transcript.
5. Add tests for the uncovered branches: a mixed-map legacy case (two artifacts, `original_filename` popped from only one) and a direct `resolve_session_file` case with two entries claiming one advertised name asserting the ambiguous refusal (`atlas/tests/test_chat_download_tool_filenames.py:154`).
6. Add a client test for `downloadFile` at `frontend/src/contexts/ChatContext.jsx:929` asserting the emitted `download_file` frame carries `filename`, `conversation_id` and `run_id` — restoring the removed pre-filter currently leaves the suite green.
7. Normalize the advertised name before storing it at `atlas/modules/file_storage/manager.py:257` — basename it, strip control and bidi characters, and cap its length, since it now reaches the browser's save-as name and the UI label unfiltered.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — a deterministic before/after reproduction, 15 targeted backend cases and frontend coverage, with full suites quoted; the one missing artifact is a rendered view of the new download-failure message in the transcript, including repeat clicks.

---

garland3:
@codex review

garland3:
Last P1 was right too, and my earlier collision tests missed it by ingesting each artifact in its own tool result. Fixed in `2f55bb94`.

`upload_files_from_base64` builds its result dict keyed by the sanitized name, so two artifacts in **one** MCP result collapsed there — before `_merge_without_displacing` ever saw them — and the first was simply gone from the session and the canvas. It now keeps both: a key already held in the same batch by a different advertised original gets suffixed, via a `FileManager.unique_key` helper that the session-level merge now shares instead of open-coding its own suffix loop.

`test_one_result_carrying_both_colliding_names_keeps_both` ingests `a b.txt` and `a_b.txt` as a single tool result and asserts two session entries and the right bytes for each name; it fails on `ae2c7ec5`.

Re-ran: `bash test/atlas_tests.sh` → 3402 passed, 17 skipped; `npx vitest run` → 917 passed; `npx eslint src` → 0 errors; `bash scripts/check-docs.sh` → OK.

garland3:
## Summary
Fixes downloads of tool-produced files whose display names differ from their sanitized storage keys, by recording the advertised name on session file refs, adding a `resolve_session_file` lookup plus non-displacing collision keying, removing the client-side pre-filter, and surfacing download failures in the transcript.

## Verdict
Request changes.

## Required changes
1. Scope the legacy sanitized-form fallback per entry at `atlas/application/chat/utilities/file_processor.py:864` — the map-wide `any(... original_filename ...)` gate closes for the whole session as soon as one new artifact is ingested, so pre-upgrade entries stop resolving by their advertised name. Compare sanitized forms only against entries that themselves record no `original_filename`, and add a mixed-map test (two refs, field popped from one only).
2. Make a re-keyed artifact reachable by its advertised name at `atlas/application/chat/utilities/file_processor.py:838` — an incumbent whose `original_filename` is `None` (written by `handle_attach_file`, `service.py:994`) counts as "different", so the artifact is exiled to `data_1.csv` while the download control labelled `data.csv` returns the attachment's bytes. Backfill the advertised name on attachment/legacy writes, or resolve downloads by stored key/file id rather than name.
3. Route the other writers of `session.context["files"]` through the same non-displacing constructor — `handle_attach_file` (`atlas/application/chat/service.py:994`) and the legacy `returned_file_names` path (`file_processor.py:591`) still assign outright and can erase an artifact entry.
4. Return a fixed user-facing string from the download catch-all at `atlas/application/chat/service.py:1105` instead of `str(e)` — the client interpolates `data.error` verbatim into a persisted transcript message (`frontend/src/handlers/chat/websocketHandlers.js:641`), writing bucket names, object keys and endpoint URLs into conversation history. Keep detail in `logger.error`, and relax the test assertion at `frontend/src/test/chat-file-download.test.js:51` that currently pins the raw text.
5. Normalize the tool-advertised name before storing it at `atlas/modules/file_storage/manager.py:257` — it is now echoed to the browser as the `a.download` attribute, so `basename` it, strip Unicode control/format characters including bidi overrides, drop `..` segments and bound the length; an MCP tool is a separate trust domain.
6. Add a test for `ChatContext.downloadFile` (`frontend/src/contexts/ChatContext.jsx:929`) asserting `sendMessage` is called with `{type:'download_file', filename, conversation_id, run_id}` when `sessionFiles` holds only the sanitized key — the client half is the regression this PR is named for and no test currently covers it.
7. Render the failure with the established error styling at `frontend/src/handlers/chat/websocketHandlers.js:639` — emit `{role:'system', type:'system', subtype:'file-download-error'}` and handle it beside `file-attach-error` in `Message.jsx`, so a failed download does not read as a neutral notice.

## Evidence
Evidence: 8/10 — the backend is well demonstrated for a change of this size, with a named before/after reproduction, a new 370-line test file, and per-revision suite results. The main gap is the browser side: no captured outgoing `download_file` frame and no screenshot of the rendered failure message (covered by item 6 above).

## Notes
- `original_filename` means the pre-sanitization name here but the sanitized name in S3 metadata (`atlas/modules/file_storage/s3_client.py:193`); rename the new field to `advertised_filename` while it has one writer.
- Repeated clicks on an unresolvable download append one permanent transcript turn each (`frontend/src/handlers/chat/websocketHandlers.js:639`).
- `unique_key` loses the stem for dotfiles: `.env` → `_1.env` (`atlas/modules/file_storage/manager.py:44`).

---

garland3:
Rebuilt on `main` after #950 landed. The only conflict was `CHANGELOG.md` — both PRs added an entry under `## [Unreleased]`; kept both, #950's first. `frontend/src/contexts/ChatContext.jsx` auto-merged (#950 touched the composer/new-chat paths, this PR touches `downloadFile`); verified `downloadFile` came through intact with its `conversation_id`/`run_id` frame, and the diff against `main` is still only this PR's eight files.

Re-ran everything on the merged tree: `bash test/atlas_tests.sh` → 3402 passed, 17 skipped; `npx vitest run` → 83 files, 936 tests passed (includes #950's new composer tests); `npx eslint src` → 0 errors; `bash scripts/check-docs.sh` → OK.

garland3:
## Summary
Makes tool-produced files downloadable from chat by recording the tool-advertised filename on session file refs and resolving download requests against it, with non-displacing collision keying on ingest and a new client-visible error message on failure.

## Verdict
Request changes.

## Required changes
1. Fix name-collision resolution between attachments and artifacts at `atlas/application/chat/utilities/file_processor.py:840` — step 1 returns the exact-key entry whenever it records no `original_filename`, so an attachment named `report final.csv` shadows a tool artifact advertising the same name and every chat download and canvas preview of that artifact serves the attachment's bytes under the artifact's label. Either resolve by stored key / file id on the download control, or backfill `original_filename` on the attachment and user-upload writes.
2. Route all writers of `session.context["files"]` through the same non-displacing insert — `handle_attach_file` (`atlas/application/chat/service.py:994`) and the legacy `returned_file_names` path (`file_processor.py:592`) assign outright and can erase a v2 artifact entry that this PR's merge was added to protect.
3. Make the sanitized-form fallback per entry at `atlas/application/chat/utilities/file_processor.py:867` — the `any(...)` gate is evaluated over the whole map, so ingesting one new artifact makes previously resolvable pre-upgrade entries return `File not found in session`. Compare sanitized forms only against candidates that themselves record no original.
4. Add a `downloadFile` test for `frontend/src/contexts/ChatContext.jsx:1063` — the client half of the fix (dropping the `sessionFiles` pre-filter, adding `conversation_id`/`run_id`) is entirely uncovered; assert `sendMessage` is called once with the full frame when `sessionFiles` holds only the sanitized key, plus the empty-filename no-send case.
5. Add a mixed-map legacy test at `atlas/tests/test_chat_download_tool_filenames.py:154` — the existing test pops `original_filename` from every ref, so the gate is only ever exercised in its false state; seed two refs, strip the field from one, and assert both resolve to their own bytes.
6. Guard repeat download clicks in `downloadFile` (`frontend/src/contexts/ChatContext.jsx:1057`) — with no in-flight tracking, each click round-trips the whole base64-encoded object over the websocket, and each failure appends another permanent transcript turn (`frontend/src/handlers/chat/websocketHandlers.js:639`). Ignore a repeat while one is pending and collapse duplicate error notices.
7. Normalize the advertised name where it is recorded at `atlas/modules/file_storage/manager.py:257` — the raw pre-sanitization name now reaches `a.download` and chat copy, so Unicode bidi overrides permit extension spoofing (`invoice‮fdp.exe`) and length is unbounded. Strip `Cc`/`Cf` characters, take the basename, drop `..`, reject empty, and cap length.

## Evidence
Evidence: 9/10. The change is well-demonstrated for its size — a named deterministic reproduction, layered backend and frontend tests, and per-revision suite/lint results; the only notable gap is that the browser-side half is asserted in prose rather than shown (no captured outbound `download_file` frame or screenshot of the new failure message).

## Notes
- Failed-download message lacks `type`/`subtype`/`id`, so it renders as neutral gray system chatter rather than an error — `frontend/src/handlers/chat/websocketHandlers.js:639`.
- Download catch-all returns `str(e)` into rendered, persisted chat content, potentially leaking bucket/key/endpoint detail — `atlas/application/chat/service.py:1105`.
- `original_filename` means the pre-sanitization name here but the post-sanitization name in S3 metadata (`s3_client.py:193`); rename now, while there is one writer — `atlas/modules/file_storage/manager.py:257`.

---

garland3:
@codex review

garland3:
All three were real; fixed in `2f756e0e`, and each of the four new tests fails on `0ba4edeb`.

**Stored-key rows could download the wrong bytes.** Right, and the suggested remedy is the right one, so I took it rather than adding another name heuristic: a control that knows which stored file it stands for now **sends the storage key**, and the key decides. `downloadFile(filename, s3Key)` carries it, the File Manager rows (`SessionFilesView`, including Download All) and the canvas pass `file.s3_key`, and `handle_download_file` accepts an optional `s3_key`. The key must still name an entry of *this* session — it disambiguates, it is not a way to reach arbitrary storage (`test_storage_key_outside_the_session_is_refused`). A frame without a key resolves by name exactly as before, which is what the chat message buttons still do, since all they ever have is a name.

**The canvas could label an attachment's key as the new artifact.** Right. `ingest_v2_artifacts` now returns the references it actually stored and `notify_canvas_files_v2` describes those, instead of looking the advertised names back up in a session map that may hold someone else's entry under one of them.

**Legacy matching was switched off globally.** Right — P2 but a real compatibility hole. The fallback is now scoped to the entries that record no advertised original, so ingesting one new-style artifact no longer stops every older control in that session from resolving.

Re-ran: `bash test/atlas_tests.sh` → 3406 passed, 17 skipped; `npx vitest run` → 83 files, 936 tests passed; `npx eslint src` → 0 errors; `bash scripts/check-docs.sh` → OK.

One process note: `CanvasPanel.jsx` and `FileManagerPanel.jsx` are CRLF in the tree, and my first pass rewrote them as LF, which showed as whole-file diffs. Line endings restored before committing — every file in this PR now diffs by the lines actually changed.

garland3:
@codex review

PR #952, opened 2026-09-17
fix(ui): remove New Chat confirmation popup

garland3:
## Summary
- remove the remaining native confirmation from the New Chat path
- stop untracked active turns before resetting the session
- preserve the existing Undo behavior for idle transcripts
- update generation behavior tests

## Verification
- `npm test -- --run src/test/new-chat-stops-generation.test.js src/test/new-chat-undo.test.jsx`
- `npm test -- --run`
- `npm run lint` (0 errors; 4 pre-existing warnings)
- `npm run build`


garland3:
## Summary
Removes the New Chat confirmation dialog in `ChatContext.clearChat` so New Chat stops any in-flight turn and clears immediately. The diff also carries the `atlas_get_runs` / `atlas_result` sub-conversation observation tools, conversation-scoped child tracking in the run registry, and the related feature-gate and CHANGELOG changes.

## Verdict
Request changes.

## Required changes
1. Distinguish "reaped" from "not yours" in `get_child_result` at `atlas/application/chat/runs/launcher.py:581` — once a terminal run ages out (30 min) or is trimmed past 50 retained runs per user, a valid `run_id` raises `LaunchRefused("That run is not a sub-conversation of this conversation.")`, which reads as an authorization failure and will make the caller re-launch the work instead of retrying. Resolve the child against a durable record, or return a distinct `result_status` (e.g. `expired`).
2. Drop the unconditional `registry.reap_terminal()` from `_scoped_children` at `atlas/application/chat/runs/launcher.py:582` — these tools are documented as poll-in-a-loop, so every poll from any conversation triggers a full-registry sweep and shrinks the result window described above.
3. Offer Undo (or at minimum a non-blocking toast) for the stop-and-clear case at `frontend/src/contexts/ChatContext.jsx:846` — `canUndo` excludes `mustStopCurrentTurn`, so the one path the removed confirm actually guarded now discards a partially streamed response with no dialog, no toast, and no recovery.
4. Test the shipped `clearChat` in `frontend/src/test/new-chat-stops-generation.test.js:36` by rendering the real `ChatContext` provider (as `new-chat-undo.test.jsx` does) with `window.confirm` stubbed and asserted not called — the file currently asserts against a local re-implementation, so this PR's headline behavior is untested.
5. Sanitize the `conversation_id` mismatch log at `atlas/application/chat/runs/context.py:98` with `sanitize_for_logging(str(...))` on both arguments, matching the adjacent `run_id` branch — the value is producer-supplied and can inject newlines or control sequences.
6. Guard the mismatch diagnostics at `atlas/application/chat/runs/context.py:91` with `logger.isEnabledFor(logging.DEBUG)` — the sanitizer arguments are built eagerly on `tag_event`, which every per-token frame passes through.
7. Replace the short-circuiting assertion at `atlas/tests/test_atlas_launch_tool.py:925` (`tool_name == GET_RUNS_TOOL_NAME or ...` is constant-true for that parametrization) with per-tool assertions on the parsed payload, and add coverage for the `MAX_RESULT_CHARS` truncation boundary at `launcher.py:668`.

## Evidence
Evidence: 5/10. The frontend change is reasonably described, but thirteen of the fifteen changed files — the entire backend tool surface — are undescribed, and verification lists only `npm test`, `npm run lint`, and `npm run build`; the most important missing artifact is a pytest run over the three edited Python test modules.

## Notes
- `result_status: "unavailable"` conflates terminal-without-result, unwired repository, and missing conversation — `atlas/application/chat/runs/launcher.py:612`.
- No `### PR #952` CHANGELOG entry for the user-visible New Chat behavior change — `CHANGELOG.md:25`.
- The launch feature gate tuple is duplicated at four sites; a missed site is a silent gate bypass — `atlas/modules/mcp_tools/atlas_server.py:426`.

---

garland3:
## Summary
Removes the last `window.confirm` from the New Chat flow in `ChatContext.clearChat`, so starting a new chat while an untracked reply is generating now stops the turn and clears the transcript without prompting. Tests and one changelog sentence are updated to match.

## Verdict
Request changes.

## Required changes
1. Give the `mustStopCurrentTurn` path some feedback or recovery at `frontend/src/contexts/ChatContext.jsx:857` — it is now the only New Chat path that discards work with no confirm, no Undo toast, and no on-screen acknowledgement, while the reversible idle path still gets both. Either build `undoSnapshot` whenever `hasContent` (drop `&& !mustStopCurrentTurn`), or at minimum emit a non-actionable `toast.info('Response stopped. New chat started.')`.
2. Cover the shipped code path in `frontend/src/test/new-chat-undo.test.jsx` — the headline behavior is currently only asserted against the local `clearChat` re-implementation in `new-chat-stops-generation.test.js:25`, which would pass unchanged even if the confirm were still in `ChatContext.jsx`. Add a real-provider case: drive the hook into generating-and-untracked, spy on `window.confirm`, then assert it was not called, `stop_streaming` was sent before `reset_session`, and no Undo action was offered.
3. Resolve the now-dead `=== false` guards at `frontend/src/components/Header.jsx:119` and `:196` — `clearChat` returns `true` on every path since the cancel branch was removed, and the comment explaining the contract was deleted from `ChatContext.jsx`.
4. Rename the `skipConfirm` option at `frontend/src/contexts/ChatContext.jsx:837` to `skipUndo` (updating `Sidebar.jsx:231` and the tests) — its only remaining effect is suppressing the Undo offer, so the current name will lead a future caller to lose that affordance unintentionally.
5. Restore the original `### PR #950` sentence in `CHANGELOG.md:9` and add a `### PR #952 - 2026-09-17` section — the file is one section per PR, and rewriting #950 erases what that release actually shipped.

## Evidence
Evidence: 6/10. Adequate in volume for a change this small, but the central claim is demonstrated against a stand-in rather than the shipped component; the missing artifact is a real-provider test with `window.confirm` spied and asserted uncalled (see required change 2).

## Notes
- Stale comment claims mid-generation "is the one path that still confirms" and keeps a now-dead confirm spy — `frontend/src/test/new-chat-undo.test.jsx:352`.
- Reflowed comment left mid-sentence after the "so Undo can put it back" clause was deleted — `frontend/src/contexts/ChatContext.jsx:847`.

---

garland3:
## Summary
Removes the remaining `window.confirm` from the New Chat flow so stopping an in-flight, untracked reply clears immediately, replacing the dialog with a `toast.info('Response stopped. New chat started.')` acknowledgement and dropping the now-dead `clearChat() === false` guards in `Header.jsx`.

## Verdict
Approve with nits.

## Evidence
Evidence: 6/10. Proportionate for a small, mostly-deletion change (5 files, +21/-56) with a stated verification run, but the headline behavior is asserted against a local re-implementation of `clearChat` in `frontend/src/test/new-chat-stops-generation.test.js` rather than the shipped provider — the single most valuable missing artifact is a case in `new-chat-undo.test.jsx` that drives the real provider into generating-and-untracked, asserts `window.confirm` was never called, and asserts the new toast fired.

## Notes
- The stop-and-clear path is the least recoverable New Chat case yet is the only one with no Undo — consider building `undoSnapshot` whenever `hasContent` (`frontend/src/contexts/ChatContext.jsx:855`).
- `skipConfirm` now only suppresses Undo; rename to `skipUndo` and update `Sidebar.jsx:231` (`frontend/src/contexts/ChatContext.jsx:837`). Relatedly, the toast fires on the conversation-delete path where the user never pressed New Chat (`ChatContext.jsx:901`), and is emitted even when `sendMessage` is falsy and no stop frame was sent.
- Stale artifacts of the removed dialog: the CHANGELOG entry for PR #950 still describes the surviving confirm (`CHANGELOG.md:9`), and `new-chat-undo.test.jsx:352` keeps a dead `window.confirm` spy and a comment describing it.

---

PR #954, opened 2026-09-18
fix: keep run-produced files downloadable after the run ends

### LINKED ISSUE #953: Files produced by a tracked run cannot be downloaded after the run ends: the run's session (and its file map) is deleted
## Symptom

A file produced by a tool during an agent-mode turn (e.g. an image artifact named `mcp_image_0.jpeg`) cannot be downloaded from chat or from the session file list. Both controls fail with:

```
Error: could not download mcp_image_0.jpeg -- Session or file manager not available
```

The same file downloads fine from the **File Library**, which fetches by S3 key and does not go through the session file map. So the bytes are stored correctly; only the websocket `download_file` path fails.

## Cause

A tracked run executes against its own `Session` (`run_record.session_id`), separate from the connection session. Tool artifacts are registered into **that** session's `context["files"]` by `ChatService._update_session_from_tool_results`.

When the run finishes, `_release_finished_run` (`atlas/main.py:182`) ends **and deletes** the run session:

```python
await chat_service.end_session(session_id)
delete = getattr(chat_service.session_repository, "delete", None)
if delete is not None:
    await delete(session_id)
```

Nothing copies the run session's file map back to the connection session — `_seed_run_session_files` (`atlas/main.py:282`) only copies connection → run, at the start of the turn.

So once the run completes, the only session that ever knew about the file is gone. This directly contradicts the stated intent of `RunRegistry.records_for_user` (`atlas/application/chat/runs/registry.py:275`):

> The download path needs the terminal ones: a file is usually fetched just *after* the run that produced it finished, and that run's session is where the file is registered.

The run *record* is retained; its *session* is not.

## Why the error message is misleading

`_download_session_candidates` (`atlas/main.py:262`) yields the connection session first, then the user's run sessions, and `main.py:1592`'s loop keeps the **last** response when every candidate fails:

- connection session exists, file not in its map → `"File not found in session"`
- terminal run's session was deleted → `handle_download_file` hits the `not session` branch → `"Session or file manager not available"`

The second overwrites the first, so the user sees an error about a missing session instead of a missing file. Even after the root cause is fixed, the loop should surface the most informative error (or the first candidate's) rather than the last one.

## Repro

1. Run an agent-mode turn where an MCP tool returns an image artifact (any image-generating MCP server).
2. After the run completes, click the download control on the image in chat, or in the session files panel.
3. Observe the error above. Download the same file from the File Library — it works.

## Suggested fix

Pick one (or both):

1. **Merge run-session files back to the connection session** when the run finishes, before deleting the run session — the mirror of `_seed_run_session_files`. Cheapest and keeps the session-map contract intact.
2. **Don't delete a terminal run's session while its record is retained** — instead let it be reaped together with the record, so `records_for_user`'s documented promise actually holds.

Also worth doing regardless: make the `download_file` loop in `atlas/main.py:1592` report the most useful error across candidates, not the last.

Affected files: `atlas/main.py` (`_release_finished_run`, `_download_session_candidates`, `download_file` branch), `atlas/application/chat/service.py:1065` (`handle_download_file`).

garland3:
Fixes #953.

## Problem

A file produced by a tool during a tracked (agent-mode) run could not be downloaded from chat or the session file list:

```
Error: could not download mcp_image_0.jpeg -- Session or file manager not available
```

The same file downloaded fine from the File Library, which fetches by S3 key and never touches the session file map — so the bytes were stored correctly.

## Root cause

A tracked run executes against its own `Session`, and `ChatService._update_session_from_tool_results` registers tool artifacts into **that** session's `context["files"]`. When the run finishes, `_release_finished_run` ends *and deletes* that session. `_seed_run_session_files` only copies connection → run at the start of the turn; nothing copied back. So once the run completed, the only session that ever knew about the file was gone.

## Fix

- **`_merge_run_session_files`** — the mirror of `_seed_run_session_files`. Before the run session is deleted, its file map is merged onto the connection session, which outlives the run and is the first candidate a download frame searches. Existing entries win, so a finished run cannot overwrite a name the connection is still using for something else.
- **Most informative error wins.** The `download_file` loop kept the *last* candidate's error, so a reaped run session's "Session or file manager not available" masked the connection session's accurate "File not found in session". `_download_error_rank` now ranks failures (a storage-layer error beats "file not found" beats "no session") and the loop keeps the best one.

## Tests

Six new tests in `atlas/tests/test_parallel_run_transport.py`: artifacts merge back, the connection's own files win a name collision, untracked turns are a no-op, a repository failure does not break release, the merge happens *before* the delete (asserted from inside a fake `delete`), and the error ranking.

`bash test/atlas_tests.sh` → 3412 passed, 17 skipped.


garland3:
## Summary
Fixes issue #953 by merging a finished tracked run's `context["files"]` back into the websocket connection session before the run session is deleted, and by ranking candidate download errors so the most informative one is returned instead of the last one.

## Verdict
Request changes.

## Required changes
1. Stop dropping the run's artifact on a name clash at `atlas/main.py:213` — `setdefault` discards the entry (and its storage key), so a client sending the correct `s3_key` still gets "File not found in session" after the run session is deleted; insert under a deduplicated name or index merged entries by key.
2. Rank on a machine-readable discriminator rather than display copy at `atlas/main.py:325-338` — `_DOWNLOAD_ERROR_PRIORITY` hard-codes two strings authored in `atlas/application/chat/service.py:1088,1095`; reword either and every error ties at rank 0, reinstating #953 with a green suite. Return an `error_code` (`no_session` / `not_found` / `storage`) and rank on that, or at minimum export the literals as shared constants and import them.
3. Do not forward unclassified exception text to the client at `atlas/main.py:336` — rank 0 now promotes raw storage-layer strings, which commonly carry bucket name, object key, endpoint host and principal details; return a fixed user-facing message and log the exception server-side.
4. Give unmapped errors a middle rank rather than the top one at `atlas/main.py:333` — a transient storage failure on an unrelated candidate currently outranks the accurate "File not found in session" from the connection session.
5. Make `connection_session_id` a required keyword-only parameter of `_release_finished_run` (`atlas/main.py:216-223`) and pass it by name at `atlas/main.py:1630` — an omitted or mis-ordered positional silently disables the merge with no error and no log.
6. Handle the post-disconnect case at `atlas/main.py:238` — `_handle_disconnect` ends the connection session while detached runs keep executing, so a run finishing after the tab closes merges into a dead session that no reconnect will consult; anchor the merge on the conversation (or skip and log when the target session is inactive) so the loss is visible.
7. Add a test that drives the candidate-selection loop (`atlas/main.py:1668`) end to end — extract it into a helper and assert `[session-missing, file-not-found] → file-not-found`, `[file-not-found, success] → success` — plus a test binding `_DOWNLOAD_ERROR_PRIORITY` to the strings `handle_download_file` actually produces.

## Evidence
Evidence: 8/10. Well-evidenced for a backend fix — unit tests cover the merge helper, the release ordering and the rank function, with a full-suite pass — but nothing records the original reported symptom ("could not download mcp_image_0.jpeg") being re-run successfully through the real websocket download path.

## Notes
- Merge silently returns when the connection session is missing (`atlas/main.py:206`); log a warning so artifact loss is distinguishable from "no files produced".
- The connection session's file map now grows unbounded for the life of the socket and is copied wholesale into each new run session (`atlas/main.py:207`, `:365`) — worth a follow-up cap.
- `# pragma: no cover` at `atlas/main.py:212` marks a branch that `test_merge_failure_does_not_break_release` now exercises.

---

garland3:
Thanks — addressed all seven in d5c8e538.

1. **Name clash no longer drops the artifact.** `_merge_one_file` keeps the connection's entry under the plain name and files the run's alongside it (`name (1).ext`), so its storage key survives and an `s3_key` lookup resolves. Re-merging the *same* stored object is idempotent rather than building a chain.
2. **Ranking is on a machine-readable code.** New `DownloadError` enum in `atlas/application/chat/service.py` (`bad_request` / `no_session` / `not_found` / `storage`); every reply carries `error_code` and `_download_error_rank` compares that. Display copy is now free to change.
3. **Storage exception text no longer reaches the client.** It returns the fixed "Unable to retrieve file content" with code `storage`; the exception keeps going to the server log. A test asserts the bucket name stays out of the reply.
4. **Unmapped codes sort in the middle** — level with `no_session`, below `not_found` and `storage`, so a transient failure elsewhere can't outrank the connection session's accurate answer.
5. **`connection_session_id` is required and keyword-only**, passed by name at the call site.
6. **Post-disconnect case is visible.** The merge checks `session.active` and, when the socket is gone, logs a warning naming the run and the file count and saying they remain reachable from the File Library. Anchoring on the conversation so the files survive a *reconnect* is a real improvement but a larger change than this fix — flagged below as follow-up rather than folded in.
7. **The loop is extracted and tested.** `_resolve_download` now owns the candidate loop; tests cover `[not_found, no_session] → not_found`, `[not_found, success] → success` (and that the search stops there), and `[not_found, unmapped] → not_found`. Separate tests in `test_chat_download_tool_filenames.py` bind each code to the reply `handle_download_file` actually produces, against a real `ChatService`.

Notes: the stale `# pragma: no cover` is gone, and the missing-connection-session return now logs. Left as follow-ups: capping the connection file map's growth over the life of a socket, and conversation-anchored persistence (item 6).

On evidence — you're right that no run exercises the original symptom through the live websocket path; verification is the unit tests plus code reading. `bash test/atlas_tests.sh` → **3436 passed, 17 skipped**.

garland3:
## Summary
Merges a finished tracked run's `context["files"]` back onto the connection session before the run session is reaped, so tool-produced artifacts stay downloadable from chat (issue #953). Also adds machine-readable `error_code` values to `file_download` replies, ranks download failures so the most informative one wins, and stops relaying raw storage exception text to the client.

## Verdict
Approve with nits.

## Evidence
Evidence: 8/10. The change is well demonstrated for its size — reproduction, root-cause mapping, targeted unit tests per failure mode, and a full-suite run; the main gap is that the originally reported symptom is never closed end to end (no test drives a tool artifact through run completion, session reaping, and a real `handle_download_file` returning bytes).

## Notes
- Keyless entries compare equal in `_file_key`, so two different files with no `key` collapse and the run's artifact is dropped — `atlas/main.py:249`.
- On a name collision the run's file is filed as `name (1).ext` with no `files_update`, so a name-only download request silently returns the connection file's bytes — `atlas/main.py:240`.
- The merged connection file map is never pruned and is copied wholesale into every subsequent run session, growing per turn for the life of the socket — `atlas/main.py:222`.

---

garland3:
Thanks. Took two of the three nits in 7bc7a294, plus the evidence gap.

- **Keyless entries no longer collapse.** `_same_stored_file` now requires a shared non-empty `key`; two keyless entries are *unknown*, not equal, so the run's artifact is filed alongside rather than mistaken for a duplicate.
- **End-to-end test added**, closing the gap both reviews flagged: a tool artifact is produced on a tracked run's own session, `_release_finished_run` merges and reaps that session, and the bytes come back through the ordinary `handle_download_file` path under the reported name `mcp_image_0.jpeg`. Verified it **fails** with the merge disabled, so it pins the actual regression rather than passing vacuously.
- **Name-only request after a collision** — left as is. The connection session's entry keeps the plain name deliberately: it is the file the user attached and the one a name-only request has always meant. The chat and file-panel controls send `s3_key` (#951), which resolves unambiguously, so the run's artifact is reachable; a name-only client hitting a collision gets the attached file, which is the same answer it got before this PR. Adding a `files_update` frame to surface the suffixed name is a UI-visible change worth its own PR.
- **Unbounded file map growth** — agreed, and still a follow-up: it needs a cap plus a matching change to what `_seed_run_session_files` copies forward, which is broader than this fix.

`bash test/atlas_tests.sh` → **3437 passed, 17 skipped**.

garland3:
## Summary
Merges a finished run session's file map back into the connection session before the run is reaped, so tool-produced artifacts stay downloadable, and adds ranked `error_code` values to `file_download` failure replies while no longer relaying raw storage exception text to the client.

## Verdict
Request changes.

## Required changes
1. Make the seed/merge round trip idempotent for keyless entries at `atlas/main.py:240` — `_same_stored_file` returns "unknown" for two entries lacking `key`, so a file seeded into the run session by `_seed_run_session_files` is merged back as `name (1).ext` on every tracked turn, roughly doubling phantom copies per turn and growing per-turn copy cost quadratically. Either skip entries the target already holds identically (`existing == meta`) or stamp seeded entries and skip them on merge-back.
2. Handle suffix exhaustion at `atlas/main.py:245` — the `range(1, 1000)` loop falls off the end and silently discards the artifact with no log, which the defect above makes reachable by ordinary use; fall back to a guaranteed-unique key and log a warning.
3. Reuse the existing collision helper instead of the second scheme at `atlas/main.py:230-249` — `file_processor._merge_without_displacing` / `FileManager.unique_key` already do this with a different identity rule and suffix format (`name_1.ext`), and `name (1).ext` is a session key no other path can produce or round-trip through `sanitize_filename`.
4. Give unmapped download error codes their own rank at `atlas/main.py:383` — `_DOWNLOAD_UNKNOWN_RANK` currently equals the `no_session` rank, so which reply the user sees depends on candidate order rather than informativeness; add a test asserting both orders agree.
5. Assert the collision name directly in `atlas/tests/test_parallel_run_transport.py:503` and cover reachability-by-key after a collision — the current `in merged.values()` check passes under any key, leaving the documented `name (1).ext` contract and the "still reachable by `s3_key`" justification unverified.

## Evidence
Evidence: 9/10. The change is well demonstrated — reproduction, root cause, per-failure-mode tests, and a regression test confirmed failing without the fix; the only gap is that the knowingly-deferred unbounded-growth path has no characterization test recording how large the map gets per turn.

## Notes
- Download failure copy is internal jargon with no recovery hint, and the File Library fallback is logged server-side but never told to the user (`atlas/application/chat/service.py:1107`).
- Sanitized storage failures give the user no correlation id to quote in a report (`atlas/application/chat/service.py:1144`); the client filename also reaches `logger.error` unsanitized at `:1140`.
- The new `error_code` field is undocumented and unconsumed by the frontend, which still branches on free-text `data.error` (`frontend/src/handlers/chat/websocketHandlers.js:637`).

---

garland3:
Good catch on #1 — that was a regression I introduced with the previous nit fix, and it was reachable by ordinary use. Fixed in 64ba373f.

1. **Seed/merge round trip is idempotent again.** `_same_stored_file` now also treats *equal* entries as the same entry, which is exactly what the seeded copy coming home looks like. Two genuinely different keyless entries stay distinct, so the original nit stays fixed. A run re-emitting an artifact that already has a suffixed key refreshes that entry rather than adding another, so repeated merges converge. Test drives three seed/merge turns over a keyed and a keyless file and asserts the map is byte-identical to where it started.
2. **Suffix exhaustion is gone** — collisions now use `FileManager.unique_key`, whose counter is unbounded, so there is no fall-off path to log.
3. **One collision scheme.** Adopted `FileManager.unique_key`, giving `name_1.ext` — the format the artifact ingest path produces and one `sanitize_filename` round-trips. I kept the merge's own identity rule rather than calling `_merge_without_displacing`: that helper keys on `original_filename` and *replaces* the matching entry, which for this path would let a finished run displace a file the user attached. Same suffix helper, same key format, different precedence — deliberately.
4. **Unmapped codes have their own rank**, between `storage` and `no_session`. Added tests asserting both candidate orders agree, for unmapped-vs-`not_found` and unmapped-vs-`no_session`.
5. **Collision name asserted directly** (`{"a.txt": conn, "a_1.txt": run}`, exact equality), plus a test against a real `ChatService` that downloads the collided run artifact **by its `s3_key`** and checks the bytes are the run's, not the other file's.

Also took the log note: the client-supplied filename now goes through `sanitize_for_logging`. Left alone: the `error_code` field is additive and the frontend's free-text branching still works — wiring the UI to it is a frontend change worth its own PR, as is a correlation id in storage failures.

On the remaining evidence gap — a characterization test for the deferred unbounded-growth path: with #1 fixed, the per-turn growth it would have characterized is gone. What remains is genuine accumulation of distinct artifacts over a long socket, which is the follow-up cap.

`bash test/atlas_tests.sh` → **3441 passed, 17 skipped**. CI was green on 7bc7a294.

garland3:
## Summary
Merges a finished parallel run's session file map back into the connection session before the run session is reaped, so artifacts produced by tracked runs stay downloadable by name; also stops relaying raw storage exception text to the client and adds an `error_code` field to download error frames.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — a regression test drives the original symptom through the production release path and the ordinary download handler, collision and idempotence cases have explicit expected maps, and the leak fix is proven negatively. No significant artifact is missing; CI status is reported against an earlier commit rather than the head.

## Notes
- Collision-suffixed names are never pushed to the client, so a name-only download after a collision can return the other file's bytes (`atlas/main.py:245`) — emit a `files_update` or note the deferral.
- `connection_session.active is False` also holds briefly during New Chat / conversation restore, so a run releasing in that window takes the "socket gone" branch and drops its artifacts (`atlas/main.py:213`).
- `_merge_one_file`/`_same_stored_file` re-implement `file_processor._merge_without_displacing`; the two write the same map and will drift (`atlas/main.py:230-260`).

---

garland3:
The New Chat nit was a real bug — fixed in 82a25159.

Confirmed the window: `handle_reset_session` calls `end_session(session_id)` and then `create_session(session_id, ...)` under the *same* id, so `active is False` holds briefly on a live socket during both New Chat and conversation restore. A run releasing in that window was taking the "socket gone" branch and discarding artifacts the user could still see on screen.

Now only a session that is genuinely **absent** short-circuits the merge (that stays a warning naming the File Library). An **inactive** one is merged into, with the ambiguity logged at info: writing into a session nobody reads costs nothing, whereas skipping loses files for certain. Test covers an inactive connection session receiving the merge.

On the other two:

- **Collision names never reaching the client** — noting the deferral, as you suggested. The plain name deliberately stays with the connection's file (the one the user attached, and what a name-only request has always meant), and both chat controls send `s3_key`, which resolves unambiguously — there is a test downloading the collided run artifact by key. Emitting `files_update` from the release path is a UI-visible change and belongs in its own PR.
- **Drift with `_merge_without_displacing`** — documented in the docstring rather than merged. That helper keys on `original_filename` and *replaces* the entry it matches, which on this path would let a finished run displace a file the user attached. It now says so explicitly, and points at the `FileManager.unique_key` suffix helper and key format the two do share, so the difference reads as a choice.

`bash test/atlas_tests.sh` → **3442 passed, 17 skipped**. CI was green on 7bc7a294 and on 64ba373f; watching it on this head.

garland3:
## Summary
Merges a finished run session's `context["files"]` back onto the connection session so tool-produced artifacts stay downloadable after the run session is reaped, and reworks download failure handling to rank candidate errors, add an `error_code` field, and stop leaking storage detail into the user-facing message.

## Verdict
Request changes.

## Required changes
1. Make the merge survive session replacement at `atlas/main.py:213` — `handle_reset_session` (`atlas/application/chat/service.py:938-958`) ends the session, awaits, then calls `create_session`, which installs a *new* `Session` object; a merge landing in that window writes into the discarded object and the artifacts are silently lost. Re-fetch and re-apply after the write, or carry the file map forward onto the new session on reset. Add a test that interleaves the reset path with a run release rather than hand-setting `active = False`.
2. Accept `original_filename` as an identity signal in `_same_stored_file` (`atlas/main.py:258`) — the ingest path refreshes an artifact in place under a *new* storage key when a tool re-emits the same name, so the merge sees a key mismatch and files `a_1.txt`, `a_2.txt`, … one per turn, while name-only downloads keep returning the first turn's bytes. Keep the existing precedence rule so user attachments are still not displaced, and add a three-turn test asserting one converged entry bound to the newest key.
3. Give `_run_tool_producing` a distinct body per file in `atlas/tests/test_chat_download_tool_filenames.py:643` — both fixtures write `b"a,b\n1,2\n"`, so the collision test passes whichever entry the key resolves to and cannot verify the claim it exists to verify. Assert the returned bytes, filename, and key are the run artifact's.

## Evidence
Evidence: 9/10. The change is well-demonstrated for its size — reported symptom, traced root cause, a regression test confirmed failing without the fix, and per-edge tests with exact expected maps; the main gap is that CI is reported green only on earlier commits, not on the head.

## Notes
- Seeded run artifacts cross conversation boundaries on a shared socket and become model- and tool-visible in unrelated conversations (`atlas/main.py:293`).
- Sanitize the exception text as well as the filename in the download failure log (`atlas/application/chat/service.py:1140`); the sibling attach handler already strips CR/LF.
- User-facing download errors are internal jargon with no correlation id to quote (`atlas/application/chat/service.py:1105,1137-1147`).

---

garland3:
All three fixed in 2a559dc8, and both of the first two turned out to be real losses, each now pinned by a test confirmed failing without its fix.

1. **Merge survives session replacement.** The merge re-fetches the connection session after writing and re-applies to the replacement when the object changed under it. The test drives a real `handle_reset_session` interleaved into the merge (patched `get` triggers the reset exactly once, after the merge has read the session it is about to write into), asserts the live session is a *different* object, and then downloads the artifact through the ordinary handler.
2. **Re-emitted artifacts converge.** `_same_stored_file` now also accepts a shared non-empty `original_filename` — the advertised name `_merge_without_displacing` already treats as an artifact's identity — and a matched entry is refreshed with the newer ref, so it stops pointing at superseded bytes. Three-turn test asserts a single entry and that a name-only download returns the *newest* bytes, not the first turn's. Entries sharing none of key / equality / advertised name are still unknown rather than equal, so two genuinely different files are still both kept.
3. **The collision test no longer passes either way.** `_run_tool_producing` takes a per-file `body`; the test asserts the two keys actually differ, then downloads by the run's key and checks the run's bytes come back.

Also sanitized the exception text in the download failure log, alongside the filename.

On the remaining notes: seeded artifacts crossing conversation boundaries on a shared socket is a pre-existing property of `_seed_run_session_files` (it copies the whole connection map into every run session) rather than something this PR introduces — worth its own issue, and it belongs with the file-map cap already deferred. A correlation id in user-facing download errors is a reasonable improvement I'd rather not fold into a fix this size.

`bash test/atlas_tests.sh` → **3444 passed, 17 skipped**. On CI: it settled green on **82a25159** (all checks, 0 non-passing) — that was the head when I wrote the last comment. Watching it on this head now.

garland3:
## Summary
Fixes loss of tool-produced artifacts by merging a finished run session's file map back into the connection session before the run session is deleted, and adds a structured `error_code` to `file_download` failure frames with priority-ranked candidate resolution.

## Verdict
Request changes.

## Required changes
1. Scope the advertised-name identity rule in `_same_stored_file`/`_merge_one_file` (`atlas/main.py:271`) to entries of matching provenance — today any run artifact whose `original_filename` matches a user-attached file overwrites that entry's storage ref, so a name-addressed download silently returns different bytes and the attached file's ref is dropped from the session map.
2. Add a test in `atlas/tests/test_parallel_run_transport.py` with two entries sharing the same `original_filename` and different `key`s, asserting `merged == {"a.txt": conn, "a_1.txt": run}` — the existing collision test at `atlas/tests/test_chat_download_tool_filenames.py:611` rewrites the run entry's advertised name, so the natural collision is never exercised.
3. Gate the session-replacement replay at `atlas/main.py:246` on conversation identity — after New Chat, the merge re-applies the old run's files onto the freshly created session, so the previous conversation's attachments and artifacts reappear in an unrelated conversation and are then seeded into its run sessions.
4. Do the same-file scan over `target` before taking a free name in `_merge_one_file` (`atlas/main.py:262`) — otherwise one artifact can occupy both a plain and a suffixed key, and `_resolve_session_file` hits its ambiguity guard and reports "File not found in session" for a file that is present twice.
5. Log at info in the suffix branch of `_merge_one_file` (`atlas/main.py:264`) naming the run, requested name and assigned key — the only merge outcome that changes what the user sees currently leaves no server-side trace, and the `files_update` frame for it is deferred.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size, with per-failure-mode tests (merge-back, collision, untracked turn, release failure, ordering) plus regression tests confirmed failing without their fix, and an end-to-end path through run release and session reaping. The one missing artifact is a CI result on the head commit itself rather than on its ancestors.

## Notes
- Merge is O(N²) in the connection file map, which grows for the life of the socket; build key/name indexes once per merge (`atlas/main.py:282`).
- `error_code` is added to the protocol but undocumented and unread by the frontend, which still branches on free-text `error` (`atlas/application/chat/service.py:52-67`).
- Merge/reconciliation helpers are domain policy living in the websocket entrypoint; they belong under `application/chat` (`atlas/main.py:183-318`).

---

garland3:
All five in 315fa2de. Item 1 was my round-5 fix overshooting — you were right that it displaced attachments.

1. **Advertised-name identity is scoped by provenance.** Entries the merge writes carry a stamp, and the `original_filename` signal applies only *between stamped entries*. A re-emitted artifact still converges onto its own entry; a user's attachment carries no stamp, so a run artifact sharing its advertised name is a different file and the attachment keeps both its name and its storage ref. Equality comparison ignores the stamp, so the seed/merge round trip stays idempotent.
2. **Test added** with two entries sharing `original_filename` and differing `key`s, asserting exactly `{"a.txt": attached, "a_1.txt": produced}` — the natural collision, no rewriting of the advertised name.
3. **Replay is gated on conversation identity.** New Chat replaces the session *and* the conversation; replaying there leaked the old conversation's files into an unrelated one. Two tests now: a same-conversation object swap still replays, and a real `handle_reset_session` interleaved into the merge leaves the new conversation's file map empty.
4. **The same-file scan runs before the name check**, so an artifact already held under a suffixed key cannot also claim the plain name. Test: after the colliding attachment is removed and the plain name is free again, the next merge still refreshes `a_1.txt` and the map has exactly one entry.
5. **The suffix branch logs** run, requested name and assigned key, all sanitized.

Took the O(N²) note too — the map is indexed by key and advertised name once per merge and kept current as entries land, instead of rescanned per artifact. Left: the `error_code`/frontend wiring, and moving these helpers out of `main.py` into `application/chat` — the latter is a sensible home but a refactor of a file this PR is already touching heavily, so I'd rather it not ride along.

`bash test/atlas_tests.sh` → **3447 passed, 17 skipped**. CI settled green on **2a559dc8** (0 non-passing, all checks) — that was the head when I last commented; watching this one now.

garland3:
## Summary
Merges a finished parallel run's session file map back into the connection session so tool-produced artifacts stay downloadable, adds an identity/collision scheme (`_FileIndex`, `_MERGED_FROM_RUN` provenance stamp), and ranks download failures across candidate sessions via a new `DownloadError` code.

## Verdict
Request changes.

## Required changes
1. Stop stamping `_MERGED_FROM_RUN` on entries refreshed in place at `atlas/main.py:262` — stamp only when the merge takes a genuinely new slot (`index.put` for a new name), or carry the held entry's provenance forward. As written, the seed/merge round trip stamps user attachments, which admits them to advertised-name matching so the next same-named run artifact overwrites the attachment's storage ref and both name- and key-addressed downloads fail.
2. Add a regression test covering that path: run one seed/merge round trip before merging a colliding artifact and assert `{"a.txt": attached, "a_1.txt": produced}`. The existing `test_a_run_artifact_does_not_displace_an_attachment_of_the_same_name` passes only because it merges into a fresh, unstamped map.
3. Apply the conversation-identity check to the first write in `_merge_run_session_files` (`atlas/main.py:213`), not just the replay — capture the run's `conversation_id` at registration and skip+log on mismatch. If a reset/restore completes before the merge reads the connection session, artifacts from conversation A are merged into B and then seeded into B's runs. Cover it with a test where the reset fully completes before the merge begins.
4. Drop the replaced entry's old `key` / `original_filename` from `_FileIndex` in `put` (`atlas/main.py:293`) — `setdefault` leaves a superseded key indexed, so a later entry in the same run map carrying that key refreshes the already-updated name and is discarded.

## Evidence
Evidence: 9/10 — the change is well demonstrated for its size: the reported symptom is reproduced end-to-end through the real release and download handlers, and per-edge tests assert exact expected maps. The only thin artifact is that CI is confirmed green on an ancestor (`2a559dc8`) rather than on this head, which is the most structurally different commit in the series.

## Notes
- Move the merge policy (`_merge_run_session_files`, `_FileIndex`, `_merge_one_file`) out of the websocket entrypoint into `atlas/application/chat/`, where `file_processor._merge_without_displacing` already implements a near-identical rule — `atlas/main.py:183-400`.
- Keep `_DOWNLOAD_ERROR_PRIORITY` next to the `DownloadError` enum it orders, with an exhaustiveness test, so a new member can't silently fall to unknown rank — `atlas/main.py:494`.
- A collision-suffixed artifact (`a_1.txt`) is only logged server-side; no `files_update` frame tells the user the name exists — `atlas/main.py:262`.

---

garland3:
All four in 94505d27. Item 1 was a real hole — the round trip was re-admitting attachments to exactly the matching I'd just excluded them from.

1. **Provenance is carried forward, not re-stamped.** A refresh keeps the held entry's provenance; only a genuinely new slot gets stamped. So one ordinary turn no longer marks the user's attachments as run output.
2. **Regression test** does a full seed/merge round trip first, *then* merges a different artifact advertising the same name, asserting `{"a.txt": attached, "a_1.txt": produced}`. Confirmed failing with the carry-forward reverted.
3. **The conversation check moved to the first write**, with the run's `conversation_id` threaded from `_release_finished_run`. Unknown on either side does not block, so runs registered without a conversation behave as before. Test: the connection is already on conversation B when the merge for a run from A begins, and B's file map stays empty; a matching-conversation test covers the other side. Confirmed failing without the gate.
4. **`_FileIndex.put` retires the outgoing entry's key and advertised name.** Test: turn one produces `a.txt`, turn two re-emits it under a fresh key *and* produces a second file reusing the old one — without eviction the second file overwrites the just-refreshed entry and loses its own name. Confirmed failing without the eviction. (My first attempt at this test passed either way; it's rewritten to actually reach the stale-index path.)

Also took the exhaustiveness note: a test binds `_DOWNLOAD_ERROR_PRIORITY` to the `DownloadError` members, so a new code can't quietly fall to unknown rank.

Still deferred, and I'd rather they were separate PRs than fold more into this one: moving the merge policy into `application/chat` (a file-level refactor of `main.py`, which this PR already touches heavily), the `files_update` frame for a suffixed name, and the frontend's `error_code` wiring.

`bash test/atlas_tests.sh` → **3452 passed, 17 skipped**. On CI: green on **2a559dc8**; `315fa2de` was still running when this went up, and I'm watching both it and this head.

garland3:
## Summary
Merges a finished tracked run's session file map back into the live connection session so run artifacts stay downloadable by name, with collision suffixing, provenance stamping and a conversation-isolation gate. Also adds a machine-readable `error_code` to `file_download` failure frames and extracts candidate ranking into `_resolve_download`.

## Verdict
Request changes.

## Required changes
1. Carry per-entry provenance through the session-replacement replay at `atlas/main.py:246` — the replay builds a `_FileIndex` over the empty map of the new `Session` and re-writes every entry via `_stamp(meta, True)`, so seeded user attachments get marked as run output and the next artifact advertising the same name overwrites the attachment's storage ref. Merge the already-correct result map (`connection_session.context["files"]`) into the replacement instead of re-running `_merge_one_file` over the raw run map, and add a test that interleaves a same-conversation session swap with a seeded attachment then merges a colliding artifact, asserting `{"a.txt": attached, "a_1.txt": produced}`.
2. Require the incoming entry to be run output too before accepting the advertised-name match in `_FileIndex.find` at `atlas/main.py:271` — `_seed_run_session_files` copies attachments into the run map, so a keyless attachment whose `original_filename` matches a previously merged artifact's advertised name silently replaces that artifact.
3. Make `_same_conversation` at `atlas/main.py:290` deny (or at minimum log at warning) when either side has no `conversation_id`, using the run's id captured at registration — as written the only cross-conversation control fails open for any `Session` built without `context["conversation_id"]`, and the merged entries are then seeded into later runs.
4. Apply `sanitize_for_logging` to `str(e)` and `run_session_id` in the catch-all, "connection session is gone" and "inactive session" branches around `atlas/main.py:270` — tool-chosen filenames and keys reach those strings raw, allowing CR/LF log forging that sibling branches already guard against.
5. Include `run_session_id`, `connection_session_id` and the file count in the catch-all warning at `atlas/main.py:271`, and extend `test_merge_failure_does_not_break_release` with a `caplog` assertion — today the one branch that fires on unexpected artifact loss is uncorrelatable to a user report.
6. Track `s3_key` in `_FakeDownloads.calls` and assert each candidate received the caller's key (`atlas/tests/test_parallel_run_transport.py:735`) — key-addressed lookup is the stated justification for not dropping a collided artifact, and the extraction dropping that argument would leave every `_resolve_download` test green.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — the reported symptom is reproduced end to end, per-edge tests assert exact expected maps, and each regression fix is confirmed failing without it; the only gap is a posted CI conclusion for this head commit rather than an ancestor.

## Notes
- Merge policy (`_merge_run_session_files`, `_FileIndex`, `_merge_one_file`) is domain logic sitting in the websocket entrypoint alongside a second, divergent identity rule in `atlas/application/chat/utilities/file_processor.py:889`; worth tracking as a follow-up rather than leaving in a comment thread.
- The keyless fallback in `_FileIndex.find` (`atlas/main.py:300`) is a linear scan with per-comparison dict copies over an unbounded, never-evicted connection file map — quadratic per turn on the miss path.
- User-facing download errors at `atlas/application/chat/service.py:1107,1147` are server-side jargon with no recovery step; now that `error_code` carries the discriminator, the copy can point at the File Library.

---

garland3:
All six in 8d04c019. Item 1 went deeper than the replay: while writing your test I found the same misreading on the **first** write, so I changed where provenance comes from rather than patching the replay alone.

**Root of items 1 and 2.** Provenance was inferred from the target map — an entry matching nothing there was taken for run output. But every connection file is seeded into each run and rides back, so whenever the target couldn't contradict it (an empty one, because the session object had been replaced) the user's attachments got marked as run output. `_seed_run_session_files` now **marks the copies it makes**, and the merge reads provenance off the entry itself. A seeded entry is additionally barred from advertised-name matching outright — it is the user's file riding back out of a run that did not produce it. Neither marker survives into the connection map.

That covers item 2 directly, and it is the reason your suggested test passes: reverting the seed marker fails both `test_a_round_trip_does_not_mark_attachments_as_run_output` and `test_the_replay_does_not_stamp_seeded_attachments`.

Also, as you asked: equality is now checked **before** the advertised name, so a seeded attachment settles against its own slot before the name ever gets a say.

3. **`_same_conversation` logs at warning when it cannot compare.** I did not make it deny: `create_session` does not set a `conversation_id`, so denying would drop the artifacts of a first turn that never left its conversation. A session with no conversation recorded also has no conversation to leak *into*. The unprovable case is now in the record.
4. **Sanitized** `str(e)`, `run_session_id` and `connection_session_id` in the catch-all, "session is gone" and "inactive session" branches.
5. **The catch-all carries** run id, connection session id and file count; `test_merge_failure_does_not_break_release` asserts both ids reach the log.
6. **`_FakeDownloads` records `(session_id, s3_key)`**, with a test asserting every candidate was asked for the caller's key.

Took the quadratic note too — `find` is backed by a content-signature index rather than a linear scan with a dict copy per comparison.

`bash test/atlas_tests.sh` → **3455 passed, 17 skipped**. CI settled **green on 315fa2de** (0 non-passing, all checks); watching this head.

PR #955, opened 2026-09-18
fix: keep run artifacts downloadable when every turn is a tracked run

garland3:
Follow-up to #954 / issue #953. The merge landed, but on the connection where it matters most there was nothing to merge into, so the reported symptom never went away.

## What was still broken

The connection session is created lazily, by the first turn that runs *against* it. A tracked run executes against its own session, so a connection whose every turn is a tracked run never creates one.

1. `_merge_run_session_files` found no connection session, read the absence as "the socket closed", logged *"they remain downloadable from the File Library only"* and dropped the artifacts. The download frame's first candidate is that same missing session, so the user saw **"Session or file manager not available"**.
2. With the session created, the isolation check failed next. Only a turn on the connection session sets its `conversation_id`, so it stayed at whatever `handle_reset_session` minted at the last New Chat, while the run carried the id the transport minted for this turn. Two ids that were never meant to match, so a connection sitting on this very conversation read as one that had moved away — **"File not found in session"**.

## The fix

`_seed_run_session_files` now creates the connection session when it is missing, and records the run's conversation on it — the latter only when no turn has ever run on the connection session itself, because an untracked turn reads that id back when it saves.

## Verification

Reproduced in a browser (Playwright) against a real tracked run — chat history enabled, save mode `server`, agent mode with tools, which is what `turn_is_eligible_for_background_run` requires:

- before: `markdown_to_pptx` output gave "Session or file manager not available"; after creating the session only, "File not found in session"
- after: a 29KB `new-mexico-one-slide.pptx`, its `.html` preview, and an MCP `ImageContent` artifact (`mcp_image_0.png`) all download, and the merge logs no warning

Four regression tests, each verified to fail with its own fix reverted. Backend suite: 3459 passed, 17 skipped.

## Not in scope

The three download failure codes still share one connection-session-shaped message. Worth separating so "the session does not exist" and "the file is not in it" read differently, but that is a separate change.


garland3:
## Summary
Fixes lost tracked-run artifacts by having `_seed_run_session_files` create the connection session when it is missing and stamp the run's `conversation_id` onto it, so the later merge in `_merge_run_session_files` finds a matching target. Adds four regression tests.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well-demonstrated for its size — reproduction, staged before/after observation, targeted regression tests each verified failing with its own half reverted, and a full backend suite run. A pasted merge-log excerpt (warning present, then absent) would be a nice strengthening artifact, not a gap.

## Notes
- Stamp is write-on-every-run, not write-once: two tracked runs in different conversations on one connection are last-writer-wins, so the earlier run's artifacts are dropped — consider only setting `conversation_id` when unset (`atlas/main.py:730`), and add a test for the two-run case.
- `conversation_id` is a fifth positional, unannotated parameter passed positionally at the only call site; make it keyword-only and typed (`atlas/main.py:686`, call site `atlas/main.py:1737`).
- `create_session` can raise `DomainError` inside the shared `try`, so a hook-denied connection session now also skips run-session seeding; also drop the now-dead `if connection_session else None` (`atlas/main.py:714`, `atlas/main.py:735`).

---

garland3:
Thanks — all six points read as real. Pushed `83d36026`.

**Last-writer-wins stamp (BOT-3) / racy guard (Copilot).** Confirmed, and the suggested fix — set `conversation_id` only when unset — puts back the bug this PR opened with. `handle_reset_session` installs a fresh `Session` and sets `conversation_id` on it (`service.py:958-960`), so after New Chat the field is always set and always a value no run will ever match; that is exactly the second failure I reproduced.

So I stopped overwriting it. The connection session now records `run_conversations` — the conversations it started runs in since the user last navigated — and `_same_conversation` accepts membership. Nothing overwrites anything, so two runs in two conversations both come home. The record clears on navigation: New Chat gets it for free (fresh `Session`), restore does it by hand. Test added for the two-run case, verified failing without the membership check.

This also removes the `history.messages` guard: the connection's own `conversation_id` is no longer touched, so there is nothing for an in-flight untracked turn to lose.

**Race on creation (Copilot, 2 votes).** Real — `create_session` overwrites, so two runs starting at once each install a session and the loser merges into a discarded object. Get-or-create now runs under a module lock. Pinned by a test whose fake repository suspends between reading and returning, which is the window; it fails without the lock.

**`create_session` inside the shared try (BOT-3).** Real regression, fixed — creation moved into `_ensure_connection_session` with its own handler, so a hook-denied connection session no longer also skips seeding the run's own files. Test added. The `if connection_session else None` is now load-bearing rather than dead: that helper returns `None` on denial.

**Keyword-only and typed (BOT-3).** Done.

**Stale lifecycle docstring (Copilot).** Reworded, and the creating helper now carries its own docstring.

**PR validation script (Copilot).** Added `test/pr-validation/test_pr955_tracked_run_artifact_download.sh`. The chat turn itself needs a model, so it drives the endpoint's own helpers against a real `ChatService`, a real `FileManager` and the real session repository — produce an artifact in the run's session, release the run, reap the session, then issue the real download — plus the two-conversation and navigation cases and the eligibility gates. Nothing on the path under test is stubbed. Passes 3/3 plus the backend suite.

**Session retention (Copilot, moderate).** Not new to this PR, so not fixed here. `cleanup_disconnected_session` has only ever called `end_session`, and `InMemorySessionRepository.delete` is called only for run sessions — so *every* socket, tracked or not, already leaves one inactive session behind, since an ordinary turn creates one the same way. This change extends that to tracked-only sockets at the same one-per-socket rate; it does not introduce the leak or change its shape. A reaper for connection sessions is worth doing, and is its own change.

Backend suite: 3462 passed, 17 skipped. Re-verified in the browser after the redesign — pptx + html + `mcp_image_0.png` all download, merge logs clean.

garland3:
## Summary
Fixes tracked (background) runs losing their produced artifacts by ensuring a connection session exists before a run is seeded, and by recording the conversations a connection's runs belong to so `_same_conversation` can authorize the merge. Adds a lock around connection-session get-or-create, a `forget_run_conversations` clear on the restore path, regression tests, and a PR-validation script.

## Verdict
Approve with nits.

## Evidence
Evidence: 9/10. The change is well demonstrated for its size — staged before/after failure messages, named success artifacts, seven regression tests each verified failing with its half reverted, and a validation script that stubs nothing on the path under test. The only missing artifact is a pasted excerpt of the merge log showing the warning present before and absent after.

## Notes
- `forget_run_conversations(session)` runs before `handle_restore_conversation` and regardless of its outcome, so a denied/failed restore clears the record while the connection has not moved and an in-flight run's artifacts are then dropped — `atlas/main.py:2148`.
- The incognito-interlude re-mint of `conversation_id` (`atlas/application/chat/service.py:1384`) does not clear the record, so a prior conversation's run still merges; either clear there or restate the invariant in the `_RUN_CONVERSATIONS` comment.
- `_record_run_conversation` evicts oldest-first with no liveness check and no log line, and the cap/eviction is untested — log on eviction and pin the contract with a test at `atlas/main.py:704`.

---

